PR #14, opened 2025-08-29
[WIP] Support min latency MoE mode for trtllm

Atream:
#### Overview:

Support min latency MoE mode for trtllm

#### Details:

I noticed that existing profilers only benchmarked the standard MoE mode; the min-latency mode, which is only available on Blackwell with FP8/FP4, was never profiled. At small batch sizes the performance of min-latency mode diverges markedly from that of the standard mode, causing simulation results to skew. This PR adds profiling support for the min-latency mode under the latest TRT-LLM (1.1.0rc1) with the Triton MoE backend.

- closes GitHub issue: #xxx

### DIFF
--- a/collector/trtllm/collect_moe.py
+++ b/collector/trtllm/collect_moe.py
@@ -284,10 +284,14 @@ def run_moe_torch(moe_type, num_tokens_lists, hidden_size, inter_size, topk, num
     moe.w3_w1_weight = ffn1_weights
     moe.w2_weight = ffn2_weights
 
+    do_finalize = not min_latency_mode
+    if min_latency_mode:
+        do_finalize = True # triton and trtllm backend does not support no_finalize
+
     torch.cuda.synchronize()
     AutoTuner.get().clear_cache()
     with torch.inference_mode(), autotune():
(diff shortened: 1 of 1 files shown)


Atream:
I have some error need to fix and code to clean.
```
File "/code/triton/python/triton_kernels/triton_kernels/matmul_ogs.py", line 412, in matmul_ogs
    if is_hopper_fp8: assert w.stride(-2) == 1, "`w` must be column-major when it has data-type FP8 on capability < 10"
                             ^^^^^^^^^^^^^^^^^
AssertionError: `w` must be column-major when it has data-type FP8 on capability < 10
```

Atream:
another bug:
```
AttributeError: 'NoneType' object has no attribute 'create_weights'
```

Atream:
I have merged the main branch. But I found that there may be some bugs when running min_latency triton kernels TritonFP8QDQFusedMoEMethod on H100.
I raise an issue in trtllm https://github.com/NVIDIA/TensorRT-LLM/issues/7418

tianhaox:
Let's close this for now. Reopen once we need it. Thank you for the effort.

PR #19, opened 2025-09-04
fix: update moe tune

YijiaZhao:
#### Overview:

Tune smaller tokens for moe if autotune failed


YijiaZhao:
> why we need introduce a env to print debug info, how about printing limited info by default?

why we need introduce a env to print debug info, how about printing limited info by default?
This is for convenient debugging. If set to default, each test case would require printing, making it difficult to view the progress bar.

YijiaZhao:
> let's use aic instead of pet, and please make the env specific to this test case with longer name

done

PR #26, opened 2025-09-14
docs: add guidance for adding a new model

tianhaox:
#### Overview:

Guidance for adding a new model in aiconfigurator

#### Details:

Guidance for adding a new model in aiconfigurator

Guidance for adding a new model in aiconfigurator, the readme under docs.

NA

### DIFF
--- a/docs/add_a_new_model.md
+++ b/docs/add_a_new_model.md
@@ -86,6 +86,9 @@ This defines a new model similar to 'LLAMA' with:
 - Vocabulary size = 152064
 - Context size = 32768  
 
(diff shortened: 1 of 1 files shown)


Arsene12358:
Looks good, except for one point, I think there might be a Situation 4: Model needs a new model family. For instance, in the case of nemotron model,  although it uses basic operations without the need to collect new perf data, we need to define a new model class in the sdk/models.py due to its unique way to organize its layers, i.e. permitting each layer to have freedom to choose its own inter_size or n_kv_heads. Also for Situation 3, if the model already requires some brand new operations, user probably will also manually define a new model class in sdk/models.py as well

Arsene12358 (approved):
I think in the current state, it's good to go. we can always improve the doc along the way

PR #42, opened 2025-10-03
chore: update versions and relax versions for numpy and pydantic

biswapanda:
#### Overview:

 we cant install aiperf==0.1.1  and aiconfigurator==0.2.0  together because of pinned numpy version conflict

 > [base 14/14] RUN --mount=type=bind,source=./container/deps/requirements.txt,target=/tmp/requirements.txt     --mount=type=bind,source=./container/deps/requirements.test.txt,target=/tmp/requirements.test.txt     uv pip install --requirement /tmp/requirements.txt --requirement /tmp/requirements.test.txt:
0.304 Using Python 3.12.3 environment at: venv
0.644   × No solution found when resolving dependencies:
0.644   ╰─▶ Because aiconfigurator==0.2.0 depends on numpy==1.26.4 and aiperf==0.1.1
0.644       depends on numpy==2.2.6, we can conclude that aiconfigurator==0.2.0 and
0.644       aiperf==0.1.1 are incompatible.
0.644       And because you require aiconfigurator==0.2.0 and aiperf==0.1.1, we can
0.644       conclude that your requirements are unsatisfiable.

- closes GitHub issue: #xxx


pyproject.toml
```diff
@@ -45,13 +45,13 @@ dependencies = [
     "gradio==5.47.1",
     "jinja2>=3.1.0",
     "matplotlib>=3.9.4",
-    "numpy==1.26.4",
+    "numpy>=2.2.6",
```
biswapanda: @saturley-hall is this fine or you'd recommend using `~=` for compatible release

saturley-hall: This is the convention that is used around it. If we decide to change to `~=` we will do it in bulk.

PR #44, opened 2025-10-03
feat: new cli and sdk to support presets for search space

tianhaox:
#### Overview:

new cli and sdk to support presets for search space

#### Details:

new cli and sdk to support presets for search space
1. now we will populate default search space based on system, model, etc. New class TaskConfig and TaskRunner are introduced to support this feature.
2. support heterogenous analysis in cli
3. connect with new aiconfigurator generator module to generate configs automatically
4. adapt test cases to support new workflow
5. new readme to introduce new workflow and benchmark methodology
6. fix minor issues: injection risk in collector; disable kimi k2 for now due to collector error

### DIFF
--- a/tests/sdk/database/test_database_helpers.py
+++ b/tests/sdk/database/test_database_helpers.py
@@ -1,4 +1,5 @@
-tests/sdk/database/test_database_helpers.py
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
 import os
 import sys
 import yaml


src/aiconfigurator/cli/report_and_save.py
```diff
@@ -0,0 +1,253 @@
...
+                    top_config_dir = os.path.join(exp_dir, f'top{i+1}')
+                    safe_mkdir(top_config_dir, exist_ok=True)
+                    with open(os.path.join(top_config_dir, 'generator_config.yaml'), 'w') as f:
+                        yaml.safe_dump(cfg, f, sort_keys=False)
+                    
+                    artifacts = generate_backend_config.from_runtime(
+                        cfg=cfg,
+                        backend=exp_task_config.backend_name,
+                        version=generated_backend_version or exp_task_config.backend_version,
+                        overrides=build_dynamo_config(args),
```
Ethan-ES, line 248: I think we can move build_dynamo_config(args) outside the loop. I’ll handle it after the merge, as I also need to add the k8s guide section.

src/aiconfigurator/generator/cli_args.py
```diff
@@ -95,8 +95,7 @@ def add_config_generation_cli(parser: argparse.ArgumentParser, default_backend:
         "--generated_config_version",
         type=str,
         default=None,
-        help="Force the target engine version for generated configs. "
-             "(default: fallback to --version)",
+        help="Target engine version for generated configs. If none is given, will skip the generation",
```
Ethan-ES, line 98: When generated_config_version is None, the config is still created from the default template. I can adjust it so it doesn’t generate anything if needed.

Ethan-ES (approved):

src/aiconfigurator/generator/cli_args.py
tianhaox: I think we can just fall back to the latest version or a version same as database. the default version can align with Dynamo major release for better alignment

src/aiconfigurator/cli/report_and_save.py
tianhaox: Yes we need to discuss this design to ensure a smooth UX. Configs can be more algined across different components of AIC.

PR #48, opened 2025-10-07
fix: collector 1.0.0 compatibility

tianhaox:
#### Overview:

to fix
- [x] attn q,kv tensor dimension
- [x] trtllm moe 1.0.0 compatibility
- [x] escaped runtime error capture
- [x] blackwell attn unsupported


collector/trtllm/collect_attn.py
```diff
@@ -260,8 +260,11 @@ def get_context_attention_test_cases():
                         if b*s*num_kv_heads*128*2 >= 2147483647:
                             continue
                         if getSMVersion() >= 100:
+                            # though it's a precheck of gen kernels during the attention op init, this cannot be skipped for now
                             # TLLM_CHECK_WITH_INFO((params.mNumHeadsQPerKv < maxNumHeadsQPerKvInCta || params.mNumHeadsQPerKv % maxNumHeadsQPerKvInCta == 0),
-                            if n >= 32 and n % 32 != 0:
+                            mNumHeadsQPerKv = 1 if n_kv == 0 else n//n_kv
+                            maxNumHeadsQPerKvInCta = 32
```
simone-chen, line 266: Out of curiosity, how's the value 32 decided here? (Asking because TRTLLM seems to have a complex logic deciding the value of maxNumHeadsQPerKvInCta [here](https://github.com/NVIDIA/TensorRT-LLM/blob/017583a94922048703c1c4346a67c84449eb23c6/cpp/tensorrt_llm/kernels/trtllmGenKernels/fmha/fmhaKernels.h#L525).

Is it possible for the value to change across different trtllm versions? If that's the case shall we make it a const value instead for easier tracking?

collector/trtllm/collect_moe_pre_1_0.py
```diff
@@ -0,0 +1,341 @@
...
+        except Exception as e:
+            if aic_debug == 1:
+               print("tune failed for tokens size {}, fallback to tokens size {}".format(num_tokens_lists[max_index], num_tokens_lists[max_index-1]))
+            max_index -= 1
+            if max_index == -len(num_tokens_lists):
+                raise ValueError("tune failed for all tokens sizes")
+            continue
+
+    for num_tokens in num_tokens_lists:
+        hidden_states = torch.randn([num_tokens, hidden_size]).bfloat16().to(torch.device(device))
```
simone-chen: Nit: Think we may not need `.to()` calls as `torch.set_default_device(device)` is already called? Same applies to the rest of the `.to()` calls in this file.
If we wanna make sure the tensor is on device, would `hidden_states = torch.randn([num_tokens, hidden_size], device=device).bfloat16()` be better as it saves the time to copy from CPU to GPU?

collector/trtllm/collect_attn.py
tianhaox: It's a workaround. The value and params naming are all aligned with TRTLLM 1.0.0/1.1.0 source code.
attention op in trtllm will do a initializie() to verify the support matrix both for context operation and generation operation. This will trigger this runtime error throw. It only goes into this execution path on sm100+, thus we're disabling the corresponding combinations.

collector/trtllm/collect_moe_pre_1_0.py
tianhaox: Yes, let me update

Ethan-ES (approved):

PR #51, opened 2025-10-07
fix: Initialize random tensors directly on device

ilyasher:
Some data collection functions (specifically gemm_trt and attention) were very slow because they were creating random input data on CPU and copying it to the GPU, instead of creating it on the GPU directly.

This PR sets adds `torch.set_default_device` to the beginning of all trtllm data collection functions which results in large speedups. gemm_trt collection on L40S went from 10h+ to only 45 mins.


tianhaox:
Have you verified whether all these new modified collectors are functioning in the parallel collection script?

PR #55, opened 2025-10-10
feat: A100 support

simone-chen:
#### Overview:

Add data for A100 collected on NVIDIA A100-SXM4-80GB cluster.

#### Details:

- [ ] Validate data with sanity check

- closes GitHub issue: #xxx


src/aiconfigurator/systems/a100_sxm.yaml
tianhaox, line 1: Have you verified the intra_node_bw of the spec, I'm assuming it's 300000000000, please correct me if I'm wrong.

simone-chen: You're right. I filled in bidirectional bandwidth. Thanks for pointing out!

PR #56, opened 2025-10-12
feat: support gptoss in aiconfigurator sdk

Arsene12358:
#### Waht Does This PR do

- Added gptoss model definitions
- Added window_size as a configurable parameter to attention op
- Updated SOL estimation to support windowed attention 
- Fixed the issue in which some attn test cases would unintentionally be skipped when head_size=64

### DIFF
--- a/tests/sdk/database/test_interpolation.py
+++ b/tests/sdk/database/test_interpolation.py
@@ -317,14 +317,14 @@ class TestCorrectData:
         )
         
         # Set an artificially low value
-        comprehensive_perf_db._generation_attention_data[kv_cache_quant_mode][n_kv][n][b][s] = sol_value * 0.5
+        comprehensive_perf_db._generation_attention_data[kv_cache_quant_mode][n_kv][128][0][n][b][s] = sol_value * 0.5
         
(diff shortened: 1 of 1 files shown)


collector/trtllm/collect_attn.py
tianhaox, line 1: Have you verified whether the new collect attn is working with version 1.0.0 and 1.1.0rc?

collector/trtllm/collect_moe.py
tianhaox, line 1: Same for this, what sm 80 and 86, we're adding a100/l40s support. Do they ahve corresponding moe backend?

collector/trtllm/collect_attn.py
Arsene12358: tested on 1.0.0rc5, and it's able to collect data successfully

collector/trtllm/collect_moe.py
Arsene12358: for gptoss, we only use triton backend when on hopper, in all other cases, the normal logic should apply, i.e. trtllm backend for min latency mode, cutlass for all other cases

PR #65, opened 2025-10-16
feat(ci): add linting and formatting with Ruff, create a developer guide

anish-shanbhag:
#### Overview:

This change adds Python formatting and linting via Ruff, which is also enforced in CI and run as a pre-commit hook. Also adds a developer guide (`DEVELOPMENT.md`) that covers how to set up the project and get started with development.   

#### Details:

The diff is very large but the vast majority are automated formatting changes, and the intent is to introduce zero differences in functionality. A follow-up PR can ignore this commit when checking `git blame` (https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view)


pyproject.toml
```diff
@@ -94,4 +96,58 @@ where = ["src"]
 include = ["*"]
 exclude = ["tests", "docker", "collector", "systems"]
 
+[tool.ruff]
+line-length = 100
```
simone-chen: Shall we set the line-length to a larger value like 120 / 150? From my experience 100 can be way too restrictive and can be troublesome to fix to. :(

collector/collect_NCCL.py
```diff
@@ -1,83 +0,0 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ilyasher, line 1: Do you know why this file got deleted?

ilyasher: Oh it's renamed

DEVELOPMENT.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Run Pre-commit Manually
+
+```bash
+pre-commit run --all-files
+```
+
+### Running Tests
+
+TODO: documentation for testing
```
ilyasher: I think you can just run `pytest` or `pytests tests/sdk`

pyproject.toml
```diff
@@ -94,4 +96,58 @@ where = ["src"]
...
+    "RUF015",  # prefer next(iter()) over list/tuple indexing
+    "N812",  # lowercase imported as non-lowercase
+    "PGH003",  # use specific rule codes for type ignore/noqa
+    "RUF059",  # unpacked variable is never used
+    "UP007",  # require using X | Y for type annotations
+    "UP045", # require using X | None for type annotations
+]
+
+[tool.ruff.lint.isort]
+known-first-party = ["aiconfigurator"]
```
ilyasher: nit: can also add `utils` and `helper`

DEVELOPMENT.md
anish-shanbhag: Added

pyproject.toml
anish-shanbhag: Good point, I updated to 120. Personally I think 150 might be a bit too high since it makes reading on split screen a bit more difficult, but we can always revisit this if needed.

PR #77, opened 2025-10-25
feat: add_prefix_len_modeling

tianhaox:
#### Overview:

Adding `prefix` to both cli and webapp as one of the inputs. Now we can consider prefix cache hit.

#### Details:

Prefix cache impacts the context phase computation both in agg and disagg. We use `prefix` to represent the avg cache hit of context phase to model this averaged state.

In current modeling, the input num_tokens will become isl-prefix for context phase. For the context attention operation, it further uses prefix to compute the right attention scores while other operations don't have interactions with prefix tokens any more.

Theoretically, our modeling of prefix is not trying to collect any new data for attention with prefix cache, but to do a correction to the latency. Considering causality,
1. Assume full len equals s,  compute ∝ s^2
2. For the prefix compute part, compute ∝ prefix^2
3. the remaining compute part is, compute ∝ s^2 - prefix^2
4. Thus we use (s^2-prefix^2)/s^2 to correct the full len attention to the prefix-cached attention.

#### Note
Please help to review
1. the sol computation of context attention and mla.
2. the prefix_correction to latency of context attention
3. whether the impact on SGLang backend is all covered
4. is the modeling reasonable. I have collected context attention data in trtllm and the study shows that it's working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,22 +258,20 @@ To estimate performance, we take the following steps:
   - MOE
   - QWEN
   - DEEPSEEK_V3
+  - Support using huggingface model id if falls into these model family and not MoE models.
 - Operations:
   - Attention
     - MHA/GQA (FP8, FP16)
     - MLA (FP8, FP16)
   - KV Cache (FP16, FP8, INT8)
   - GEMM (FP16, FP8, FP8-Block, FP8-OOTB, SQ, INT8 WO, INT4 WO, NVFP4)
-  - AllReduce (FP16)
+  - CustomAllReduce (FP16)
   - Embedding
   - P2P
   - ElementWise
   - NCCL (all_reduce, all_gather, all-to-all, reduce_scatter)
   - MoE (FP16, FP8, FP8-Block, W4A-FP8, INT4 WO, NVFP4)
   - MLA BMM (FP16, FP8)
-- TRTLLM Versions
-  - 0.20.0
-  - 1.0.0rc3
 - Parallel modes:
   - Tensor-parallel
   - Pipeline-parallel
@@ -284,6 +282,10 @@ To estimate performance, we take the following steps:
   - Aggregated serving (continuous batching)
   - Disaggregated serving
   - MTP (for DEEPSEEK)
+- Backends:
+  - TRTLLM (regular)
+  - SGLang (WideEP and regular)
+  - vLLM (regular with dense models; Non-moe support for now)
 
 ### Data Collection
 
@@ -296,8 +298,8 @@ To go through the process, refer to the [guidance](collector/README.md) under th
 
 | System | Framework(Version) | Status |
 |--------|-------------------|--------|
-| h100_sxm | TRTLLM(0.20.0, 1.0.0rc3) | ✅ |
-| h200_sxm | TRTLLM(0.20.0, 1.0.0rc3) | ✅ |
+| h100_sxm | TRTLLM(0.20.0, 1.0.0rc3), SGLang(0.5.1.post1), vLLM(0.11.0) | ✅ |
+| h200_sxm | TRTLLM(0.20.0, 1.0.0rc3), SGLang(0.5.1.post1), vLLM(0.11.0) | ✅ |
 | b200_sxm | TRTLLM(1.0.0rc6) | ✅ |
 | gb200_sxm | TRTLLM(1.0.0rc6) | ✅ |
 | a100_sxm | TRTLLM(1.0.0) | ✅ |
@@ -316,7 +318,8 @@ Adding a new model will require modifying the source code and perhaps collecting
 
 ## Known Issues
 
-1. MoE memory estimation for the `trtllm` backend needs to consider workspace.
+1. Memory estimation for the backends needs to be studied more.
 2. Results can be overly optimistic in the low-speed, high-throughput region.
+3. SGLang and vLLM support is the initial version, needs more alignment in future.
 
 > **Note**: The results are not final or absolute. They can be inaccurate due to modeling gaps or indicate performance improvement opportunities. The tool aims to align with the framework's current implementation and to provide configuration suggestions. Verify results in real benchmarks with the generated configurations and perform follow-up tuning.


src/aiconfigurator/sdk/perf_database.py
```diff
@@ -1904,20 +1906,23 @@ def get_sol(
             """
             Get the sol time, sol math and sol mem
             """
+            full_s = s + prefix
             if w > 0 and s > w:
```
Arsene12358: Shouldn't we change the condition to check for whether full_s > w instead?

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -1904,20 +1906,23 @@ def get_sol(
...
                 # Sliding window attention
                 # Each position attends to at most w previous positions
-                ops = 2 * b * s * w * n * h * 2
+                ops = 2 * b * (full_s - prefix) * w * n * h * 2
```
Arsene12358, line 2007: also not quite sure about this part. maybe we need to have two seperate cases here, one for when w <= prefix and another for w > prefix?

README.md
```diff
@@ -128,35 +128,31 @@ tokens/s/gpu                         tokens/s/user
...
+INFO 2025-10-25 23:40:24,396 main.py:340] All experiments completed in 13.90 seconds
 ```
 
-These results indicate that deploying Qwen3-32B on h200_sxm in FP8 can achieve **1.70x** higher tokens/s/gpu for disaggregated versus aggregated deployment **under the SLA targets TTFT ≤ 300 ms and TPOT ≤ 10 ms**, with ISL:OSL of 4000:500.
+These results indicate that deploying Qwen3-32B on h200_sxm in FP8 can achieve **1.64x** higher tokens/s/gpu for disaggregated versus aggregated deployment **under the SLA targets TTFT ≤ 300 ms and TPOT ≤ 10 ms**, with ISL:OSL of 4000:500 (with prefix len: 500).
 Try different ISL:OSL values and SLA limits to fit your use case, for example:
 
 ```bash
-aiconfigurator cli default --model QWEN3_32B --total_gpus 32 --system h200_sxm --ttft 200 --tpot 10 --isl 8000 --osl 200
+aiconfigurator cli default --model QWEN3_32B --total_gpus 32 --system h200_sxm --ttft 200 --tpot 10 --isl 8000 --osl 200 --prefix 0
```
jasonqinzhou: --prefix 500?

src/aiconfigurator/sdk/backends/trtllm_backend.py
```diff
@@ -84,35 +85,41 @@ def run_agg(
...
             ) -> float:
                 num_tokens = ctx_tokens + gen_tokens
+                # treat this as a combined single batch inference, extract non-attention latency
                 summary = self.run_static(
                     model,
                     database,
-                    RuntimeConfig(batch_size=1, beam_width=1, isl=num_tokens, osl=1),
+                    # num tokens for gemm needs to be adjusted for prefix, depends on the avg prefix len per request
+                    RuntimeConfig(
+                        batch_size=1, beam_width=1, isl=num_tokens, osl=1, prefix=prefix * np.floor(ctx_tokens / isl)
```
jasonqinzhou, line 105: why is prefix=prefix * np.floor(ctx_tokens / isl) ? Should prefix=prefix?

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -1904,20 +1906,23 @@ def get_sol(
...
+                )  # 2 for fma, 2 for q*k^t+*v, /2 for causality.
             mem_bytes = (
                 2
                 * b
                 * (
-                    n * s * h  # Q read, assuming 16 bits
-                    + 2 * n_kv * s * h  # K,V read
-                    + n * s * h
+                    n * (full_s - prefix) * h  # Q read, assuming 16 bits
+                    + 2 * n_kv * full_s * h  # K,V read
```
AichenF, line 2018: is full_s - prefix equals to _s_ here?

tests/sdk/database/test_attention.py
```diff
@@ -13,17 +13,18 @@ class TestContextAttention:
...
         result = comprehensive_perf_db.query_context_attention(
-            b, s, n, n_kv, kv_cache_quant_mode, fmha_quant_mode, sol_mode=common.SOLMode.SOL
+            b, s, prefix, n, n_kv, kv_cache_quant_mode, fmha_quant_mode, sol_mode=common.SOLMode.SOL
         )
 
         # Calculate expected SOL result
-        ops = 2 * b * s * s * n * 128 * 2 / 2  # 2 for fma, 2 for q*k^t+*v, 2 for causality
-        mem_bytes = 2 * b * (n * s * 128 + 2 * n_kv * s * 128 + n * s * 128)
+        ops = 2 * b * (s * s - prefix * prefix) * n * 128 * 2 / 2  # 2 for fma, 2 for q*k^t+*v, 2 for causality
+        mem_bytes = 2 * b * (n * (s - prefix) * 128 + 2 * n_kv * s * 128 + n * (s - prefix) * 128)
```
AichenF: s here represent the previous full_s?

src/aiconfigurator/sdk/perf_database.py
tianhaox: exactly, no check for now. Do you have any fix suggestion. can we treat w as the full_s, or the s without prefix?

src/aiconfigurator/sdk/perf_database.py
tianhaox: same as the previous one, i'm also not sure about the case where there's a window size

README.md
tianhaox: yes, a typo here. fixed in the latest commit.

src/aiconfigurator/sdk/perf_database.py
tianhaox: yes, i thinks so. just to make it more readable.

src/aiconfigurator/sdk/backends/trtllm_backend.py
tianhaox: we are trying to get non-attn op's perf with this delegate inference where bs=1 and ctx_tokens=bs*isl. However, it's actually bs requests, each has a perfix. Thus the total cached prefix tokens for the actual requests are prefix_len_in_each_request * how_many_requests (np.floor(ctx_tokens/isl))

tests/sdk/database/test_attention.py
```diff
@@ -153,17 +155,20 @@ class TestContextMLA:
...
         )
 
         # Calculate expected SOL result
-        ops = b * num_heads * 2 / 2 * (s * s * 192 + s * s * 128)
-        mem_bytes = b * num_heads * 2 * (2 * s * 192 + 2 * s * 128)
+        ops = (
+            b * num_heads * 2 / 2 * (192 + 128) * (s * s - prefix * prefix)
+        )  # 2 for fma, 2 for causality. num_heads, for local heads
+        # s * 192 for q read, full_s * 192 for k read, full_s * 128 for v read, s * 192 for write.
+        mem_bytes = b * num_heads * 2 * (s * (192 + 128) + (s - prefix) * (192 + 128))  # 2 for fp16, TODO
```
jasonqinzhou, line 192: what is the TODO?

tests/cli/e2e_validation/test_e2e_sweep.py
```diff
@@ -24,13 +24,53 @@
...
 from aiconfigurator.sdk.perf_database import get_latest_database_version
 
-_LATEST_VERSIONS_CACHE = {}
-
-
-def get_all_backends():
-    """Get all backends to test."""
-    # In the future, this could come from common.BackendName
-    return ["trtllm"]
+_SUPPORTED_MODELS = [model for model in common.SupportedModels if model not in ["QWEN3_480B", "KIMI_K2"]]
```
jasonqinzhou, line 27: Why do we exclude them? Shall we use skip in pytest instead?

src/aiconfigurator/sdk/perf_database.py
Arsene12358: with current implementaion ,  "s" essentially becomes uncached_tokens. To determine whether window attention should  be applied, I think we can just use "if w >0 and full_s > w: "

src/aiconfigurator/sdk/perf_database.py
Arsene12358: for windowed attention, I think making uncached _tokens attend to at most w previous postions should be good enough, which means it's ok to keep the original formula "ops = 2 * b * s * w * n * h * 2" , since "s" is basically the uncached_tokens in this implementation

src/aiconfigurator/sdk/backends/trtllm_backend.py
```diff
@@ -84,35 +85,41 @@ def run_agg(
...
                 )
                 latency_dict = summary.get_context_latency_dict()
                 non_attention_latency = 0.0
-                # TODO, fix for DS. DS has different ops for attn in ctx and gen.
                 for layer_name, latency in latency_dict.items():
                     if layer_name != "context_attention":
                         non_attention_latency += latency
 
                 # second pass to get ctx attn, split full isl over
                 # num_steps(=np.ceil(isl/ctx_tokens)), average the ctx attn latency
```
jasonqinzhou: just curious: shall we remove this num_steps comment?

src/aiconfigurator/sdk/inference_session.py
jasonqinzhou: shall we remove with diversity here? and update the comment below?

src/aiconfigurator/sdk/backends/trtllm_backend.py
tianhaox: this is for the code interpretation

src/aiconfigurator/sdk/inference_session.py
tianhaox: Yes, let me update this

src/aiconfigurator/sdk/backends/trtllm_backend.py
tianhaox: This is for the explanation of the code

src/aiconfigurator/sdk/backends/trtllm_backend.py
```diff
@@ -42,6 +42,7 @@ def run_agg(
         """
```
jasonqinzhou, line 42: Since we add sglang and vllm support, we need to add prefix support to sglang/vllm backend specific functions too. 
We can put that in another PR.

tianhaox:
image
vllm qwen3 32b fp8, isl8000 osl1000 ttft2000, h200, prefix 0 vs 6000

tianhaox:
image
trtllm qwen3 235b fp8 block, h200, long seq: isl 8000 osl 1000 ttft 1000 limit. Prefix 0 vs 6000, prefix enables much larger throughput to mitigate the ttft limit.

PR #86, opened 2025-10-31
feat: support using --hf_id as an alternative to --model

simone-chen:
#### Overview:

Added a new cmd option `--hf_id` as a mutually exclusive alternative to `--model`. 

When a hf id is provided, AIC will:
- First try to load pre-downloaded configs saved in `aiconfigurator/model_configs`.
- If the config is not pre-downloaded, try to fetch `config.json` from huggingface, using the user's hf token stored in `~/.cache/huggingface/token`.
- Raise an error if both failed.

The model's architecture will be parsed from `config.json` if `--hf_id` is used. It will still use the hard-coded values in `SupportedModels` if using `--model`.

#### Details:
- Accepted model variation `Qwen/Qwen3-32B-FP8`. Fetched config from HF as it's not pre-downloaded.
Screenshot 2025-11-03 at 2 52 57 PM

- Accepted model `Qwen/Qwen3-32B` whose config is pre-downloaded. Even if the user doesn't have a valid HF token locally.
Screenshot 2025-11-03 at 2 56 32 PM

- Reject models whose architecture is not supported by AIC
Screenshot 2025-11-03 at 2 50 36 PM

- User must have a valid HF token saved locally to use model configs that's not cached by AIC
Screenshot 2025-11-03 at 2 54 52 PM

- closes GitHub issue: #xxx

### DIFF
--- a/tests/cli/workflow/test_argument_parsing.py
+++ b/tests/cli/workflow/test_argument_parsing.py
@@ -55,7 +55,6 @@ class TestCLIArgumentParsing:
         "param_name,expected_choices",
         [
             ("model", list(common.SupportedModels.keys())),
-            ("hf_id", list(common.CachedHFModels)),
             ("backend", [backend.value for backend in common.BackendName]),
         ],
     )


src/aiconfigurator/sdk/common.py
```diff
@@ -125,6 +125,38 @@ class BlockConfig:
     "GPT_OSS_120B": ["MOE", 36, 64, 8, 64, 2880, 2880, 201088, 131072, 4, 128, 2880, None],
     "GPT_OSS_20B": ["MOE", 24, 64, 8, 64, 2880, 2880, 201088, 131072, 4, 32, 2880, None],
 }
+SupportedHFModels = {
```
jasonqinzhou: How about keeping this as a set? We check if hf_id is supported, then find the config.json according to hf_id, for example, we store deepseek-ai.DeepSeek-V3.config.json instead of DEEPSEEK_V3_config.json. Then we found architecture, and map ARCHITECTURE_TO_MODEL_FAMILY.

src/aiconfigurator/sdk/utils.py
```diff
@@ -110,3 +118,112 @@ def _sanitize_path_component(component: str) -> str:
...
+    try:
+        config = _download_hf_config(hf_id)
+        logger.info(f"Fetched config.json for {hf_id} from HuggingFace.")
+        return _parse_hf_config_json(config)
+    except Exception as e:
+        logger.info(
+            f"Failed to download from HuggingFace using user's HF token saved in ~/.cache/huggingface/token, "
+            f"trying to use cached config: {e}"
+        )
+        if hf_id not in SupportedHFModels:
```
jasonqinzhou: CachedHFModels?

src/aiconfigurator/sdk/common.py
simone-chen: That makes sense. Fixed in the latest commit 🙏

PR #88, opened 2025-11-03
fix: update for mla collect and sol

YijiaZhao:
#### Overview:

fix sol compute for MLA

update collector for trtllm 1.2.0rc2

### DIFF
--- a/collector/trtllm/collect_moe.py
+++ b/collector/trtllm/collect_moe.py
@@ -20,8 +20,10 @@ from helper import get_sm_version, log_perf
 
 aic_debug = int(os.getenv("aic_moe_debug", "0"))  # noqa: SIM112
 
+moe_tune_path = os.path.dirname(os.path.abspath(__file__))
 
-def cleanup_empty_json_files(directory="/tmp/moe_tune_path"):
+
+def cleanup_empty_json_files(directory):
     if not os.path.exists(directory):
         return
 
@@ -466,9 +468,9 @@ def run_moe_torch(
     torch.cuda.synchronize()
 
     if moe_type != "w4a16_mxfp4":
-        cleanup_empty_json_files("/tmp/moe_tune_path")
+        cleanup_empty_json_files(moe_tune_path)
         cache_path = (
-            f"/tmp/moe_tune_path/{moe_type}_{hidden_size}_{inter_size // moe_tp_size}_{num_experts // moe_ep_size}"
+            f"{moe_tune_path}/{moe_type}_{hidden_size}_{inter_size // moe_tp_size}_{num_experts // moe_ep_size}"
         )
         existing_files = glob.glob(f"{cache_path}*")
         cache_loaded = False
--- a/src/aiconfigurator/sdk/perf_database.py
+++ b/src/aiconfigurator/sdk/perf_database.py
@@ -1490,14 +1490,13 @@ class PerfDatabase:
                     )  # s
                     target_z_list = [1, 2, 4, 8, 16, 32, 64, 128, 256, 384, 512, 1024, 2048]  # b
 
-                self._extrapolate_data_grid(
-                    data_dict=data_dict,  # tpsize,sb
-                    target_x_list=target_x_list,
-                    target_y_list=target_y_list,
-                    target_z_list=target_z_list,
-                    sqrt_y_value=True,
-                )
-
+                    self._extrapolate_data_grid(
+                        data_dict=data_dict,  # tpsize,sb
+                        target_x_list=target_x_list,
+                        target_y_list=target_y_list,
+                        target_z_list=target_z_list,
+                        sqrt_y_value=True,
+                    )
         # wideep generation mla
         if getattr(self, "_wideep_generation_mla_data", None) is not None:
             for kernel_source in self._wideep_generation_mla_data:
@@ -1581,12 +1580,12 @@ class PerfDatabase:
                     2097152 * 8,
                 ]  # s
 
-            self._extrapolate_data_grid(
-                data_dict=data_dict,  # tpsize, bs
-                target_x_list=target_x_list,
-                target_y_list=target_y_list,
-                target_z_list=target_z_list,
-            )
+                self._extrapolate_data_grid(
+                    data_dict=data_dict,  # tpsize, bs
+                    target_x_list=target_x_list,
+                    target_y_list=target_y_list,
+                    target_z_list=target_z_list,
+                )
 
         # post-correction
         self._correct_data()


collector/trtllm/collect_mla_1_1rc2.py
tianhaox, line 1: we use float16 to unify fp16 and bf16 for now. same for kvcache, float16 or fp8, no other precision. We can use specific bfloat16 to call into trtllm, but log the result as float16 or fp8 to keep aligned previous previous defintion.

src/aiconfigurator/sdk/perf_database.py
tianhaox, line 1: can we fix the fmha quant mode's memory and kvcache quant mode's compute in common.py and leverage the value here

collector/trtllm/collect_mla_1_1rc2.py
YijiaZhao: done

src/aiconfigurator/sdk/perf_database.py
YijiaZhao: already reuse FMHAQuantMode.value.compute here

collector/trtllm/collect_moe.py
```diff
@@ -148,8 +148,8 @@ def get_moe_test_cases():
         12288,
         16384,
         20480,
-        32768,
-        65536,
+        # 32768,
```
jasonqinzhou: we don't need these data points anymore?

collector/collect.py
```diff
@@ -422,7 +427,7 @@ def collect_trtllm(num_processes: int, ops: list[str] | None = None):
             else "trtllm.collect_moe_pre_1_0"
             if v.startswith(("0.21.0", "1.0.0"))
             else "trtllm.collect_moe"
-            if v.startswith("1.1.0")
+            if v.startswith(("1.1.0", "1.2.0"))
```
jasonqinzhou, line 575: Just curious that are you going to collect some data points for trtllm 1.2.x ? Will that be in this PR or in another PR?

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -2036,7 +2035,7 @@ def get_sol(
             ops = (
                 b * num_heads * 2 / 2 * (s * s * 192 + s * s * 128)
             )  # 2 for fma, 2 for causality. num_heads, for local heads
-            mem_bytes = b * num_heads * 2 * (2 * s * 192 + 2 * s * 128)  # 2 for fp16, TODO
+            mem_bytes = b * num_heads * (kvcache_quant_mode.value.memory * (s * 192 + s * 128) + 2 * (s * 192 + s * 128)) # fp16 io + fp16/fp8 kv cache, TODO fp8 io
```
jasonqinzhou: Just curious: what is 192 and 128 in this context?

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -2071,13 +2070,17 @@ def get_sol(
...
+            if kvcache_quant_mode == common.KVCacheQuantMode.fp8:
+                quant_mode_gen = common.FMHAQuantMode.fp8
+            else:
+                quant_mode_gen = common.FMHAQuantMode.float16
             # only consider fp16 mmha
             ops = 2 * b * num_heads * 1088 * s  # 2 for fma
             # kvcache load bytes will depend on kvcache quant.
             # while input q and output might be in fp16.
-            mem_bytes = b * (num_heads * 1088 * 2 + (s - 1) * 1088 * kvcache_quant_mode.value.memory)
+            mem_bytes = b * (num_heads * 1088 * 2 + (s - 1) * 512 * kvcache_quant_mode.value.memory) # fp16 io + fp16/fp8 kv cache, TODO fp8 io
```
jasonqinzhou: what is 1088 and 512 here?

YijiaZhao:
h200 sanity_check results:
Gemm
image

ctx attention:
image

gen attn:
image
image

ctx mla:
image
image
image

moe:

YijiaZhao:
moe:
WeChat09ebf2980fec929f17d528af57fd228c

src/aiconfigurator/sdk/perf_database.py
YijiaZhao: 192 for dims of Q and K, 128 for v and output

src/aiconfigurator/sdk/perf_database.py
YijiaZhao: q dim 576, output dim 512,  576 + 512 = 1088

collector/collect.py
YijiaZhao: will upload on this PR for sanity check

collector/trtllm/collect_moe.py
YijiaZhao: Moe will use chunk_forward for large tokens.  besides, 32768 and 65536 is too slow to collect.

collector/collect.py
tianhaox: 1.2 is needed to fix the mla sol problem

collector/collect.py
```diff
@@ -222,8 +222,10 @@ def create_process_exit_error(device_id, exit_code):
             # Stall detection unchanged...
             if progress_value.value == last_progress:
                 stall_count += 1
-                if stall_count > 30:
+                if stall_count > 30 and "moe" not in func.__name__:
```
tianhaox, line 225: what will happen if we don't add this specific logic? Looks like sglang moe collector is very fast without tuning.

collector/trtllm/collect_moe.py
tianhaox: if you remove this here, where you add the logic to collect data with num_tokens == 65536?

collector/trtllm/collect_moe.py
tianhaox, line 1: maybe we need to check when will the moe illegal mem be fixed and find a closest RC.

src/aiconfigurator/sdk/perf_database.py
tianhaox: can you help to fix FMHAQuantMode.value.memory in this pr as well. Is that used in context attention and mla?

src/aiconfigurator/systems/data/h200_sxm/trtllm/1.2.0rc1/context_mla_perf.txt
tianhaox, line 1: do we have fp8 compute for context mla?

src/aiconfigurator/systems/data/h200_sxm/trtllm/1.2.0rc1/context_attention_perf.txt
tianhaox, line 1: I remember that you said in 1.2.0, there're only two options: fp8 compute+fp8kvcache, or fp16 compute+fp16 kvcache, is this for context or generation?

collector/trtllm/collect_moe.py
AichenF: maybe we can use _try_ to handle potential oom issues

collector/collect.py
YijiaZhao: To reuse the tuned results, we packed all the tokens for same n,k in one test case. Therefore it needs more time but has fewer cases

collector/collect.py
YijiaZhao: I've fixed sol

collector/trtllm/collect_moe.py
YijiaZhao: I added it back.

src/aiconfigurator/systems/data/h200_sxm/trtllm/1.2.0rc1/context_attention_perf.txt
YijiaZhao: It could receive fp16 input, but cvt to fp8 by apply rope. 
It would be better to update the fp8 io for attn when we do it for gemm

src/aiconfigurator/sdk/perf_database.py
YijiaZhao: done

src/aiconfigurator/systems/data/h200_sxm/trtllm/1.2.0rc1/context_mla_perf.txt
YijiaZhao: no. I think it would be better to update it when we collect low-bit io gemm

collector/trtllm/collect_moe.py
YijiaZhao: I've checked it without w4afp8 which has TMA bug

collector/collect_all_reduce.py
tianhaox, line 1: why we need this?

collector/trtllm/collect_moe.py
```diff
@@ -416,66 +418,85 @@ def run_moe_torch(
...
-            )
-            logits_max_tokens = (
-                torch.randn([num_tokens_lists[max_index], num_experts]).to(router_logits_dtype).to(torch.device(device))
-            )
+    # dty run
+    torch.cuda.synchronize()
+    moe.forward(hidden_states_max_tokens, logits_max_tokens, do_finalize=not min_latency_mode)
+    torch.cuda.synchronize()
+
+    if moe_type != "w4a16_mxfp4":
```
tianhaox, line 470: why we need cache here, how it's reused? I'm concerned about file write and read if you allow customized path. It will be tagged as vulnerability by static risk check. Can you fix moe_tune_path to tmp, or somewhere you have write access

tianhaox, line 470: Is this code working with older trtllm version without w4a16_mxfp4? If not, we need a new moe collect file.

YijiaZhao: Not for older versions.  W4a16_mxfp4 use triton kerne. It doesn't need autotune.

YijiaZhao: Autotune process will cost lots of time. We could reuse the cache file if 
1). The new cases which has same inter_size // moe_tp and expert_num // moe_ep as anyone of the cache files.
2). If MoE collecting is not finished, next running process could reuse current cache files.

I've modified the cache file path to /tmp/moe_tune_path

collector/collect_all_reduce.py
YijiaZhao: I've deleted it

collector/trtllm/collect_moe.py
YijiaZhao: I use moe_tune_path = os.path.dirname(os.path.abspath(__file__)) instead of  /tmp/moe_tune_path

PR #89, opened 2025-11-04
feat: add vllm support (h100 only, dense models only)

ilyasher:
#### Overview:

Add gemm and attention data for vllm + H100

CLI passes for 12/20 models. Failures are either due to model being MoE (not supported in this PR) or due to non-vllm related AIC errors.

### DIFF
--- a/collector/collect.py
+++ b/collector/collect.py
@@ -281,6 +281,7 @@ def parallel_run(tasks, func, num_processes, module_name="unknown"):
 
     return errors
 
+
 def collect_ops(
     num_processes: int,
     collections: list[dict],


collector/vllm/fake_hf_model/config.json
```diff
@@ -0,0 +1,30 @@
+{
```
jasonqinzhou, line 1: Can we change the folder to vllm? vllm_v1 is confusing.

src/aiconfigurator/sdk/task.py
```diff
@@ -415,16 +417,23 @@ def _get_quant_mode(
 
         sm_version = database.system_spec["gpu"]["sm_version"]
 
+        if backend == "vllm":
+            fp8_gemm_quant = "fp8"
```
jasonqinzhou, line 448: Could you add comments to this if for vllm?

collector/collect.py
```diff
@@ -299,34 +346,40 @@ def collect_vllm(num_processes: int):
         logger.exception("VLLM is not installed. Please install it from https://github.com/vllm-project/vllm")
         return
 
-    # supported vllm v1 GEMM collection wich supports fp16,fp8,fp8_block_wise,awq and gptq
-    try:
-        import vllm_v1.collect_gemm
+    collections = [
+        # GEMM collections
+        # vllm v1 GEMM collection for fp16, fp8, fp8_block, nvfp4, awq, and gptq
```
jasonqinzhou, line 437: remove v1 in the comment.

collector/vllm/fake_hf_model/config.json
tianhaox: what's the vllm version that we're supporting in this PR

jasonqinzhou: 0.11.0, mentioned in system folder.

src/aiconfigurator/sdk/task.py
tianhaox: do we have fp_block gemm or only per-tensor fp8 gemm?

collector/collect.py
tianhaox: the community is using v1 to represent the re-architected vllm https://docs.vllm.ai/en/stable/usage/v1_guide.html
either one works. Since v0 will be deprecated in future, use vllm directly is also working.

jasonqinzhou: Ilya and I agreed that this comment doesn't need update.

jasonqinzhou:
> I've done the sanity check for the data, all make sense. One thing is that, do you have h200 or h100 sxm access? Current data is mixing h100 pcie, sxm and nvl. I am concerned about the consistency.

We do have h100 sxm and h200 sxm access on computelab. However, since the nodes are limited, sometimes we can only get pcie nodes instead.
Can we update that in another PR? We should have more nodes in the weekend.

src/aiconfigurator/sdk/task.py
ilyasher: Here I only collected per-tensor fp8 gemm, we can add fp8_block later

.github/workflows/copyright-check.ps1
```diff
@@ -72,7 +72,7 @@ $global:copyright_results = @{
 
 $ignored_files = @('.clang-format', '.gitattributes', '.gitignore', '.gitkeep', '.patch', 'Cargo.lock', 'LICENSE', 'uv.lock', 'rust-toolchain.toml', 'codespell.txt', 'exclusions.txt')
 write-debug "<copyright-check> ignored_files = ['$($ignored_files -join "','")']."
-$ignored_paths = @('.github', '.mypy_cache', '.pytest_cache', 'lib/llm/tests/data/sample-models')
+$ignored_paths = @('.github', '.mypy_cache', '.pytest_cache', 'lib/llm/tests/data/sample-models', 'collector/vllm/utils.py')
```
jasonqinzhou, line 75: just curious: why do we add utils.py here?

ilyasher: The copyright header doesn't match the expected format because the file is taken from vllm

PR #90, opened 2025-11-04
fix: Move gradio to optional dependencies

Arsene12358:
#### Overview:

1. Move gradio to optional dependencies. `pip install .` will no longer install the package
2. Disable webapp related imports in aiconfigurator/main.py, so that cli can still work normally

### DIFF
--- a/src/aiconfigurator/main.py
+++ b/src/aiconfigurator/main.py
@@ -3,6 +3,7 @@
 
 import argparse
 import importlib.util
+import logging
 import sys
 
 from aiconfigurator import __version__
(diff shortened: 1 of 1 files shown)


src/aiconfigurator/main.py
tianhaox, line 1: can you use logger module instead of print

jasonqinzhou:
just curious: why the title is temporarily disable webapp? Are we going to add it back?

src/aiconfigurator/main.py
Arsene12358: sure

jasonqinzhou (approved):

tianhaox:
> To unblock dynamo I have gone ahead and merged this PR. There are a handful of shortcomings that we will want to address relating to the Dockerfile and (possibly) documentation. I will call them out more explicitly in a future comment.

Thank you. I will expect to get it back to default once there's an update of brotli 1.2.0 on pypi

PR #91, opened 2025-11-04
feat: Add L40S data

ilyasher:
Collected TRTLLM data for L40S.

Note: also changed the default ttft and tpot values in the CLI to infinity. Many models had all configurations go over the previous ttft and tpot values on L40S. Rather than bumping the values slightly or trying to pick a reasonable value for the given model & system, we should let the user define the SLA if they require it.

Model support:
Pass (15) :
QWEN2.5_1.5B
QWEN2.5_32B
QWEN3_0.6B
QWEN2.5_7B
QWEN3_1.7B
QWEN3_8B
LLAMA3.1_8B
MOE_Mixtral8x22B
MOE_Mixtral8x7B
QWEN3_235B
LLAMA3.1_70B
QWEN2.5_72B
LLAMA2_70B
QWEN3_32B
Nemotron_super_v1.1

Failed due to limited MoE support (3):
DEEPSEEK_V3
QWEN3_480B
LLAMA3.1_405B

Failed due to "No successful experiment runs to compare" (2)
LLAMA2_7B
LLAMA2_13B

### DIFF
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -81,8 +81,8 @@ def _add_default_mode_arguments(parser):
     )
     parser.add_argument("--isl", type=int, default=4000, help="Input sequence length.")
     parser.add_argument("--osl", type=int, default=1000, help="Output sequence length.")
-    parser.add_argument("--ttft", type=float, default=float("inf"), help="Time to first token in ms.")
-    parser.add_argument("--tpot", type=float, default=float("inf"), help="Time per output token in ms.")
+    parser.add_argument("--ttft", type=float, default=2000.0, help="Time to first token in ms.")
+    parser.add_argument("--tpot", type=float, default=30.0, help="Time per output token in ms.")
 
 
 def _add_experiments_mode_arguments(parser):
@@ -311,7 +311,9 @@ def _execute_task_configs(
                 results[exp_name] = task_result
                 logger.info("Experiment %s completed with %d results.", exp_name, len(pareto_frontier_df))
             else:
-                logger.warning("Experiment %s returned no results.", exp_name)
+                logger.warning(
+                    "Experiment %s returned no results. The TTFT and TPOT constraints may need to be relaxed.", exp_name
+                )
         except Exception:
             logger.exception("Error running experiment %s", exp_name)
 
--- a/src/aiconfigurator/sdk/task.py
+++ b/src/aiconfigurator/sdk/task.py
@@ -733,20 +733,6 @@ class TaskConfig:
         if check_is_moe(self.model_name) and self.backend_name == "vllm":
             raise NotImplementedError("AIConfigurator does not yet support MOE models for VLLM backend.")
 
-        if (
-            self.backend_name == "trtllm"
-            and self.system_name == "l40s"
-            and self.model_name
-            in [
-                "DEEPSEEK_V3",
-                "LLAMA3.1_405B",
-                "QWEN3_480B",
-            ]
-        ):
-            raise NotImplementedError(
-                f"AIConfigurator does not yet support {self.model_name} + {self.system_name} + {self.backend_name}"
-            )
-
     def pretty(self) -> str:
         def _convert(obj: Any) -> Any:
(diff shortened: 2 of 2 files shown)


src/aiconfigurator/sdk/task.py
```diff
@@ -687,6 +687,15 @@ def __init__(
         else:
             raise ValueError(f"Invalid serving mode: {serving_mode}")
 
+        self.validate()
+
+    def validate(self):
+        """
+        Check that the task can be run by AIC.
+        """
+        if check_is_moe(self.model_name) and self.system_name == "l40s":
```
jasonqinzhou: I saw moe_perf.txt is there. Shall we remove this check for l40s?

ilyasher: Fixed, thanks

ilyasher:
> Just found the reason why qwen3 32b doesn't have output for l40s: the default SLA is too restricted for slow l40s parser.add_argument("--ttft", type=float, default=1000.0, help="Time to first token in ms.") parser.add_argument("--tpot", type=float, default=20.0, help="Time per output token in ms.") [2:16](https://nvidia.slack.com/archives/D086Y6J31L3/p1762294613752279) you can try to * 20 to 20000 and 400 in the PR.

Thanks, updated the default max ttft and tpot to infinity

src/aiconfigurator/cli/main.py
tianhaox, line 1: can we set to a reasonable value like ttft 2000 and tpot 50. Otherwise, there won't be perf gain with these defaults for disagg.
In order to pass the tests, perhaps we can set a specific ttft and tpot to the cli command for that platform+model

jasonqinzhou: tpot has to be 150 for Qwen3 32B to succeed for default settings.

Then I think we don't change here, but update the test_model_system_combinations.py instead.

ilyasher:
@AliceChenyy I ran the Jupyter notebook and all results were <=100% of SOL except for query_generation_mla, which was up to 1.2x faster than the calculated SOL. I think there is a bug with the query_generation_mla SOL mode, because the same thing happens with the existing H100 data too.

tianhaox:
> @AliceChenyy I ran the Jupyter notebook and all results were <=100% of SOL except for query_generation_mla, which was up to 1.2x faster than the calculated SOL. I think there is a bug with the query_generation_mla SOL mode, because the same thing happens with the existing H100 data too.

I think so, https://github.com/ai-dynamo/aiconfigurator/pull/88 we have this pr open. Hope it can fix the problem. But it's okay, I think we need to do the sanity check automatically in the future.

src/aiconfigurator/sdk/task.py
```diff
@@ -687,6 +687,26 @@ def __init__(
...
+        """
+        Check that the task can be run by AIC.
+        """
+        if (
+            self.backend_name == "trtllm"
+            and self.system_name == "l40s"
+            and self.model_name
+            in [
+                "DEEPSEEK_V3",
+                "LLAMA3.1_405B",
```
jasonqinzhou: LLAMA3.1_405B won't fail for L40S if we give enough GPUs.

src/aiconfigurator/sdk/task.py
```diff
@@ -687,6 +687,26 @@ def __init__(
         else:
             raise ValueError(f"Invalid serving mode: {serving_mode}")
 
+        self.validate()
+
+    def validate(self):
```
jasonqinzhou, line 729: do not recommend adding this validate function as some models might pass under different constraints.

src/aiconfigurator/cli/main.py
tianhaox: Yes, agree. Need to keep CLI with reasonable defaults. The TPOT could also be caused by abnormal comm data.

ilyasher: Changed to ttft 2000 tpot 30

src/aiconfigurator/sdk/task.py
ilyasher: removed

PR #97, opened 2025-11-06
fix: LLAMA2_7B and LLAMA2_13B errors

ilyasher:
#### Overview:
This PR fixes a bug where AIC would break when running LLAMA2_7B and LLAMA2_13B.

## Bug overview:
This happens for any model where `num_kv_heads == num_heads` (standard MHA rather than GQA). The dicts `_generation_attention_data` and `_context_attention_data` stored data points with `num_kv_heads=0` to actually mean `num_kv_heads=num_heads`. The bug was that when the database was queried, `num_kv_heads` was not converted to 0 in the `num_kv_heads=num_heads`, leading to seemingly missing data.

## Results
Now LLAMA2_7B and LLAMA2_13B work for any system:

```bash
aiconfigurator cli default --model LLAMA2_7B --system h100_sxm --backend trtllm --total_gpus 16
```

### DIFF
--- a/tests/sdk/database/test_attention.py
+++ b/tests/sdk/database/test_attention.py
@@ -139,6 +139,21 @@ class TestGenerationAttention:
         assert isinstance(result, float)
         assert result > 0
 
+    def test_query_generation_attention_non_sol_mode_mha(self, comprehensive_perf_db):
+        """Test non-SOL mode with MHA (n_kv == n)."""
+        b, s, n = 2, 64, 16
+        n_kv = n  # MHA case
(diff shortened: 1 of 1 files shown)


src/aiconfigurator/sdk/perf_database.py
```diff
@@ -643,7 +643,8 @@ def load_context_attention_data(context_attention_file):
         window_size = int(window_size)
```
jasonqinzhou, line 643: Just curious: Does this fix work for vllm/sglang as well?

ilyasher: Yes, I tested it

PR #100, opened 2025-11-09
fix: fix webapp with sglang and vllm

tianhaox:
new backends have different support matrix of models, paralle, quant, etc. Need to get it working generally and allow possible failures.
add tests for cli with new backends.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,22 +262,20 @@ To estimate performance, we take the following steps:
   - MOE
   - QWEN
   - DEEPSEEK_V3
+  - Support using huggingface model id if falls into these model family and not MoE models.
 - Operations:
   - Attention
     - MHA/GQA (FP8, FP16)
     - MLA (FP8, FP16)
   - KV Cache (FP16, FP8, INT8)
   - GEMM (FP16, FP8, FP8-Block, FP8-OOTB, SQ, INT8 WO, INT4 WO, NVFP4)
-  - AllReduce (FP16)
+  - CustomAllReduce (FP16)
   - Embedding
   - P2P
   - ElementWise
   - NCCL (all_reduce, all_gather, all-to-all, reduce_scatter)
   - MoE (FP16, FP8, FP8-Block, W4A-FP8, INT4 WO, NVFP4)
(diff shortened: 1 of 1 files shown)


davilu-nvidia (approved):
LGTM

src/aiconfigurator/webapp/events/event_fn.py
```diff
@@ -1080,28 +1094,67 @@ def update_quant_mode_choices(model_name, system_name, backend_name, version):
...
-        gemm_quant_mode_choices = sorted(database_dict[system_name][backend_name][version].supported_quant_mode["gemm"])
+        supported_quant_mode = database_dict[system_name][backend_name][version].supported_quant_mode
+
+        if get_model_family(model_name) != "DEEPSEEK":
+            gemm_quant_mode_choices = sorted(supported_quant_mode["gemm"])
+            kvcache_quant_mode_choices = sorted(supported_quant_mode["generation_attention"])
+            fmha_quant_mode_choices = sorted(supported_quant_mode["context_attention"])
+            moe_quant_mode_choices = sorted(supported_quant_mode["moe"])
+        else:
+            if backend_name == "sglang":
```
jasonqinzhou, line 1105: use backend_name == "sglang" and enable_wideep to reduce a repetition ?

tianhaox: I'm leaving this for future backend-specific logic

README.md
```diff
@@ -262,22 +262,20 @@ To estimate performance, we take the following steps:
   - MOE
   - QWEN
   - DEEPSEEK_V3
+  - Support using huggingface model id if falls into these model family and not MoE models.
```
jasonqinzhou, line 265: shall we remove `and not MoE models` ?

tianhaox: Yes, perhaps needs to refine this doc together with our next pr about support matrix? Here 'not MoE models' means that even though you can pass a hd_id of the moe model, you cannot support it if we don't have the moe op data for that as we are actually collect moe data model by model.

tianhaox: Ambiguous here.

PR #101, opened 2025-11-09
fix: fix collector minor problems

tianhaox:
fix: fix collector minor problems
#### Overview:

1. fix sglang collector attn module, add causal and fix fp8 gen attn.
2. renaming to unify the wideep path of sglang
3. collect qheads=1,2 for attn data of all 3 backends
4. update readme a little bit to reflect the naming.


AichenF (approved):
LGTM

PR #104, opened 2025-11-13
feat: [Collector] Kernel power measurement support [1/N]

jaywonchung:
#### Overview:

The broader context is enabling power-aware planning in power-constrained scenarios. For that, we want to start with kernel power benchmarking.

This PR is the first step towards kernel power measurement support. It begins with `trtllm` backend and the basic GEMM and attention ops, and NCCL kernels.

Apologies for the delay. OSDI deadline and other work left little bandwidth for me :(

#### Details:

I'm hoping the changes in `collector/README.md` should provide a bit more detail. I'm using Zeus because it's just convenient (like obeying `CUDA_VISIBLE_DEVICES`), it's not very heavy of a dependency, and it also works out of the box for AMD GPUs, which is nice to have as an open-source project.

This is a cleaned up version of what did run before, but I no longer have the GPU node to test this refactored version properly. But when I'll get any GPUs is unclear, so I decided to post this anyway as a draft PR to receive feedback on general structure.

`README.md` -> Structure changes in `collector.py` -> Each op.

cc. @jasonqinzhou


collector/helper.py
```diff
@@ -264,3 +268,118 @@ def log_perf(
...
+
+def get_gpu_specs_from_device(device_name: str) -> dict:
+    """Load GPU specifications from system YAML files.
+
+    Dictionary keys are float16_tflops, fp8_tflops, int8_tflops, mem_bw_gbs, power_max.
+    Keys follow system YAML files, except for power_max (which is just 'power' in YAML).
+    """
+    # Map device name to system file
+    device_upper = device_name.upper()
+    if "H100" in device_upper:
```
jasonqinzhou: Can we reduce the number of if else with a prefix match?

collector/README.md
```diff
@@ -106,6 +105,45 @@ See `deep_collector/README.md` for complete multi-node setup instructions.
...
+## Kernel Power Profiling
+
+The collector can measure kernel power consumption during benchmarking, which is used for power-aware configuration generation.
+
+```bash
+python collector/collect.py \
+  --backend trtllm \
+  --ops gemm attention_context attention_generation \
+  --measure-power \
+  --power-limits 700 500 300 \
```
jasonqinzhou: What are the typical power limits? Shall we provide them as default so we the collected data points can cover majority of cases?

collector/trtllm/collect_attn.py
tianhaox, line 1: how would you like to use compute_bound in aiconfigurator.sdk? We have a SOL_FULL mode of the database for query query_op func. E.g., you can get sol_time, sol_math, sol_mem = db.query_gemm(SOL_MODE=SOL_FULL)
you can then check weather it's compute bound or mem bound by checking sol_time equals sol_math or sol_mem

collector/README.md
kaim-eng: GPUs are shipped in float-clock-cap-power mode, we have a default power limit (e.g., 700W for H100), which is on the product spec sheet.

collector/README.md
```diff
@@ -105,6 +105,45 @@ See `deep_collector/README.md` for complete multi-node setup instructions.
...
+
+- Calculate arithmetic intensity for each kernel configuration
+- **Memory-bound kernels**: Measures actual power over a default 3-second window (duration configured with `--kernel-power-measurement-duration`)
+- **Compute-bound kernels**: Power assumed to be equal power limit (power not measured)
+
+NVML power/energy counters only update once 100 ms, which is significantly longer than most kernels. Thus, kernels are run repeatedly for `--kernel-power-measurement-duration` seconds for power measurement.
+
+### Requirements
+
+- Zeus (`pip install zeus`) for measurement & GPU power limit control
```
kaim-eng: @jaywonchung , I took a brief look at Zeus codebase. Zeus seems to be an abstraction layer on top of NVML. Conceptually, I kind of agree with this extra abstraction layer idea because it's easier to add other venders' device support. My key question here: do we commit to maintain this lib for a long-term horizon? On the other end, NVML is part of Nvidia's official library suite that enable customers to interact with GPU. Maybe we want to directly use NVML here?
@tianhaox for Viz and suggestion on introducing new dependency here.

collector/README.md
jaywonchung: Also, different GPU models have quite different default power limits. 400W for A100, 700W for H100, etc. If not specified, we could query the GPU's default (max) power limit and just use that as the default.

I think the challenging part is whether or not to make power measurement the default. It does inflate benchmarking time since each memory-bound kernel is measured for three seconds (configurable). Though if someone decided to benchmark anyway, spending some extra time for a more complete measurement could make sense.

collector/trtllm/collect_attn.py
kaim-eng: If we already have a way of deciding boundness, I do agree that we want to re-use the code.

jaywonchung: Yeah I will use the database query! I wasn't aware that this existed.

collector/README.md
jaywonchung: Yeah, for GPUs, it's an abstraction layer over NVML/AMDSMI + `CUDA_VISIBLE_DEVICES`/`HIP_VISIBLE_DEVICES` index remapping (because NVML is not on the CUDA application layer, it ignores it). Zeus's measurement API hasn't changed for very long since we know that's the API people use most often.

But I totally understand the concerns around having a new dependency. I'd be happy to do what the maintainers decide towards. I'll pause testing in case we decide to switch to pure NVML, since that requires some extra code.

kaim-eng: Just come up with a patchset to migrate to NVML. I run some simple test on my end it seems working fine on my end. 
git apply kaim-zeus2nvml.patch to reply the changes on top of this branch will do

[kaim-zeus2nvml.patch](https://github.com/user-attachments/files/23592633/kaim-zeus2nvml.patch)
[zeus2nvml_MIGRATION_NOTES.md](https://github.com/user-attachments/files/23592634/zeus2nvml_MIGRATION_NOTES.md)

src/aiconfigurator/cli/main.py
```diff
@@ -29,6 +29,11 @@ def _build_common_cli_parser() -> argparse.ArgumentParser:
     common_parser = argparse.ArgumentParser(add_help=False)
     common_parser.add_argument("--save_dir", type=str, default=None, help="Directory to save the results.")
     common_parser.add_argument("--debug", action="store_true", help="Enable debug mode.")
+    common_parser.add_argument(
```
jasonqinzhou, line 32: Don't have to add this option. Just check if _power.txt exists and add the power column to the output table.

jasonqinzhou: We can add power prioritized Pareto curve as a new option in another PR. this one is big enough.

collector/collect.py
```diff
@@ -44,9 +44,21 @@ def setup_warning_filters():
 
 logger = None
 
-
-def collect_module_safe(module_name, test_type, get_test_cases_func, run_func, num_processes):
-    """Safely collect module with comprehensive error handling"""
+# Power measurement defaults
+DEFAULT_KERNEL_POWER_MEASUREMENT_DURATION = 3.0  # seconds
```
jasonqinzhou, line 48: Can we try 1.0s? The faster the better. We need to control an operation to be finished within 10h which is the limit from computelab.

PR #117, opened 2025-11-20
feat: introduce profiler UI

Harrilee:


Arsene12358 (approved):

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -2212,6 +2217,11 @@ def get_sol(
 
         if sol_mode is None:
             sol_mode = self._default_sol_mode
+
+        # If MLA data is not available (e.g., for vllm backend), force SOL mode
+        if self._generation_mla_data is None:
+            sol_mode = common.SOLMode.SOL
+
```
jasonqinzhou: # If MLA BMM data is not available (e.g., for vllm backend), force SOL mode
        if self._mla_bmm_data is None:
            sol_mode = common.SOLMode.SOL

I think we need this too for query_mla_bmm(), if we want to add this condition to all queries that might miss silicon data.

Harrilee:
> FYI: I have updated this PR with hiding profiling tab by default.

Have you considered using the existing `--experimental` flag? @jasonqinzhou

jasonqinzhou:
> > FYI: I have updated this PR with hiding profiling tab by default.
> 
> Have you considered using the existing `--experimental` flag? @jasonqinzhou

That's for enabling SOL Mode box.

src/aiconfigurator/sdk/perf_database.py
tianhaox: i think we need to remove this logic and force failure. We need to handle this carefully instead of setting to sol because this will introduce gap silently. We have a pr named hybrid modeling to handle this more systematically.

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -2721,6 +2731,11 @@ def get_sol(
 
         if sol_mode is None:
             sol_mode = self._default_sol_mode
+
```
tianhaox, line 2726: same here. What will happen if we don't have this for now?

src/aiconfigurator/webapp/components/base.py
```diff
@@ -176,19 +205,35 @@ def create_model_misc_config(app_config):
...
 
     with gr.Accordion("Runtime config"):
         with gr.Row():
-            isl = gr.Number(value=2048, label="input sequence length", interactive=True)
-            osl = gr.Number(value=128, label="output sequence length", interactive=True)
-            prefix = gr.Number(value=0, label="prefix cache length", interactive=True)
+            gr.HTML(
+                "<span style='color: var(--body-text-color-subdued);'>More inputs = more precise profiling results.</span>"
+            )
+        with gr.Row():
```
tianhaox: i will suggest prefix len can always be true, where we don't need it?

src/aiconfigurator/webapp/components/profiling/js/cdn_loader.js
tianhaox, line 1: what this file for?

src/aiconfigurator/webapp/components/profiling/sdk/config.py
tianhaox, line 1: I will leave a reminder here but not a problem to resolve for now. We'll soon have a pr about the api change for generator module. We need to refactor this part after that pr being merged.

src/aiconfigurator/webapp/components/profiling/sdk/config.py
```diff
@@ -0,0 +1,195 @@
...
+    if conc_per_dp < granularity:
+        ans = list(range(attn_dp_size, conc_per_dp * attn_dp_size + 1, attn_dp_size))
+    else:
+        step = (conc_per_dp - 1) * attn_dp_size / (granularity - 1)
+        ans = [attn_dp_size + int(i * step) * attn_dp_size for i in range(granularity)]
+
+    return ans
+
+
+def enumerate_moe_configs(tp_size, attention_dp_size=1):
```
tianhaox, line 172: in future, we need to refactor this part about parallel enumeration to use the unified one defined in sdk.  If it's wokring like this, I'm okay with that for now. Workable MoE config is framework specific.

src/aiconfigurator/webapp/main.py
```diff
@@ -55,6 +58,7 @@ def main(args):
         )
 
     with gr.Blocks(
+        title="Dynamo aiconfigurator for Disaggregated Serving Deployment",
```
tianhaox, line 61: will this be duplicate with line 79?

src/aiconfigurator/sdk/perf_database.py
Harrilee: Removed

src/aiconfigurator/sdk/perf_database.py
Harrilee: Answer to the question: there will be a less meaningful err msg, users are hard to find out that VLLM does not support MoE. Something like this:

```
Error Type: TypeError
Error Message: 'NoneType' object is not subscriptable
Location: File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/sdk/perf_database.py", line 2176, in query_context_mla

Full Stack Trace:
  File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/webapp/components/profiling/core/orchestrator.py", line 123, in generate_profiling_plots
    prefill_results = profile_prefill_performance(database, backend_instance, model_name, profile_num_gpus, isl)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/webapp/components/profiling/core/profiling.py", line 57, in profile_prefill_performance
    perf_dict = estimate_prefill_perf(
                ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/webapp/components/profiling/sdk/estimation.py", line 93, in estimate_prefill_perf
    return estimate_perf(
           ^^^^^^^^^^^^^^
  File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/webapp/components/profiling/sdk/estimation.py", line 66, in estimate_perf
    summary = session.run_static(mode=mode_to_sdk_mode[mode], runtime_config=runtime_config, stride=32)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/sdk/inference_session.py", line 65, in run_static
    return self._backend.run_static(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/sdk/backends/base_backend.py", line 115, in run_static
    context_latency_dict = _run_context(batch_size, isl, prefix)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/sdk/backends/base_backend.py", line 73, in _run_context
    latency = op.query(database, x=x, batch_size=batch_size, beam_width=1, s=isl, prefix=prefix)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/sdk/operations.py", line 591, in query
    database.query_context_mla(
  File "/home/harrli/Projects/aiconfigurator/src/aiconfigurator/sdk/perf_database.py", line 2176, in query_context_mla
    mla_dict = self._context_mla_data[fmha_quant_mode][kvcache_quant_mode]
               ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

Harrilee: However, since we are not releasing the profiler UI to users now, I have moved this SOL mode trick as discussed in the thread above.

src/aiconfigurator/webapp/components/base.py
Harrilee: In the profiler tab, we don't need "prefix len". @tedzhouhk could you confirm that we do not need prefix length in the profiler UI?

PixPin_2025-12-02_14-26-28

src/aiconfigurator/webapp/components/profiling/sdk/config.py
Harrilee: A loader module for JavaScript libraries.
Since we do not have a JS package manager, and I do not want to copy-paste the entire external JS library to our code base, I used their public CDN endpoint to load the JS scripts. The libraries we needed include:

- jQuery: for manipulating DOM, which is also the prereq of other libraries below
- Chart.js: for rendering plot. Note that we cannot use plotly.js provided by Gradio backend as we need JS to control the interaction between the plot and the datatables
- datatables: to display the data in tables. Note that the table provided by Gradio backend does not support adding `<button>` as content, and does not support dynamic interaction with the Chart, nor does it support custom CSS (for highlight purposes)
- highlight.js: to hilight `<pre></pre>` yaml content.

src/aiconfigurator/webapp/main.py
Harrilee: Line 61 is the `<title>` tag, otherwise the browser tab will only show "Gradio".
PixPin_2025-12-02_14-38-33
PixPin_2025-12-02_14-39-55

src/aiconfigurator/webapp/components/base.py
tedzhouhk: yeah, current profiler doesn't support prefix cache

src/aiconfigurator/webapp/components/profiling/js/cdn_loader.js
Harrilee: A loader module for JavaScript libraries.
Since we do not have a JS package manager, and I do not want to copy-paste the entire external JS library to our code base, I used their public CDN endpoint to load the JS scripts. The libraries we needed include:

- jQuery: for manipulating DOM, which is also the prereq of other libraries below
- Chart.js: for rendering plot. Note that we cannot use plotly.js provided by Gradio backend as we need JS to control the interaction between the plot and the datatables
- datatables: to display the data in tables. Note that the table provided by Gradio backend does not support adding `<button>` as content, and does not support dynamic interaction with the Chart, nor does it support custom CSS (for highlight purposes)
- highlight.js: to hilight `<pre></pre>` yaml content.

PR #118, opened 2025-11-21
feat: Add MoE support for VLLM

ilyasher:
Add MoE collector for VLLM and collect data for H100

# Perf results
I only collected h100 data for now. It is poor compared to TRTLLM. While TRTLLM is about ~30% of calculated SOL, the collected VLLM data is <10% of SOL. I suspect this is because the TRTLLM data collector ran autotune for MoE kernels, but vllm did not. vllm does have a autotuner for MoE, but it takes hours per test case, so it is not feasible to run it for our data collection.

VLLM MoE sanity check
moe_updated

TRTLLM for reference
Screenshot 2025-11-20 at 7 54 50 PM

### DIFF
--- a/collector/common_test_cases.py
+++ b/collector/common_test_cases.py
@@ -1,0 +1,132 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import dataclasses
+import itertools
+from typing import Optional
+
+
+@dataclasses.dataclass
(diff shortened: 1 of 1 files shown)


collector/collect.py
```diff
@@ -329,6 +335,7 @@ def collect_ops(
                     "traceback": traceback.format_exc(),
                 }
             )
+    return all_errors
```
tianhaox: duplicate return

ilyasher:
I found my mistake when ep>1, so I updated the collector and re-collected the h100 moe perf. I also updated enumerate_parallel_config() as suggested by @tianhaox

PR #122, opened 2025-11-22
fix: sglang gemm quantization method alignment

tianhaox:
#### Overview:

need to align the quantize method in real inference.
For fp8 blockwise, use per-token quantize
For fp8, assume dynamic by default, use per-token quantize. 
Note, 1, we didn't enable dynamic per-tensor. 2. we didn't enable static per-tensor. If needed, we shall add more.
For int4_wo, the function to be called in sglang is expected to be awq. We remove int4_wo support for now. Can implement in future.
Tested in 0.5.5.post2 and post3. For SGLang, in future, we need to manage these versions as well to make api compatible.

### DIFF
--- a/collector/sglang/collect_gemm.py
+++ b/collector/sglang/collect_gemm.py
@@ -4,6 +4,7 @@
 import pkg_resources
 import torch
 import torch.nn.functional as F
+from common_test_cases import get_gemm_common_test_cases
 from sgl_kernel import fp8_scaled_mm, int8_scaled_mm, sgl_per_tensor_quant_fp8, sgl_per_token_quant_fp8
 from sglang.srt.layers.deep_gemm_wrapper import (
     DEEPGEMM_SCALE_UE8M0,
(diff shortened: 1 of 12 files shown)


collector/sglang/collect_gemm.py
```diff
@@ -159,33 +177,41 @@ def sglang_scaled_fp8_quant(
...
             b_fp8, scale_b_fp8 = sglang_scaled_fp8_quant(b_fp32, scale_b)
             b_fp8 = b_fp8.t()
 
             repeat_n = 5
 
             def gemm_op():
+                # Dynamic quantization for A (per-token by default)
+                scale_a = None
+                # Uncomment to use static per-tensor quantization
+                # scale_a = torch.ones(1, device=device, dtype=torch.float32)
```
xutizhou, line 154: uncomment or parse the parameter `per_tensor` or `per_token`, which one is better

collector/sglang/collect_gemm.py
```diff
@@ -159,33 +177,41 @@ def sglang_scaled_fp8_quant(
...
 
-            scale_a = torch.randn((M,), device="cuda", dtype=torch.float32)
-            scale_b = torch.randn((N,), device="cuda", dtype=torch.float32)
-            a_fp8, scale_a_fp8 = sglang_scaled_fp8_quant(a_fp32, scale_a)
+            scale_b = torch.randn((N,), device=device, dtype=torch.float32)
+            # Quantize B (weights) outside
             b_fp8, scale_b_fp8 = sglang_scaled_fp8_quant(b_fp32, scale_b)
             b_fp8 = b_fp8.t()
 
             repeat_n = 5
```
xutizhou: set repeat_n out of the if else branch only once for simplicity.

collector/sglang/collect_gemm.py
tianhaox: Depends on whether we need per-tensor static quantize data. Looks like for qwen it's dynamic per-token. We have another ongoing effort to study the quantization impact. Perhaps we can improve this in that PR once we have more thorough study of this.

collector/sglang/collect_gemm.py
tianhaox: done

PR #125, opened 2025-11-24
feat: hybrid modeling

tianhaox:
#### Overview:

- rename sol mode of database to silicon mode.
- add empirical mode, where the latency = sol / empirical factor. Current emprical factor is a constant number, could be more complicated in future. Add each op should design its own empirical factors.
- add hybrid mode, when missing silicon data, in this mode, allow to fallback to empirical mode.

So in future, when you add a new op, you can support running configuration without actual data if you use sol/empirical/hybrid mode. The hybrid mode should give you the best guess of the overal model performance.

fbd50242b800a82f013bce57711a6323
This is the agg mode of running qwen3 32b on h200. Pure emprical mode has much smaller diff than sol mode, comparing with silicon mode. Imagine if part of the model is missing, the hybrid mode can give you a very close result unless the missing part is taking the major portion.

### DIFF
--- a/tests/cli/e2e_validation/test_e2e_sweep.py
+++ b/tests/cli/e2e_validation/test_e2e_sweep.py
@@ -480,6 +480,7 @@ class TestE2ESweep:
                 args.tpot = float(tpot)
                 args.backend = backend
                 args.backend_version = version
+                args.database_mode = None  # Use default (SILICON)
                 args.save_dir = temp_dir
                 args.debug = False
                 args.generated_config_version = None


src/aiconfigurator/sdk/perf_database.py
```diff
@@ -1968,14 +1968,32 @@ def get_sol(m: int, n: int, k: int, quant_mode: common.GEMMQuantMode) -> tuple[f
             sol_time = max(sol_math, sol_mem)
             return sol_time, sol_math, sol_mem
 
-        if sol_mode is None:
-            sol_mode = self._default_sol_mode
-        if sol_mode == common.SOLMode.SOL:
+        def get_empirical(m: int, n: int, k: int, quant_mode: common.GEMMQuantMode) -> tuple[float, float, float]:
+            """
+            Get the hybrid time, hybrid math and hybrid mem
```
Arsene12358: the docstring is a bit misleading, by the look of it, this function seems to only return the empirical time

tianhaox: done

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -1968,14 +1968,32 @@ def get_sol(m: int, n: int, k: int, quant_mode: common.GEMMQuantMode) -> tuple[f
...
         else:
-            result = self._interp_3d(m, n, k, self._gemm_data[quant_mode], "cubic")
+            try:
+                result = self._interp_3d(m, n, k, self._gemm_data[quant_mode], "cubic")
+            except Exception:
+                if database_mode == common.DatabaseMode.HYBRID:
+                    logger.debug(f"Failed to query gemm data for {m=}, {n=}, {k=}, {quant_mode=}, using empirical mode")
+                    result = get_empirical(m, n, k, quant_mode)
+                else:
+                    logger.exception(f"Failed to query gemm data for {m=}, {n=}, {k=}, {quant_mode=}")
```
Arsene12358: maybe we also add a message to advise users to switch to empirical mode, when they encounter error in the silicon mode due to missing data?

src/aiconfigurator/sdk/perf_database.py
davilu91, line 1: I saw many hardcoded empirical scale_factors, how confident are we with these numbers (only saw one example with H200, it's better than sol for sure)? Do these apply to all models and SKUs?

Arsene12358: yeah, I don't think we distinguish between models or SKUs. it's op-level , so I don't  believe we need to distinguish between models anyway. As for SKUs, it's possible that different sku might warrant a different scaling factor, but more investigations are definitely needed.

tianhaox: i think it's a starting point. need to make more complicated modeling for the empical factor. e.g., for gemm, it should be of high mem% when it's mem-bound, and high math% when it's compute-bound.

davilu-nvidia (approved):
I'll approve first to unblock merging.

src/aiconfigurator/sdk/perf_database.py
tianhaox: done

PR #126, opened 2025-11-24
feat: add comprehensive support matrix testing framework

Harrilee:
Implement a complete support matrix testing system that validates all model/system/backend/version combinations for both aggregated and disaggregated serving modes.


src/aiconfigurator/systems/support_matrix.csv
tianhaox, line 1: shall we add a commit to this support_matrix file? or something else to map it to a specific code status, or we can do only for release tag of course.

tools/support_matrix/generate_support_matrix.py
tianhaox, line 1: can you create a subdirectory for this under tools/

Harrilee: Done

src/aiconfigurator/systems/support_matrix.csv
Harrilee: Could you elaborate on the use case for this mapping? In my opinion, I think tracking git history would be adequate.

tianhaox: if we modify some logic which makes the support matrix smaller, will the support matrix changes automatically?

tests/sdk/support_matrix/suppport_matrix.py
tianhaox, line 1: This source file looks like a sweep test. Why we need it here in sdk? Who will use this class, is that better to locate in test or tools/support_matrix?

Which func is used to query the csv file's support matrix?

Harrilee: In the design doc, we mentioned another function regarding checking the csv via cli (or even webapp), for that functionality, this path would work. Do you have any ideas?

src/aiconfigurator/systems/support_matrix.csv
Harrilee: No. In that case `test_system_and_backend_matches_database` will not pass

Harrilee: We should manually update the csv then.

tianhaox: Yes, that's what I mean. The csv file's support matrix only works for a specific commit. We won't run the costly support matrix regression upon every pr. Or we can just say, it's mapped to a specific release.

tests/sdk/support_matrix/suppport_matrix.py
tianhaox: Sorry I still didn't understand
1. Have you added that into cli already? I didn't see relevant code to use aic cli suportmatrix xxx.
2. back to this impl, it's all for enumerating and testing, is this better in tests? What do you mean by checking the csv via cli, it's just reading the csv and check, right?

src/aiconfigurator/systems/support_matrix.csv
Harrilee: In the design doc, we planned to run it daily at 11 PM China time. On default branch only.
https://github.com/ai-dynamo/aiconfigurator/pull/126/files#diff-1b0d895a03b46976028731cb51b3ff81f8b3ccbfc570fa0f311e88e5dd49caf1R5-R6

tests/sdk/support_matrix/suppport_matrix.py
Harrilee: 1. Not added to CLI yet. But we are planning to do so.
2. "checking the csv via cli" means something like `aiconfigurator cli support-matrix` related commands.

If we want to support support matrix via cli, this `src/aiconfigurator/sdk/suppport_matrix.py` shall be a production code, otherwise it would be a pure test code.

.github/CODEOWNERS
```diff
@@ -85,4 +87,4 @@
 /.pre-commit-config.yaml @saturley-hall
 
 # CI/CD and workflows
-/.github/workflows @saturley-hall @tianhaox @jasonqinzhou @Arsene12358
\ No newline at end of file
+/.github/workflows @saturley-hall @tianhaox @jasonqinzhou @Arsene12358  @Harrilee
```
jasonqinzhou: minor: redundant space?

Harrilee: Removed. Thx!

tests/sdk/support_matrix/suppport_matrix.py
Harrilee: Will put in `tests`. And add another PR for csv query.

src/aiconfigurator/systems/support_matrix.csv
Harrilee: As discussed in the meeting, we will be running daily tests.

tests/sdk/support_matrix/suppport_matrix.py
Harrilee: Done.

PR #128, opened 2025-11-25
perf: cache repeated queries to perf database and match_workers

anish-shanbhag:
#### Overview:

This change updates the perf database and `match_workers` to cache repeated queries, which should be fine since these operations don't appear to have any side effects.

#### Details:

In my local testing, this speeds up `aiconfigurator cli default --model QWEN3_32B --total_gpus 32 --system h200_sxm` from `206s` --> `9s`

- closes GitHub issue: #xxx

### DIFF
--- a/src/aiconfigurator/sdk/perf_database.py
+++ b/src/aiconfigurator/sdk/perf_database.py
@@ -1930,7 +1930,13 @@ class PerfDatabase:
         """
         Set the default sol mode
         """
-        self._default_sol_mode = mode
+        if mode != self._default_sol_mode:
+            # Clear cached query methods since default sol mode affects the results
+            for attr_name in dir(self):
+                attr = getattr(self, attr_name)
+                if hasattr(attr, "cache_clear") and callable(attr):
+                    attr.cache_clear()
+            self._default_sol_mode = mode
 
     def get_default_sol_mode(self) -> common.SOLMode:
         """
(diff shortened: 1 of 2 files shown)


src/aiconfigurator/sdk/perf_database.py
```diff
@@ -2042,6 +2045,7 @@ def get_sol(
             latency = self._interp_3d(n, full_s, b, attention_dict, "cubic") * prefix_correction
             return latency
 
+    @functools.lru_cache(maxsize=32768)
```
tianhaox, line 2054: There's one member of database is editable self._default_sol_mode. In most cases, we will not pass in the sol mode. However, we will possibly set the database default mode from default to sol or any other mode. (especially when we want to add hybrid and power mode in future). Is there any quick solution to this?

anish-shanbhag: Good catch - just pushed https://github.com/ai-dynamo/aiconfigurator/pull/128/commits/18958fe70d70227c3134973df5a4037ea099a980 which addresses this in the simplest way possible (invalidating the cache when default SOL mode is changed). I also updated tests to check this behavior.

I did consider a few other approaches, such as:
1. Caching interpolation methods instead of query methods
2. Requiring default SOL mode to be specified when `PerfDatabase` is created, meaning it can't be changed later
3. Splitting query methods into one method that resolves the SOL mode, and another that does the actual query logic

But all of these would require a decent amount of refactoring.

tianhaox: Yes, good idea. In near future we can follow current logic. But do a planning in upcoming cycles about the refactoring. Previously I want to resue db obj in webapp as users might want to compare different modes, that's why we have this design.

Arsene12358 (approved):

PR #131, opened 2025-11-25
fix: moe memory fix, nvfp4 gemm fix for 1.x trtllm

YijiaZhao:
fix: moe memory fix, nvfp4 gemm fix for 1.x trtllm

### DIFF
--- a/collector/collect.py
+++ b/collector/collect.py
@@ -40,7 +40,7 @@ from datetime import datetime
 import torch
 from tqdm import tqdm
 
-from helper import create_test_case_id, save_error_report, setup_logging, setup_signal_handlers
+from helper import EXIT_CODE_RESTART, create_test_case_id, save_error_report, setup_logging, setup_signal_handlers
 
 logger = None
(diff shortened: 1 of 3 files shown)


tianhaox:
do you mean that if it's moe, we need to restart the process for every test case?

YijiaZhao:
> do you mean that if it's moe, we need to restart the process for every test case?

yeah, Every moe case has 30 cases for different tokens which means it is a big case!

YijiaZhao:
> > > do you mean that if it's moe, we need to restart the process for every test case?
> > 
> > 
> > yeah, Every moe case has 30 cases for different tokens which means it is a big case!
> 
> Can you try to midify run_moe_torch() instead of modify the persistent task process? We can have a delegate func def run_moe_torch_delegate(device): run subprocess onto the same device to handle the moe collecting. Then you don't have to restart the 8 task process again and again.

sorry. For current codes,  only moe need restart. other collectors don't.

PR #139, opened 2025-11-26
feat: Add MoE support for VLLM

ilyasher:
#### Overview:

Copy of https://github.com/ai-dynamo/aiconfigurator/pull/118, which I merged into the wrong branch. This PR merges into `main`.

### DIFF
--- a/collector/collect.py
+++ b/collector/collect.py
@@ -336,7 +336,6 @@ def collect_ops(
                     "traceback": traceback.format_exc(),
                 }
             )
-    return all_errors
 
     return all_errors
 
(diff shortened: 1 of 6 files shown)


davilu-nvidia:
I'm a little worried about the perf - vllm moe can only achive ~10% of SOL. 
1. Did you run nsys (say qwen3 32b) and allign with actual moe kernel execution time? please refer to AIC Performance Alignment Criteria doc as reference.
2. Regarding autotuner for MoE, if it takes a few hours, how can deleloper adopt it in real production, is it a pre-deployment process? any reference link

PR #140, opened 2025-11-26
chore: Factor out common moe and gemm test cases

ilyasher:
#### Overview:

The perf collectors for trtllm, sglang, vllm all duplicate roughly the same test cases for all the ops. This PR factors out the common MoE and gemm sizes into a separate file.


collector/common_test_cases.py
```diff
@@ -0,0 +1,195 @@
+import dataclasses
```
jasonqinzhou, line 3: copyright check failed.

PR #143, opened 2025-11-27
feat: sglang wideep tp attention modeling

AichenF:
In wide for sglang, when TP attention is enabled in the prefill phase, it affects the input to the shared experts as num_tokens/tp_size, and introduces two additional NCCL operators to handle the TP attention outputs(all gather and reduce scatter). Here, the modeling and MLP logic for WideEP DeepSeek has been modified accordingly.

### DIFF
--- a/src/aiconfigurator/sdk/models.py
+++ b/src/aiconfigurator/sdk/models.py
@@ -1323,7 +1323,7 @@ class WideEPDeepSeekModel(BaseModel):
                     h,
                     self._moe_inter_size,
                     moe_quant_mode,
-                    tp_attn_size=tp_size,
+                    tp_size=tp_size,
                 )
             ]
         )
--- a/src/aiconfigurator/sdk/operations.py
+++ b/src/aiconfigurator/sdk/operations.py
@@ -747,11 +747,11 @@ class WideEPMLP(Operation):
         self._quant_mode = quant_mode
         self._weights = (3 * self._hidden_size * self._intermediate_size) * quant_mode.value.memory
         self.is_context = kwargs.get("is_context", True)  # Default to context mode
-        self.tp_attn_size = kwargs.get("tp_attn_size", 1)
+        self.tp_size = kwargs.get("tp_size", 1)
 
     def query(self, database: PerfDatabase, **kwargs):
         x = kwargs.get("x")  # num_tokens
-        x /= self.tp_attn_size
+        x /= self.tp_size
         overwrite_quant_mode = kwargs.get("quant_mode")
         quant_mode = self._quant_mode if overwrite_quant_mode is None else overwrite_quant_mode
 


src/aiconfigurator/sdk/operations.py
```diff
@@ -747,9 +747,11 @@ def __init__(
         self._quant_mode = quant_mode
         self._weights = (3 * self._hidden_size * self._intermediate_size) * quant_mode.value.memory
         self.is_context = kwargs.get("is_context", True)  # Default to context mode
+        self.tp_attn_size = kwargs.get("tp_attn_size", 1)
 
     def query(self, database: PerfDatabase, **kwargs):
         x = kwargs.get("x")  # num_tokens
+        x /= self.tp_attn_size
```
tianhaox: what's the x here? in non-wideep path, x is local num tokens, and then the tokens fed into moe module is x*attention_dp_size, what about here?

AichenF: x is local num tokens, after tp attention and reduce scatter, each shared expert processes x/tp_size local tokens.

Arsene12358: Maybe it's better not to use "x" to name variable or dict keys. It is just not very descriptive

xutizhou: should we consider the tp of shared expert?

src/aiconfigurator/sdk/models.py
```diff
@@ -1295,6 +1323,7 @@ def __init__(self, topk: int, num_experts: int, moe_inter_size: int, *args) -> N
                     h,
                     self._moe_inter_size,
                     moe_quant_mode,
+                    tp_attn_size=tp_size,
```
xutizhou: does tp_attn_size equal to tp_size? should we use two params to distinguish these.

src/aiconfigurator/sdk/operations.py
AichenF: > Maybe it's better not to use "x" to name variable or dict keys. It is just not very descriptive

sure.

AichenF: > should we consider the tp of shared expert?

If we replace the MLP with GEMM, it should also cover the TP scenario.

src/aiconfigurator/sdk/models.py
AichenF: yes, they are equal

src/aiconfigurator/sdk/operations.py
tianhaox: all code uses x. it can be refactored in another PR to rename this as num_tokens, or something.

src/aiconfigurator/sdk/models.py
tianhaox: can we leverage tp_size directly, any risk?

PR #155, opened 2025-12-05
feat: replace deepseek mlp with gemm

AichenF:


src/aiconfigurator/sdk/operations.py
```diff
@@ -111,15 +111,25 @@ class GEMM(Operation):
...
         super().__init__(name, scale_factor)
         self._n = n
         self._k = k
         self._quant_mode = quant_mode
         self._weights = self._n * self._k * quant_mode.value.memory
+        self._tp_size = kwargs.get("tp_size", 1)
 
     def query(self, database: PerfDatabase, **kwargs):
         x = kwargs.get("x")
+        x //= self._tp_size
```
tianhaox: I'm still not quite sure about why we need to divide x by tp_size. x typically is the local rank num_tokens. is shared moe having same parallel as attention?

AichenF: TP attention﻿ changes the number of tokens that the shared expert﻿ has to process. The current modeling does not take the TP shared expert﻿ into account. In sglang﻿, the shared expert﻿ TP﻿ is controlled by a separate parameter and is independent of attention﻿. Modeling for TP shared expert﻿ can be further extended in the future.

tianhaox: is tp size, attn dp size, moe tp size and moe ep size enough for sglang wideep?
tp size/attn dp size in current modeling is used to control the attn module while moe tp size and moe ep size are for moe module. what's the tp size here, is it typically same as attn module's tp size or moe's tp size or neither?

AichenF: Same as the TP size of the attention module, TP MoE and TP shared expert are not yet modeled. The moe-dense-tp-size argument is used to control TP shared experts and is an independent parameter from the moe_tp_size.

PR #169, opened 2025-12-13
fix: use ttft to filter prefill candidates

tianhaox:
use ttft to filter out prefill candidates rather than context_latency


Arsene12358:
I see this is for AIC-259. LGTM

Arsene12358 (approved):

PR #172, opened 2025-12-14
fix: Update DeepEP interface for SGLang 0.5.6+ compatibility

xutizhou:
## Summary
Refactor and optimize the DeepEP MoE benchmark in `collect_wideep_deepep_moe.py` with improved configuration handling and performance optimizations.

## Changes
- **Dynamic `deepep_mode` configuration**: Set `deepep_mode="normal"` for prefill and `deepep_mode="low_latency"` for decode to match optimal settings for each phase
- **Performance optimization**: Pre-compute `expected_m` values and clone `masked_m` tensors outside kernel function to reduce overhead
- **Subprocess MoE initialization**: Add proper `_set_envs_and_config` and `initialize_moe_config` calls in subprocesses for correct DeepEP + DeepGEMM backend setup
- **Configuration clarity**: Update `top_k` assignment to use `moe_layer.topk.topk_config.top_k` for improved readability
- **Backend specification**: Explicitly set `moe_runner_backend="deep_gemm"` for DeepGEMM MoE runner

- **SGLang**: = 0.5.6

### DIFF
--- a/collector/sglang/collect_wideep_mlp.py
+++ b/collector/sglang/collect_wideep_mlp.py
@@ -19,6 +19,7 @@ from sglang.srt.distributed import (
 )
 from sglang.srt.layers.quantization import Fp8Config
 from sglang.srt.models.deepseek_v2 import DeepseekV2MLP
+from sglang.srt.server_args import ServerArgs, set_global_server_args_for_scheduler
 
 DEEPSEEK_MODEL_PATH = os.environ.get("DEEPSEEK_MODEL_PATH", "/deepseek-v3")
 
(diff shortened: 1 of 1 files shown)


collector/sglang/collect_wideep_deepep_moe.py
```diff
@@ -845,8 +887,9 @@ def run_moe(
         tp_size=2,
         trust_remote_code=True,
         mem_fraction_static=0.3,
-        enable_deepep_moe=True,
-        enable_ep_moe=True,
+        moe_a2a_backend="deepep",  # replaced enable_deepep_moe=True
```
jasonqinzhou, line 890: Will these changes break aic sglang < 0.5.6 ?

Shall we collect one gpu to make sure new data points work e2e?

xutizhou: Yes, this update is not compatible with original version.  I have checked the diff of new data and original data. the variance is within 20% which looks reasonable.

Arsene12358:
Discussed with @xutizhou ,  I think we can merge this one first to start updating database. Any follow-up issues can be separate PRs

Arsene12358 (approved):

PR #178, opened 2025-12-16
fix: make SOL_FULL mode return a tuple like before

ilyasher:
#### Overview:

Recent MR changed SOL_MODE to just return a single number instead of a tuple of (sol_total, sol_math, sol_mem). This broke the code in the `validate_database.ipynb` validation notebook. This PR reverts the change and makes SOL_MODE return a tuple as before (using the new `PerformanceResult` class).


tianhaox:
https://github.com/ai-dynamo/aiconfigurator/pull/170 i think it's duplicate with this pr

ilyasher:
@tianhaox `validate_database.ipynb` is working now on `main` 👍 . Should we still make SOL_FULL return a tuple of `PerformanceResult` instead of a tuple of `float`?

Arsene12358:
I prefer the old way of just returning tuple of float. tuple of PerformanceResult seems a bit over complicated

PR #193, opened 2025-12-19
fix: config generator missed moe parallel config if using huggingface_id

jasonqinzhou:
#### Overview:

fix: config generator missed moe parallel config if using huggingface_id #193

#### Details:

aic command
```
aiconfigurator cli exp --yaml_path src/aiconfigurator/cli/exps/moe.yaml --save_dir moe_out
```

Yes, the MoE parallel config IS included in generator_config.yaml:
generator_config.yamlLines 34-39
    tensor_parallel_size: 1    pipeline_parallel_size: 1    data_parallel_size: 2    moe_tensor_parallel_size: 1    moe_expert_parallel_size: 2    max_batch_size: 1
For prefill: moe_tensor_parallel_size: 1, moe_expert_parallel_size: 2
For decode: moe_tensor_parallel_size: 1, moe_expert_parallel_size: 4
However, I notice a potential issue: the actual engine configs (prefill_config.yaml, decode_config.yaml) and k8s_deploy.yaml don't include the MoE-specific parameters - they only have tensor_parallel_size and pipeline_parallel_size:
prefill_config.yamlLines 2-3
tensor_parallel_size: 1pipeline_parallel_size: 1
The moe_tensor_parallel_size and moe_expert_parallel_size parameters are calculated but not being written to the engine configs that get deployed. This might be intentional (if TRT-LLM infers them) or a bug in the config generation.

AFTER THE FIX:
I found the issue! The is_moe flag is stored in ModelConfig but not being passed to the template context, so the MoE section in the engine config template is never rendered.

The fix works! The MoE parallel config is now included in the engine configs:
prefill_config.yaml:
moe_expert_parallel_size: 2moe_tensor_parallel_size: 1moe_config:    backend: CUTLASS
decode_config.yaml:
moe_expert_parallel_size: 4moe_tensor_parallel_size: 1moe_config:    backend: CUTLASS
These values match what was already in generator_config.yaml but were previously missing from the actual engine configs. The k8s_deploy.yaml should also include these now.

- closes GitHub issue: #xxx


simone-chen (approved):

src/aiconfigurator/generator/rendering/engine.py
```diff
@@ -347,6 +347,16 @@ def prepare_template_context(param_values: dict[str, Any], backend: str) -> dict
     """
     context = {}
 
+    # Extract ModelConfig (is_moe, nextn, etc.)
+    model_config = param_values.get("ModelConfig", {})
+    if model_config.get("is_moe"):
+        context["is_moe"] = model_config["is_moe"]
+    if model_config.get("nextn"):
+        context["is_speculative"] = True
+        context["nextn"] = model_config["nextn"]
```
ilyasher: Is there an edge case for `nextn == 1` or no?

ilyasher (approved):

Arsene12358 (approved):

src/aiconfigurator/generator/rule_plugin/trtllm.rule
```diff
@@ -18,3 +18,8 @@ agg_prefill_decode enable_attention_dp = ((data_parallel_size or 1) > 1) and Mod
 
 when (ModelConfig.prefix or 0) > 0:
     agg_prefill_decode disable_prefix_cache = false
+
+# Speculative decoding
+when (ModelConfig.nextn or 0) > 0:
+    agg_decode speculative_decoding_type = "MTP"
+    agg_decode num_nextn_predict_layers = ModelConfig.nextn
```
Ethan-ES, line 25: Need to confirm to see if the settings are correct:
1.  speculative decoding is only enabled when nextn>0, and only `agg` and `decode` will include this part.
2. num_nextn_predict_layers==nextn (num_nextn_predict_layers represents for the [speculative_config.num_nextn_predict_layers](https://github.com/NVIDIA/TensorRT-LLM/blob/v1.2.0rc2/tensorrt_llm/llmapi/llm_args.py#L813) in trtllm)

tianhaox (approved):

src/aiconfigurator/generator/rule_plugin/trtllm.rule
tianhaox: what's our previous logic for deepseek with mtp=1?
aic core doesn't support eagle now. It only works with deepseek v3. but will soon support that.
And IIRC, there's some difference between trtllm versions for nextn config. we also need to think about how we support this in future, say, some feature matrix tells people what's verifie. We cannot maintain everything by us, especially to support previous versions once we add a new keyword.

PR #229, opened 2026-01-07
fix: enable hybrid mode of TP/DP/EP in wideep SGLang

AichenF:
Added a new parameter to the MoEDispatch operation to scale the number of tokens when using Tensor Parallelism with attention in DeepEP. In WideEP, the search space of xTP–yDP–xyEP configurations can now be explored.

### DIFF
--- a/src/aiconfigurator/sdk/operations.py
+++ b/src/aiconfigurator/sdk/operations.py
@@ -440,7 +440,7 @@ class MoEDispatch(Operation):
                         hidden_size=self._hidden_size,
                     )
             else:
-                assert self._attention_tp_size == 1 and self._attention_dp_size == 1, (
+                assert self._attention_tp_size == 1 or self._attention_dp_size == 1, (
                     "We don't enable the path for non-wideep SGLang to support TP>1 and DP>1 for attn simultaneously"
                 )
                 # TODO: support TP+DP


src/aiconfigurator/sdk/operations.py
```diff
@@ -275,6 +275,7 @@ def __init__(
         self._sms = kwargs.get("sms", 12)
         self._moe_backend = kwargs.get("moe_backend")
         self._is_context = kwargs.get("is_context", True)
+        self._scale_num_tokens = kwargs.get("scale_num_tokens", 1)
```
xutizhou, line 278: Should we rename _scale_num_tokens for clarity?

src/aiconfigurator/sdk/models.py
xutizhou, line 1415: should we apply the scale at decode phase

src/aiconfigurator/sdk/operations.py
AichenF: The name is kept consistent with the one used previously in GEMM

src/aiconfigurator/sdk/models.py
AichenF: When using TP in the decode phase, the same issue would also occur. It hasn’t been added for now because in practical deployments TP is generally not used during decode.

xutizhou (approved):

PR #240, opened 2026-01-11
fix: add w4a16_mxfp4 moe data and set proper moe_quant_mode default for gpt-oss

Arsene12358:
#### Overview:

- Add missing w4a16_mxfp4 perf data for trtllm backend
- Set default moe_quant_mode of gpt-oss models to w4a16_mxfp4


tianhaox:
is collector collecting this precision by default, why the db missed the data.

jasonqinzhou (approved):

PR #246, opened 2026-01-15
feat: support AIConfigurator on Intel XPU

Spycsh:
#### Overview:

Support AIConfigurator on Intel XPU

#### Details:

Thanks for this great project! Here is the initial support for AIConfigurator on Intel B60 XPU. Most of the code is currently hardcoded for CUDA, so I have addressed those gaps and made it XPU-compatible. The operation performance data were measured on my end (B60 XPU), and this PR includes a partial release of those results. Any feedback is very welcome!

None

## Release Notes

* **New Features**
  * Added support for Intel XPU accelerators as an alternative to CUDA.
  * Introduced `--device` flag to select between CUDA and XPU backends for data collection.
  * Added B60 system configuration and performance profiles.

* **Documentation**
  * Updated README with examples for running collectors with XPU backend on Intel hardware.


collector/collect_all_reduce.py
```diff
@@ -337,21 +353,25 @@ def benchmark_vllm_allreduce(
...
         # Test both graph capture and eager mode
-        for use_graph in [True, False]:
+        flag_lst = [False] if torch.xpu.is_available() else [True, False]
+        for use_graph in flag_lst:
             mode_str = "graph" if use_graph else "eager"
 
             if use_graph:
                 # Graph capture mode
-                with vllm_mods["graph_capture"](device=torch.cuda.current_device()) as graph_capture_context:
+                with vllm_mods["graph_capture"](device=torch.xpu.current_device()) as graph_capture_context:
```
Arsene12358: wouldn't this change just make graph cpature mode crash on cuda devices? And by the look of it, I think the correct intention is that only cuda devices should run under graph mode, and eager mode should be available for both cuda and xpu devices

collector/collect_all_reduce.py
```diff
@@ -400,22 +420,22 @@ def benchmark_vllm_allreduce(
...
+                torch.xpu.synchronize()
 
                 # Stop power monitoring
                 if power_monitor:
                     power_stats = power_monitor.stop_sampling()
 
             else:
                 # Eager mode
-                input_tensor = torch.ones(input_shape, dtype=torch_dtype, device="cuda")
+                input_tensor = torch.ones(input_shape, dtype=torch_dtype, device="xpu")
```
Arsene12358: Similar logic here, with this change, eager mode will just crash on cuda devices

collector/vllm/collect_attn.py
```diff
@@ -348,6 +362,14 @@ def run():
     dtype_str = "float16"
     kernel_source = f"vllm_{backend_name_str}".lower()
 
+    if torch.cuda.is_available():
+        device_name = torch.cuda.get_device_name(device)
+    elif torch.xpu.is_available():
+        device_name = torch.xpu.get_device_name(device)
```
Arsene12358: The if cuda /elif xpu patterns appears rather often, would it be better to define some helper functions in helper.py to avoid repetitions. Maybe something like
```python
def get_device_module():
    if torch.cuda.is_available():
        return torch.cuda
    elif torch.xpu.is_available():
        return torch.xpu
    raise RuntimeError("No supported device")

def get_device_str():
    if torch.cuda.is_available():
        return "cuda"
    elif torch.xpu.is_available():
        return "xpu"
    raise RuntimeError("No supported device")
```

collector/collect_all_reduce.py
Spycsh: Exactly, the current intention is to use eager mode for xpu and cuda can use both modes. Thanks, fixed.

collector/collect_all_reduce.py
Spycsh: Thanks, fixed.

collector/vllm/collect_attn.py
Spycsh: @Arsene12358  Yes I agree moving some shared code to some common place would be good. However, I think currently in `helper.py` , `import torch` is clarified in every function when needed. To follow the current code, I need to write functions like 
```py
def get_device_module():
    import torch
    if torch.cuda.is_available():
        return torch.cuda
    elif torch.xpu.is_available():
        return torch.xpu
    raise RuntimeError("No supported device")

def get_device_str():
    import torch
    if torch.cuda.is_available():
        return "cuda"
    elif torch.xpu.is_available():
        return "xpu"
    raise RuntimeError("No supported device")
```

I think it could be somewhat not elegant to import torch every time when sync or get device. What do you think?

Arsene12358: @Spycsh I think it's fine to do import torch inside the helper functions. We can also apply a lru cache to make the function only exectute once

```python
@functools.lru_cache(maxsize=1)
def get_device_module():
    import torch
    if torch.cuda.is_available():
        return torch.cuda
    if torch.xpu.is_available():
        return torch.xpu
    raise RuntimeError("No supported device (need CUDA or XPU)")
```

Spycsh: @Arsene12358  thanks, updated.

Arsene12358:
@Spycsh Thanks for the fixes. Could you also add a few test cases under /tests to make sure that the e2e workflow is functionally working and this can also help us guard against future changes that might break things.

collector/vllm/collect_attn.py
tianhaox, line 1: I have a quick discussion with team, suggest maintaining XPU collector in a seperate file, say, collect_attn_xpu.py, so that you can have your own cadence to update the version because we currently don't have XPU to be included in automation. Do you think that makes sense? I will have concern that our modification to the collector might break your code

src/aiconfigurator/systems/b60.yaml
```diff
@@ -0,0 +1,30 @@
...
+  mem_empirical_constant_latency: 0.000003 # 3us some nonofficial correction based on observations, you should try to modify based on your own observations
+  mem_capacity: 25769803776 # 24GiB
+  float32_tc_flops: 12280000000000 # 12.28TFLOPS
+  float16_tc_flops: 98304000000000 # 98.3TFLOPS
+  int8_tc_flops: 197000000000000 # 197TOPS
+  power: 200 # Watt (Total Board Power)
+
+node:
+  num_gpus_per_node: 8 # check your # cards with sycl-ls
+  inter_node_bw: 25000000000 # Byte/s per GPU, single direction, 1:1 CX7 per node, 25GB/s
```
tianhaox: we have a fix in another PR that CX7 should be 50GB/s per card. What's the typical networking setup for B60?

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -1884,12 +1884,15 @@ def __init__(self, system: str, backend: str, version: str, systems_root: str =
         self._extracted_metrics_cache = {}
 
         data_dir = os.path.join(systems_root, self.system_spec["data_dir"], backend, version)
-        nccl_data_dir = os.path.join(
-            systems_root,
-            self.system_spec["data_dir"],
-            "nccl",
-            self.system_spec["misc"]["nccl_version"],
-        )
+        if "misc" in self.system_spec:
```
tianhaox: looks like your system yaml file has 'misc' section. what's this for?

collector/collect_comm.sh
tianhaox, line 114: is trtllm path also working for xpu?

Spycsh: No, currently vllm is the only mature backend supported on Intel XPU, and sglang will possibly be supported on XPU in the future. Trtllm path is cuda specific and is not supported.

src/aiconfigurator/systems/b60.yaml
Spycsh: We also use CX7 and find this should be 50GB/s per card. Thanks, will update.

collector/vllm/collect_attn.py
Spycsh: @tianhaox  I'm considering whether we should introduce a BaseCollector, with GPUCollector and XPUCollector inheriting from it to handle operator collection. This follows the same pattern used in vLLM (similar to its ModelRunner design), allowing us to reuse as much common logic as possible in the base class—i.e., encapsulating device-agnostic behaviors.

This abstraction involves extra work, such as creating a BaseCollector. What do you think?

src/aiconfigurator/sdk/perf_database.py
Spycsh: Thanks for pointing out, will update.

collector/vllm/collect_attn.py
tianhaox: Can you say more about this design? We now have some platform specific code of each component. cc @ilyasher for dicussion as well.

hshen14: Basically, it's device abstraction which allows the common operation interfaces in BaseCollector and the different implements in DeviceCollector. This may help benefit the future code maintenance in case more devices would be supported in Dynamo.

Spycsh: @tianhaox @ilyasher this design is like following, 2 files `collect_gemm.py` and `collect_gemm_xpu.py`:

* `collect_gemm.py`
```py
# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0

__compat__ = "vllm>=0.11.0"

import os
import torch

from vllm.config import VllmConfig, set_current_vllm_config
from vllm.model_executor.layers.linear import RowParallelLinear
from vllm.model_executor.layers.quantization.fp8 import Fp8Config

try:
    from vllm.model_executor.layers.quantization.utils.fp8_utils import (
        maybe_post_process_fp8_weight_block,
    )
except Exception:
    print("No maybe_post_process_fp8_weight_block found, please check your vLLM version.")

from vllm.utils.deep_gemm import per_block_cast_to_fp8
from vllm.version import __version__ as vllm_version

from collector.common_test_cases import get_gemm_common_test_cases
from collector.helper import benchmark_with_power, get_device_module, get_sm_version, log_perf
from collector.vllm.utils import setup_distributed, with_exit_stack


FP8_BLOCK_SHAPE = (128, 128)


class GPUGemmCollector:
    device_type = "cuda"

    def __init__(self, device=None):
        self.device = device or f"{self.device_type}:0"

    # ----------------------------
    # Device helpers (overrideable)
    # ----------------------------

    def is_gpu_available(self):
        return torch.cuda.is_available()

    def setup_device(self):
        setup_distributed(self.device)

        dtype = torch.float16
        torch.set_default_dtype(dtype)

        get_device_module().set_device(self.device)

        if self.is_gpu_available():
            torch.set_default_device(self.device)

        return dtype

    # ----------------------------
    # Test case generation
    # ----------------------------

    def get_gemm_test_cases(self):
        sm = get_sm_version()

        gemm_list = ["float16"]
        if sm > 86:
            gemm_list += ["fp8"]

        if sm >= 90:
            gemm_list += ["fp8_block"]

        test_cases = []

        for gemm_common_testcase in get_gemm_common_test_cases():
            x = gemm_common_testcase.x
            n = gemm_common_testcase.n
            k = gemm_common_testcase.k

            for gemm_type in gemm_list:
                if gemm_type in ("nvfp4", "fp8_block") and (n < 128 or k < 128):
                    continue

                if gemm_type == "fp8_block":
                    block_n, block_k = FP8_BLOCK_SHAPE

                    if (n % block_n) != 0 or (k % block_k) != 0:
                        continue

                    if sm >= 100 and (x % 4) != 0:
                        continue

                test_cases.append([gemm_type, x, n, k, "gemm_perf.txt"])

        return test_cases

    # ----------------------------
    # Quant config
    # ----------------------------

    def create_quant_config(self, gemm_type):
        if gemm_type == "fp8":
            return Fp8Config(
                is_checkpoint_fp8_serialized=True,
                activation_scheme="static",
                ignored_layers=None,
                weight_block_size=None,
            )

        if gemm_type == "fp8_block":
            return Fp8Config(
                is_checkpoint_fp8_serialized=True,
                activation_scheme="dynamic",
                weight_block_size=list(FP8_BLOCK_SHAPE),
            )

        return None

    # ----------------------------
    # GEMM creation
    # ----------------------------

    def create_gemm(self, gemm_type, m, n, k, dtype, x):
        qc = self.create_quant_config(gemm_type)

        gemm = RowParallelLinear(
            input_size=k,
            output_size=n,
            bias=False,
            skip_bias_add=True,
            params_dtype=dtype,
            quant_config=qc,
            prefix="",
            return_bias=True,
            disable_tp=True,
        )

        gemm.to(torch.device(self.device))

        if gemm_type == "fp8" and hasattr(gemm, "weight"):
            new_weight = gemm.weight.data.t()
            gemm.weight = torch.nn.Parameter(new_weight)

        elif gemm_type == "fp8_block":
            self.setup_block_fp8_weight(gemm, n, k)

        gemm.forward(x)  # dry run
        return gemm

    def setup_block_fp8_weight(self, gemm, n, k):
        block_n, block_k = FP8_BLOCK_SHAPE

        with torch.no_grad():
            raw_weight = torch.randn((n, k), dtype=torch.float32, device=self.device)

            q_weight, weight_scale = per_block_cast_to_fp8(
                raw_weight,
                [block_n, block_k],
                use_ue8m0=False,
            )

            if hasattr(gemm, "weight"):
                gemm.weight.copy_(q_weight)

            if hasattr(gemm, "weight_scale_inv"):
                gemm.weight_scale_inv.copy_(
                    weight_scale.contiguous().to(torch.float32)
                )

                if not hasattr(gemm, "weight_scale"):
                    gemm.weight_scale = gemm.weight_scale_inv

            try:
                maybe_post_process_fp8_weight_block(gemm)
            except TypeError:
                maybe_post_process_fp8_weight_block(
                    gemm, cutlass_block_fp8_supported=True
                )

    # ----------------------------
    # GEMM execution
    # ----------------------------

    @with_exit_stack
    def run_gemm(self, exit_stack, gemm_type, m, n, k, perf_filename):
        os.environ["VLLM_USE_DEEP_GEMM"] = "1"

        dtype = self.setup_device()

        x = torch.randn((m, k), dtype=dtype, device=torch.device(self.device))

        exit_stack.enter_context(set_current_vllm_config(VllmConfig()))

        outside_loop_count = 6
        op_list = [
            self.create_gemm(gemm_type, m, n, k, dtype, x)
            for _ in range(outside_loop_count)
        ]

        def kernel_func():
            for op in op_list:
                op.forward(x)

        with benchmark_with_power(
            device=self.device,
            kernel_func=kernel_func,
            num_warmups=3,
            num_runs=6,
            repeat_n=1,
        ) as results:
            pass

        log_perf(
            item_list=[
                {
                    "gemm_dtype": gemm_type,
                    "m": m,
                    "n": n,
                    "k": k,
                    "latency": results["latency_ms"] / outside_loop_count,
                }
            ],
            framework="VLLM",
            version=vllm_version,
            device_name=get_device_module().get_device_name(self.device),
            op_name="gemm",
            kernel_source="vllm_default",
            perf_filename=perf_filename,
            power_stats=results["power_stats"] if self.is_gpu_available() else None,
        )

if __name__ == "__main__":
    collector = GPUGemmCollector()

    test_cases = collector.get_gemm_test_cases()

    for test_case in test_cases[:10]:
        collector.run_gemm(*test_case)

```

*  `collect_gemm_xpu.py`
```py
# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES.
# SPDX-License-Identifier: Apache-2.0

import torch

from collector.collect_gemm_gpu import GPUGemmCollector
from collector.vllm.utils import setup_distributed


class XPUGemmCollector(GPUGemmCollector):
    device_type = "xpu"

    def is_gpu_available(self):
        return hasattr(torch, "xpu") and torch.xpu.is_available()

    def setup_device(self):
        setup_distributed(self.device)

        dtype = torch.float16
        torch.set_default_dtype(dtype)

        if hasattr(torch, "xpu"):
            torch.xpu.set_device(self.device)

        return dtype


if __name__ == "__main__":
    collector = XPUGemmCollector()

    test_cases = collector.get_gemm_test_cases()

    for test_case in test_cases[:10]:
        collector.run_gemm(*test_case)
```

Spycsh: In this design, we use a class `GPUGemmCollector` that encapsulate the common behaviors `create_gemm` and `run_gemm`. However, for XPU, as you know, b60 has no `sm_version` and also small memory compared to cuda devices so we may need to narrow down some test cases (remove some op shapes), so we can rewrite `get_gemm_test_cases` method e.g.:

```bash
class XPUGemmCollector(GPUGemmCollector):
    device_type = "xpu"

    def is_gpu_available(self):
        return hasattr(torch, "xpu") and torch.xpu.is_available()

    def get_gemm_test_cases():
        # !!!!!!!!!!!!!!! Rewrite here !!!!!!!!!!

        gemm_list = ["float16"]

        test_cases = []

        # Do a filter 
        test_cases  = get_gemm_common_test_cases()
        test_cases_xpu = self.filter_gemm_test_cases_xpu(test_cases)

        for gemm_common_testcase in test_cases_xpu:
            x = gemm_common_testcase.x
            n = gemm_common_testcase.n
            k = gemm_common_testcase.k
            for gemm_type in gemm_list:
                if gemm_type in ("nvfp4", "fp8_block") and (n < 128 or k < 128):
                    continue
                if gemm_type == "fp8_block":
                    block_n, block_k = FP8_BLOCK_SHAPE
                    # Block-wise kernels expect dimensions that align with the block.
                    if (n % block_n) != 0 or (k % block_k) != 0:
                        continue

                test_cases.append([gemm_type, x, n, k, "gemm_perf.txt"])

            return test_cases
    
    def filter_gemm_test_cases_xpu(self):
        # narrow down the search space a bot, e.g. remove the test cases wih nk_dim 16384, 65536
        pass

    def setup_device(self):
        setup_distributed(self.device)

        dtype = torch.float16
        torch.set_default_dtype(dtype)

        if hasattr(torch, "xpu"):
            torch.xpu.set_device(self.device)

        return dtype
```

src/aiconfigurator/systems/b60.yaml
Spycsh: Done.

src/aiconfigurator/sdk/perf_database.py
Spycsh: Done.

src/aiconfigurator/systems/b60.yaml
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+data_dir: data/b60 # relative to systems_dir
+gpu:
+  mem_bw: 456000000000 # 456GB/s
+  mem_bw_empirical_scaling_factor: 0.8 # some nonofficial correction based on observations, you should try to modify based on your own observations
+  mem_empirical_constant_latency: 0.000003 # 3us some nonofficial correction based on observations, you should try to modify based on your own observations
+  mem_capacity: 25769803776 # 24GiB
```
hshen14, line 9: Assuming ECC disabled?

src/aiconfigurator/systems/b60.yaml
```diff
@@ -0,0 +1,30 @@
...
+  float32_tc_flops: 12280000000000 # 12.28TFLOPS
+  float16_tc_flops: 98304000000000 # 98.3TFLOPS
+  int8_tc_flops: 197000000000000 # 197TOPS
+  power: 200 # Watt (Total Board Power)
+
+node:
+  num_gpus_per_node: 8 # check your # cards with sycl-ls
+  inter_node_bw: 50000000000 # Byte/s per GPU, single direction, 1:1 CX7 per node, 50GB/s
```
hshen14, line 17: Any scaling factor for network bw?

collector/vllm/collect_attn.py
hshen14: Let's use separate file to support xpu connecter first as suggested, and then probably make a RFC to discuss the abstraction. @Spycsh

src/aiconfigurator/systems/b60.yaml
tianhaox: what kind of scaling factor?

src/aiconfigurator/systems/b60.yaml
tianhaox: For GPU, I think so. This part is not modeled. But if it's impacting the result a lot, better to consider it.

src/aiconfigurator/sdk/task.py
```diff
@@ -487,6 +489,11 @@ def _finalize_disagg(cls, config: DefaultMunch, ctx: TaskContext) -> None:
                 raise ValueError(f"total_gpus must be greater than 2 for disagg, got {ctx.total_gpus}")
             replica_cfg.max_gpu_per_replica = min(ctx.total_gpus, replica_cfg.get("max_gpu_per_replica"))
             logger.debug("Using max gpu per replica %s", replica_cfg.max_gpu_per_replica)
+            # Prefill/Decode num_gpu_per_worker should be strictly smaller than total_gpus
+            prefill_cfg.num_gpu_per_worker = [num for num in prefill_cfg.num_gpu_per_worker if num <= ctx.total_gpus]
+            logger.debug("Overwriting num gpu per prefill worker to %s", prefill_cfg.num_gpu_per_worker)
+            decode_cfg.num_gpu_per_worker = [num for num in decode_cfg.num_gpu_per_worker if num <= ctx.total_gpus]
+            logger.debug("Overwriting num gpu per decode worker to %s", decode_cfg.num_gpu_per_worker)
```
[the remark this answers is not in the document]
tianhaox: actually this will naturally be filtered out. But if we filter it here, it can be faster. Maybe wen can add some warning here

collector/helper.py
```diff
@@ -233,43 +234,47 @@ def benchmark_with_power(
...
+    if torch.cuda.is_available():
+        torch.cuda.synchronize()
+        for _ in range(num_warmups):
+            if use_graph:
+                g.replay()
+            else:
+                # Fallback: Direct execution matching actual execution path
+                for _ in range(repeat_n):
+                    kernel_func()
+        torch.cuda.synchronize()
```
tianhaox, line 277: Do you need to do warmup on XPU? this is skipped i think.

collector/helper.py
```diff
@@ -582,6 +586,8 @@ def get_sm_version():
             device = torch.cuda.current_device()
             capability = torch.cuda.get_device_capability(device)
             return capability[0] * 10 + capability[1]
+        elif torch.xpu.is_available():
+            return -1
```
tianhaox: this might be risky. Do you think it's safer to rewrite the get_xxx_cases for xpu? So that you can call different arch query by yourself and make it dedicated.

src/aiconfigurator/systems/b60.yaml
hshen14: Similar to memory bandwidth scaling factor. E.g., 0.8 for inter_node_bw.

Arsene12358: I don't think we have an efficiency scaling for network bw at the moment. We can have a follow up PR to take those into consideration across all of our supported platforms. It should be fine to use the raw value here as is, I think

collector/helper.py
Spycsh: I think warmup is already done before in https://github.com/Spycsh/aiconfigurator/blob/xpu/collector/helper.py#L238 ; think this should be enough for eager mode?

collector/helper.py
Spycsh: Agree, updated.

collector/vllm/collect_attn.py
Spycsh: @hshen14 agreed.

Arsene12358 (approved):

PR #290, opened 2026-01-25
feat: generator add benchmark mode by setting rule plugins

Ethan-ES:
#### Overview:

To ensure the generator can be used across different environments, we introduced a **benchmark mode(running with `--generator-set rule=benchmark`)** on top of the existing generator. In benchmark mode, the deployment configuration is generated strictly based on the simulation results produced by the aiconfigurator sdk, helping developers align performance more accurately.

By default, the generator runs in **production mode**, which makes trade-offs tailored for general production deployments, such as increasing batch size and reducing the CUDA Graph batch size.

In addition to these two modes, users can define their own rules by adding new folders under `rule_plugin`. When invoking the CLI, different rule sets can be selected via the corresponding `--generator-set rule=<folder_name>` option.


src/aiconfigurator/generator/rule_plugin/benchmark/vllm.rule
simone-chen, line 1: Do we want to include `cuda_graph_batch_sizes` to `vllm.rule` as well?

Ethan-ES: Thanks for the review, I've added the cuda_graph_batch_sizes to vllm.rule on line 8 to stay consistent with trtllm and sglang. Please let me know if this is not appropriate.

simone-chen (approved):

jasonqinzhou (approved):

PR #308, opened 2026-01-29
fix: sglang moe_dense_tp_size only support 1 or None

simone-chen:
#### Overview:

Limit the etp list to [1] as SGLang moe_dense_tp_size only support 1 or None right now: [code pointer](https://github.com/sgl-project/sglang/blob/b77b0ffd6021ef787c9ca7084731578c2f2f4687/python/sglang/srt/server_args.py#L4895). 

- closes GitHub issue: #xxx

### DIFF
--- a/src/aiconfigurator/generator/config/backend_config_mapping.yaml
+++ b/src/aiconfigurator/generator/config/backend_config_mapping.yaml
@@ -35,7 +35,9 @@ parameters:
 
 - param_key: moe_tensor_parallel_size
   vllm: null
-  sglang: moe-dense-tp-size
+  # https://github.com/sgl-project/sglang/issues/8530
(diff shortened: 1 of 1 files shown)


tianhaox:
this param is only for the shared expert in moe model. with it unset, it's None, which means we don't specify a different tp size for shared expert and just use the attention tp size. i think our current implementation should be correct.

tianhaox:
The new impl is correct i think. https://github.com/sgl-project/sglang/blob/b77b0ffd6021ef787c9ca7084731578c2f2f4687/python/sglang/srt/distributed/parallel_state.py#L1621

moe tp size in sgl is derived from tp_size//moe_ep_size.

tianhaox (approved):

PR #311, opened 2026-01-30
fix: ensure quantization block sizes can be divided by moe intermediate size per GPU

simone-chen:
#### Overview:

Avoid the error with AIC configs:
`ValueError: For quantized MoE models, please make sure (moe_intermediate_size=1536 / moe_tp_size=8) % weight_block_size_n=128 == 0 where moe_tp_size is equal to tp_size (8) divided by ep_size (1). You can fix this by setting arguments --tp and --ep correctly.`

- closes GitHub issue: #xxx

### DIFF
--- a/src/aiconfigurator/sdk/inference_session.py
+++ b/src/aiconfigurator/sdk/inference_session.py
@@ -518,10 +518,10 @@ class DisaggInferenceSession:
                         else:  # larger b will always OOM
                             break
                 except Exception as e:
-                    logger.exception(
+                    logger.warning(
                         f"Error getting candidate workers with parallel config: "
                         f"tp={tp_size}, pp={pp_size}, dp={dp_size}, moe_tp={moe_tp_size}, "
-                        f"moe_ep={moe_ep_size}; skipping this combination"
+                        f"moe_ep={moe_ep_size}; skipping this combination. Error: {e}"
                     )
                     exceptions.append(e)
                     continue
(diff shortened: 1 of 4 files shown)


src/aiconfigurator/sdk/task.py
```diff
@@ -1100,6 +1100,8 @@ def run_agg(self, task_config: DefaultMunch) -> dict[str, pd.DataFrame | None]:
                 is_moe=check_is_moe(task_config.model_path),
                 backend=common.BackendName(task_config.worker_config.backend_name),
                 enable_wideep=task_config.enable_wideep,
+                model_path=task_config.model_path,
+                moe_quant_mode=task_config.worker_config.moe_quant_mode,
```
ilyasher: I think instead of passing `model_path` and `moe_quant_mode` everywhere, it would be cleaner to do the `moe_intermediate_size // moe_tp_size % block_size == 0` check inside of the `MoeModel` class.

ilyasher (changes requested):

src/aiconfigurator/sdk/task.py
simone-chen: Excellent idea!

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -518,10 +518,10 @@ def _get_summary_df(
                         else:  # larger b will always OOM
                             break
                 except Exception as e:
-                    logger.exception(
```
simone-chen, line 521: This avoids the tracebacks to be printed in the console which largely increases log readability.

tianhaox (approved):

ilyasher (approved):

PR #321, opened 2026-02-01
feat: customized system path

tianhaox:
#### Overview:

 add --customized-paths to cli and webapp, allows reading additional paths

--system-paths default,/additional_path, or --system-paths /new_path


src/aiconfigurator/sdk/perf_database.py
```diff
@@ -175,60 +215,110 @@ def get_database(
...
+            data_path = os.path.join(systems_root, system_spec["data_dir"], backend, version)
             if os.path.exists(data_path):
                 try:
-                    database = PerfDatabase(system, backend, version, systems_dir)
-                    databases_cache[system][backend][version] = database
+                    database = PerfDatabase(system, backend, version, systems_root)
+                    databases_cache[cache_key][backend][version] = database
                 except Exception:
                     logger.exception(f"failed to load {system=}, {backend=}, {version=}")
                     database = None
```
Arsene12358: I think in the case of multiple system path we should continue the search loop if yaml is found, but database failed to load or does not exist instead of simply returning None and exit

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -175,60 +215,110 @@ def get_database(
...
                 database = None
-        else:
-            logger.exception(f"system yaml {os.path.join(systems_dir, system + '.yaml')} not found")
-            database = None
+            return database
+
+    logger.exception(f"system yaml {system}.yaml not found in {systems_paths}")
+    return None
 
     return database
```
Arsene12358: This return statement is not reachable, we should delete it

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -175,60 +215,110 @@ def get_database(
...
                     database = None
             else:
                 logger.exception(f"data path {data_path} not found")
                 database = None
-        else:
-            logger.exception(f"system yaml {os.path.join(systems_dir, system + '.yaml')} not found")
-            database = None
+            return database
+
+    logger.exception(f"system yaml {system}.yaml not found in {systems_paths}")
```
xutizhou: `logger.exception()` on line 253 is called outside an `except` block. `logger.exception()` automatically appends the current exception traceback via `sys.exc_info()`. Outside an `except` block there is no active exception, so it prints a spurious `NoneType: None` in the log output. Should be `logger.warning()` or `logger.error()`.

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -175,60 +215,110 @@ def get_database(
...
                 try:
-                    database = PerfDatabase(system, backend, version, systems_dir)
-                    databases_cache[system][backend][version] = database
+                    database = PerfDatabase(system, backend, version, systems_root)
+                    databases_cache[cache_key][backend][version] = database
                 except Exception:
                     logger.exception(f"failed to load {system=}, {backend=}, {version=}")
                     database = None
             else:
                 logger.exception(f"data path {data_path} not found")
```
xutizhou: Same issue here: `logger.exception()` is called outside an `except` block (this is the `else` branch of `if os.path.exists(data_path)`). It will produce a spurious `NoneType: None` line in the log. Should be `logger.warning()` or `logger.error()`.

src/aiconfigurator/sdk/perf_database.py
tianhaox: implemented in latest commit

src/aiconfigurator/sdk/perf_database.py
tianhaox: implemented in latest commit

src/aiconfigurator/sdk/perf_database.py
tianhaox: implemented in latest commit

src/aiconfigurator/sdk/perf_database.py
tianhaox: implemented in latest commit

Arsene12358 (approved):

PR #331, opened 2026-02-05
feat: add support for `--backend any` (homogenous)

simone-chen:
#### Overview:                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                     
Add support for `--backend any` to sweep across TensorRT-LLM, vLLM, and SGLang simultaneously and identify the optimal backend for a given workload. 

This PR does NOT support
- heterogenous hardware for P/D
- heterogenous backend for P/D

```
> aiconfigurator cli default --model Qwen/Qwen3-32B --backend any --total_gpus 8 --system h100_sxm --backend any --save_dir results                                                       
```
<img width="1950" height="362" alt="Screenshot 2026-02-04 at 5 18 07 PM" src="https://github.com/user-attachments/assets/f6484752-3dfe-4706-88e1-3956398c2842" />

### DIFF
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -15,17 +15,13 @@ import yaml
 
 from aiconfigurator import __version__
 from aiconfigurator.cli.report_and_save import log_final_summary, save_results
+from aiconfigurator.cli.utils import merge_experiment_results_by_mode, process_experiment_result
 from aiconfigurator.generator.api import (
     add_generator_override_arguments,
     generate_naive_config,
     generator_cli_helper,
 )
 from aiconfigurator.sdk import common, perf_database
-from aiconfigurator.sdk.pareto_analysis import (
-    get_best_configs_under_request_latency_constraint,
-    get_best_configs_under_tpot_constraint,
-    get_pareto_front,
-)
 from aiconfigurator.sdk.task import TaskConfig, TaskRunner
 from aiconfigurator.sdk.utils import get_model_config_from_model_path
 
@@ -100,10 +96,10 @@ def _add_default_mode_arguments(parser):
     )
     parser.add_argument(
         "--backend",
-        choices=[backend.value for backend in common.BackendName],
+        choices=[backend.value for backend in common.BackendName] + ["any"],
         type=str,
         default=common.BackendName.trtllm.value,
-        help="Backend name.",
+        help="Backend name. Use 'any' to sweep across all backends (trtllm, vllm, sglang) and compare results.",
     )
     parser.add_argument(
         "--backend_version",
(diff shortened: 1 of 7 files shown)


src/aiconfigurator/cli/main.py
```diff
@@ -522,13 +538,169 @@ def build_experiment_task_configs(
...
+            backend_name = exp_name.split("_", 1)[1]  # Extract backend from "agg_trtllm"
+            df = best_configs[exp_name].copy()
+            if not df.empty:
+                df["backend"] = backend_name
+                agg_dfs.append(df)
+
+        if agg_dfs:
+            # Combine all agg results and take top N
+            combined_agg = pd.concat(agg_dfs, ignore_index=True)
+            combined_agg = combined_agg.sort_values("tokens/s/gpu_cluster", ascending=False).head(5)
```
Arsene12358: by the look of it, this is taking top 5 not top N?

src/aiconfigurator/cli/main.py
```diff
@@ -522,13 +538,169 @@ def build_experiment_task_configs(
...
+
+    When backend="any", we have 6 experiments: agg_trtllm, agg_vllm, agg_sglang,
+    disagg_trtllm, disagg_vllm, disagg_sglang. This function merges them into 2:
+    agg (with top N from all backends) and disagg (with top N from all backends).
+
+    Returns merged: (best_configs, best_throughputs, pareto_fronts, pareto_x_axis, task_configs)
+    """
+    # Detect multi-backend mode
+    exp_names = list(results.keys())
+    is_multi_backend = any("_" in name for name in exp_names)
```
Arsene12358: I think the check is too fragile. In exp model user can define their experiment in yaml and it will be extremely common to have "_" in the exp_names. I think we should try to check whether backend is "any" explicitly

src/aiconfigurator/cli/main.py
```diff
@@ -522,13 +538,169 @@ def build_experiment_task_configs(
...
+                backends = (
+                    merged_pareto_fronts["agg"]["backend"].unique()
+                    if "backend" in merged_pareto_fronts["agg"].columns
+                    else "N/A"
+                )
+                logger.info(f"After recomputing Pareto: {num_points} points, backends: {backends}")
+                merged_pareto_x_axis["agg"] = x_axis
+
+    # Merge disagg experiments
+    if disagg_exps:
```
Arsene12358: it seems the merging logics of disagg_exps are very similar to the logics used to merge the agg results. If possible it might be more clean to define a helper function and use it in both places

simone-chen: Thanks, it makes a lot of sense. I refactored the function and moved many utilities into `cli/utils.py`

src/aiconfigurator/cli/main.py
simone-chen: Yes, this has been fixed.

Arsene12358 (approved):

PR #333, opened 2026-02-05
feat: make generator support both dynamo version and backend version

Ethan-ES:
#### Overview:

Let generator support both dynamo version and backend version

  - Use `--generator-dynamo-version v0.8.0` to select the Dynamo release. This affects both the generated backend config version and the default K8s image tag.
  - If `--generator-dynamo-version` is not provided, the default is the first entry in `generator/config/backend_version_matrix.yaml`.
  - If `--generated_config_version` is provided, it overrides the generated backend version, but the default K8s image tag still follows the first entry in `backend_version_matrix.yaml`.

- Added example usage to `aiconfigurator cli default --help`
Screenshot 2026-02-09 at 6 05 08 PM

Will create a separate PR to unify dash and underscores in CLI flags.

### DIFF
--- a/docs/cli_user_guide.md
+++ b/docs/cli_user_guide.md
@@ -283,6 +283,11 @@ By default, we output top 5 configs we have found. You can get the configs and s
 
 `--save_dir DIR` allows you to specify more information such as generating the config for a different version of the backend, say estimating the performance using trtllm 1.0.0rc3 but generate config for 1.0.0rc6. This is allowed and feasible. By passing `--generated_config_version 1.0.0rc6` can give you the right result.
 
+**Generator Dynamo version**
+- Use `--generator-dynamo-version v0.7.1` to select the Dynamo release. This affects both the generated backend config version and the default K8s image tag.
+- If `--generator-dynamo-version` is not provided, the default is the first entry in `generator/config/backend_version_matrix.yaml`.
+- If `--generated_config_version` is provided, it overrides the generated backend version, but the default K8s image tag still follows the first entry in `backend_version_matrix.yaml`.
+
 Use `--generator-config path/to/file.yaml` to provide ServiceConfig/K8sConfig/DynConfig/WorkerConfig/Workers.<role> sections, or add inline overrides via `--generator-set KEY=VALUE`. Examples:
 
 - `--generator-set ServiceConfig.model_path=Qwen/Qwen3-32B-FP8`
--- a/src/aiconfigurator/cli/report_and_save.py
+++ b/src/aiconfigurator/cli/report_and_save.py
@@ -14,7 +14,7 @@ from prettytable import PrettyTable
 
 from aiconfigurator.generator.api import (
     generate_backend_artifacts,
-    get_default_backend_version_entry,
+    get_default_dynamo_version_mapping,
     load_generator_overrides_from_args,
     resolve_backend_version_for_dynamo,
 )
--- a/src/aiconfigurator/cli/report_and_save.py
+++ b/src/aiconfigurator/cli/report_and_save.py
@@ -14,7 +14,6 @@ from prettytable import PrettyTable
 
 from aiconfigurator.generator.api import (
     generate_backend_artifacts,
-    get_default_dynamo_version_mapping,
     load_generator_overrides_from_args,
     resolve_backend_version_for_dynamo,
 )
@@ -543,27 +542,31 @@ def save_results(
                 pareto_df.to_csv(os.path.join(exp_dir, "pareto.csv"), index=False)
 
             # 3. Save the config for this experiment
-            exp_task_config = task_configs[exp_name]
-            dynamo_version = (generator_overrides or {}).get("generator_dynamo_version")
-            
-            # there could be multiple backends in the same experiment if backend == "any" as the result is merged
-            if backend == "any":
+            if backend != "any":
+                exp_task_config = task_configs[exp_name]
+                backend_version_str = exp_task_config.backend_version
+            else:
+                # There could be multiple backends in the same experiment if backend == "any" as the result is merged
                 actual_backend_versions = {
-                    task_config.backend_name: task_config.backend_version
-                    for task_config in task_configs.values()
+                    task_config.backend_name: task_config.backend_version for task_config in task_configs.values()
                 }
                 backend_version_str = ", ".join(
-                    f"({backend_name}){backend_version}" for backend_name, backend_version in actual_backend_versions.items()
+                    f"({backend_name}){backend_version}"
+                    for backend_name, backend_version in actual_backend_versions.items()
                 )
-            else:
-                backend_version_str = exp_task_config.backend_version
+                exp_task_configs = {
+                    f"{exp_name}_{backend_name}": task_configs[f"{exp_name}_{backend_name}"]
+                    for backend_name in actual_backend_versions
+                }
+                # generated backend versions for each backend, empty unless --generator_dynamo_version is provided
+                generated_backend_versions = {}
 
-            # case #1: --generated_config_version is provided            
+            # case #1: --generated_config_version is provided
             if generated_backend_version:
                 effective_generated_version = generated_backend_version
                 logger.warning(
                     "\n" + "=" * 80 + "\n"
-                    "  ⚠️  IMPORTANT: Config Generation Version\n" + "=" * 80 + "\n"
+                    "  🟢  IMPORTANT: Config Generation Version\n" + "=" * 80 + "\n"
                     "  Experiment: %s\n"
                     "  Using generated_config_version: %s\n"
                     "\n"
@@ -574,13 +577,14 @@ def save_results(
                 )
             # case #2: --generator_dynamo_version is provided, generating config matching the dynamo version,
             # but the data used for prediction may not match dynamo version due to imperfect coverage.
-            elif dynamo_version:
+            elif dynamo_version := (generator_overrides or {}).get("generator_dynamo_version"):
                 if backend != "any":
                     try:
                         effective_generated_version = resolve_backend_version_for_dynamo(
                             dynamo_version,
                             exp_task_config.backend_name,
                         )
+                        backend_version_str = f"({exp_task_config.backend_name}){effective_generated_version}"
                     except ValueError as exc:
                         logger.exception(
                             "Failed to resolve backend version for generator_dynamo_version=%s.",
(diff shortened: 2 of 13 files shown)


src/aiconfigurator/generator/api.py
```diff
@@ -33,6 +34,67 @@ def _load_yaml_payload(path: str) -> Any:
...
+
+    Args:
+        dynamo_version: Dynamo release string (e.g., "v0.8.1").
+        backend: Backend name (e.g., "trtllm", "vllm", "sglang").
+        matrix_path: Optional backend version matrix YAML file.
+    """
+    version_key = str(dynamo_version).strip()
+    if not version_key:
+        raise ValueError("dynamo_version must be a non-empty string.")
+    matrix = _load_backend_version_matrix(matrix_path)
```
hhzhang16: When would the user not use the default backend version matrix?

docs/cli_user_guide.md
```diff
@@ -265,6 +265,11 @@ By default, we output top 5 configs we have found. You can get the configs and s
 
 `--save_dir DIR` allows you to specify more information such as generating the config for a different version of the backend, say estimating the performance using trtllm 1.0.0rc3 but generate config for 1.0.0rc6. This is allowed and feasible. By passing `--generated_config_version 1.0.0rc6` can give you the right result.
 
+**Generator Dynamo version**
+- Use `--generator-dynamo-version v0.7.1` to select the Dynamo release. This affects both the generated backend config version and the default K8s image tag.
+- If `--generator-dynamo-version` is not provided, the default is the first entry in `generator/config/backend_version_matrix.yaml`.
+- If `--generated_config_version` is provided, it overrides the generated backend version, but the default K8s image tag still follows the first entry in `backend_version_matrix.yaml`.
```
hhzhang16, line 289: What is this mix of dashes and underscores?

src/aiconfigurator/cli/report_and_save.py
```diff
@@ -453,9 +455,10 @@ def save_results(
 
             # 3. Save the config for this experiment
             exp_task_config = task_configs[exp_name]
-            effective_generated_version = generated_backend_version or exp_task_config.backend_version
-
+            default_dynamo_version, default_backend_versions = get_default_backend_version_entry()
+            dynamo_version = (generator_overrides or {}).get("generator_dynamo_version") or default_dynamo_version
```
hhzhang16: What if you do `dynamo_version = (generator_overrides or {}).get("generator_dynamo_version")` and have it be none if not provided, then you can check for that later

src/aiconfigurator/cli/report_and_save.py
```diff
@@ -467,7 +470,39 @@ def save_results(
                     exp_name,
                     generated_backend_version,
                 )
+            elif dynamo_version:
+                if dynamo_version == default_dynamo_version:
+                    backend_key = exp_task_config.backend_name
+                    effective_generated_version = default_backend_versions.get(backend_key)
+                else:
+                    effective_generated_version = None
```
hhzhang16: This can be the check for None/whether the user passes it in -- if not, use default dynamo version and lines 475-476

src/aiconfigurator/cli/report_and_save.py
```diff
@@ -467,7 +470,39 @@ def save_results(
...
+                    "  Using generator_dynamo_version: %s\n"
+                    "  Resolved backend version: %s\n"
+                    "\n"
+                    "  Config formats differ across backend releases. Ensure the Dynamo version\n"
+                    "  matches your deployment target!\n" + "=" * 80,
+                    exp_name,
+                    dynamo_version,
+                    effective_generated_version,
+                )
             else:
```
hhzhang16, line 614: When does this else case get triggered (in high level terms)?

src/aiconfigurator/generator/rendering/schemas.py
```diff
@@ -36,6 +38,30 @@ def _load_schema_inputs(schema_path: str) -> list[dict[str, Any]]:
...
+        matrix = {}
+    _MATRIX_CACHE[path] = matrix
+    return matrix
+
+
+def get_default_dynamo_version() -> Optional[str]:
+    matrix = _load_backend_version_matrix(_BACKEND_VERSION_MATRIX_FILE)
+    if not matrix:
+        return None
+    return str(next(iter(matrix.keys())))
```
hhzhang16: Should default be the first key?

src/aiconfigurator/generator/api.py
```diff
@@ -33,6 +34,67 @@ def _load_yaml_payload(path: str) -> Any:
         return yaml.safe_load(fh) or {}
 
 
+def _load_backend_version_matrix(
```
hhzhang16: What makes this different from `_load_backend_version_matrix` in schemas.py?

src/aiconfigurator/generator/api.py
```diff
@@ -33,6 +34,67 @@ def _load_yaml_payload(path: str) -> Any:
...
+    if backend_version is None:
+        supported_backends = ", ".join(sorted(entry.keys()))
+        raise ValueError(
+            f"No backend version mapping for backend '{backend_key}' in dynamo '{version_key}'. "
+            f"Supported backends: {supported_backends or 'none'}."
+        )
+    return str(backend_version)
+
+
+def get_default_backend_version_entry(
```
hhzhang16: What makes this different than `get_default_dynamo_version` in schema.py?

src/aiconfigurator/generator/module_bridge.py
```diff
@@ -141,16 +141,22 @@ def _build_worker_params(prefix: str, extra_overrides: dict | None) -> tuple[dic
...
-    k8s_cfg = apply_defaults("K8sConfig", k8s_cfg, backend=backend_name)
-
     dyn_cfg = {
         "mode": task_config.serving_mode,
     }
     dyn_cfg = _deep_merge(dyn_cfg, overrides.get("DynConfig"))
     dyn_cfg = apply_defaults("DynConfig", dyn_cfg, backend=backend_name)
 
+    generator_dynamo_version = overrides.get("generator_dynamo_version")
+    k8s_cfg = {}
```
hhzhang16, line 168: Why set this to empty if you're going to update it in the next line?

docs/cli_user_guide.md
simone-chen: +1, it looks in AIC cli default parser the args are underscores, but in generator is using dashes. 
Dash might be the more standard way but we need to unify it. 
This could end in a separate PR.

tianhaox: let's make it in a seperate pr to unify it to dash.

simone-chen: Created https://linear.app/nvidia/issue/TRTC-266/cli-unify-underscores-and-dashes-in-cli-flags

src/aiconfigurator/generator/rendering/schemas.py
simone-chen: Yes, this line is the same as `str(matrix.keys()[0])`

src/aiconfigurator/generator/api.py
simone-chen: When `--generator_dynamo_version` is passed in the CLI.

src/aiconfigurator/generator/module_bridge.py
simone-chen: I think the goal here is for readability and flexibility. In lines above there are other `*_cfg` like `model_cfg` and `dyn_cfg`. They are all initialized with some values and then followed by a `_deep_merge` for value overrides.

Albeit functionally unnecessary here, the format aligns with the other code and makes it easier for us to add default values in the future.

src/aiconfigurator/generator/api.py
simone-chen: Makes sense. Refactored by moving `load_backend_version_matrix` to utils.py.

src/aiconfigurator/generator/api.py
simone-chen: This one returns both dynamo version and the backend versions associated with the dynamo version. But I agree there is some code duplications here.

src/aiconfigurator/cli/report_and_save.py
simone-chen: Makes sense. Refactored.

src/aiconfigurator/cli/report_and_save.py
simone-chen: Added comment to describe these 3 branches.
This else case is triggered when neither `--generated_config_version` or `--generator-dynamo-version` is passed.  AIC will use the backend version as the generated config version.

simone-chen (approved):

tests/unit/cli/test_argument_parsing.py
```diff
@@ -110,6 +110,7 @@ def test_default_values_are_set(self, cli_parser):
         assert args.debug is False
         assert args.decode_system is None
         assert args.generated_config_version is None
+        assert args.generator_dynamo_version is None
```
xutizhou, line 113: Currently only argument parsing is tested. It would be good to have at least one test that passes --generator-dynamo-version end-to-end and asserts the resulting output YAML contains the expected image tags (e.g., nvcr.io/.../tensorrtllm-runtime:v0.8.0). This would catch any breakage in the matrix lookup → Jinja rendering → output pipeline.

Arsene12358 (approved):

src/aiconfigurator/generator/utils.py
```diff
@@ -40,3 +46,80 @@ def coerce_int(value: Optional[Any]) -> Optional[int]:
...
+    dynamo_version: str,
+    backend: str | None = None,
+    matrix_path: str = DEFAULT_BACKEND_VERSION_MATRIX_PATH,
+) -> str:
+    """
+    Given a Dynamo (generator) version, look up the corresponding backend version for a specified backend.
+
+    Parameters:
+        dynamo_version (str): The target Dynamo generator release (e.g., "v0.8.1").
+        backend (str | None): Name of the backend to look up ("trtllm", "vllm", "sglang", or "any").
```
hhzhang16, line 94: We're changing this to "auto", can you change it within your code as well?

hhzhang16 (approved):

src/aiconfigurator/generator/utils.py
hhzhang16: Either in this MR or in a following one, but would be best to get it in before code freeze!

tests/unit/cli/test_argument_parsing.py
simone-chen: Yes that's a great idea. Will add in a separate PR. Thanks for the suggestion!

src/aiconfigurator/generator/utils.py
simone-chen: Sure, will open a separate PR.

PR #334, opened 2026-02-05
feat: generator will generate benchmark command

Ethan-ES:
#### Overview:

Generator will generate benchmark command.

Note: the current concurrency is set to a power of 2, and the generator needs to receive the concurrency parameter from upstream.


src/aiconfigurator/generator/rule_plugin/benchmark/sglang.rule
```diff
@@ -1,3 +1,4 @@
+BenchConfig.estimated_concurrency = ((WorkerConfig.agg_workers or 0) * (params.agg.max_batch_size or 0) if DynConfig.mode == 'agg' else (WorkerConfig.decode_workers or 0) * (params.decode.max_batch_size or 0))
```
tianhaox, line 1: how we handle replica here? same questions for other two backends

tianhaox: checked the results, seems working. closing

src/aiconfigurator/generator/rendering/engine.py
```diff
@@ -486,6 +512,38 @@ def prepare_template_context(param_values: dict[str, Any], backend: str) -> dict
...
+    context["bench_model"] = bench_config.get("model")
+    context["bench_tokenizer"] = bench_config.get("tokenizer")
+    context["bench_endpoint_type"] = bench_config.get("endpoint_type")
+    context["bench_endpoint_url"] = bench_config.get("endpoint_url")
+    context["bench_isl"] = _safe_int(bench_config.get("isl"), 0)
+    context["bench_osl"] = _safe_int(bench_config.get("osl"), 0)
+    context["bench_isl_stddev"] = _safe_int(bench_config.get("isl_stddev"), 0)
+    context["bench_osl_stddev"] = _safe_int(bench_config.get("osl_stddev"), 0)
+    context["bench_estimated_concurrency"] = bench_estimated_concurrency
+    context["bench_ui"] = bench_config.get("ui")
```
Arsene12358: where are these variables actually used? it seems in the backend_templates we always use the BenchConfig dict instead of these variables

src/aiconfigurator/generator/config/backend_templates/benchmark/k8s_bench.yaml.j2
```diff
@@ -0,0 +1,49 @@
...
+              {% set eff_low = (BenchConfig.estimated_concurrency * 0.95) | int %}
+              {% set eff_high = (BenchConfig.estimated_concurrency * 1.05) | int %}
+              concurrency_array=({{ (base_concurrency + [eff_low, eff, eff_high]) | join(' ') }})
+              {% else %}
+              concurrency_array=({{ base_concurrency | join(' ') }})
+              {% endif %}
+              for concurrency in "${concurrency_array[@]}"; do
+                echo "Run concurrency: $concurrency"
+                aiperf profile \
+                  -m {{ (BenchConfig.model.split('/')[-1] if BenchConfig.model else '') }} \
```
Arsene12358: if the user don't set ServiceConfig.served_model_name explicitly when using cli to generate these artifacts, I think the benchmark will fail due to model name mismatch here

tianhaox: since the model name is all hf-model now, is this still happening?

Arsene12358: An example perhaps
`aiconfigurator cli default --model Qwen/Qwen3-32B-FP8 \
	--total_gpus 8 --system h200_sxm --backend trtllm \
	--isl 4000 --osl 500 --ttft 1000 --tpot 20 \
	--generated_config_version 1.2.0rc6 \
	--generator-set K8sConfig.k8s_namespace=yimingl-aic-test \
  --generator-set ServiceConfig.model_path=Qwen/Qwen3-32B-FP8 \
	--save_dir results/`

with the above command, My understanding of the current logic is **if the user don't set ServiceConfig.served_model_name**, then the server will have a served-model-name "None", but the aiperf profile here will fallback to use **Qwen3-32B-FP8** derived from --model Qwen/Qwen3-32B-FP8. Thus resulting in a failure

Ethan-ES: Fixed, commands without ServiceConfig.served_model_name will work now

PR #320, opened 2026-02-01
feat: Trtllm wideep pipeline

liyuanzhe1991:
#### Overview:

Add TensorRT-LLM WideEP pipeline support for DeepSeek V3/R1 models, enabling EPLB-aware performance prediction with automatic kernel selection.

#### Details:

**Data Loading:**
- Add `load_wideep_moe_compute_data()` to load MoE EPLB compute data from `moe_eplb_perf.txt`
- Add `load_wideep_alltoall_data()` to load All2All communication data from `wideep_alltoall_perf.txt`
- Support `kernel_source`, `num_nodes`, and `num_slots` dimensions in data structure

**SDK Pipeline:**
- Add `TrtllmWideEPDeepSeekModel` class for WideEP-enabled DeepSeek models
- Add `query_wideep_moe_eplb_compute()` for EPLB MoE query with automatic kernel selection
- Add `query_wideep_alltoall()` for All2All query with automatic kernel selection
- Add `wideep_num_slots` config parameter for EPLB slots
- Add WideEP configuration validation
- Update `MoE` and `MoEDispatch` operations to support WideEP path

**CLI:**
- Add `--wideep_num_slots` parameter

#### Where should the reviewer start?

1. `src/aiconfigurator/sdk/models.py` - `TrtllmWideEPDeepSeekModel` class
2. `src/aiconfigurator/sdk/perf_database.py` - `load_wideep_*` and `query_wideep_*` methods
3. `src/aiconfigurator/sdk/operations.py` - WideEP path in `MoE` and `MoEDispatch`

PR #335, opened 2026-02-06
feat: add Trtllm wideep moe collector

liyuanzhe1991:
#### Overview:

Add a dedicated TensorRT-LLM WideEP MoE compute-only collector that benchmarks MoE computation performance excluding All-to-All communication. This collector supports three EPLB modes and uses accurate simulation mode with DP-split and rank0 filtering.

#### Details:

**New file: `collector/trtllm/collect_wideep_moe_compute.py`**
- `WideEPMoEComputeSimulator`: Simulates WideEPMoE computation on a single GPU using the same kernel selection logic (MoEOpSelector) as TensorRT-LLM's WideEPMoE
- Supports three EPLB configurations:
  - EPLB OFF: `num_slots = num_experts` (standard routing)
  - EPLB ON (baseline): `num_slots = num_experts` (load-balanced routing)
  - EPLB ON (redundant): `num_slots = 288` (redundant expert)
- Accurate WideEP simulation mode: router uses `dp_num_tokens = num_tokens / ep_size` (single DP rank), MoE computes only tokens routed to EP rank 0
- Scope: DeepSeek-V3 model only, tp=1, ep>1, power_law distribution only

**Modified: `collector/helper.py`**
- Added `power_law_logits_v3()`: generates power-law distributed router logits with EPLB slot assignment support
- Added helper functions for greedy expert assignment and EPLB simulation

**Modified: `collector/collect.py`**
- Registered `moe_eplb` and `trtllm_moe_wideep` collection types
- Added corresponding entries to the CLI ops list

- `collector/trtllm/collect_wideep_moe_compute.py` — the main new collector, start with `WideEPMoEComputeSimulator` class and `run_wideep_moe_compute()` function
- `collector/helper.py` — the `power_law_logits_v3()` function for EPLB-aware logit generation

#### Related Issues:

- Relates to #320 (TrtLLM WideEP pipeline support)


collector/trtllm/collect_wideep_moe_compute.py
```diff
@@ -0,0 +1,1043 @@
...
+# =============================================================================
+# WideEPMoE Compute Simulator - simulates WideEPMoE computation on single GPU
+# =============================================================================
+class WideEPMoEComputeSimulator(nn.Module):
+    """
+    Simulates WideEPMoE's computation path on SINGLE GPU.
+
+    Uses the SAME kernel selection logic as WideEPMoE (MoEOpSelector.select_op).
+
+    模拟 EP size > 1 场景:
```
tianhaox: let's write comment in english

collector/collect.py
```diff
@@ -675,6 +674,26 @@ def collect_trtllm(num_processes: int, ops: list[str] | None = None):
             if v.startswith(("1.1.0", "1.2.0", "1.3.0"))
             else None,
         },
+        # MOE with EPLB (Expert Parallel Load Balancer) collection
+        {
```
tianhaox, line 687: do you think eplb can be part of wideep?

liyuanzhe1991: Yes, in fact, EPLB here is just a post-processing step for the power law – it’s simply a plugin for the MoE collector. This standalone collector was largely created because we revised the previous collection mode: we separated each individual OP within the module and assigned to each OP an input that is as close to real-world scenarios as possible.

PR #338, opened 2026-02-08
feat: infer quantization from model info

tianhaox:
#### Overview:

in our previous design, in order to do more quantization study, the quant mode of each component of the model is not derived from the model config. As we gradually connect our model definition with huggingface id, we can now infer the quantization from the model config and hf quant config.

This pr will infer the quantization mode of each model and use that to set up the simulation.
The previous quant mode setting will still take effect but will be treated as overrides options.

In future, if you want to do simulation directly and don't need to override the quantization, you no longer need to set that in the yaml definition. A better way to trigger low-precision inference is to feed the quantized model on huggingface hub.

e.g.,
```
aiconfigurator cli default --model nvidia/DeepSeek-V3.1-NVFP4 --total_gpus 128 --system b200_sxm --isl 8192 --osl 1024 --ttft 10000 --tpot 50
```

webapp needs to be redesigned to align the usage. Will implement that in another PR.

- [x] update previous commands and yamls
- [x] doc
- [x] tests
- [x] more robust tests


src/aiconfigurator/sdk/models.py
```diff
@@ -46,6 +44,96 @@ def _architecture_to_model_family(architecture: str) -> str:
...
+        "Model config (final quant modes): gemm=%s moe=%s kvcache=%s fmha=%s comm=%s",
+        model_config.gemm_quant_mode,
+        model_config.moe_quant_mode,
+        model_config.kvcache_quant_mode,
+        model_config.fmha_quant_mode,
+        model_config.comm_quant_mode,
+    )
+
+
 def _architecture_to_model_family(architecture: str) -> str:
```
Arsene12358: Not related to the feature of this PR, but can we remove this duplicated function? it's already defined at line 30 above

tianhaox: great catch. done

Arsene12358 (approved):

simone-chen (approved):

PR #351, opened 2026-02-11
fix: Speed up slow trtllm moe collector

ilyasher:
1. Fix removed "rank" kwarg to autotune()
2. Parallelize file operations where possible
3. Filter out some unsupported test cases
4. Replace for loops in balanced_logits() with vectorized torch ops
5. Shuffle test cases to increase number of MOE autotune cache hits

Speeds up moe collection from 8+ hours to ~2 hours on gb300

trtllm.moe: 100%|█████████▉| 2245/2247 [2:16:53<00:07,  3.66s/it]

### DIFF
--- a/collector/helper.py
+++ b/collector/helper.py
@@ -663,7 +663,7 @@ def balanced_logits(num_tokens, num_experts, topk):
     stride = math.ceil(num_experts / topk)
 
     token_indices = torch.arange(num_tokens).unsqueeze(1)  # [num_tokens, 1]
-    topk_indices = torch.arange(topk).unsqueeze(0)         # [1, topk]
+    topk_indices = torch.arange(topk).unsqueeze(0)  # [1, topk]
 
     if num_tokens >= stride:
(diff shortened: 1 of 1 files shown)


collector/trtllm/collect_moe.py
```diff
@@ -472,21 +503,12 @@ def run_forward_pass():
...
-
-    for _ in range(5):
-        gc.collect()
-        torch.cuda.empty_cache()
-    AutoTuner.get().clear_cache()
-
     # Exit the worker process after completing MOE task to ensure complete resource cleanup
     # This forces OS to reclaim all GPU memory, CUDA context, and other resources
-    import sys
+    if AIC_RESTART_WORKER:
```
YijiaZhao, line 509: how to reclaim GPU MEM without sys.exit if AIC_RESTART_WORKER = 0?

ilyasher: It runs gc_collect() (which includes torch.cuda.empty_cache()) at the beginning of the function

PR #355, opened 2026-02-11
fix: fix validator service key mismatch

Ethan-ES:
#### Overview:

VLLMWorker -> VllmDecodeWorker
VLLMPrefillWorker -> VllmPrefillWorker
VLLMDecodeWorker -> VllmDecodeWorker


tools/generator_validator/backend/vllm.py
Arsene12358: Shall we change this to `worker = services.get(service_key or "VllmDecodeWorker", {})`  as well?

Ethan-ES: Fixed, thx.

PR #357, opened 2026-02-11
feat: Enahnce generator to generate aiperf benchmark command

Arsene12358:
#### Overview:

- Generator now will be able to generate benchmark command.
Note: the current concurrency is set to a power of 2, and the generator needs to receive the concurrency parameter from upstream.

- use “-” instead of “_”  for cli arguments


src/aiconfigurator/cli/main.py
simone-chen, line 240: These are still using underscores which is fixed in https://github.com/ai-dynamo/aiconfigurator/pull/345

Arsene12358: yeah, I think we should also cherry-pick #345. Otherwise it would not make sense to use "-" in the relase/v0.7.0

Arsene12358: maybe we can have a separate cherry-pick for #345 afterwards, what do you think @simone-chen ?

simone-chen: Sure, both ways work for me as long as they get in to 0.7.0 👍

Arsene12358: In that case, I will resolve this conversation for now

Arsene12358:
I will include #345 in this cp, I think they may as well go hand in hand. you can have another look once it's ready @simone-chen

simone-chen (approved):

Ethan-ES (approved):

PR #361, opened 2026-02-12
feat: hook to dynamo planner profiler's config gen

tedzhouhk:
support both path for minimum intrusion to AIC code, default behavior remains the same.

### DIFF
--- a/src/aiconfigurator/generator/rendering/engine.py
+++ b/src/aiconfigurator/generator/rendering/engine.py
@@ -99,7 +99,7 @@ def render_backend_templates(
     backend: str,
     templates_dir: Optional[str] = None,
     version: Optional[str] = None,
-    use_dynamo_generator: bool = True,
+    use_dynamo_generator: bool = False,
 ) -> dict[str, str]:
     """
     Render templates for a specific backend with version-specific template selection.


src/aiconfigurator/generator/rendering/engine.py
```diff
@@ -31,11 +31,75 @@
...
+    Args:
+        param_values: The full AIC parameter dict (ServiceConfig, K8sConfig, params, etc.)
+        backend: Backend name (e.g., 'vllm', 'sglang', 'trtllm')
+        context: The template context dict with pre-computed CLI args and worker info
+
+    Returns:
+        Rendered k8s_deploy.yaml content as a string
+    """
+    # for now, we import the config modifiers from dynamo, will migrate to aiconfigurator later
+    from dynamo.profiler.utils.config_modifiers import CONFIG_MODIFIERS
```
simone-chen, line 55: Should we add dynamo to `pyproject.toml` as a dependency?

The version backward compatibility could be an issue if we use this import, as we can only pin to 1 dynamo version as dependency right? But since it's a temporary solution I think we should be good for now.

src/aiconfigurator/generator/rendering/engine.py
```diff
@@ -31,11 +31,75 @@
...
+    if model_cache:
+        kwargs["pvc_name"] = model_cache
+        kwargs["pvc_mount_path"] = "/workspace/model_cache"
+        kwargs["model_path"] = k8s_cfg.get("k8s_hf_home") or "/workspace/model_cache"
+
+    config_dict = modifier.build_dgd_config(**kwargs)
+    return yaml.dump(config_dict, sort_keys=False)
+
+
 def render_backend_templates(
```
simone-chen, line 97: Will DGDR directly use this API with `use_dynamo_generator==True`?

tedzhouhk: Yes, the shim python layer in dynamo will call this

src/aiconfigurator/generator/rendering/engine.py
tedzhouhk: this will introduce tons of dependencies (dynamo is too heavy) and we will likely be AIC release <> dynamo release deadlocked. 

There is already a try catch block wrapping this function I think we should be good for now.

PR #373, opened 2026-02-13
feat: expose real-gpu enumeration logic

tedzhouhk:
## Summary

Adds a new `enumerate_profiling_configs()` API that enumerates disaggregated prefill/decode parallelization candidates and returns aggregated DGD configs for real-silicon profiling sweeps. This is the primary entry point for Dynamo's profiler to obtain sweep candidates from AIC.

Additionally, unifies the PVC parameter naming across the codebase (`k8s_pvc_name`, `k8s_pvc_mount_path`, `k8s_model_path_in_pvc`).

Test locally and should not break any existing functionality.

### New file: `src/aiconfigurator/generator/enumerate.py`

Core enumeration module with two public functions:

- **`check_model_hardware_support()`** - Checks if AIC supports a model x hardware combo for disagg mode
- **`enumerate_profiling_configs()`** - Main entry point that:
  1. Checks AIC disagg support for the model x hardware combination
  2. Computes memory-based minimum GPUs per engine (reuses `_calculate_min_tp` from `naive.py`)
  3. Enumerates valid parallelization configs (TP / TEP / DEP, no PP) using AIC's existing `enumerate_parallel_config` with a new `real_silicon_sweep=True` filter
  4. For each config, uses AIC's rendering engine (with rule plugins) to compute CLI args, then builds an aggregated DGD via dynamo's `build_dgd_config` + `convert_config`
  5. Applies AIC heuristics for `max_batch_size` / `max_num_tokens` when the combo is supported; otherwise uses safe defaults
  6. Returns `(prefill_configs, decode_configs)` - two lists of aggregated DGD config dicts

### Modified: `src/aiconfigurator/sdk/utils.py`

Added parameters to `enumerate_parallel_config()`:
- `real_silicon_sweep` (default `False`) - When `True`: forces `pp_list=[1]`, filters by `min_num_gpus`/`max_num_gpus`, and for MoE models only allows pure TEP, pure DEP, or pure TP (no mixed configs)
- `min_num_gpus` / `max_num_gpus` - GPU count bounds (only effective when `real_silicon_sweep=True`)
- `allow_moe_pure_tp` (default `True`) - Set to `False` for MLA+MoE models (e.g. DeepSeek) to exclude pure TP

All new parameters have backward-compatible defaults. Existing callers are unaffected.

### Modified: `src/aiconfigurator/sdk/task.py`

Extracted `build_disagg_parallel_lists()` as a standalone module-level function from `_disagg_defaults_layer()`. This is the single source of truth for TP/PP/DP/MoE-TP/MoE-EP search-space lists, shared by both the existing CLI sweep and the new enumeration.

`_disagg_defaults_layer()` now delegates to `build_disagg_parallel_lists()` and adds runtime metadata (system_name, backend_name, backend_version) afterwards. Behavior is identical.

### PVC param unification (multiple files)

Unified PVC naming across the codebase:

| New name | Old name(s) | Meaning |
|---|---|---|
| `k8s_pvc_name` | `k8s_model_cache` | PVC claim name |
| `k8s_pvc_mount_path` | (hardcoded `/workspace/model_cache`) | Where PVC mounts in pod |
| `k8s_model_path_in_pvc` | `k8s_hf_home`, `k8s_pvc_model_path` | Relative model path inside PVC |

Final model path = `{k8s_pvc_mount_path}/{k8s_model_path_in_pvc}`

**Backward compatible**: All code reads new names first, falls back to old names. Jinja2 templates still receive `k8s_model_cache` and `k8s_hf_home` via context aliases.

Files changed:
- `generator/rendering/engine.py` - `_generate_k8s_via_dynamo` and `prepare_template_context`
- `generator/aggregators.py` - PVC variable handling
- `generator/config/deployment_config.yaml` - Added new keys, kept old as aliases
- `cli/main.py` - Updated CLI help example
- `webapp/components/profiling/sdk/config.py` - Updated default config key

## MoE Parallelization Strategy

For `real_silicon_sweep=True`, the MoE filter enforces pure parallelization strategies:

| Strategy | Conditions | Example |
|---|---|---|
| Pure TEP | `tp > 1, dp == 1, moe_tp == 1, moe_ep > 1` | `tp=8, moe_ep=8` |
| Pure DEP | `tp == 1, dp > 1, moe_tp == 1, moe_ep > 1` | `dp=32, moe_ep=32` |
| Pure TP | `tp > 1, dp == 1, moe_tp > 1, moe_ep == 1` | `tp=8, moe_tp=8` |

- **MLA+MoE models** (e.g. DeepSeek-R1): Only TEP + DEP (pure TP disabled)
- **GQA+MoE models** (e.g. Qwen3-235B-A22B): TEP + DEP + pure TP

WideEP is auto-enabled for all MoE models. For GQA+MoE, a second enumeration pass with `enable_wideep=False` picks up pure TP configs.

## Backward Compatibility

- `enumerate_parallel_config()`: All new params have safe defaults (`real_silicon_sweep=False`, etc.). Existing callers in `task.py` (3 sites) and `webapp/events/event_fn.py` (4 sites) are unaffected.
- `_disagg_defaults_layer()`: Refactored to delegate to `build_disagg_parallel_lists()`. Output dict is structurally identical.
- PVC params: Old names (`k8s_model_cache`, `k8s_hf_home`) still accepted via fallback logic.
- New files (`enumerate.py`) are purely additive.

## How to review

1. **`src/aiconfigurator/sdk/utils.py`** - Check the `real_silicon_sweep` filter block (lines ~103-135). Verify the MoE whitelist logic matches the strategy table above.

2. **`src/aiconfigurator/sdk/task.py`** - Verify `build_disagg_parallel_lists()` produces identical output to the old inline code in `_disagg_defaults_layer()`. The parallel lists should be byte-for-byte the same.

3. **`src/aiconfigurator/generator/enumerate.py`** - Main new code. Key things to check:
   - `_build_param_values()` - Does it construct param_values correctly for the rendering engine?
   - `_build_agg_dgd_for_candidate()` - Does the flow (render CLI args -> build_dgd_config -> update_model_from_pvc -> convert_config) match `engine.py`'s approach?
   - `enumerate_profiling_configs()` - Is the min/max GPU logic correct? Is the GQA+MoE second-pass for pure TP correct?

4. **PVC changes** - Verify backward compat by checking that old param names still work (fallback chains in `engine.py` and `aggregators.py`).


src/aiconfigurator/cli/main.py
simone-chen: Could you also help change the `--model_path` to `--model` in this PR? I think it's a typo unfixed 😱

src/aiconfigurator/sdk/task.py
```diff
@@ -94,6 +94,145 @@ def _ensure_munch(obj: dict | DefaultMunch | Munch) -> DefaultMunch:
...
+    """Build the TP/PP/DP/MoE-TP/MoE-EP search-space lists for disagg enumeration.
+
+    This is the single source of truth shared by :class:`TaskConfigFactory` (for the
+    default CLI sweep) and the profiling enumeration in
+    ``aiconfigurator.generator.enumerate``.
+
+    Args:
+        backend_name: Backend identifier (``"trtllm"``, ``"sglang"``, ``"vllm"``).
+        prefill_system: System name for the prefill worker (e.g. ``"h200_sxm"``).
+        decode_system: System name for the decode worker.
```
simone-chen, line 114: Do we want to support heterogenous backends?

src/aiconfigurator/sdk/task.py
```diff
@@ -94,6 +94,145 @@ def _ensure_munch(obj: dict | DefaultMunch | Munch) -> DefaultMunch:
     return DefaultMunch.fromDict(obj, DefaultMunch)
 
 
+def build_disagg_parallel_lists(
```
tianhaox, line 97: is this duplicate with line 416?

tedzhouhk: sorry, cursor messed it up solving a merge conflict, should be good now

src/aiconfigurator/generator/aggregators.py
```diff
@@ -67,17 +67,36 @@ def collect_generator_params(
...
-    k8s_hf_home = _hf_home_raw.strip() if isinstance(_hf_home_raw, str) else ""
+    _pvc_mount_raw = k8s.get("k8s_pvc_mount_path")
+    k8s_pvc_mount_path = (
+        _pvc_mount_raw.strip()
+        if isinstance(_pvc_mount_raw, str) and _pvc_mount_raw.strip()
+        else "/workspace/model_cache"
+    )
+
+    _model_in_pvc_raw = k8s.get("k8s_model_path_in_pvc") or k8s.get("k8s_pvc_model_path")
+    k8s_model_path_in_pvc = _model_in_pvc_raw.strip() if isinstance(_model_in_pvc_raw, str) else ""
```
simone-chen: Should we do `_model_in_pvc_raw.strip(" /")` to strip both space and slash - in case a path like `/Qwen3-32B` is pased as `K8sConfig.k8s_model_path_in_pvc`?

src/aiconfigurator/generator/enumerate.py
```diff
@@ -0,0 +1,633 @@
...
+    Args:
+        model_path: HuggingFace model path or local path.
+        system: System name (GPU type), e.g. ``"h200_sxm"``.
+        backend: Backend name (``"trtllm"``, ``"sglang"``, ``"vllm"``).
+        backend_version: Optional backend database version.
+
+    Returns:
+        ``True`` if disaggregated mode is supported, ``False`` otherwise.
+    """
+    from aiconfigurator.sdk.utils import get_model_config_from_model_path
```
simone-chen: Out of curiosity, why do we use function-scope import here? It looks `get_model_config_from_model_path` is also used elsewhere in this file.

src/aiconfigurator/sdk/utils.py
```diff
@@ -90,6 +105,37 @@ def enumerate_parallel_config(
                 if tp * pp in num_gpu_list:
                     parallel_config_list.append([tp, pp, 1, 1, 1])
 
+    # Apply real silicon sweep filters to reduce sweep time on real silicon
```
simone-chen, line 163: Should we create a separate function for real_silicon sweep? The original `enumerate_parallel_config` can be reused.

src/aiconfigurator/generator/enumerate.py
```diff
@@ -0,0 +1,633 @@
...
+        "backend": backend,
+    }
+
+
+# ---------------------------------------------------------------------------
+# Step 3: Build a single aggregated DGD for one candidate
+# ---------------------------------------------------------------------------
+
+
+def _build_agg_dgd_for_candidate(
```
simone-chen, line 181: Nit(n00b): calling it agg_dgd can be a bit confusing as it may sound like it's for creating DGD for aggregated serving, but essentially this function is used for DGD generate for disagg mode right?

src/aiconfigurator/cli/main.py
tedzhouhk: done

src/aiconfigurator/sdk/task.py
tedzhouhk: not for now, we can add later

src/aiconfigurator/generator/aggregators.py
tedzhouhk: good catch, updated. also updated `engine.py`

src/aiconfigurator/generator/enumerate.py
tedzhouhk: good catch. Most AI coding agents like to lazy import, my cursor rule is set in dynamo and not applied to AIC repo

src/aiconfigurator/sdk/utils.py
tedzhouhk: this flag default false and should not affect this function. I've moved the filtering logic to a standalone function to make it cleaner.

src/aiconfigurator/generator/enumerate.py
tedzhouhk: this part is indeed confusing. To profiler over P/D engine independently, we need an "agg" DGD that simulate a P/D engine in a disagg deploy. Let me update comment

simone-chen: Thanks!

PR #386, opened 2026-02-15
feat: Add picking modes and expose standalone picking API

tedzhouhk:
# PR: Add picking modes and expose standalone picking API

## Summary

Adds three configuration picking modes and exposes them as standalone functions callable from external scripts (e.g. real-GPU sweep pipelines). Also fixes a bug in the Dynamo DGD generator for agg mode.

### New file: `src/aiconfigurator/sdk/picking.py`

Standalone picking functions for configuration selection, decoupled from `TaskConfig`. Three public functions:

- **`pick_default(pareto_df, total_gpus, serving_mode, ...)`** -- Maximize throughput for a fixed GPU budget under SLA. Same logic as the existing default CLI path.
- **`pick_load_match(pareto_df, serving_mode, target_request_rate=..., target_concurrency=..., ...)`** -- Minimize GPUs for a target load under SLA. Computes `replicas_needed` per config row, ranks by `total_gpus_needed`. When GPU budget is exceeded, switches to maximizing throughput and reports `load_served_pct`.
- **`pick_autoscale(prefill_df, decode_df, target_ttft, target_tpot, ...)`** -- Pick prefill and decode engines independently (no rate matching), return disagg config with 1 replica each. Designed for external autoscaler integration.
- **`build_disagg_summary_dict()`** -- Standalone helper to combine prefill/decode dicts into a disagg summary row (extracted from `DisaggInferenceSession._get_disagg_summary_dict`).

All functions accept perf DataFrames directly (ColumnsAgg/ColumnsDisagg/ColumnsStatic schemas) without requiring a `TaskConfig`. Usage:

```python
from aiconfigurator.sdk.picking import pick_default, pick_load_match, pick_autoscale

result = pick_default(pareto_df=my_agg_df, total_gpus=8, serving_mode="agg", target_tpot=30.0)
result = pick_load_match(pareto_df=my_df, serving_mode="disagg", target_tpot=30.0, target_request_rate=10.0)
result = pick_autoscale(prefill_df=my_p_df, decode_df=my_d_df, target_ttft=2000.0, target_tpot=30.0)
```

### Modified: `src/aiconfigurator/sdk/task.py`

- Added `autoscale: bool = False` to `TaskRunner.run()` and `run_disagg()`. When `True`, skips rate matching and picks P/D independently.

### Modified: `src/aiconfigurator/sdk/inference_session.py`

- Extracted `get_worker_candidates()` as a public method on `DisaggInferenceSession` (was nested `_get_summary_df`).
- `_get_disagg_summary_dict` now delegates to `picking.build_disagg_summary_dict`.
- `_pick_autoscale` now delegates to `picking.pick_autoscale`.
- Rate-matching constants moved to `picking.py` as single source of truth, imported here.
- Added `autoscale` and `target_tpot` params to `find_best_disagg_result_under_constraints()`.

### Modified: `src/aiconfigurator/sdk/pareto_analysis.py`

- `_get_best_configs_under_constraint()` now returns closest-to-target configs instead of empty when SLA can't be met.
- Added `get_best_configs_for_target_load()` for load-match picking: computes replicas from `seq/s` or `concurrency` column, ranks by `total_gpus_needed`. Handles GPU-capped scenario.
- Added `autoscale` and `target_tpot` kwargs passthrough in `disagg_pareto()`.

### Modified: `src/aiconfigurator/cli/main.py`

- `_execute_task_configs()` now returns a 5-tuple (added `best_latencies` dict with `ttft`/`tpot`/`request_latency`/`replicas_needed`/`total_gpus_needed`/`load_served_pct` from the rank-1 config per experiment).
- Added `target_request_rate`, `target_concurrency`, `max_total_gpus` params for load-match mode.

### Modified: `src/aiconfigurator/cli/utils.py`

- `process_experiment_result()` refactored to delegate to `picking.pick_default` / `picking.pick_load_match`.

### Modified: `src/aiconfigurator/cli/api.py`

- `CLIResult` dataclass: added `best_latencies` field.
- `_execute_and_wrap_result()` updated for 5-tuple unpacking.

### Modified: `src/aiconfigurator/cli/report_and_save.py`

- `log_final_summary()`: added load-match summary (target load, GPUs needed per mode, `load_served_pct` warning when capacity exceeded).
- `save_results()`: added `use_dynamo_generator` param, threaded to `generate_backend_artifacts()`.

### Modified: `src/aiconfigurator/generator/api.py`

- `generate_backend_artifacts()`: added `use_dynamo_generator: bool = False`, threaded to `render_backend_templates()`.

### Modified: `src/aiconfigurator/generator/module_bridge.py`

- `task_config_to_generator_config()`: added `num_gpus_per_node: int | None = None` override. When `None`, auto-detects from system config (existing behavior).

### Modified: `dynamo/components/src/dynamo/profiler/utils/config_modifiers/protocol.py`

- Fixed `_apply_agg_worker()`: added fallback to find worker service by `componentType` when `subComponentType` lookup fails (agg templates use generic names like `TRTLLMWorker`).
- Fixed `_apply_model_update_to_cfg()`: added fallback to patch any non-Frontend/Planner worker when `PREFILL`/`DECODE` subtype lookup misses (agg mode).

## Backward Compatibility

All new parameters have safe defaults (`None`, `False`, `field(default_factory=dict)`). Existing callers are unaffected:

- `_execute_task_configs()` return changed from 4-tuple to 5-tuple. All call sites updated (cli/main.py, cli/api.py, test mocks).
- `process_experiment_result()` refactored to delegate to `picking.py` -- same return signature (4-tuple).
- `_get_disagg_summary_dict` delegates to `picking.build_disagg_summary_dict` -- same arithmetic, same constants.
- `_pick_autoscale` delegates to `picking.pick_autoscale` -- same filtering/ranking logic.
- `_get_best_configs_under_constraint` now returns closest matches instead of empty when SLA not met -- more useful behavior, no breakage.
- New files (`picking.py`, `test_picking.py`) are purely additive.

## How to review

1. **`src/aiconfigurator/sdk/picking.py`** -- New standalone picking module. Check the three `pick_*` functions and `build_disagg_summary_dict` helper.

2. **`src/aiconfigurator/sdk/pareto_analysis.py`** -- Check `get_best_configs_for_target_load()` (load-match picking) and the closest-match fallback in `_get_best_configs_under_constraint()`. Check GPU-capped behavior.

3. **`src/aiconfigurator/sdk/inference_session.py`** -- Verify `get_worker_candidates()` extraction and delegation patterns for `_get_disagg_summary_dict` and `_pick_autoscale`.

4. **`src/aiconfigurator/sdk/task.py`** -- Check `autoscale` param threading.

5. **`dynamo/.../protocol.py`** -- Check the agg-mode fix in `_apply_agg_worker()` and `_apply_model_update_to_cfg()` worker-lookup fallback.

6. **`test_picking.py`** -- E2E tests. Run with `python test_picking.py` to generate DGD configs in `picking_output/`.

### DIFF
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -675,11 +675,12 @@ def _execute_task_configs(
 
     best_configs: dict[str, pd.DataFrame] = {}
     best_throughputs: dict[str, float] = {}
+    best_latencies: dict[str, dict[str, float]] = {}
     pareto_fronts: dict[str, pd.DataFrame | None] = {}
     pareto_x_axis: dict[str, str] = {}
     for name, task_result in results.items():
         task_config = task_configs[name]
-        best_config_df, best_throughput, pareto_frontier_df, x_axis_col = process_experiment_result(
+        best_config_df, best_throughput, pareto_frontier_df, x_axis_col, latencies = process_experiment_result(
             task_config,
             task_result,
             top_n,
@@ -689,6 +690,7 @@ def _execute_task_configs(
         )
         best_configs[name] = best_config_df
         best_throughputs[name] = best_throughput
+        best_latencies[name] = latencies
         pareto_fronts[name] = pareto_frontier_df
         pareto_x_axis[name] = x_axis_col
 
@@ -711,27 +713,6 @@ def _execute_task_configs(
         target_request_rate=target_request_rate,
         target_concurrency=target_concurrency,
     )
-
-    # Build best_latencies from rank-1 config of each experiment
-    best_latencies: dict[str, dict[str, float]] = {}
-    for name, df in best_configs.items():
-        if df is not None and not df.empty:
-            row = df.iloc[0]
-            entry: dict[str, float] = {
-                "ttft": float(row["ttft"]),
-                "tpot": float(row["tpot"]),
-                "request_latency": float(row["request_latency"]),
-            }
-            # Load-match fields (present when target_request_rate/target_concurrency is set)
-            if "replicas_needed" in row.index:
-                entry["replicas_needed"] = int(row["replicas_needed"])
-            if "total_gpus_needed" in row.index:
-                entry["total_gpus_needed"] = int(row["total_gpus_needed"])
-            if "load_served_pct" in row.index:
-                entry["load_served_pct"] = float(row["load_served_pct"])
-            best_latencies[name] = entry
-        else:
-            best_latencies[name] = {"ttft": 0.0, "tpot": 0.0, "request_latency": 0.0}
 
     end_time = time.time()
     logger.info("All experiments completed in %.2f seconds", end_time - start_time)
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -620,15 +620,23 @@ def _execute_task_configs(
     task_configs: dict[str, TaskConfig],
     mode: str,
     top_n: int = 5,
-) -> tuple[str, dict[str, pd.DataFrame], dict[str, pd.DataFrame], dict[str, float]]:
+    target_request_rate: float | None = None,
+    target_concurrency: float | None = None,
+    max_total_gpus: int | None = None,
+) -> tuple[str, dict[str, pd.DataFrame], dict[str, pd.DataFrame], dict[str, float], dict[str, dict[str, float]]]:
     """
-    Execute task configs and return the chosen experiment, best configs, results, and best
-    throughputs.
+    Execute task configs and return the chosen experiment, best configs, results, best
+    throughputs, and estimated latencies.
 
     Args:
         task_configs: Dictionary mapping experiment names to TaskConfig objects to execute.
         mode: Execution mode ('default' or 'exp').
         top_n: Number of top configurations to return for each experiment.
+        target_request_rate: If set, activates load-match picking (minimize
+            GPUs for the given request rate in req/s).
+        target_concurrency: If set, activates load-match picking (minimize
+            GPUs for the given number of concurrent requests).
+        max_total_gpus: Optional upper bound on total GPUs for load-match.
 
     Returns:
         tuple:
@@ -636,6 +644,9 @@ def _execute_task_configs(
             - Dictionary of best config DataFrames per experiment (or per serving mode if merged).
             - Dictionary of Pareto frontier DataFrames per experiment (or mode).
             - Dictionary of best throughput values per experiment (or mode).
+            - Dictionary of estimated latencies per experiment. Each value is a
+              dict with keys ``"ttft"``, ``"tpot"``, ``"request_latency"``
+              extracted from the rank-1 config.
     """
     results: dict[str, dict[str, pd.DataFrame]] = {}
     start_time = time.time()
@@ -669,7 +680,12 @@ def _execute_task_configs(
     for name, task_result in results.items():
         task_config = task_configs[name]
         best_config_df, best_throughput, pareto_frontier_df, x_axis_col = process_experiment_result(
-            task_config, task_result, top_n
+            task_config,
+            task_result,
+            top_n,
+            target_request_rate=target_request_rate,
+            target_concurrency=target_concurrency,
+            max_total_gpus=max_total_gpus,
         )
         best_configs[name] = best_config_df
         best_throughputs[name] = best_throughput
@@ -692,12 +708,35 @@ def _execute_task_configs(
         mode=mode,
         pareto_x_axis=pareto_x_axis,
         top_n=top_n,
+        target_request_rate=target_request_rate,
+        target_concurrency=target_concurrency,
     )
+
+    # Build best_latencies from rank-1 config of each experiment
+    best_latencies: dict[str, dict[str, float]] = {}
+    for name, df in best_configs.items():
+        if df is not None and not df.empty:
+            row = df.iloc[0]
+            entry: dict[str, float] = {
+                "ttft": float(row["ttft"]),
+                "tpot": float(row["tpot"]),
+                "request_latency": float(row["request_latency"]),
+            }
+            # Load-match fields (present when target_request_rate/target_concurrency is set)
+            if "replicas_needed" in row.index:
+                entry["replicas_needed"] = int(row["replicas_needed"])
+            if "total_gpus_needed" in row.index:
+                entry["total_gpus_needed"] = int(row["total_gpus_needed"])
+            if "load_served_pct" in row.index:
+                entry["load_served_pct"] = float(row["load_served_pct"])
+            best_latencies[name] = entry
+        else:
+            best_latencies[name] = {"ttft": 0.0, "tpot": 0.0, "request_latency": 0.0}
 
     end_time = time.time()
     logger.info("All experiments completed in %.2f seconds", end_time - start_time)
 
-    return chosen_exp, best_configs, pareto_fronts, best_throughputs
+    return chosen_exp, best_configs, pareto_fronts, best_throughputs, best_latencies
 
 
 def _run_generate_mode(args):
@@ -864,7 +903,7 @@ def main(args):
     else:
         raise SystemExit(f"Unsupported mode: {args.mode}")
 
-    _, best_configs, pareto_fronts, _ = _execute_task_configs(
+    _, best_configs, pareto_fronts, _, _ = _execute_task_configs(
         task_configs,
         args.mode,
         top_n=args.top_n,
--- a/src/aiconfigurator/cli/utils.py
+++ b/src/aiconfigurator/cli/utils.py
@@ -6,78 +6,83 @@ import logging
 import pandas as pd
 
 from aiconfigurator.sdk.pareto_analysis import (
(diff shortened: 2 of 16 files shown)


src/aiconfigurator/sdk/inference_session.py
```diff
@@ -189,108 +193,21 @@ def _get_disagg_summary_dict(
...
-        total_time = ttft + decode_time
 
-        if total_time > 0:
-            disagg_power_avg = (prefill_power * ttft + decode_power * decode_time) / total_time
-        else:
-            disagg_power_avg = 0.0
-
-        logger.debug(
-            f"DISAGG Power Result: {disagg_power_avg}W (time-weighted from {prefill_power}W and {decode_power}W)"
+        Delegates to :func:`aiconfigurator.sdk.picking.build_disagg_summary_dict`
```
simone-chen: Out of curiosity, why don't we just use the `build_disagg_summary_dict` function and remove the `_get_disagg_summary_dict`? I'm fine with the current approach though.

src/aiconfigurator/sdk/pareto_analysis.py
```diff
@@ -466,32 +470,46 @@ def _get_best_configs_under_constraint(
...
-        # Optionally, one could return the one closest to target_tpot if no strict candidates exist.
-        # For now, return empty if no config meets the criteria.
-        logger.info("No config found on Pareto front with %s <= %s.", constraint_col, target_value)
-        return pd.DataFrame()
+
+    candidate_configs = (
+        candidate_configs.sort_values(by=sort_columns, ascending=sort_ascending).head(top_n).reset_index(drop=True)
+    )
+    candidate_configs.drop(columns=["_sla_exceeded"], inplace=True)
+    return candidate_configs
```
simone-chen, line 512: Just to confirm - in the previous implementation it returned an empty dataframe when `candidate_configs.empty`. Now it seems we are still returning candidate configs when sla exceeds. Is it expected behavior?

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -456,82 +520,6 @@ def _match_workers(
 
             return prefill_opt_num_worker, decode_opt_num_worker
 
-        def _get_summary_df(
```
simone-chen, line 459: Is this change purely a renaming to `get_worker_candidates`?

src/aiconfigurator/sdk/picking.py
```diff
@@ -0,0 +1,476 @@
...
+        "pareto_frontier_df": pareto_frontier_df,
+    }
+
+
+# ---------------------------------------------------------------------------
+# pick_autoscale
+# ---------------------------------------------------------------------------
+
+
+def pick_autoscale(
```
simone-chen, line 370: It would be nice to have some unit tests for each pick function. It could also help developers understand the picking logic better.

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -166,10 +171,9 @@ def __init__(
         self._prefill_latency_correction_scale = 1.0
         self._decode_latency_correction_scale = 1.0
 
-        # comes from pipeline bubble, especially when benchmarking with concurrency
-        self._RATE_MATCHING_PREFILL_DEGRADATION_FACTOR = 0.9
-        # comes from not saturating the batchsize slot of decode worker
-        self._RATE_MATCHING_DECODE_DEGRADATION_FACTOR = 0.92
+        # See picking.py for rationale
+        self._RATE_MATCHING_PREFILL_DEGRADATION_FACTOR = _RATE_MATCHING_PREFILL_DEGRADATION_FACTOR
```
simone-chen: Why do we take an extra step to make them instance variables if their values are hardcoded?

src/aiconfigurator/cli/main.py
```diff
@@ -692,12 +708,35 @@ def _execute_task_configs(
...
+            # Load-match fields (present when target_request_rate/target_concurrency is set)
+            if "replicas_needed" in row.index:
+                entry["replicas_needed"] = int(row["replicas_needed"])
+            if "total_gpus_needed" in row.index:
+                entry["total_gpus_needed"] = int(row["total_gpus_needed"])
+            if "load_served_pct" in row.index:
+                entry["load_served_pct"] = float(row["load_served_pct"])
+            best_latencies[name] = entry
+        else:
+            best_latencies[name] = {"ttft": 0.0, "tpot": 0.0, "request_latency": 0.0}
```
simone-chen: Is it possible to put the logic within `process_experiment_result`? So we can set `best_latencies` the same way as `best_throughputs` in the loop on line 691.

src/aiconfigurator/sdk/inference_session.py
tedzhouhk: good catch, this was due to a historical circular dependency issue. removed this wrapper

src/aiconfigurator/sdk/pareto_analysis.py
tedzhouhk: yes, we want to still return user something even if it is impossible to adhere the SLA

src/aiconfigurator/sdk/inference_session.py
tedzhouhk: from cursor (sorry i'm lazy

> Yes, it's a pure extract + rename. The only differences from the original nested _get_summary_df are:
> - Signature: _get_summary_df didn't take model_path (captured from closure). get_worker_candidates takes it as an explicit parameter.
> - Scope: moved from a nested function inside find_best_disagg_result_under_constraints to a class method on DisaggInferenceSession.
> - Log format: changed from f-strings to %-style (minor style cleanup).
> 
> The actual logic -- iterating parallel configs, creating models, running sess.run_static, checking OOM, concatenating results -- is identical line for line.

src/aiconfigurator/sdk/picking.py
tedzhouhk: added e2e tests at `tests/e2e/cli/test_picking.py`

src/aiconfigurator/sdk/inference_session.py
tedzhouhk: good catch, previously there are two sources for this factor, I merged them by importing one from another but forgot to clean these. Now removed

src/aiconfigurator/cli/main.py
tedzhouhk: done

src/aiconfigurator/sdk/picking.py
simone-chen: Do you mind adding a `tests/integration` folder and put `test_picking.py` there? Thank you!

simone-chen (approved):
LGTM, thanks for the PR! Left one final nit-picking comment. 
Also it seems there is a test failure to be fixed.

src/aiconfigurator/sdk/picking.py
```diff
@@ -0,0 +1,476 @@
...
+# local concurrency lc=15-20, formula lc/20+0.95 gives ~1.8
+_AUTOSCALE_TTFT_CORRECTION_FACTOR = 1.8
+
+
+# ---------------------------------------------------------------------------
+# Helper: build a disagg summary dict from prefill + decode dicts
+# ---------------------------------------------------------------------------
+
+
+def build_disagg_summary_dict(
```
jasonqinzhou: Shall we keep this internal?

src/aiconfigurator/cli/utils.py
```diff
@@ -6,80 +6,88 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-def process_experiment_result(task_config: TaskConfig, result: dict[str, pd.DataFrame], top_n: int = 5) -> tuple:
+def process_experiment_result(
+    task_config: TaskConfig,
+    result: dict[str, pd.DataFrame],
+    top_n: int = 5,
+    target_request_rate: float | None = None,
```
jasonqinzhou, line 21: Shall we expose 3 picking options through cli? I think it is helpful for users other than DGDR and in this way the cli and sdk are aligned.

tedzhouhk: sure, we can expose them in a future PR

src/aiconfigurator/sdk/picking.py
tedzhouhk: `inference_session.py` is also using this function, can't make it internal

src/aiconfigurator/cli/utils.py
jasonqinzhou: Agreed. This PR is big enough.

src/aiconfigurator/sdk/picking.py
jasonqinzhou: I think it is fine to rename it to _build_disagg_summary_dict(). It can still be used in inference_session.py since they are in the same repo. 
It is a bit confusing to have build_disagg_summary_dict() as a public function for other users.

tedzhouhk: from cursor:

> In Python, functions starting with a single underscore (_) are a naming convention indicating they are intended for internal use within their module (weak "internal use" indicator). By convention, they should not be accessed from other files, though they are not technically private and can still be imported.

jasonqinzhou (approved):

PR #400, opened 2026-02-19
feat: Add github workflow which auto creates sanity check charts for new perf data

ilyasher:
#### Overview:

When PRs are opened with new perf data, this workflow will run the sanity check jupyter notebook and export the sanity check charts, which can be downloaded as a github artifact. It will then post a comment on the PR with the download link and chart generation report.

Unfortunately I can't display the chart images directly in the comments, since the download URL for github artifacts only lets you download a zip file instead of the individual image files.

Screenshot 2026-02-18 at 5 23 42 PM


ilyasher:
converting to draft while I test some things

.github/workflows/create-charts.yml
```diff
@@ -0,0 +1,125 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+    
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ".[dev]"
```
simone-chen, line 37: It's not in a venv - would the runner's complain about "This environment is externally managed"?

.github/workflows/create-charts.yml
```diff
@@ -0,0 +1,125 @@
...
+      - name: Create charts
+        id: create-charts
+        run: |
+          BASE_REF="${{ github.event.pull_request.base.ref }}"
+          python tools/sanity_check/create_charts.py --base-ref origin/${BASE_REF} --head-ref HEAD --output-dir charts_output --output-md-file comment.md
+        continue-on-error: true
+    
+      - name: Upload charts as artifacts
+        if: steps.create-charts.outcome == 'success'
+        uses: actions/upload-artifact@v4
```
simone-chen, line 50: Out of curiosity - is there a way to view the files like how we view files in S3?

tools/sanity_check/create_charts.py
```diff
@@ -0,0 +1,213 @@
...
+import subprocess
+import sys
+import textwrap
+from collections import defaultdict
+
+from aiconfigurator.sdk.perf_database import get_database
+
+# Disable interactive backend
+os.environ["MPLBACKEND"] = "agg"
+import matplotlib.pyplot as plt
```
simone-chen, line 22: Why do we set the env var before an import?

tools/sanity_check/create_charts.py
```diff
@@ -0,0 +1,213 @@
...
+
+# Import validate_database.ipynb jupyter notebook
+old_cwd = os.getcwd()
+os.chdir(os.path.abspath(os.path.dirname(__file__)))
+import import_ipynb  # noqa: F401
```
simone-chen, line 27: Is `import_ipynb` a local file on your own machine? There doesn't seem to be a `import_ipynb` file in the codebase 🤔

tools/sanity_check/create_charts.py
```diff
@@ -0,0 +1,213 @@
...
+
+            Below is a report of whether the chart generation was successful for each op.
+            If doesn't validate whether the perf data itself is sane.
+        """)
+        )
+
+    changed_files = get_changed_files(args.base_ref, args.head_ref)
+
+    # Organize changed files by (system, backend, backend_version) so that
+    # they are grouped togather in the output text.
```
simone-chen:
```suggestion
    # they are grouped together in the output text.
```

tools/sanity_check/create_charts.py
```diff
@@ -0,0 +1,213 @@
...
+    # map (system, backend, backend_version) -> [changed_file]
+    system_backend_version_to_changed_files = defaultdict(list)
+
```
simone-chen, line 149: Shall we add a docstring documenting the expected directory tree structure for files in `src/aiconfigurator/systems/` which could be handled by this script?

.github/workflows/create-charts.yml
ilyasher: What do you mean by this?

tools/sanity_check/create_charts.py
ilyasher: It's an external package: https://pypi.org/project/import-ipynb/

.github/workflows/create-charts.yml
simone-chen: Like is there a way to navigate our github artifacts via UI/CLI? Can we only download them via API calls?

ilyasher: Oh yeah it's possible. Here's an example: https://github.com/ilyasher/aiconfigurator/actions/runs/22164290457

.github/workflows/create-charts.yml
ilyasher: Good question, it worked fine when I tested it: https://github.com/ilyasher/aiconfigurator/actions/runs/22164290457/job/64088062539

Maybe the python installed by `actions/setup-python` doesn't give this error unlike the OS-provided python

tools/sanity_check/create_charts.py
ilyasher: I don't think it's necessary, it's just looking at the AIC perf data directory, nothing custom

tools/sanity_check/create_charts.py
ilyasher: When we import `validate_database` below, it will run the entire jupyter notebook. If the matplotlib backend is interactive, it will fail. So we need to set the environment variable before importing the jupyter notebook. I don't remember if it's also necessary to set it before importing matplotlib, but it doesn't hurt

simone-chen (approved):

PR #407, opened 2026-02-20
feat: support DeepSeek V3.2

jasonqinzhou:

- closes GitHub issue: #xxx


src/aiconfigurator/model_configs/nvidia--DeepSeek-V3-0324-NVFP4_hf_quant_config.json
tianhaox, line 1: Is this duplicate with
src/aiconfigurator/model_configs/nvidia--DeepSeek-V3.1-NVFP4_hf_quant_config.json ?

tianhaox: but it's okay with slight difference if users want.

src/aiconfigurator/sdk/common.py
```diff
@@ -284,6 +288,7 @@ def get_default_models() -> set[str]:
     "Qwen3ForCausalLM": "LLAMA",
     "DeepSeekForCausalLM": "DEEPSEEK",
     "DeepseekV3ForCausalLM": "DEEPSEEK",
+    "DeepseekV32ForCausalLM": "DEEPSEEK",
```
tianhaox, line 291: i think this is a risk. Can we comment this out and get it back when we do support DS V3.2. I'm afraid that we will forget about this if we don't do it.

PR #408, opened 2026-02-20
feat: unified HYBRIDMOE family (Llama 4 Scout/Maverick + MiMo-V2-Flash)

jasonqinzhou:
## Summary

Introduces a unified `HYBRIDMOE` model family and `HybridMoEConfig` dataclass that covers all hybrid-attention + mixed-FFN models, replacing two separate implementations.

**Problem:** MiMo-V2-Flash and Llama 4 Scout/Maverick share the same fundamental structure — interleaved SWA/local + global attention, interleaved MoE + dense FFN layers — but were being treated as separate concerns requiring separate config classes and model implementations.

**Solution:** One `HybridMoEConfig`, one `HybridMoEModel`, one `HYBRIDMOE` family.

---

### `HybridMoEConfig` design

Normalizes both pattern styles into per-layer tuples:

| Field | Description |
|---|---|
| `attn_layer_pattern` | per-layer `0=SWA/local, 1=global`, len=`num_layers` |
| `moe_layer_freq` | per-layer `0=dense, 1=MoE`, len=`num_layers` |
| `swa_*` / `global_v_head_dim` | 0 = fall back to model defaults (Llama 4 has same dims; MiMo differs) |
| `sliding_window_size` | local attention window |
| `dense_inter_size` | 0 = use model-level `inter_size` |

MiMo-V2-Flash passes explicit per-layer lists from config. Llama 4 generates the lists at parse time from `interleave_moe_layer_step` and alternating attention.

---

### `HybridMoEModel` — 4 layer types

| Type | Attention | FFN |
|---|---|---|
| `global_moe` | global (full) | MoE |
| `swa_moe` | SWA/local | MoE |
| `swa_dense` | SWA/local | dense SwiGLU |
| `global_dense` | global (full) | dense SwiGLU |

---

### Layer type decomposition (validated)

| Model | global_moe | swa_moe | swa_dense | global_dense |
|---|---|---|---|---|
| Llama 4 Scout (16E, step=1) | 24 | 24 | 0 | 0 |
| Llama 4 Maverick (128E, step=2) | 24 | 0 | 24 | 0 |
| MiMo-V2-Flash | 39 | 8 | 1 | 0 |

---

### Files changed

- `sdk/common.py`: `HybridMoEConfig` dataclass; `HYBRIDMOE` added to `ModelFamily`; `MiMoV2FlashForCausalLM` + `Llama4ForConditionalGeneration` registered in `HYBRIDMOE`
- `sdk/utils.py`: unified parsing for both architectures → both produce `HybridMoEConfig`; input validation (pattern length, values ∈ {0,1}, step > 0)
- `sdk/models.py`: `HybridMoEModel` covers all 4 layer types; wired into `get_model()` and `check_is_moe()`; MTP/nextn scaling on generation ops (mirrors `MOEModel`); `set_hybrid_config()` validates pattern length == `num_layers` and 0/1 values
- `model_configs/`: Llama 4 Scout + Maverick + MiMo-V2-Flash + MiMo-7B-Base configs added
- `tests/`: parsing tests (Scout, Maverick, MiMo-V2-Flash), negative tests (pattern mismatch, invalid step), builder-level tests (all three model variants)

## Test plan

- [ ] `pytest tests/unit/sdk/test_utils.py` — all hybrid parsing + builder tests pass
- [ ] `pytest tests/unit/sdk/` — full suite passes
- [ ] Manually verify layer counts for Scout / Maverick / MiMo-V2-Flash via `get_model()`

## Summary by CodeRabbit

* **New Features**
  * Added support for four new models: Meta Llama‑4 Scout & Maverick (17B variants) and Xiaomi MiMo‑7B Base & MiMo V2 Flash.
  * Introduced Hybrid Mixture-of-Experts (HybridMoE) model runtime and registry integration.

* **Enhancements**
  * Improved model config handling with per-layer hybrid patterns, sliding-window/MoE settings, and enhanced quantization inference/normalization.
  * Expanded default model listings and architecture mappings.

* **Tests**
  * Added unit tests for HF config parsing and HybridMoE model wiring and validation.

### DIFF
--- a/src/aiconfigurator/model_configs/XiaomiMiMo--MiMo-7B-Base_config.json
+++ b/src/aiconfigurator/model_configs/XiaomiMiMo--MiMo-7B-Base_config.json
@@ -1,0 +1,28 @@
+{
+  "architectures": [
+    "MiMoForCausalLM"
+  ],
+  "model_type": "mimo",
+  "num_hidden_layers": 36,
+  "hidden_size": 4096,
+  "num_attention_heads": 32,
+  "num_key_value_heads": 8,
+  "head_dim": 128,
+  "intermediate_size": 11008,
+  "vocab_size": 151680,
+  "max_position_embeddings": 32768,
+  "rope_theta": 640000,
+  "rms_norm_eps": 1e-05,
+  "hidden_act": "silu",
+  "attention_bias": true,
+  "attention_dropout": 0.0,
+  "sliding_window": 32768,
+  "max_window_layers": 32,
+  "use_sliding_window": false,
+  "use_mrope": false,
+  "num_nextn_predict_layers": 1,
+  "tie_word_embeddings": false,
+  "torch_dtype": "bfloat16",
+  "transformers_version": "4.40.1",
+  "use_cache": true
+}
--- a/src/aiconfigurator/model_configs/XiaomiMiMo--MiMo-V2-Flash_config.json
+++ b/src/aiconfigurator/model_configs/XiaomiMiMo--MiMo-V2-Flash_config.json
@@ -1,0 +1,60 @@
+{
+  "architectures": [
+    "MiMoV2FlashForCausalLM"
+  ],
+  "model_type": "mimo_v2_flash",
+  "num_hidden_layers": 48,
+  "hidden_size": 4096,
+  "num_attention_heads": 64,
+  "num_key_value_heads": 4,
+  "head_dim": 192,
+  "v_head_dim": 128,
+  "intermediate_size": 16384,
+  "vocab_size": 152576,
+  "max_position_embeddings": 262144,
+  "rope_theta": 5000000,
+  "swa_rope_theta": 10000,
+  "partial_rotary_factor": 0.334,
+  "sliding_window": 128,
+  "sliding_window_size": 128,
+  "attention_chunk_size": 128,
+  "swa_num_attention_heads": 64,
+  "swa_num_key_value_heads": 8,
+  "swa_head_dim": 192,
+  "swa_v_head_dim": 128,
+  "hybrid_layer_pattern": [
+    0, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 0,
+    1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 0
+  ],
+  "moe_layer_freq": [
+    0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1
+  ],
+  "moe_intermediate_size": 2048,
+  "n_routed_experts": 256,
+  "n_shared_experts": null,
+  "num_experts_per_tok": 8,
+  "norm_topk_prob": true,
+  "scoring_func": "sigmoid",
+  "n_group": 1,
+  "topk_group": 1,
+  "topk_method": "noaux_tc",
+  "routed_scaling_factor": null,
+  "hidden_act": "silu",
+  "layernorm_epsilon": 1e-05,
+  "attention_dropout": 0.0,
+  "attention_bias": false,
+  "attention_value_scale": 0.707,
+  "add_swa_attention_sink_bias": true,
+  "add_full_attention_sink_bias": false,
+  "tie_word_embeddings": false,
+  "torch_dtype": "bfloat16",
+  "transformers_version": "4.40.1",
+  "use_cache": true,
+  "quantization_config": {
+    "activation_scheme": "dynamic",
+    "fmt": "e4m3",
+    "quant_method": "fp8",
+    "weight_block_size": [128, 128]
+  }
+}
--- a/src/aiconfigurator/sdk/models.py
+++ b/src/aiconfigurator/sdk/models.py
@@ -3122,13 +3122,13 @@ class HybridMoEModel(BaseModel):
     """
     Hybrid attention + mixed FFN model (MiMo-V2-Flash, Llama 4 Scout/Maverick, and similar).
 
-    Handles four layer types derived from HybridConfig.attn_layer_pattern and moe_layer_freq:
+    Handles four layer types derived from HybridMoEConfig.attn_layer_pattern and moe_layer_freq:
     - global_moe:  global (full) attention + MoE FFN
     - swa_moe:     SWA/local attention + MoE FFN
     - swa_dense:   SWA/local attention + dense SwiGLU FFN
     - global_dense: global attention + dense SwiGLU FFN (rare but supported)
 
-    SWA/local attention dims fall back to model-level defaults when HybridConfig fields are 0.
+    SWA/local attention dims fall back to model-level defaults when HybridMoEConfig fields are 0.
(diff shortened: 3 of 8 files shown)


src/aiconfigurator/model_configs/XiaomiMiMo--MiMo-7B-Base_config.json
```diff
@@ -0,0 +1,28 @@
+{
```
tianhaox, line 1: https://huggingface.co/XiaomiMiMo/MiMo-7B-Base/blob/main/config.json

src/aiconfigurator/model_configs/XiaomiMiMo--MiMo-V2-Flash_config.json
```diff
@@ -0,0 +1,60 @@
+{
```
tianhaox, line 1: https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash/blob/main/config.json

src/aiconfigurator/sdk/models.py
```diff
@@ -2991,6 +3011,471 @@ def _build_generation_ops(self) -> None:
         )
 
 
+class HybridMoEModel(BaseModel):
```
tianhaox, line 3121: is gptoss following this model arch?

jasonqinzhou: Good question. Technically it *could* approximate the attention, but it would break on **three other things** that GPT-OSS doesn't have. Let me walk through exactly why.

## Why `attn_scale_factor` works for GPT-OSS

GPT-OSS only differs in **window size** per layer. Everything else is uniform:
- Same `num_kv_heads` for both SWA and global layers
- Same `head_size` for both
- Same `v_head_dim` for both
- **All** layers are MoE (no dense FFN layers)

So the only thing `attn_scale_factor` needs to capture is how many layers use `window_size=128` vs `window_size=0`. The QKV GEMM, proj GEMM, and MoE ops can still use `num_layers` uniformly.

## Why MiMo-V2-Flash can't do the same

Looking at `MiMoConfig` and what `HybridMoEModel` actually models:

```63:84:src/aiconfigurator/sdk/common.py
@dataclass(frozen=True)
class MiMoConfig:
    hybrid_layer_pattern: tuple[int, ...]
    moe_layer_freq: tuple[int, ...]
    swa_num_kv_heads: int
    swa_head_dim: int
    swa_v_head_dim: int
    global_v_head_dim: int
    sliding_window_size: int
```

There are three problems that `attn_scale_factor` alone can't fix:

**Problem 1 — Different KV head counts per layer type**

SWA layers have `swa_num_kv_heads` (fewer KV heads), global layers have `num_kv_heads` (more KV heads). In `MOEModel`, the QKV GEMM output size is fixed:
```833:833:src/aiconfigurator/sdk/models.py
                    self._num_heads * self._head_size // tp_size + self._head_size * num_kv_heads_per_gpu * 2,
```
You'd need two separate QKV GEMM ops with different output sizes. `attn_scale_factor` only scales the attention op, not the preceding GEMM.

**Problem 2 — Different `v_head_dim` per layer type**

`swa_v_head_dim ≠ global_v_head_dim` in MiMo. This affects both the FMHA compute and the output proj GEMM input size. `MOEModel` uses a single proj GEMM:
```850:850:src/aiconfigurator/sdk/models.py
                    self._num_heads * self._head_size // tp_size,
```
You'd need `num_layers/6 * global_v_head_dim` and `5*num_layers/6 * swa_v_head_dim` variants — again not capturable with just a scale factor.

**Problem 3 — Mixed dense + MoE FFN layers**

This is the biggest one. `moe_layer_freq` means some MiMo layers have dense FFN, not MoE. `MOEModel` applies MoE to **all** `num_layers`:
```888:888:src/aiconfigurator/sdk/models.py
                    self._num_layers,
```
There's no mechanism in `MOEModel` to say "only N of these layers actually have MoE, the rest are dense." You'd need separate dense FFN ops for those layers — which is exactly what `HybridMoEModel._count_layer_types()` does.

---

## Summary

| What needs modeling | GPT-OSS | MiMo-V2-Flash |
|---|---|---|
| Attention window differs per layer | ✓ `attn_scale_factor` handles it | ✓ but also needs... |
| KV head count differs per layer | ✗ same for all | ✗ `swa_num_kv_heads` ≠ `num_kv_heads` |
| v_head_dim differs per layer | ✗ same for all | ✗ `swa_v_head_dim` ≠ `global_v_head_dim` |
| Some layers are dense FFN (not MoE) | ✗ all MoE | ✗ `moe_layer_freq` has zeros |

You could use `attn_scale_factor = 6` to get the attention ratio right (1 global for every 5 SWA), but you'd still be wrong on the GEMM sizes and dense layer count. The errors would compound in a non-obvious way — which is precisely why `HybridMoEModel` was the right call.

src/aiconfigurator/sdk/models.py
```diff
@@ -2991,6 +3011,471 @@ def _build_generation_ops(self) -> None:
...
+    """
```
jasonqinzhou, line 3122: Should GPT-OSS move to HYBRIDMOE?

**No — and this actually corrects something I said earlier.** Looking at the configs:

- **GPT-OSS-120B** (36 layers): strictly `sliding, full, sliding, full...` × 18 = **exactly 18 SWA + 18 full**
- **GPT-OSS-20B** (24 layers): strictly `sliding, full, sliding, full...` × 12 = **exactly 12 SWA + 12 full**

Both are a **perfect 50/50 split**. So `attn_scale_factor = 2` is not a rough approximation — it's **exactly right**. I was wrong earlier when I suggested 5:1; that was guessing from web search, not the actual model config.

GPT-OSS does not need HYBRIDMOE because:

| | GPT-OSS | MiMo-V2-Flash (why it needs HYBRIDMOE) |
|---|---|---|
| SWA vs Global KV head count | Same (`num_key_value_heads=8` for both) | Different (`swa_num_kv_heads=8` vs `num_kv_heads=4`) |
| SWA vs Global head/v dim | Same (`head_dim=64` for both) | Different (`swa_v_head_dim` vs `v_head_dim`) |
| Mixed dense FFN layers | None (all MoE) | Yes (layer 0 is dense, `moe_layer_freq[0]=0`) |
| Attention ratio | Exact 1:1 alternating, `attn_scale_factor=2` captures it perfectly | Per-layer pattern with 9 SWA + 39 global |

## The one thing worth improving

The `attn_scale_factor = 2` is currently a magic number in code. A cleaner fix would be to derive it dynamically from the `layer_types` field that already exists in the config:

```python
if self.architecture == "GptOssForCausalLM":
    sliding_count = sum(1 for t in layer_types if t == "sliding_attention")
    full_count = self._num_layers - sliding_count
    # model sliding layers at window_size=128, full layers at window_size=0
```

But that's a code quality improvement, not a correctness fix, and certainly not a reason to move to HYBRIDMOE. Keep GPT-OSS in `MOE` — the current approach is architecturally sound.

src/aiconfigurator/sdk/utils.py
```diff
@@ -505,6 +506,51 @@ def _parse_hf_config_json(config: dict) -> dict:
...
+        extra_params = HybridConfig(
+            attn_layer_pattern=tuple(config.get("hybrid_layer_pattern", [])),
+            moe_layer_freq=moe_layer_freq,
+            swa_num_kv_heads=config.get("swa_num_key_value_heads", 0),
+            swa_head_dim=config.get("swa_head_dim", 0),
+            swa_v_head_dim=config.get("swa_v_head_dim", 0),
+            global_v_head_dim=config.get("v_head_dim", 0),
+            sliding_window_size=config.get("sliding_window_size", 0),
+            dense_inter_size=0,  # dense layers use model-level inter_size
+        )
```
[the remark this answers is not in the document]
jasonqinzhou: is it good now

Arsene12358 (approved):

jasonqinzhou:
https://github.com/coderabbitai review

PR #409, opened 2026-02-20
feat: support Qwen3-VL-235B-A22B-Thinking in HYBRID mode

jasonqinzhou:
● Yes, Qwen/Qwen3-VL-235B-A22B-Thinking is now fully supported in HYBRID mode. All 26 unit tests pass.                   
                                                                                                                         
  Here's what makes it work:                                                                                             
  - The user added VisionEncoderConfig, VLM_ARCHITECTURES, and text_config flattening in utils.py — so the nested VL
  config format is handled automatically                                                                                 
  - Qwen3VLMoeForConditionalGeneration → MOE family (text backbone is identical to Qwen3-235B-A22B: 94 layers, 128       
  experts, topk=8)
  - Local config JSON created at Qwen--Qwen3-VL-235B-A22B-Thinking_config.json for offline HYBRID mode
  - Model added to DefaultHFModels

  The vision encoder metadata (VisionEncoderConfig) is also parsed and stored, ready for future use if you want to model
  prefill cost with image tokens.

### DIFF
--- a/src/aiconfigurator/sdk/backends/base_backend.py
+++ b/src/aiconfigurator/sdk/backends/base_backend.py
@@ -75,6 +75,34 @@ class BaseBackend(ABC):
             if isl <= 0:
                 raise ValueError(f"isl must be greater than 0 after removing prefix, but got {isl}")
 
+            # --- Vision encoder phase ---
+            num_images = getattr(runtime_config, "num_images", 0)
+            if model.vision_ops and num_images > 0:
+                ve_cfg = model.vision_encoder.config
+                img_h = getattr(runtime_config, "image_height", 384)
+                img_w = getattr(runtime_config, "image_width", 384)
+                num_patches = ve_cfg.num_patches(img_h, img_w)
+                vision_x = batch_size * num_images * num_patches
+
+                for op in model.vision_ops:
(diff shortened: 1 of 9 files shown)


src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -75,6 +75,34 @@ def _run_context(batch_size: int, isl: int, prefix) -> tuple[dict[str, float], d
             if isl <= 0:
                 raise ValueError(f"isl must be greater than 0 after removing prefix, but got {isl}")
 
+            # --- Vision encoder phase ---
```
tianhaox, line 78: looks like this design is in conflict with kimi k2.5's vision module?

src/aiconfigurator/sdk/backends/sglang_backend.py
```diff
@@ -481,6 +481,8 @@ def _get_memory_usage(
         # Calculate weights memory - same as TRTLLM
         for op in model.context_ops:
             weights += op.get_weights()
+        for op in model.vision_ops:
```
tianhaox, line 484: in kimi k2.5's design, I remember vision ops are part of context ops?

PR #441, opened 2026-02-25
feat: Hide/deduplicate spammy logs

ilyasher:
#### Overview:

Improve logging based on VDR feedback. By default logs should be informative rather than overwhelming.

1. Remove duplicate logs (especially "apply final quant mode" and "loaded model architecture")

2. Move some logs from info to debug

Together with https://github.com/ai-dynamo/aiconfigurator/pull/452, the log output is much cleaner than before:

Screenshot 2026-02-24 at 4 47 38 PM

### DIFF
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -853,7 +853,7 @@ def _execute_task_configs(
     for exp_name, task_config in task_configs.items():
         try:
             logger.info("Starting experiment: %s", exp_name)
-            logger.info("Task config: %s", task_config.pretty())
+            logger.debug("Task config: \n%s", task_config.pretty())
             task_result = runner.run(task_config)
             pareto_df = task_result["pareto_df"]
             if pareto_df is not None and not pareto_df.empty:
--- a/src/aiconfigurator/sdk/models.py
+++ b/src/aiconfigurator/sdk/models.py
@@ -3,8 +3,10 @@
 
 from __future__ import annotations
 
+import dataclasses
 import logging
 from functools import cache
+from typing import Optional
 
 import aiconfigurator.sdk.operations as ops
 from aiconfigurator.sdk import common, config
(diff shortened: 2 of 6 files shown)


src/aiconfigurator/sdk/models.py
```diff
@@ -160,8 +160,6 @@ def get_model(
     # Convert architecture (e.g., 'LlamaForCausalLM') to model family (e.g., 'LLAMA')
     model_family = _architecture_to_model_family(architecture)
 
-    _apply_model_quant_defaults(model_config, raw_config, architecture, backend_name)
```
ilyasher, line 163: Removing because quant modes will be resolved by this point

src/aiconfigurator/cli/main.py
```diff
@@ -852,7 +852,7 @@ def _execute_task_configs(
     for exp_name, task_config in task_configs.items():
         try:
             logger.info("Starting experiment: %s", exp_name)
-            logger.info("Task config: %s", task_config.pretty())
+            logger.debug("Task config: \n%s", task_config.pretty())
```
simone-chen, line 856: What's the diff in `info` and `debug` here? Wouldn't `debug` spit more lines?

simone-chen (approved):

src/aiconfigurator/cli/main.py
ilyasher: Now the task config will only be logged to stderr if AIC is run with the `--debug` flag.

src/aiconfigurator/sdk/models.py
tianhaox: what's the action here?

tianhaox (approved):

src/aiconfigurator/sdk/models.py
ilyasher: No action, just wanted to explain to reviewers why I removed this line. i actually ended up adding it back

PR #449, opened 2026-02-25
fix: update support cli wording

Harrilee:


Harrilee:
Confirmed that @jasonqinzhou made the change on purpose.

PR #452, opened 2026-02-25
feat: Add colored logging

ilyasher:
#### Overview:

Trying to address from feedback from the VDR to improve logging. We want the logs to be more readable and look less like a debug stream. Colored logging helps with this

1.  log header (timestamp / filename / lineno) appear fainter than the log message, so the message itself is easier to read
2. Add color to severity label (`[I]` is blue, `[W]` is yellow, `[E]` is red)

Screenshot 2026-02-24 at 4 47 38 PM

### DIFF
--- a/src/aiconfigurator/logging_utils.py
+++ b/src/aiconfigurator/logging_utils.py
@@ -44,12 +44,7 @@ class ColoredFormatter(logging.Formatter):
         if not self.use_colors:
             return log_message
 
-        # Handle different log formats
-        # Format 1: "HH:MM:SS [aiconfigurator] [L] [filename:line] message"
-        # Format 2: "LEVELNAME YYYY-MM-DD HH:MM:SS,mmm filename:line] message"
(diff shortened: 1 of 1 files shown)


src/aiconfigurator/webapp/main.py
```diff
@@ -6,6 +6,7 @@
 from collections import defaultdict
 
 import aiconfigurator
+from aiconfigurator.logging_utils import setup_logging
```
simone-chen, line 9: Nit: do you think it will be cleaner if we create a `utils` module instead of leaving `logging_utils` in the top-level module? So we would have `aiconfigurator.utils.*` in parallel with `aiconfigurator.sdk.*`, `aiconfigurator.cli.*` etc.

simone-chen (approved):

src/aiconfigurator/webapp/main.py
ilyasher: I'm not really sure. We can rename later

PR #469, opened 2026-02-27
feat: add resume capability of collector

tianhaox:
#### Overview:

Add resume/checkpoint to collector so interrupted data collection jobs can pick up where they left off.

#### Details:

Every run writes per-module checkpoint files to `.collector_resume/<backend>/<module>.json`, recording which tasks are done (regardless of success/failure). Pass `--resume` to skip completed tasks; omit it to start fresh (overwrites old checkpoint).

```bash
python3 collect.py --backend trtllm                # normal run, writes checkpoint
python3 collect.py --backend trtllm --resume       # skip already-attempted tasks
python3 collect.py --backend trtllm --resume --resume-dir /path/to/ckpt
```

Key design choices:
- Binary done/not-done model — no state machine, no retry logic
- `create_test_case_id` uses SHA-256 instead of Python `hash()` for cross-session stability
- No behavioral change to existing worker crash → restart handling

Changes: `collector/collect.py`, `collector/helper.py` , `collector/README.md` , `.gitignore`
Tests: `tests/unit/collector/test_resume_mechanism.py` — 10 cases covering fresh/resume/corrupt/mismatch/all-success/all-fail/short-circuit

Fix 1 minor issue:
- No need to spawn new process if task queue is empty to avoid endless waiting

`ResumeCheckpoint` class in `collector/collect.py` (L57–139, ~90 lines) and its integration in `parallel_run()`.

#### Related Issues:

- closes GitHub issue: #xxx


Harrilee (approved):

Harrilee:
cc @ilyasher PIC for collector for a second review if necessary.

collector/collect.py
```diff
@@ -291,17 +441,25 @@ def create_process_exit_error(device_id, exit_code):
                         logger.error(f"Process {i} exceeded restart limit, not restarting")
                         continue
 
-                    processes[i] = start_process(i)
+                    if remaining > 0:
+                        processes[i] = start_process(i)
+                        queue.put(None)  # ensure the new worker can terminate
+                    else:
+                        processes[i] = p  # keep dead process ref, skip restart
```
xutizhou, line 473: Bug: All `num_processes` sentinels are already enqueued at lines 377-378. When a dead worker is restarted here, adding another `None` creates an **extra** sentinel — one of the other live workers will receive it and terminate prematurely. This compounds with each restart.

The dead worker either:
- Already consumed its sentinel (so the count is balanced without this line), or
- Died mid-task before reaching a sentinel (so its original sentinel is still in the queue for the new worker)

Either way, this extra `queue.put(None)` is a net surplus. Consider removing it, or tracking sentinel count explicitly.

collector/collect.py
```diff
@@ -94,6 +195,15 @@ def worker(queue, device_id: int, func, progress_value, lock, error_queue=None,
...
     worker_logger.info(f"Worker {device_id} initialized for {module_name}")
 
+    def emit_done(task_id: str):
+        """Notify the main process that a task was attempted (success or failure)."""
+        if not result_queue:
+            return
+        try:
+            result_queue.put(task_id)
+        except Exception:
+            pass
```
xutizhou, line 209: Nit: The bare `except Exception: pass` means if `result_queue.put()` fails during process teardown (e.g., after `exit(0)` for `AcceleratorError`), the task silently won't be checkpointed and will be re-run on resume. Safe given the binary done/not-done model, but a `logger.debug` here would help operators understand why a task was repeated:
```python
except Exception as e:
    worker_logger.debug(f"emit_done failed for {task_id}: {e}")
```

collector/helper.py
```diff
@@ -613,10 +614,14 @@ def get_sm_version():
...
-    test_str = str(test_case)
-    return f"{module_name}_{test_type}_{abs(hash(test_str)) % 100000}_{test_str}"
+    """Create a stable, cross-session identifier for a test case.
+
+    Uses SHA-256 instead of Python's built-in hash() so that IDs are
+    deterministic across interpreter sessions (PYTHONHASHSEED) and machines.
+    """
+    raw = f"{module_name}:{test_type}:{test_case}"
+    digest = hashlib.sha256(raw.encode("utf-8")).hexdigest()[:16]
+    return f"{module_name}_{test_type}_{digest}"
```
xutizhou: The new ID format drops the human-readable test params suffix. Old: `module_type_12345_(param1, param2, ...)`, new: `module_type_a1b2c3d4e5f6g7h8`. This makes error reports harder to debug since you can't see which params failed at a glance.

Consider keeping a truncated suffix:
```python
return f"{module_name}_{test_type}_{digest}_{str(test_case)[:80]}"
```

xutizhou (commented):
**Missing tests**: PR description mentions `tests/unit/collector/test_resume_mechanism.py` with 10 test cases, but this file is not in the diff and doesn't exist on the branch. Are the tests missing from the commit?

collector/collect.py
```diff
@@ -818,6 +1025,12 @@ def main():
 
     num_processes = torch.cuda.device_count()
     logger.info(f"Starting collection with {num_processes} GPU processes")
+    resume_options = {
+        "resume": args.resume,
+        "resume_dir": args.resume_dir,
+    }
+    if args.resume:
+        logger.info(f"Resume enabled: dir={Path(args.resume_dir).expanduser()}")
```
ilyasher: nit: Instead of passing a `resume_options` dict, cleaner to just pass `checkpoint_path: str | None`

collector/helper.py
ilyasher: Why bother with the SHA at all? Why not just `f"{module_name}:{test_type}:{test_case}"`?

collector/collect.py
```diff
@@ -806,6 +1002,17 @@ def main():
         default=1.0,
         help="Minimum duration for kernel runs when power measurement is enabled (default: 1.0s)",
     )
+    parser.add_argument(
+        "--resume",
+        action="store_true",
+        help="Resume collection from checkpoint, skipping already-attempted tasks",
+    )
+    parser.add_argument(
+        "--resume-dir",
```
ilyasher: `--checkpoint-dir` sounds better

collector/collect.py
```diff
@@ -46,13 +46,99 @@ def setup_warning_filters():
...
+            "schema": RESUME_SCHEMA_VERSION,
+            "backend": backend,
+            "module": module_name,
+            "run_func": run_func_name,
+        }
+        self._done: set[str] = set()
+
+        safe_name = module_name.replace("/", "_").replace(":", "_")
+        self._path = Path(resume_dir).expanduser().resolve() / backend / f"{safe_name}.json"
+        self._path.parent.mkdir(parents=True, exist_ok=True)
```
ilyasher, line 83: If it is 1 file for each module, then does the checkpoint really need to be a JSON? Can't it just be a growing list of `task_id` separated by newlines? Then you don't need to re-write the entire file all the time, just append new task_ids to the end.

ilyasher (approved):

collector/collect.py
tianhaox: good catch. fixed in latest commit and added several tests to handle the multi-process task execution

collector/collect.py
tianhaox: This pass is used to give a promise to the finally block that we never throw exception from this block.

tianhaox: the worker's debug module can be in an undefined status if it's crashed. This is in conflict with 'pass'

collector/helper.py
tianhaox: Followed Ilya's suggestion, using `f"{module_name}:{test_type}:{test_case}"` directly

collector/collect.py
tianhaox: maybe we can keep it here for now, leaving some room for future extension.

collector/collect.py
tianhaox: accepted

collector/collect.py
tianhaox: can we keep it for now, atomic read/write make things easier, otherwise need to handle different data structure of the flush and append. Claude said it's not a bottleneck for now.

PR #472, opened 2026-02-27
feat: add MTP simulation support for Qwen series models

harryjing:
Extend Multi-Token Prediction (MTP) speculative decoding simulation support from DeepSeek V3/R1 to Qwen series models (LLAMAModel and MOEModel). This enables performance prediction for deployments using speculative decoding with Qwen models.

Changes:
- LLAMAModel: Remove nextn assertion, add MTP scale factor calculation
- MOEModel: Remove nextn assertion, add MTP scale factor calculation
- Apply MTP scale factor to generation ops and P2P communication
- Update trtllm_backend comment for clarity

Backward compatible: nextn=0 produces identical results as before.

- closes GitHub issue: #xxx

## Summary by CodeRabbit

* **New Features**
  * Added Multi-Token Prediction (MTP) speculative decoding support across multiple model architectures for improved inference efficiency.

* **Documentation**
  * Updated comments to clarify memory calculations for draft token handling.


xutizhou (commented):
**Review Summary**: Request Changes

核心问题：
1. DeepSeek `nextn=1→0` 是 breaking behavior change，需要明确说明意图
2. `nextn_accept_rates` 默认值在三处定义不一致，会导致不同入口路径行为不同
3. 测试中 bare `except` + skip 会掩盖真实 bug

整体设计方向（将 MTP 从 DeepSeek-only 泛化到所有模型）合理，DeepSeek `context_p2p` 的 bug fix 也是正确的。

src/aiconfigurator/sdk/task.py
```diff
@@ -303,7 +303,7 @@ def _mode_layers(cls, ctx: TaskContext) -> list[ConfigLayer]:
 
     @staticmethod
     def _base_common_layer(ctx: TaskContext) -> dict:
```
xutizhou, line 305: 🔴 **Breaking behavior change** — DeepSeek default nextn changed from 1 to 0

On the current main branch, DeepSeek models default to `nextn=1`, reflecting the fact that DeepSeek V3/R1 natively supports MTP. Changing this to 0 means all existing DeepSeek users will see different performance predictions (lower generation throughput estimates) unless they explicitly pass `--nextn 1`.

Suggestion: keep the `nextn=1` default for DeepSeek, or explicitly call out this breaking change in the PR description with justification for why DeepSeek should also default to MTP-off.

xutizhou (commented):
**Review Summary**: Request Changes

Core issues:
1. DeepSeek `nextn=1→0` is a breaking behavior change — needs explicit justification
2. `nextn_accept_rates` default values are inconsistent across three locations
3. Bare `except` + `pytest.skip` in tests will silently mask real bugs

Overall design direction (generalizing MTP from DeepSeek-only to all models) is sound, and the DeepSeek `context_p2p` bug fix is correct.

src/aiconfigurator/cli/main.py
```diff
@@ -145,6 +145,19 @@ def _add_default_mode_arguments(parser):
...
     parser.add_argument("--prefix", type=int, default=0, help="Prefix cache length. Default to 0.")
+    parser.add_argument(
+        "--nextn",
+        type=int,
+        default=0,
+        help="Number of draft tokens for MTP (Multi-Token Prediction) speculative decoding. Default is 0 (disabled).",
+    )
+    parser.add_argument(
+        "--nextn-accept-rates",
+        type=str,
```
xutizhou, line 156: 🔴 **Inconsistent `nextn_accept_rates` defaults across three locations**

| Location | Default value |
|----------|---------------|
| CLI `--nextn-accept-rates` (here) | `0.85,0.3,0,0,0` |
| `build_default_task_configs` | `[0.85, 0.3, 0.0, 0.0, 0.0]` |
| `task.py` `_base_common_layer` | `[0.85, 0, 0, 0, 0]` |

The CLI path has **0.3** as the 2nd accept rate, while `task.py` has **0**. This means the CLI path and the YAML/webapp path will produce different MTP behavior for the same model. Please unify all defaults to a single source of truth.

src/aiconfigurator/cli/main.py
```diff
@@ -624,12 +642,24 @@ def build_default_task_configs(
         "database_mode": database_mode,
     }
 
+    # Create yaml_config to pass nextn and nextn_accept_rates if specified
+    yaml_config = None
+    if nextn > 0:
+        yaml_config = {
```
xutizhou, line 647: 🟡 **Redundant double fallback**

Line 613 already does `nextn_accept_rates = nextn_accept_rates or [0.85, 0.3, 0.0, 0.0, 0.0]`, so by this point `nextn_accept_rates` is guaranteed non-empty. The `or` here will never trigger. Suggest removing the redundant fallback.

src/aiconfigurator/sdk/models.py
```diff
@@ -571,12 +571,19 @@ class LLAMAModel(BaseModel):
...
     Due to implementation, attn layer name needs to be context_attention or generation_attention,
     exact match is required. Same for logits_gemm.
-    Other than DS V3, all other models don't support mtp
+    Supports MTP (Multi-Token Prediction) speculative decoding simulation.
     """
 
     def __init__(self, *args) -> None:
         super().__init__(*args)
-        assert self._nextn == 0, "Only DS V3 supports mtp"
+
```
xutizhou, line 594: 🟡 **Missing `nextn == 0` fast path (inconsistent with DeepSeekModel)**

DeepSeekModel has an `if self._nextn > 0 else 1.0` guard, but LLAMAModel and MOEModel don't. While the formula does evaluate to 1.0 when `nextn=0` (mathematically correct), adding the same guard would be more consistent and avoid unnecessary computation:

```python
self._mtp_scale_factor = (
    1.0
    / (1 + calc_expectation(self._nextn, self._nextn_accept_rates))
    * (self._nextn + self._num_layers)
    / self._num_layers
    if self._nextn > 0
    else 1.0
)
```

tests/unit/sdk/models/test_mtp_scaling.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        try:
+            model_config = self._create_model_config(nextn=2)
+            # Use a known MOE model or skip if not available
+            model = models.get_model("mistralai/Mixtral-8x7B-v0.1", model_config, "trtllm")
+
+            # Verify model was created successfully with nextn > 0
+            assert model is not None
+            assert model.config.nextn == 2
+            assert hasattr(model, '_mtp_scale_factor')
```
xutizhou: 🟡 **Bare `except Exception` + `pytest.skip` will mask real MTP bugs**

If the MOE model fails to instantiate due to an actual MTP-related bug, this will silently skip instead of failing the test. Suggest catching only specific environment/config exceptions (e.g. `FileNotFoundError`, `KeyError`) rather than all `Exception`.

Same concern for `test_mtp_scale_factor_exists_for_all_models` where the DeepSeek block uses `except Exception: pass` — this completely swallows errors. At minimum, narrow the exception type or log a warning.

tests/unit/sdk/models/test_mtp_scaling.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Test that mtp_scale_factor is 1.0 when nextn=0 (MTP disabled).
+
+        Backward compatibility: nextn=0 should produce identical results as before.
+        """
+        model_config_zero = self._create_model_config(nextn=0)
+        model_zero = models.get_model("Qwen/Qwen3-32B", model_config_zero, "trtllm")
+
+        # When nextn=0, mtp_scale_factor should be 1.0 (no scaling)
+        assert model_zero._mtp_scale_factor == 1.0, \
```
xutizhou: 🟡 **Insufficient test coverage — missing verification of actual generation ops scaling**

Current tests only verify that `_mtp_scale_factor` attribute exists and `calc_expectation` computes correctly. They don't verify that generation ops are actually scaled. Suggest adding:

- Create the same model with `nextn=1` and `nextn=0`, compare a generation op's `num_layers` parameter to verify it's correctly multiplied by `_mtp_scale_factor`
- Verify that context ops are **not** scaled by MTP (regression test for the P2P bug fix)

src/aiconfigurator/sdk/task.py
tianhaox: i think it's safe to set  this to 0. but we need to derive the nextn from the huggingface config. for deepseek, it's 1. just like what we did for quantization.

tianhaox: https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/config.json#L31

src/aiconfigurator/cli/main.py
tianhaox: suggest using [0.85, 0.3, 0.0, 0.0, 0.0]

src/aiconfigurator/sdk/task.py
tianhaox: a quick example, 

Image

process: apply_base_common_layer -> yaml config patch (this is treated as the overide option, set other value to overide the default nextn inferred from huggingface config)

FYI

xutizhou (approved):
LGTM. All previous review comments have been addressed:

- DeepSeek retains default `nextn=1`; other models default to 0
- `nextn_accept_rates` unified to `[0.85, 0.3, 0.0, 0.0, 0.0]` across all locations
- Redundant fallback removed
- `nextn==0` fast path added for LLAMAModel/MOEModel
- Test exception handling narrowed to specific types
- New tests verify generation ops scaling and context ops non-scaling (P2P regression test)

CI all green. Nice work on the fix commit.

Arsene12358 (approved):

PR #507, opened 2026-03-04
feat: Refine per-phase KV cache length estimation in run_agg

joshuayao:
#### Overview:

We would like to propose a small refinement to the KV cache length proxy used when estimating
decode-step latency inside `VLLMBackend.run_agg()`. The current approach uses a single
**global midpoint** (`isl + osl // 2`) for both the mixed-phase and the generation-only-phase. 
It would be great to use a **per-phase midpoint** could bring the predicted TPOT closer
to measured values.

#### Details:

To estimate the decode latency for one step in each phase, both
`_get_mix_step_latency` and `_get_genonly_step_latency` each call `run_static` with
`osl=2` at a representative KV cache length. The choice of
that representative length directly affects TPOT accuracy.

---

##### Current behavior

Both phases use the same proxy:

```python
RuntimeConfig(..., isl=isl + osl // 2, osl=2)
```

`osl // 2` is the midpoint of the **entire** output sequence, regardless of which phase
is being estimated.

---

##### Proposed refinement

Use the midpoint of each respective phase instead:

**`_get_mix_step_latency`** — mix steps cover decode steps `1..num_mix_steps`, so the
representative KV length is `isl + num_mix_steps // 2`:

```python
# before
RuntimeConfig(..., isl=isl + osl // 2, osl=2)

# after
RuntimeConfig(..., isl=isl + num_mix_steps // 2, osl=2)
```

**`_get_genonly_step_latency`** — genonly steps cover decode steps
`num_mix_steps+1..osl`, so the representative KV length is
`isl + num_mix_steps + num_genonly_steps // 2`:

```python
# before
RuntimeConfig(..., isl=isl + osl // 2, osl=2)

# after
RuntimeConfig(..., isl=isl + num_mix_steps + num_genonly_steps // 2, osl=2)
```

---

- src/aiconfigurator/sdk/backends/*_backend.py, and search for ", osl=2".

- closes GitHub issue: n/a

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved latency calculation accuracy across backend implementations for more reliable performance estimation.

### DIFF
--- a/src/aiconfigurator/sdk/backends/trtllm_backend.py
+++ b/src/aiconfigurator/sdk/backends/trtllm_backend.py
@@ -146,7 +146,7 @@ class TRTLLMBackend(BaseBackend):
                     summary = self.run_static(
                         model,
                         database,
-                        RuntimeConfig(batch_size=num_tokens, beam_width=1, isl=isl + isl + num_mix_steps // 2, osl=2),
+                        RuntimeConfig(batch_size=num_tokens, beam_width=1, isl=isl + num_mix_steps // 2, osl=2),
                         mode="static_gen",
                     )
                     latency_dict = summary.get_generation_latency_dict()


src/aiconfigurator/sdk/backends/trtllm_backend.py
```diff
@@ -138,15 +138,15 @@ def _get_mix_step_latency(
...
+                # third pass to get generation attn. use isl + num_mix_steps // 2 for avg generation attn latency.
                 gen_attention_latency_ms = 0.0
                 gen_attention_energy_wms = 0.0  # RENAMED from gen_attention_power_weighted
                 if gen_tokens > 0:
                     num_tokens = gen_tokens
                     summary = self.run_static(
                         model,
                         database,
-                        RuntimeConfig(batch_size=num_tokens, beam_width=1, isl=isl + osl // 2, osl=2),
+                        RuntimeConfig(batch_size=num_tokens, beam_width=1, isl=isl + isl + num_mix_steps // 2, osl=2),
```
Arsene12358: I think this is incorrectly resolved, shouldn't it be
```pytohn
RuntimeConfig(batch_size=num_tokens, beam_width=1, isl=isl + num_mix_steps // 2, osl=2)
```
based on the intent of the PR ?

joshuayao: Thanks for the review.  Fixed the issue.

PR #533, opened 2026-03-06
feat: vLLM 0.16.0 compatibility

ilyasher:
Also enable collect_comm.sh to be run from any cwd

### DIFF
--- a/collector/vllm/collect_attn.py
+++ b/collector/vllm/collect_attn.py
@@ -331,7 +331,6 @@ def run_attention_torch(
             output=output,
         )
 
-    # Use benchmark_with_power context manager.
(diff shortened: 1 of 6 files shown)


collector/vllm/collect_attn.py
```diff
@@ -410,6 +421,10 @@ def get_context_attention_test_cases(if_unit_test=False):
     kv_cache_dtype_list = [False]
     if get_sm_version() > 86:
         kv_cache_dtype_list.append(True)
+    # Flash Attention 2 lacks SM100+ (Blackwell) kernels and crashes/hangs.
```
tianhaox, line 424: not a problem here, but a question, is attention using trtllm kernels from flashinfer?

tianhaox (approved):

PR #537, opened 2026-03-06
feat: moe extrapolation according to MFU

YijiaZhao:
MOE extrapolation according to MFU
reduce max_tokens for MOE collector

<= 16384 tokens
16384_tokens

65546 tokens
output_origin

## Summary by CodeRabbit

* **New Features**
  * Added automatic performance estimation for queries exceeding collected token data.

* **Improvements**
  * Optimized memory management during MOE computations.
  * Enhanced profiling cache compatibility with rank parameters.

* **Tests**
  * Expanded test coverage for boundary and overflow scenarios.


tools/sanity_check/validate_database.ipynb
```diff
@@ -691,6 +691,15 @@
     "                    num_experts = 256\n",
     "                    hidden_size = 7168\n",
     "                    inter_size = 2048\n",
+    "\n",
```
tianhaox, line 694: why not leverage quantmode.mem

collector/trtllm/collect_moe_v3.py
tianhaox, line 4: is this really compatible with <1.2.0rc7, you are specifying rank to load cache, when is this introduced? can you check when this api change is introduced and make it compatible. if we think this compatible code is easy enough, you can keep it in v3, otherwise, let's write a new v4 collector

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -4476,6 +4476,74 @@ def get_empirical(
...
+            if isinstance(last_point, dict):
+                last_latency = float(last_point["latency"])
+                last_power = float(last_point.get("power", 0.0))
+                last_energy = float(last_point.get("energy", 0.0))
+            else:
+                last_latency = float(last_point)
+                last_power = 0.0
+                last_energy = 0.0
+
+            if last_latency <= 0:
```
tianhaox: why this can happen?

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -4539,6 +4607,20 @@ def get_silicon():
                     moe_dict = moe_data[quant_mode][used_workload_distribution][topk][num_experts][hidden_size][
                         inter_size
                     ][moe_tp_size][moe_ep_size]
+                    overflow_estimate = _estimate_overflow_with_last_token_util(
```
tianhaox: can we do 
```
result = xxx
if result is none:
    previous path

return result finally.
```
so that you don't shortcut the return path.

tianhaox: same for antoher two

ilyasher: why is it bad to return early? it prevents indenting the previous path.

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -4476,6 +4476,74 @@ def get_empirical(
...
+            moe_dict: dict,
+            hidden_size: int,
+            inter_size: int,
+            topk: int,
+            num_experts: int,
+            moe_tp_size: int,
+            moe_ep_size: int,
+            quant_mode: common.MoEQuantMode,
+            workload_distribution: str,
+        ) -> PerformanceResult | None:
```
ilyasher: Nit: I think it makes more sense to just return `PerformanceResult` (never `None`), and only call this function when `query_tokens > token_points`

ilyasher (approved):
Can you also add unit tests for the new `query_moe` behavior? ?

src/aiconfigurator/sdk/perf_database.py
tianhaox: it's more like a correction. but either is okay. just my personal opinion. Perhaps the more reasonable way is to make it in the extrapolation func by passing sol func. But it can be in a seperate PR.

src/aiconfigurator/sdk/perf_database.py
tianhaox: looks like current token count verification is inside this func. return none if no need to do the correction or fail to do the correction

tools/sanity_check/validate_database.ipynb
YijiaZhao: already use weight_bits = int(quant_mode.value.memory * 8)

collector/trtllm/collect_moe_v3.py
YijiaZhao: it could check "rank" parameters after 2rd commit
  if "rank" in inspect.signature(load_cache).parameters:
      load_cache(json_path, rank=device.index)
  else:
      load_cache(json_path)

src/aiconfigurator/sdk/perf_database.py
YijiaZhao: already deleted it

src/aiconfigurator/sdk/perf_database.py
YijiaZhao: I use query_tokens > token_points in the 2rd commit

src/aiconfigurator/sdk/perf_database.py
YijiaZhao: I use a early return branch for num_tokens_corrected > token_points[-1] in 2rd commit

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -4476,6 +4476,67 @@ def get_empirical(
...
+            )[0]
+
+            util = max(sol_last / last_latency, 1e-8)
+            est_latency = sol_query / util
+
+            est_energy = 0.0
+            if last_power > 0:
+                est_energy = last_power * est_latency
+            elif last_energy > 0:
+                est_energy = last_energy * (est_latency / last_latency)
```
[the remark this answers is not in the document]
YijiaZhao: it never possible for  last_latency <0

Arsene12358 (approved):

collector/trtllm/collect_moe_v2.py
```diff
@@ -239,11 +239,14 @@ def run_moe_torch(
...
-                power_law_logits_v3(num_tokens, num_experts, topk, moe_ep_size, power_law_alpha).to(router_logits_dtype)
+                power_law_logits_v3(num_tokens, num_experts, topk, moe_ep_size, power_law_alpha)
+                .to(device=device, dtype=router_logits_dtype)
                 for _ in range(num_iter)
             ]
         elif distributed == "balanced":
-            actual_logits = balanced_logits(num_tokens, num_experts, topk).to(router_logits_dtype)
+            actual_logits = balanced_logits(num_tokens, num_experts, topk).to(
+                device=device, dtype=router_logits_dtype
+            )
```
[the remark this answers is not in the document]
YijiaZhao: done

src/aiconfigurator/sdk/perf_database.py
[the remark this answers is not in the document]
YijiaZhao: done

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -4539,6 +4599,20 @@ def get_silicon():
...
+                            moe_dict,
+                            hidden_size,
+                            inter_size,
+                            topk,
+                            num_experts,
+                            moe_tp_size,
+                            moe_ep_size,
+                            quant_mode,
+                            workload_distribution,
+                        )
```
Arsene12358, line 4615: This logic seems duplicated for each backend, can we consolidate the logic in one place like
```python
def _maybe_extrapolate(num_tokens, moe_dict, hidden_size, inter_size,
                       topk, num_experts, moe_tp_size, moe_ep_size,
                       quant_mode, workload_distribution):
    token_points = sorted(moe_dict.keys())
    if num_tokens > token_points[-1]:
        return _estimate_overflow_with_last_token_util(
            num_tokens, moe_dict, hidden_size, inter_size,
            topk, num_experts, moe_tp_size, moe_ep_size,
            quant_mode, workload_distribution,
        )
    return None
```
And in each backend branch, we can do 
```python
result = _maybe_extrapolate(...)
if result: return result
```

PR #550, opened 2026-03-07
fix: wideep decode PDL gap close 

liyuanzhe1991:
#### Overview:

Add PDL (Programmatic Dependent Launch) discount modeling for GB200 WideEP DeepSeek decode phase to correct TPOT overestimation.

SM100 (GB200) supports PDL which eliminates kernel-to-kernel launch gaps, reducing per-layer wall time by ~6.5%. Previously AIC used pure kernel-sum semantics with no PDL modeling, causing TPOT to be overestimated by ~7% vs nsys measurements.

#### Details:

- Add automatic PDL discount (factor=0.935) applied to per-layer ops during the generation phase
- Auto-activates only when **all** conditions are met: SM100+ GPU, trtllm backend, DeepSeek model family, WideEP enabled
- Factor calibrated from nsys alignment on DeepSeek-V3.1-NVFP4 (GB200, tp1pp1dp16ep16, BS32, ISL8192/OSL1024):
  `(nsys_tpot - non_layer) / (aic_tpot - non_layer) = 17.84 / 19.09 = 0.935`
- Non-layer ops (embedding, logits_gemm, p2p) are excluded from discount since they run outside the CUDA Graph or have no PDL benefit
- Supports yaml override via `gpu.pdl_discount_factor` for fine-tuning; if not set, uses built-in `_PDL_DEFAULTS` lookup table
- New defaults can be added to `_PDL_DEFAULTS` dict for other configurations without code changes elsewhere

**Validation**: TPOT deviation reduced from **+6.9%** to **+0.17%** (18.18ms predicted vs 18.15ms nsys measured).

- `src/aiconfigurator/sdk/backends/base_backend.py` — the only file changed. Look at `_PDL_DEFAULTS`, `_get_pdl_discount_factor()`, and the PDL application block inside `_run_generation()`.

#### Related Issues:

- Relates to: WideEP decode alignment with nsys measurements

### DIFF
--- a/src/aiconfigurator/sdk/backends/base_backend.py
+++ b/src/aiconfigurator/sdk/backends/base_backend.py
@@ -98,6 +98,48 @@ class BaseBackend(ABC):
 
             return context_latency_dict, context_energy_wms_dict
 
+        _NON_LAYER_OP_KEYWORDS = ("embedding", "logits_gemm", "p2p")
+
+        def _is_layer_op(op_name: str) -> bool:
+            """Check if an op is a per-layer op (not a per-step-only op like embedding/logits/p2p)."""
+            return not any(kw in op_name for kw in _NON_LAYER_OP_KEYWORDS)
+
+        # Default PDL discount factors per (sm_version, backend, model_family, wideep).
+        # Calibrated from nsys alignment: factor = nsys_layer_wall / aic_layer_prediction.
+        _PDL_DEFAULTS = {
+            # (sm_version_min, backend, model_family, wideep) -> factor
+            (100, "trtllm", "DEEPSEEK", True): 0.935,
+        }
+
+        def _get_pdl_discount_factor() -> float:
+            """Determine PDL discount factor from conditions or yaml override.
+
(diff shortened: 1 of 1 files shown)


src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -98,6 +98,48 @@ def _run_context(batch_size: int, isl: int, prefix) -> tuple[dict[str, float], d
 
             return context_latency_dict, context_energy_wms_dict
 
+        _NON_LAYER_OP_KEYWORDS = ("embedding", "logits_gemm", "p2p")
+
+        def _is_layer_op(op_name: str) -> bool:
```
tianhaox, line 103: why not just using a fixed pdl factor for trtllmwideepdeepseek model, and make it a new scale factor, in addition to numlayers and mtp.

liyuanzhe1991: got it, I'll make it simple.

PR #588, opened 2026-03-12
perf: cut Blackwell GEMM collection time via weight caching and adaptive L2 bypass

Harrilee:
perf: cut Blackwell GEMM collection time via weight caching and adaptive L2 bypass

Major:
- Adaptive outside_loop_count: replace fixed 5 weight copies with
  max(1, min(5, ceil(50MB / weight_bytes))), avoiding redundant
  allocations for shapes that already exceed GPU L2 cache
- Per-worker weight cache: single-entry LRU keyed by (gemm_type, n, k)
  reuses weight tensors across batch-size variants of the same shape
- Task reordering: sort from (x, n, k, type) to (n, k, type, x) so
  consecutive tasks share the same weight shape, maximising cache hits

Minor:
- Per-device perf files (gemm_perf_cuda{N}.txt) to avoid log_perf lock
  contention; merged back to canonical filename after collection
- Fix fp8/float16 UnboundLocalError: init group_size = None
- Stall detection: raise threshold to 120s, require all workers dead
  and queue empty before breaking the monitor loop
- SIGABRT handler uses os._exit(1) to prevent core dumps from
  DeepGEMM C++ destructor
- Monitor loop poll interval reduced from 2s to 0.5s

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Improved worker monitoring robustness with better stall detection and recovery mechanisms.
  * Enhanced signal handling for cleaner process termination.

* **Performance**
  * Faster monitoring responsiveness with more frequent status checks.
  * Optimized GEMM collection with weight caching to reduce memory overhead.

* **Features**
  * Automatic consolidation of per-device performance data into unified reports.

### DIFF
--- a/collector/trtllm/collect_gemm.py
+++ b/collector/trtllm/collect_gemm.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import ctypes
 import math
 from collections import defaultdict
 
@@ -38,6 +39,17 @@ def _compute_fp8_block_weight_scales(weight: torch.Tensor, group_size: int) -> t
 # Keyed by (gemm_type, n, k).  Only the most recently used entry is kept so
 # that GPU memory is not accumulated across (n, k) groups.
 _weight_cache: dict = {}
+
+
+def _get_l2_cache_bytes(device_id: int = 0) -> int:
+    """Query the GPU L2 cache size via the CUDA runtime API."""
+    cuda_dev_attr_l2_cache_size = 38
+    libcudart = ctypes.CDLL("libcudart.so")
+    value = ctypes.c_int()
+    ret = libcudart.cudaDeviceGetAttribute(ctypes.byref(value), cuda_dev_attr_l2_cache_size, device_id)
+    if ret != 0 or value.value <= 0:
+        raise RuntimeError(f"Failed to query L2 cache size (cudaError={ret}, value={value.value})")
+    return value.value
 
 
 def _build_weights(gemm_type: str, n: int, k: int, device, dtype, group_size, x) -> dict:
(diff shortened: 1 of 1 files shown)


collector/trtllm/collect_gemm.py
```diff
@@ -77,58 +136,44 @@ def run_gemm(gemm_type, m, n, k, perf_filename, device="cuda:0"):
         qc = QuantConfig(quant_algo=QuantAlgo.NVFP4, group_size=group_size)
     else:
         qc = None
+        group_size = None
 
-    outside_loop_count = 5  # to reduce impact of L2 cache hit
+    _l2_cache_bytes = 50 * 1024 * 1024  # H100 L2 = 50MB
```
ilyasher: What about other GPUs?

collector/helper.py
```diff
@@ -695,6 +696,52 @@ def log_perf(
             os.unlink(lock_file)
 
 
+def merge_per_device_perf_files():
```
tianhaox: Is this a critical perf issue? If not, we can keep it the same behavior as others.

Harrilee: No it's not. Just 8 processes competing to write to the same file. Just a little delay in file writing. Reverted.

collector/trtllm/collect_gemm.py
```diff
@@ -77,58 +131,44 @@ def run_gemm(gemm_type, m, n, k, perf_filename, device="cuda:0"):
...
+    _weight_bytes = n * k * 2  # BF16
+    outside_loop_count = max(1, min(5, math.ceil(_l2_cache_bytes / _weight_bytes)))
```
tianhaox, line 151: what about blackwell, is the L2 larger?

tianhaox: maybe we can set a safer value?

Harrilee: Changed to 192 MB (for GB300). 
I am comparing the latest perf data with the legacy ones.

collector/trtllm/collect_gemm.py
```diff
@@ -77,58 +131,44 @@ def run_gemm(gemm_type, m, n, k, perf_filename, device="cuda:0"):
         qc = QuantConfig(quant_algo=QuantAlgo.NVFP4, group_size=group_size)
     else:
         qc = None
+        group_size = None
 
-    outside_loop_count = 5  # to reduce impact of L2 cache hit
+    _l2_cache_bytes = 192 * 1024 * 1024  # B200/B300 L2 = 192MB (largest)
+    _weight_bytes = n * k * 2  # BF16
```
xutizhou: `_weight_bytes = n * k * 2` assumes BF16 for all types, but fp8 weights are 1 byte and nvfp4 is 0.5 — so `outside_loop_count` can end up too low to actually flush L2 for those types.

Harrilee: Great catch. Modified!

PR #619, opened 2026-03-19
fix: use single copy for activations & KV Cache, remove duplicate

YijiaZhao:
Overview:
Fix prefill memory estimation double-counting KV cache and activations.

Details:
During prefill, both activations and kvcache scale with batch_size * isl. The empirical activation coefficient (c_dict) already captures KV cache write overhead during the forward pass, so adding a separate kvcache term for the prefill portion results in over-estimation of GPU memory.

The fix splits KV cache into two components:

kvcache_prefill (batch_size * isl): overlaps with activations during prefill, so we use max(activations, kvcache_prefill) instead of summing both.
kvcache_decode (batch_size * beam_width * osl): does not overlap with activations (decode activations are tiny), added separately as before.
This change only affects the total field used for OOM checks. The breakdown fields (activations, kvcache) still report raw values for debugging.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced memory calculation precision in model configuration backends, resulting in more accurate memory allocation estimates across different inference phases.

### DIFF
--- a/src/aiconfigurator/sdk/backends/sglang_backend.py
+++ b/src/aiconfigurator/sdk/backends/sglang_backend.py
@@ -547,6 +547,17 @@ class SGLANGBackend(BaseBackend):
             * kvcache_per_token
         )
 
+        # Split kvcache into prefill and decode portions to avoid double-counting
+        # with activations. During prefill, activations and kvcache_prefill both scale
+        # with batch_size * isl and overlap in memory (the empirical activation coefficient
+        # already captures KV cache writes). During decode, activations are tiny and
+        # kvcache_decode is separate.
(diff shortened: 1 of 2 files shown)


src/aiconfigurator/sdk/backends/sglang_backend.py
```diff
@@ -547,6 +547,17 @@ def _get_memory_usage(
             * kvcache_per_token
```
tianhaox, line 547: what about vllm?

YijiaZhao: vllm call trtllm. this mr is not right, i need rewrite a pr

PR #621, opened 2026-03-19
fix: use single copy for activations & past KV Cache, remove duplicate

YijiaZhao:
Overview:
Fix memory estimation to account for prefix caching — prefix tokens were being counted in both activations and KV cache as if they were regular input tokens.

Details:
When prefix > 0, the memory estimator previously ignored it entirely, leading to over-estimation:

Activations over-counted: num_tokens defaulted to isl * batch_size, but prefix tokens have cached KV and don't need computation. Fixed to (isl - prefix) * batch_size.

KV cache over-counted: batch_size * isl allocated per-request KV for all context tokens, including the shared prefix. Added common_prefix_ratio (0-1, default 0) to control how much of the prefix is shared across the batch:

Shared portion: only 1 copy of KV (prefix * common_prefix_ratio)
Per-request portion: batch_size copies (isl - prefix * common_prefix_ratio)
Example impact (prefix=8000, isl=10000, bs=2, common_prefix_ratio=1):

Activations: 20000 tokens → 4000 tokens (-80%)
KV cache context: 20000 token-copies → 12000 token-copies (-40%)
When prefix=0 (default), behavior is identical to before.

Changes across all three backends (trtllm, sglang, vllm) and the base class.

aiconfigurator/src/aiconfigurator/sdk/backends/trtllm_backend.py — core changes in _get_memory_usage: new prefix and common_prefix_ratio parameters, updated num_tokens default and KV cache formula
aiconfigurator/src/aiconfigurator/sdk/backends/sglang_backend.py — same changes applied to SGLang backend
aiconfigurator/src/aiconfigurator/sdk/backends/base_backend.py — updated abstract method signature and all run_static callers to pass prefix
aiconfigurator/src/aiconfigurator/sdk/backends/vllm_backend.py — pass-through to TRTLLMBackend with new parameters
Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
Fixes: memory over-estimation when prefix caching is enabled — prefix tokens were fully counted in both activations and per-request KV cache

## Summary by CodeRabbit

* **Refactor**
  * Enhanced memory estimation logic across all inference backend implementations to improve accuracy of memory calculations.

### DIFF
--- a/src/aiconfigurator/sdk/backends/vllm_backend.py
+++ b/src/aiconfigurator/sdk/backends/vllm_backend.py
@@ -476,4 +476,6 @@ class VLLMBackend(BaseBackend):
         # TODO
         from aiconfigurator.sdk.backends.trtllm_backend import TRTLLMBackend
 
-        return TRTLLMBackend()._get_memory_usage(model, database, batch_size, beam_width, isl, osl, num_tokens, prefix)
+        return TRTLLMBackend()._get_memory_usage(
+            model, database, batch_size, beam_width, isl, osl, num_tokens, prefix=prefix
+        )


src/aiconfigurator/sdk/backends/vllm_backend.py
```diff
@@ -471,8 +471,9 @@ def _get_memory_usage(
         isl: int,
         osl: int,
         num_tokens: int = 0,
+        prefix: int = 0,
     ) -> dict[str, float]:
         # TODO
         from aiconfigurator.sdk.backends.trtllm_backend import TRTLLMBackend
 
-        return TRTLLMBackend()._get_memory_usage(model, database, batch_size, beam_width, isl, osl, num_tokens)
+        return TRTLLMBackend()._get_memory_usage(model, database, batch_size, beam_width, isl, osl, num_tokens, prefix)
```
Arsene12358: let's use `prefix=prefix` for consistency

Arsene12358 (changes requested):

src/aiconfigurator/sdk/backends/vllm_backend.py
YijiaZhao: done

Arsene12358 (approved):

PR #667, opened 2026-03-31
feat: QWEN3.5 support (hybrid mode)

Harrilee:
Silicon mode is also in the PR, but data collection will be in a separate PR.

## Manual Test

Test result on all 21 models https://huggingface.co/collections/Qwen/qwen35 on GB200.
17/21 models succeeded. 4/21 cases failed due to `GPTQ-INT4 not supported`.

## Note to Code Reviewers

`tests/unit/collector/test_version_resolver.py::TestRegistryIntegrity::test_module_exports_declared_functions[trtllm]` is failing because collector script (for SILICON mode) is not included.

## Release Notes

* **New Features**
  * Added support for Qwen3.5 model family including 27B, 35B-A3B, and 397B-A17B variants
  * Introduced GDN (Gated DeltaNet) operation support for efficient linear attention layers
  * Enabled hybrid attention mechanism combining linear and full attention within single models
  * Added performance profiling capabilities for GDN kernel operations across different hardware backends


src/aiconfigurator/model_configs/Qwen--Qwen3.5-27B_config.json
```diff
@@ -0,0 +1,135 @@
...
+            "linear_attention",
+            "linear_attention",
+            "linear_attention",
+            "full_attention"
+        ],
+        "linear_conv_kernel_dim": 4,
+        "linear_key_head_dim": 128,
+        "linear_num_key_heads": 16,
+        "linear_num_value_heads": 48,
+        "linear_value_head_dim": 128,
```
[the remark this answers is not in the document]
Harrilee: The finding is incorrect — both sides already use the same key name:

  - JSON (Qwen--Qwen3.5-27B_config.json:89): "linear_value_head_dim": 128
  - Parser (utils.py:602): config["linear_value_head_dim"]

  They match exactly. There is no mismatch, no zero value, and no fix needed.

src/aiconfigurator/model_configs/Qwen--Qwen3.5-35B-A3B_config.json
```diff
@@ -0,0 +1,115 @@
...
+            "linear_attention",
+            "linear_attention",
+            "linear_attention",
+            "full_attention"
+        ],
+        "linear_conv_kernel_dim": 4,
+        "linear_key_head_dim": 128,
+        "linear_num_key_heads": 16,
+        "linear_num_value_heads": 32,
+        "linear_value_head_dim": 128,
```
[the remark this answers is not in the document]
Harrilee: Same situation as the previous finding — both sides use linear_value_head_dim. The JSON (Qwen--Qwen3.5-35B-A3B_config.json:64) and the parser (utils.py:602) already agree on the key name. The finding is incorrect — no fix needed.

collector/sglang/registry.py
```diff
@@ -77,4 +77,10 @@
         get_func="get_wideep_moe_test_cases",
         run_func="run_wideep_moe",
     ),
+    OpEntry(
+        op="gdn",
+        module="collector.sglang.collect_gdn",
+        get_func="get_gdn_test_cases",
+        run_func="run_gdn_torch",
+    ),
```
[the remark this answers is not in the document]
Harrilee: But this ops registration is indeed needed right? Although we have not support SILICON model yet, we still need this ops for SOL mode. collector.sglang.collect_gdn will be ready later.

src/aiconfigurator/sdk/models.py
```diff
@@ -4362,6 +3585,394 @@ def _build_generation_ops(self) -> None:
...
+
+        if cfg.num_experts > 0:
+            assert (
+                self.config.tp_size * self.config.attention_dp_size == self.config.moe_tp_size * self.config.moe_ep_size
+            ), (
+                f"tp_size ({self.config.tp_size}) * attention_dp_size "
+                f"({self.config.attention_dp_size}) should equal moe_tp_size "
+                f"({self.config.moe_tp_size}) * moe_ep_size ({self.config.moe_ep_size})"
+            )
+            assert cfg.num_experts >= self.config.moe_ep_size
```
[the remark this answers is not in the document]
Harrilee: The <= 256 upper bound on tp_size × attention_dp_size is redundant and incorrect.

  Why redundant: The two assertions already present above it are sufficient:
  1. tp * attn_dp == moe_tp * moe_ep — balances the parallelism dimensions
  2. num_experts >= moe_ep — prevents more EP ranks than experts exist

  Together these imply tp * attn_dp = moe_tp * moe_ep ≤ moe_tp * num_experts. The product is already bounded by the model's expert count. A separate constant cap adds nothing.

  Why incorrect: 256 was the largest expert count in the codebase when the check was written (DeepSeek V3/R1). It is a coincidental data point, not a hardware or algorithmic limit. Qwen3.5-397B-A17B already has 512 experts — a valid configuration like moe_ep=512,
  moe_tp=1 would be rejected by the old check even though num_experts >= moe_ep passes cleanly. Any future model with >256 experts faces the same false rejection.

  Conclusion: The num_experts >= moe_ep assertion is the correct and principled constraint. The <= 256 constant was model-coverage folklore masquerading as a validation rule.

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -1339,6 +1339,78 @@ def load_mamba2_data(mamba2_file: str):
     return result
 
 
+def load_gdn_data(gdn_file: str):
```
tianhaox, line 1342: is there any quantization?

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -5300,6 +5374,120 @@ def _mamba2_gen_entry(val):
...
+            seq_len: sequence length (context only; unused for generation)
+            d_model, num_k_heads, head_k_dim, num_v_heads, head_v_dim, d_conv: GDN layer config
+
+        Returns:
+            PerformanceResult with latency (ms) and energy (W·ms).
+            Uses SOL-based fallback when gdn_perf data is not loaded.
+        """
+        gdn_data: dict = getattr(self, "_gdn_data", {})
+
+        def _sol_fallback() -> PerformanceResult:
```
tianhaox: I think we need to align GDN query definition with other ops, i.e., get_sol, get_empirical, etc.

src/aiconfigurator/sdk/models.py
```diff
@@ -4362,6 +4358,439 @@ def _build_generation_ops(self) -> None:
...
+    All layers share the same FFN:
+      - Dense models (27B):          SwiGLU dense FFN
+      - MoE models (35B-A3B, 397B): All-MoE FFN (num_experts > 0)
+    """
+
+    def __init__(self, *args) -> None:
+        super().__init__(*args)
+        cfg: common.Qwen35Config = self.extra_params
+        assert isinstance(cfg, common.Qwen35Config), "Qwen35Model requires Qwen35Config extra_params"
+        assert self._nextn == 0, "Qwen35Model does not support mtp"
```
tianhaox: I think qwen3.5 supports MTP. And the MTP's transformer layer is using full attention.

src/aiconfigurator/sdk/perf_database.py
Harrilee: Mamba does not have quant mode either. It might be a bug. Let me start a thread in Slack.

src/aiconfigurator/sdk/models.py
Harrilee: MTP support added.

src/aiconfigurator/sdk/perf_database.py
Harrilee: Unified. Also, renamed for Mamba.

src/aiconfigurator/sdk/perf_database.py
Harrilee: Both 3 framework use FP16/BF16 as IO for the GDN kernel (some internally use FP32 for internal calculation). But we should always use FP16 as IO to the gdn data query.

Arsene12358 (approved):

Arsene12358:
@coderabbitai review the recent commits, judge whether it's safe to merge

PR #691, opened 2026-04-03
fix: vLLM 0.17.0 collector + data (DSA, MLA, MoE)

simone-chen:
#### Overview:

Fix vLLM 0.17.0 collector compatibility for DSA module, MLA kernel, and MoE MXFP4 benchmarks on B200.

#### Details:

**DSA module collector** (`collect_mla_module.py`):

- **Deterministic weight/tensor init** — vLLM 0.17.0's FlashInfer sparse MLA backend ([vllm#33451](https://github.com/vllm-project/vllm/pull/33451)) and DSA CUDA graph support ([vllm#34457](https://github.com/vllm-project/vllm/pull/34457)) leave CUDA graph RNG offset tracking active after `DeepseekV2MLAAttention` construction. Any subsequent RNG operation crashes with `"Offset increment outside graph capture"`.
  - `enforce_eager` and `manual_seed()` do not clear the state — the corruption originates inside module construction
  - Replace all post-construction RNG (`normal_`, `uniform_`, `randn`, `randint`) with deterministic `fill_()`/`torch.full()`
  - Safe for benchmarking: kernel latency depends on shapes/dtypes, not values; dummy weights are overwritten by `process_weights_after_loading()` anyway
  - Filed upstream: [vllm#39371](https://github.com/vllm-project/vllm/issues/39371)
- **KV cache scale buffers** — vLLM registers `k_scale`/`v_scale` as buffers, not parameters. The init loop missed them, leaving sentinel values that fail `process_weights_after_loading()` (`k_scale > 0.0` assertion).
- **auto_map stripping** — DeepSeek-V3's `config.json` has `auto_map` pointing to `configuration_deepseek.py`. HuggingFace's `AutoConfig.from_pretrained()` (called by vLLM's `ModelConfig`) unconditionally tries to import it from the temp directory where it doesn't exist. Strip it; vLLM natively supports the architecture.
- **MLA backend selection** — vLLM 0.17.0 calls `get_current_vllm_config()` during `get_attn_backend_cls()`. Wrap in `set_current_vllm_config()` context.

**MLA kernel-level collector** (`collect_mla.py`):

- **Backend selection** — same `get_current_vllm_config()` issue as DSA module collector. Wrap backend selection in `set_current_vllm_config()` with a temporary config.

**MoE MXFP4 collector** (`collect_moe.py`):

- **Forward context** — vLLM 0.17.0's MoERunner abstraction ([vllm#32344](https://github.com/vllm-project/vllm/pull/32344)) routes `FusedMoE.forward()` through `get_forward_context()` → `get_layer_from_name()`, requiring the module to be registered in `static_forward_context`. Share the same `VllmConfig` between `FusedMoE.__init__` and the benchmark's `set_forward_context()` so the registration is visible.
- **pcp_size** — vLLM 0.17.0 added prefill context parallel to `FusedMoE` ([vllm#32344](https://github.com/vllm-project/vllm/pull/32344)). Pass `pcp_size=1` to avoid `get_pcp_group()` which requires distributed init.

**Data** — replace partial/dirty perf data with clean collection from [job 294842392](https://gitlab-master.nvidia.com/dl/ai-dynamo/aic-auto-collector/-/jobs/294842392) (0 DSA/MLA module errors). Adds previously missing `mla_context_module_perf.txt` and `mla_generation_module_perf.txt`.

#### Known limitations:

- 42 MoE MXFP4 `weight_scale_vec_size` errors — FlashInfer TRTLLM FP4 kernel rejects the weight format; likely needs FlashInfer-side fix
- 6 MoE MXFP4 test cases with `tp_size > 1` fail at `FusedMoE.__init__` — requires distributed init not available in standalone collector
- MLA kernel-level collector may have additional errors from builder `get_per_layer_parameters()` lookup after backend selection is fixed — to be investigated in follow-up if needed

`collector/vllm/collect_mla_module.py`


collector/common_test_cases.py
tianhaox: will this make it duplicate for moe collection?

simone-chen: Good catch. Thanks for pointing it out! Fixed.

simone-chen:
Superseded by #718 — clean v2 collector files with version routing instead of patching existing files.

PR #705, opened 2026-04-06
fix: Filter invalid vllm moe parallelism configs

mandy-li:
vLLM cannot set moe_tp independently from tp. When dp>1, only pure EP configs are valid. Filter out dp>1 + moe_tp>1 combinations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced configuration validation to eliminate incompatible parallel training setups, improving overall configuration quality and search space accuracy.

### DIFF
--- a/src/aiconfigurator/sdk/utils.py
+++ b/src/aiconfigurator/sdk/utils.py
@@ -160,8 +160,10 @@ def enumerate_parallel_config(
                                         not enable_wideep and moe_ep > 1
                                     ):  # wideep only has ep
                                         continue
-                                elif backend == common.BackendName.vllm:
-                                    pass  # TODO
+                                elif backend == common.BackendName.vllm and dp > 1 and moe_tp > 1:
+                                    # vLLM can't set moe_tp independently from tp;
+                                    # dp>1 with moe_tp>1 requires unsupported heterogeneous parallelism.
+                                    continue
                                 parallel_config_list.append([tp, pp, dp, moe_tp, moe_ep])
             else:
                 if tp * pp in num_gpu_list:


simone-chen (dismissed):
LGTM, thanks for the PR!

tianhaox:
need to double check whether it's true for GPU @Ethan-ES

mandy-li:
> need to double check whether it's true for GPU @Ethan-ES

Any feedback? 
Just provide details that this PR is to address the invalid config projection for moe model such as: tp=1, dp=2, etp=2, ep=1 which is not valid in vLLM (and there is no way to specify it as well)

mandy-li:
> LGTM, thanks for the PR!

The reason why i have `dp>1 and etp>1` check is because when dp>1, only ep>1 is allowed in vLLM. Maybe we are trying to address the same problem from different point. But if you have enough test cases to make sure moe_tp!=tp work as well, I don't mind to change it.

tianhaox:
Hi  Mandy, sorry for the late update, i think for vllm, it doesn't support moe_ep > 1 and moe_tp > 1 simultaneously.

however, it allows to set dp >1 and moe_ep >1.
if tp and dp are set to >1, for the moe part, moe_ep=tp\*dp if --enable-expert-parallel, otherwise, moe_tp=tp\*dp
this is what I have learnt from vllm source code. Is this aligned with your obersavation?

mandy-li:
> Hi Mandy, sorry for the late update, i think for vllm, it doesn't support moe_ep > 1 and moe_tp > 1 simultaneously.
> 
> however, it allows to set dp >1 and moe_ep >1. if tp and dp are set to >1, for the moe part, moe_ep=tp*dp if --enable-expert-parallel, otherwise, moe_tp=tp*dp this is what I have learnt from vllm source code. Is this aligned with your obersavation?

Hi Tianhao, your observation/understanding is perfectly correct. My filter is narrow comparing to what vllm source code indicate. I tried to modify my code, but found the problem has been fixed [by another PR](https://github.com/ai-dynamo/aiconfigurator/blob/09510979aa12ee586a7ec0041e375a24ac6e095d/src/aiconfigurator/sdk/utils.py#L165) which I agree. If you agree as well, pls feel free to close this PR. Thanks.

tianhaox:
Thanks for the update, let's close this.

PR #729, opened 2026-04-13
feat(encoder): add Qwen3-VL vision encoder modeling

sungwook-son:
Overview:

Adds end-to-end latency estimation for Qwen3-VL family vision-language models. This includes the full ViT encoder pipeline attention, FFN, PatchMerger MLP, and projection to LLM and correctly feeds post-merge image tokens into the LLM backbone ISL so that TTFT accounts for both encoder and prefill compute.

Details:

- common.py: Add VisionEncoderConfig dataclass to carry ViT architecture parameters (patch_size, depth, heads, hidden_size, spatial_merge_size, deepstack_visual_indexes). Register Qwen3VLForConditionalGeneration / Qwen3VLMoeForConditionalGeneration in MULTIMODAL_TEXT_CONFIG_KEY and ARCHITECTURE_TO_MODEL_FAMILY.
- config.py: Add image_height, image_width, num_images_per_request, num_image_tokens to RuntimeConfig.
- utils.py: Capture vision_config before the text_config unwrap so ViT parameters are not lost. Parse into VisionEncoderConfig and store in extra_params for Qwen3VL architectures, including deepstack_visual_indexes.
- models.py: Add Qwen3VLModel with 11 encoder ops covering the full ViT + PatchMerger pipeline. Fix PatchMerger from a single incorrect GEMM (k=1152, n=1) to the correct 3-op MLP (fc1/act/fc2, k=4608, n_mergers=4 for 32B).
- base_backend.py: Add _run_encoder() inside run_static() with a pre-merge / post-merge token split — ViT transformer ops use (H // patch_size)² tokens, the projection op uses post-merge count. img_ctx_tokens (post-merge × n_images) is added to  LLM ISL so context and generation phases reflect actual sequence length. TTFT = encoder latency + context latency.
- inference_summary.py: Surface encoder latency, energy, and power as first-class outputs via get_encoder_latency_dict().
- Model configs: Add cached HF config.json for Qwen3-VL 2B / 4B / 8B / 32B / 32B-Thinking / 30B-A3B / 235B-A22B.
- Tests: Add tests/unit/sdk/backends/test_encoder.py covering token count formula, PatchMerger shape, and encoder latency
integration (55 tests total, all passing).

1. src/aiconfigurator/sdk/backends/base_backend.py — _run_encoder() and the TTFT aggregation
2. src/aiconfigurator/sdk/models.py — Qwen3VLModel op list and PatchMerger shape
3. tests/unit/sdk/backends/test_encoder.py — validates token formulas and latency outputs

## Release Notes

* **New Features**
  * Added support for Qwen3-VL multimodal models (2B, 4B, 8B, 30B, 32B, and 235B variants)
  * Introduced vision encoder configuration and performance metrics for vision-language model inference
  * Added image and video input parameters to runtime configuration

* **Improvements**
  * Enhanced disaggregation reporting with improved MoE parallelism metadata
  * Extended inference summary to track encoder performance alongside context and generation phases

* **Tests**
  * Added comprehensive test coverage for vision encoder functionality and multimodal model configurations


src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -206,6 +206,81 @@ def run_static(
...
+                    runtime_config.image_width // img_stride
+                )
+                pre_merge_per_image = (runtime_config.image_height // enc_cfg.patch_size) * (
+                    runtime_config.image_width // enc_cfg.patch_size
+                )
+            elif runtime_config.num_image_tokens > 0:
+                tokens_per_image = runtime_config.num_image_tokens
+                pre_merge_per_image = runtime_config.num_image_tokens  # no merge info, use as-is
+            else:
+                tokens_per_image = isl
```
nqzhou2426: The new image workload fields are only added to `RuntimeConfig`, but the CLI/task/webapp call sites still construct `RuntimeConfig` with just `isl`/`osl`/SLA fields. For a Qwen3-VL run through the normal CLI path this falls into the `else` branch here and treats `isl` as the per-image token count, then adds that back into the LLM context length. That makes common requests unconfigurable and often wrong, e.g. a 448x448 Qwen3-VL image should be 196 post-merge tokens, but `--isl 512` is modeled as 512 image tokens. Can we plumb `image_height`/`image_width`/`num_images_per_request` or `num_image_tokens` through the public config paths, or avoid defaulting VL image tokens to `isl` when the image workload is unspecified?

sungwook-son: @nqzhou2426  Thanks for the feedback! 

Fixed in the latest commit: added --image-height, --image-width, and --num-images CLI flags that flow through build_default_task_configs() → TaskConfig → TaskContext → RuntimeConfig → backend. 

Also removed the fallback in _run_encoder().
If no image dimensions are provided, encoder modeling is now skipped cleanly rather than producing a
misleading estimate.

Let me know if a rebase is needed or if you'd like any other changes.

collector/sglang/collect_attn.py
```diff
@@ -157,30 +157,33 @@ def get_context_attention_test_cases():
...
-        for s in sorted(s_list, reverse=True):
-            for b in sorted(b_list, reverse=True):
-                for n_kv in n_kv_list:
-                    if n_kv != 0 and (n_kv >= n or n % n_kv != 0):
-                        continue
-                    num_kv_heads = n_kv if n_kv != 0 else n
-
-                    if num_kv_heads == n:
-                        if b * s > 65536 or b > 128:
+    head_dim_list = [72, 128]
```
tianhaox: I think for vision model, what's needed is
1. context phase only (no need to run decode phase)
2. MHA (n == n_kv, you can filter out all MQA and GQA cases)
3. fp16 (compute+kvcache, so you can skip some quant combination)
4. head_dim 72 (this dimension can be merged into this line)
5. **non-causal**, this dimension needs to be added, also adding is_causal (0, or 1) to the perf file to distinguish causal attention for llm and non-causal attention for ViT (and future DiT)
Do you think this makes sense?

tianhaox: better to open a new issue to track the fix of the theoretical calculation for non-causal attention in perf_database.py.

tianhaox: maybe you can build a dedicated loop to append test cases for vision part.

src/aiconfigurator/cli/main.py
```diff
@@ -151,32 +148,25 @@ def _add_default_mode_arguments(parser):
...
+        help="Image width in pixels for vision-language models. Default: 0 (disabled).",
     )
     parser.add_argument(
-        "--strict-sla",
-        action="store_true",
-        default=False,
-        help="Filter the Pareto frontier and best configs to only SLA-compliant data points "
-        "(--ttft + --tpot, or --request-latency). Without this flag, the Pareto frontier "
-        "may include configs that exceed --tpot.",
+        "--num-images", type=int, default=1, help="Number of images per request for vision-language models. Default: 1."
```
tianhaox, line 243: do you think it's better to use num-images to control disable/enable?  and giving height and width a default value, >0

tianhaox:
Is generator considered? If not, maybe we add an issue to follow up. cc @Ethan-ES  as well who is the mainainer of generator module

src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -215,58 +287,85 @@ def run_static(
...
+        context_latency_dict, context_energy_wms_dict = {}, {}
+        generation_latency_dict, generation_energy_wms_dict = {}, {}
 
         if mode == "static_ctx":
-            memory = self._get_memory_usage(model, database, batch_size, beam_width, isl, 1, prefix=prefix)
+            encoder_latency_dict, encoder_energy_wms_dict, img_ctx_tokens = _run_encoder(batch_size)
+            context_latency_dict, context_energy_wms_dict = _run_context(batch_size, isl + img_ctx_tokens, prefix)
+            memory = self._get_memory_usage(model, database, batch_size, beam_width, isl + img_ctx_tokens, 1)
         elif mode == "static_gen":
+            _, _, img_ctx_tokens = _run_encoder(batch_size)
```
tianhaox: will this path still query the db to get the latency? is there any short cut to return the img_ctx_tokens only

src/aiconfigurator/sdk/models.py
```diff
@@ -900,6 +936,126 @@ def __init__(self, *args) -> None:
...
+        tp_size = self.config.tp_size
+        depth = encoder_config.depth
+        h_vit = encoder_config.hidden_size
+        n_vit = encoder_config.num_heads
+        inter_vit = encoder_config.intermediate_size
+        h_llm = encoder_config.out_hidden_size
+        head_size_vit = h_vit // n_vit  # 1152 // 16 = 72
+        n_mergers = 1 + len(encoder_config.deepstack_visual_indexes)
+
+        if tp_size > 1:
```
tianhaox: In currrent design, is vision module using the same tp setting as LLM part? what about in real deployment.

src/aiconfigurator/sdk/models.py
```diff
@@ -900,6 +936,126 @@ def __init__(self, *args) -> None:
...
+            if n_vit % tp_size != 0:
+                raise ValueError(f"ViT num_heads ({n_vit}) must be divisible by tp_size ({tp_size})")
+            if inter_vit % tp_size != 0:
+                raise ValueError(f"ViT intermediate_size ({inter_vit}) must be divisible by tp_size ({tp_size})")
+
+        # ViT always runs in float16 regardless of LLM quantization settings
+        vit_gemm_mode = common.GEMMQuantMode.float16
```
tianhaox: latest code is using bfloat16 instead

src/aiconfigurator/sdk/models.py
```diff
@@ -1224,6 +1380,119 @@ def _validate_fp8_block_quantized_moe_config(self) -> None:
             )
 
 
+class Qwen3VLMoEModel(MOEModel):
```
tianhaox: what about we don't extend the class defintion. We attach the vision module explicitly at the end of get_model(), because BaseModel.extra_params is already used by various text architectures for their own extra parameters. In this case, we attach **encoder_config** and **encoder_ops** to the model to build. In base class, we can have these two variables defined but set to None.
In this case, you don't have to create two classes for, say, kimi 2.6 with/without vision module. What do you think?

src/aiconfigurator/webapp/events/event_fn.py
```diff
@@ -228,15 +228,17 @@ def run_estimation_static(
         stderr_text = stderr_buffer.getvalue()
         if is_error:
             return (
```
tianhaox, line 242: in static_tab.py,

Image

you need a quick fix to avoid crash.

tianhaox:
I've finished the 1st round of review. What a huge feature, thanks for the great PR. 👍 In additon to the comments in place, suggest adding a TODO list for future followup (in seperate PRs),
1. agg mode
2. EPD rate matching
3. webapp to support vision model
4. support other vision models such as KimiK2..6, etc.
5. other GPUs verification
6. Generator

sungwook-son:
Thank you @tianhaox for the thorough review.

Acknowledging all feedback:

In progress:
- `vit_gemm_mode`: will update `float16` to `bfloat16` per latest Qwen3-VL
- UI error routing (`event_fn.py`): will fix error message going to encoder box instead of summary box
- Encoder weights PP normalization: will move encoder weight addition before `weights /= pp_size` in both trtllm and sglang backends
- `deepstack_visual_indexes test assertion: will add to test_utils.py
- Placeholder encoder `None` locals in `run_disagg()`: will clean up

On TP for the vision encoder (`models.py` line 967):
In practice both vLLM and SGLang use DP for the ViT. The current modeling uses the same tp_size as the LLM, which is a simplification. The allreduce ops after encoder_proj_gemm, encoder_ffn2_gemm, and encoder_merger_fc2 are already present in the branch. For the scope of this PR, I'd like to keep the encoder TP consistent with the LLM TP as a starting point. **Post-merge, would like to discuss with the Dynamo team how to extend the P/D search to E/P/D(/G)**. 

On architecture, attaching encoder in get_model() instead of subclassing (`models.py` line 1383):
Great suggestion and aligns well with future models (Gemma3, InternVL3, etc.) that may share the same LLM backbone. Will consider restructuring in a follow-up.

On `--num-images` toggle: will address the design (with positive defaults for `--image-height`/`--image-width`) in a follow-up PR.

On `base_backend.py` line 300 : the early return exits before any op.query() calls are made, so no DB queries happen when image dims are missing.

Collector (collect_attn.py): agreed on all points. 
Opened #1028 to track:
- Adding `is_causal` column to attention perf files
- Dedicated vision attention test case loop in `collect_attn.py` (context-only, MHA, fp16, head_dim=72, non-causal)
- Fixing theoretical SOL in perf_database.py for non-causal attention

src/aiconfigurator/sdk/models.py
sungwook-son: Fixed: vit_gemm_mode corrected to bfloat16

src/aiconfigurator/cli/main.py
sungwook-son: Fixed: num_images=0 disables encoder; height/width have positive defaults

src/aiconfigurator/webapp/events/event_fn.py
sungwook-son: Fixed: error messages now route to the correct output box.

collector/sglang/collect_attn.py
sungwook-son: Tracked in #1124 (non-causal attention SOL fix in perf_database.py).

src/aiconfigurator/sdk/models.py
sungwook-son: The current approach uses build_encoder_ops() (a module-level function in vit_ops.py) which keeps the logic out of individual model classes. A get_model() attachment approach is a larger refactor.

sungwook-son:
Thanks @tianhaox for the thorough review.

Fixed in follow-up commits:
- bfloat16 correction — vit_gemm_mode updated from float16 → bfloat16 per Qwen3-VL spec
- --num-images as enable/disable toggle — num_images=0 disables encoder; image_height/image_width now have positive defaults
- Image params plumbed end-to-end — --image-height, --image-width, --num-images flow through CLI → RuntimeConfig → all backends (addresses @nqzhou2426's feedback)
- effective_isl = isl + img_ctx_tokens — context and generation phases now account for image tokens in the LLM backbone; encoder node memory computed separately via _get_encoder_node_memory()
- TP all-reduce ops — all 10 ViT transformer ops and 4 projector ops include correct CustomAllReduce placements; build_encoder_ops() in vit_ops.py raises ValueError if num_heads or intermediate_size is not divisible by tp_size
- Generic encoder op builder — replaced duplicated per-class op construction with a module-level build_encoder_ops(enc_cfg, tp_size) in vit_ops.py; extensible to Gemma3, InternVL3, Pixtral, Llama4 via projector_dims config field
Tracked as follow-up issues:
- Non-causal attention SOL fix in perf_database.py → #1028 #1092
- Generator module support for VL models → #1125

On the get_model() attachment suggestion: the current approach uses a module-level build_encoder_ops() function which keeps encoder op construction out of individual model classes. Happy to discuss whether a get_model()-level attachment is preferred — can be a follow-up refactor if so.

wenpengw-nv:
@sungwook-son  Thanks for the great contribution! I'm currently working on the non-causal attention fix in #1092 — it's still in progress, but I'd suggest we go ahead and merge your PR first (after review) to avoid blocking, and update the corresponding attention ops in a follow-up.

Also, we're planning to bring Omni and DiT support to aic. — do you have similar needs or plans on your side?
cc @tianhaox @Arsene12358

Arsene12358 (approved):

Arsene12358:
I think we are fine to merge this after the conflicts are resolved.

sungwook-son:
Hi @wenpengw-nv, DiT and Omni is in our interest.
cc. @tthakkal, @dsocek

Running final rebase for easier merge.

Arsene12358:
@sungwook-son I think it needs a quick lint auto-fix also another rebase after the merge of https://github.com/ai-dynamo/aiconfigurator/pull/1134. I will prioritize merging this PR after these are resolved, thanks

Arsene12358:
BTW @sungwook-son I saw you are also active in vLLM's dev slack channel(https://slack.vllm.ai/), I think we can also use slack to sync over stuff if needed

sungwook-son:
Appreciate the review and the pointer to #1134. That's a significant change I need to account for. Working on the rebase now and will follow up shortly.

Arsene12358:
Looks clean now, thanks for contributing. I will merge this now as a starting point for multimodal support in aiconfigurator. Looking forward to keep building on top of this. Thanks.

PR #749, opened 2026-04-14
feat: add bfloat16 as xpu default dtype

Spycsh:
#### Overview:

Intel XPU natively supports bfloat16, whereas float16 requires an additional conversion step. We need to make changes to the collector and CLI projection code to support this.

#### Details:

- [x] add bfloat16 as default for xpu collector and cli projection
- [x] add "device" abstraction in `ModelConfig` for differentiating between logits/router quant mode defaults
- [x] add vllm 0.17.0 data

NA

NA

## Summary by CodeRabbit

* **New Features**
  * Added bfloat16 quantization support for XPU systems.
  * Introduced explicit device configuration in model settings.
  * Added performance benchmarking data for b60 XPU system.

* **Chores**
  * Centralized XPU system definitions for consistency across tools and tests.
  * Updated default precision handling for XPU-based inference.


Arsene12358 (approved):

tianhaox:
i think for GPU, the actual default value is bfloat16 as well. This is a legacy problem. Do you think whether it's safe to use float16 for now in aic-sdk, while use bfloat16 for collector. We can create a seperate task to renaming all the existing float16 dtype in aic-sdk as well as the perf files so that you don't have to maintain a standalone path/if-else branch for your code.

Spycsh:
@tianhaox  if we collect bfloat16 data as default for cuda/xpu, then all the op data files collected under `systems/data` will be `bfloat16` so the aic-sdk logics should be able to capture that change. Will you also refactor all `float16` defaults to `bfloat16` under aic-sdk for cuda path, and also make it compatible to your cuda op perf data files which already has  `float16` as default? If yes, sure I can remove my changes under aic-sdk in this pr and only keep the collector part for xpu.

tianhaox:
perhaps i will rename all float16 to bfloat16 across the whole project including all collectors on GPU and XPU and maybe even remove float16. Does that make sense and solve the problem? In this case, you don't have to maintain the different precision path for XPU

Spycsh:
> perhaps i will rename all float16 to bfloat16 across the whole project including all collectors on GPU and XPU and maybe even remove float16. Does that make sense and solve the problem? In this case, you don't have to maintain the different precision path for XPU

Yes, it does solve the problem. Will close this pr after your change.

Spycsh:
> > perhaps i will rename all float16 to bfloat16 across the whole project including all collectors on GPU and XPU and maybe even remove float16. Does that make sense and solve the problem? In this case, you don't have to maintain the different precision path for XPU
> 
> Yes, it does solve the problem. Will close this pr after your change.

Link to https://github.com/ai-dynamo/aiconfigurator/pull/895.

PR #751, opened 2026-04-14
feat: enable system/backend-specific empirical correction

joshuayao:
#### Overview:

This PR introduces a configurable empirical-correction mechanism for performance projection, replacing hardcoded correction math with system/backend-specific YAML overrides while preserving safe fallback behavior. If no override exists (or an override is invalid), the existing default formulas are still used. So, no user behavior changes and no correction formula/factor changes for now.

#### Details:

- Added a reusable resolver to load and evaluate empirical correction expressions from:
`systems/empirical_correction/<system>/<backend>/<version>/empirical_correction.yaml` (this YAML file is not required, fallback to default formula if it does exist). Sample file:
```yaml
corrections:
  tpot_num_mix_steps_for_tpot_calc: "max(1, num_mix_steps - 3)"
  ttft_correction_factor: "min(2 + (steps_to_finish_ctx - 3) / 2 / 10, 4)"
```
- Implemented safe expression evaluation (AST-based) with a restricted function/operator set.
Integrated the resolver into vLLM backend estimation flow (I can apply this mechanism to other backends if needed) for the two correction points currently in use (we can apply it to more points if needed):
  - ttft_correction_factor
  - tpot_num_mix_steps_for_tpot_calc
- Kept backend-specific default formulas in code as fallback to ensure backward-compatible behavior.
- Updated advanced tuning docs to introduce this mechanism.

- src/aiconfigurator/sdk/empirical_correction.py
- src/aiconfigurator/sdk/backends/vllm_backend.py
- docs/advanced_tuning.md

- closes GitHub issue: n/a

## Summary by CodeRabbit

* **New Features**
  * Introduced customizable empirical corrections via YAML configuration files, allowing fine-tuning of correction factors per system/backend/version combination.

* **Documentation**
  * Added advanced tuning guide documenting empirical correction customization, supported expressions, and safe fallback behavior.

### DIFF
--- a/docs/advanced_tuning.md
+++ b/docs/advanced_tuning.md
@@ -158,3 +158,45 @@ Webapp is actually a UI for these stuff. The logics are the same. They're buildi
 In order to save search time, you need to reduce the search space by choosing fewer parallel options. Say for `num_gpu_per_worker` here, it's DeepSeek V3 with 671B model 
 parameters. With fp8_block, the rough estimation of the model weights is 671GB. You can not hold it on 4/2/1 gpus, you can modify it to `[8]` only. 
 Of source, in most cases, we would like to have the default set work. Ideally, users don't have to modify them. But for specific perf studies, you can try it.
+
+## Customized empirical corrections for systems and backends
+Empirical corrections can be customized per system/backend/version instead of being hard-coded.
+
+Current support:
+1. vLLM backend is supported.
+2. Other backends can adopt the same mechanism in future updates.
+
+Place a yaml file at:
+`src/aiconfigurator/systems/empirical_correction/<system>/<backend>/<version>/empirical_correction.yaml`
+
+Example:
+```yaml
+corrections:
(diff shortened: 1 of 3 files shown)


tianhaox:
Is this too heavy to resolve a func to get the correction? What's the correction you need, is that a complex function or some constant factor?

joshuayao:
> Is this too heavy to resolve a func to get the correction? What's the correction you need, is that a complex function or some constant factor?

Hi @tianhaox 
I cannot provide a definitive answer at this stage; as we are still in the early phases, the design is continuing to evolve. We're focusing on the following areas to improve the TTFT projection's accuracy and made progress:

1. Reducing the margin of error for TTFT in long-context scenarios where chunked prefill is enabled.

      - [ ] https://github.com/ai-dynamo/aiconfigurator/pull/897

2. Refactored TTFT corrections related to vLLM scheduler affinity.
3. Establishing the precise correlation between `ctx_token` (AIC) and `max_num_batched_tokens` (vLLM).

Once we have an initial design ready, I would suggest scheduling a meeting to have a discussion together.

PR #761, opened 2026-04-15
feat: make sglang collector compatible with sglang 0.5.10

Arsene12358:
## Intent

Upgrade the aiconfigurator collector to run cleanly against **sglang 0.5.10** on both **H200 SXM** and **B200 SXM**, and refresh the perf tables that drive aiconfigurator's SILICON-mode simulations. Alongside the 0.5.10 API-compat work, this PR closes several GLM-5-FP8 and fp8_block coverage holes so simulations for DSA-family models (DeepSeek-V3.2, GLM-5-FP8) are actually usable end-to-end.

Backward-compatible with sglang 0.5.5 - 0.5.9. No SDK behavior change for non-DSA models.

### 1. sglang 0.5.10 API breakages (blocked all MLA / attention / MOE collection)

| File | Fix | Upstream cause |
|---|---|---|
| `collector/sglang/collect_mla.py`, `collect_attn.py` | Add `attn_cp_size = 1` on `MockModelRunner` | `FlashAttentionBackend.__init__` now reads `model_runner.attn_cp_size` (sglang `bb737d7a8`) |
| `collector/sglang/collect_mla_module.py` | `hasattr` guard around piecewise-cuda-graph toggle | `enable_piecewise_cuda_graph` renamed to `disable_piecewise_cuda_graph` (sglang `c64274c74`) |
| `collector/sglang/collect_attn.py` | `MockServerArgs` carries both old and new field names | same rename, hand-rolled mock |
| `collector/sglang/collect_moe.py` | Set `enable_fused_moe_sum_all_reduce = False` on the `MagicMock` | `fused_moe()` now reads this flag (sglang `ee5ccde0a`); `MagicMock`'s default truthy value was activating a fused all-reduce path that wrote out of bounds and SIGABRT'd the worker |

### 2. MOE collection robustness on Blackwell

- **`collect_moe.py`**: Tightened the nvfp4 test-case filter from `K % 16 != 0` to `K % 32 != 0`. CuteDSL grouped GEMM requires 16-byte contiguous alignment for 4-bit elements (= 32 elements); the looser filter let 1,340 guaranteed-to-fail configs through on B200.
- **`collect_moe.py`**: Re-raise `TypeError` from CuteDSL `can_implement()` as `RuntimeError` so the worker treats it as a non-fatal per-task skip instead of an unknown exception. Safety net for configs that slip past the filter.
- **`collect.py`**: Treat `DSLCudaRuntimeError` (from CUTLASS DSL) as GPU-fatal and restart the worker, same as `torch.AcceleratorError`. Without this, the worker continued with a poisoned CUDA context and produced a cascade of collateral `AcceleratorError`s across unrelated float16/fp8_block tasks.
- New unit tests: `tests/unit/collector/test_moe_nvfp4_filter.py` (15 cases covering alignment + known-failing / known-passing B200 models).

### 3. Wideep MOE fixes

- **`collect_wideep_deepep_moe.py`**: `run_wideep_moe()` now passes `os.getcwd()` as `output_path` into `_run_moe_subprocess()` instead of hardcoded `None`. Previously perf files landed under the source `collector/` dir and were discarded on container cleanup, so CI reported success while `result/*.txt` was empty.
- **`collect_wideep_deepep_moe.py`**: New `_make_scale_tensor()` helper chooses `torch.int32` (ue8m0-packed) or `torch.float32` based on `sglang.srt.layers.deep_gemm_wrapper.configurer.DEEPGEMM_SCALE_UE8M0`. On Blackwell sglang 0.5.10 packs scales as ue8m0, and `ep_scatter` asserts matching dtypes; the old all-float32 path failed every prefill case.

### 4. DSA (NSA) module collection and GLM-5-FP8 coverage

These are the biggest coverage wins of the PR.

- **`collect_mla_module.py` — fp8_block precision combo.** `_get_precision_combos` now emits `(bf16, kv, fp8_block)` alongside `(bf16, kv, bf16)` for SM >= 89. `load_model_runner` takes a `gemm_type` arg and sets `server_args.quantization = "fp8"` when requested so the inner DSA projections + the `fp8_paged_mqa_logits` indexer scoring kernel are measured in their native fp8 precision instead of being forced to bf16.
- **`collect_mla_module.py` — GLM-5 Blackwell sparse-MLA prefill override.** sglang's `server_args.py` has a GLM-5-only branch that forces `SGLANG_NSA_PREFILL_DENSE_ATTN_KV_LEN_THRESHOLD=0` on Blackwell (keeps prefill on the sparse MLA path). The collector rewrites GLM-5's architecture string to `"DeepseekV3ForCausalLM"` to let HF `AutoConfig` load it, which silently disabled that gate and dispatched prefill through `flashinfer.prefill.trtllm_ragged_attention_deepseek`. That kernel asserts on GLM-5's `v_head_dim=256` and SIGABRT'd the subprocess after ~14 rows per bucket. Fix: save the original arch string and set the env var manually when it was `GlmMoeDsaForCausalLM` on SM >= 10. Cross-checked against a production nsys trace (TP=8 B200 fp8) - production uses the same sparse path and never calls `trtllm_ragged_attention_deepseek`.
- **`collect_mla_module.py` — fp8_block quant-config injection.** The arch rewrite above causes HF's `DeepseekV3Config` to drop the JSON's `quantization_config`; sglang's `_get_quantization_config` falls through to a bare `Fp8Config()` with `weight_block_size=None`, `Fp8LinearMethod.block_quant=False`, and `process_weights_after_loading` transposes `kv_b_proj.weight` from `(7168, 512)` to `(512, 7168)`. `deepseek_weight_loader.py:555` then runs `unflatten(dim0=512, 448)` and `RuntimeError`s on every GLM-5 fp8_block bucket. New `_ensure_fp8_block_quant_config(hf_cfg)` re-injects `{quant_method: fp8, weight_block_size: [128, 128], ...}` onto `hf_config.quantization_config` after `ModelConfig.from_server_args`. Handles dict / dataclass / missing cases.
- **`collect_mla_module.py` — collateral niceties.** Extended `_HEAD_NUMS` to include 8 (so GLM-5 tp=8 at 64 native heads reaches heads=8). Eager 5-iter pre-warm in `_run_decode` before entering `benchmark_with_power` to drain DeepGEMM / flashinfer JIT outside the CUDA-graph capture window.
- **`src/aiconfigurator/sdk/models.py` — FMHA quant-mode gate.** Extend the DSA FMHA float16-downgrade rule from `trtllm`-only to `trtllm`+`sglang`. This is a perf-table schema remapping, not a coverage workaround: the DSA module's `fmha_quant_mode` axis has no `fp8` enum entry, so fp8 FMHA latency — which sglang's `_forward_trtllm` does run on Blackwell decode (sglang internally quantizes q to fp8 via `mla_quantize_and_rope_for_fp8` before calling `flashinfer.decode.trtllm_batch_decode_with_kv_cache_mla`) — is carried in the table under `(fmha=bfloat16, kv_cache=fp8, gemm=*)` rows. The gate remaps a user-level `fmha_quant_mode=fp8` request to `float16` before lookup; without it, any GLM-5-FP8 or DSv3.2 DSA simulation on sglang errors out with `Unsupported dsa_context_module quant mode 'fp8' ...`. Permanent (not collection-gap-dependent).
- **`src/aiconfigurator/model_configs/zai-org--GLM-5-FP8_config.json`**: new file. Required so quant auto-detection picks `fp8_block` for GLM-5-FP8.

### 5. Refreshed sglang 0.5.10 perf tables for H200 SXM + B200 SXM

Full perf dataset under `src/aiconfigurator/systems/data/{h200,b200}_sxm/sglang/0.5.10/` (LFS-tracked). Covers gemm, moe, attention ctx/gen, mla ctx/gen, mla_bmm, wideep ctx/gen (mla + moe), dsa ctx/gen, custom_allreduce. Latest DSA rows come from auto-collect pipeline #1210.

### DSA / MLA module (pipeline #1210)

| Bucket (per system) | Before any fix | After this PR |
|---|---:|---:|
| H200 DSA context total rows | 0 (0.5.10 never collected) | **3,917** |
| H200 DSA generation total rows | 0 | **4,752** |
| B200 DSA context total rows | 0 | **4,031** |
| B200 DSA generation total rows | 0 | **4,092** |

Within those totals:

| Bucket | #1161 era | After PR | Cause of recovery |
|---|---:|---:|---|
| B200 GLM-5 ctx × heads={8,16,32,64} × kv={bf16,fp8} × gemm=bf16 | ~14 rows/bucket | ~123 rows/bucket | Blackwell sparse-MLA override (`329d6755`); ~1,800 rows back |
| H200 + B200 GLM-5 {ctx,gen} × gemm=fp8_block (16 buckets total) | 0 rows | fully populated | fp8_block quant-config injection (`cf2bd738`); ~4,800 rows back |
| B200 DSv3.2 generation coverage | sparse for reduced-heads | dense outside Issue C region | fp8 precision combo + head=8 extension |

### MOE collection (B200)

| Metric | Baseline (57,240 cases) | After PR (83,835 cases) |
|---|---:|---:|
| Total errors | 1,546 | **176** |
| nvfp4 TypeErrors (misaligned K) | 1,340 | **0** |
| AcceleratorError cascade | 128 | 78 (no more fp8_block/float16 collateral) |
| Pass rate | 97.3% | **99.79%** |

Remaining ~176 are genuine upstream flashinfer/CUTLASS kernel bugs on Blackwell, tracked separately.

### Attention / MLA collection

- **B200 sglang 0.5.10 MLA**: 89 `mla_context` errors → **0** (single missing `attn_cp_size` attribute).
- **H100 sglang 0.5.10 Attention**: 1,855 `AttributeError`s (100% failure) → **0** (attn_cp_size fix mirrored into `collect_attn.py`).
- **Wideep MOE prefill on B200**: 492/492 `ep_scatter` dtype failures → **0**; `wideep_context_moe_perf.txt` now produced.

### GLM-5-FP8 simulation

- **Before:** errored out at `Unsupported dsa_context_module quant mode 'fp8'` (SDK gate) or crashed the subprocess during collection (upstream paths).
- **After:** GLM-5-FP8 disagg simulation runs end-to-end in SILICON mode on both H200 and B200. H200 best = 110.64 tokens/s/gpu at 50 tok/s/user. GLM-5 B200 context is fully populated across all precision combos (~123 rows per `(mla_dtype, kv, gemm, heads)` bucket — the full `(batch_size × seq_len)` sweep). GLM-5 B200 decode is also fully populated at heads ∈ {8, 16, 32} (the usual tp-sharded targets on a 64-head model); the only residual gap is a mild **Issue D** truncation at heads=64 × bf16-KV × sl=4096 (55 / 123 rows — subprocess SIGABRT'd in the chunk-path kernel partway through the sweep). **Issue C** is **DSv3.2-only** and does not affect GLM-5 — see the deferred-gaps section below.

## Deferred to follow-up PRs

Two known gaps are documented rather than fixed here so this PR can ship:

1. **Issue D — chunk-path IMA at heads >= 64, seq_len > 2048.** Localised to the sglang `Indexer._get_topk_ragged` chunk loop + `deep_gemm.fp8_mqa_logits`. Affects DSv3.2 heads ∈ {64, 128} bf16-KV sl=4096 on both systems, plus B200 GLM-5 heads=64 bf16-KV sl=4096. Production path, so can't be sidestepped in the collector without polluting the perf DB with off-path kernel shapes. A dedicated sglang upstream-fix branch will carry the reproducer + patch; on fix-landing we re-collect to fill the empty cells. Full root-cause analysis + file:line in `/workspace/PR/sglang_0510/2026-04-20-issue-d-*.md` and project memory `project_issue_d_upstream_blocked.md`.

2. **Issue C — B200 DSv3.2 reduced-heads generation at `kv_len >= 256`.** Async `cudaErrorIllegalAddress` in the trtllm_decode path, only on B200 DSv3.2 at heads ∈ {8, 16, 32}. Production nsys proves the kernel itself is fine; the failure is a collector-setup divergence (likely dummy-weights + cleared-KV-pool → topk indexer picks page indices into unwritten regions). Shipped as 77/132 rows per affected bucket. Next experiment: populate KV pool via real forward before the timed decode probe.

## Backward compatibility

Every fix above is safe on sglang 0.5.5 - 0.5.9:

- Extra mock attributes (`attn_cp_size`, `disable_piecewise_cuda_graph`, `enable_fused_moe_sum_all_reduce`) are ignored by older versions that don't read them.
- `hasattr` guards select the right field name per sglang version.
- ue8m0 scale helper falls back to `float32` when the `deep_gemm_wrapper.configurer` import is absent.
- SDK FMHA gate extension and DSA arch-fallback only fire for the explicit `(architecture, backend, quant_mode)` triples described above; other models are untouched.
- GLM-5 Blackwell threshold override only fires when `original_arch == "GlmMoeDsaForCausalLM"` on SM >= 10 and leaves the env var alone if the caller already set it.

## Test plan

- [x] Unit tests pass: `pytest tests/unit/collector/` (30 passing; 4 pre-existing unrelated failures)
- [x] End-to-end pipeline #1210 on `{h200_sxm, b200_sxm} × {dsa_context_module, dsa_generation_module}` with `lmsysorg/sglang:nightly-dev-cu13-20260416-a4cf2ea1`: all 4 jobs green, row counts per-bucket above
- [x] MOE verification (B200 pipeline, 83,835 test cases): 99.79% pass rate, 0 nvfp4 TypeErrors
- [x] GLM-5-FP8 H200 SILICON-mode simulation runs end-to-end
- [x] Wideep MOE on B200: both prefill and decode perf files produced and updated

## Release Notes

* **New Features**
  * Added GLM-5 FP8 quantized model configuration for optimized inference.
  * Extended FP8 block quantization support for improved performance.
  * Integrated performance data collection for B200 and H200 systems.

* **Bug Fixes**
  * Improved CUDA fatal error detection and GPU context recovery.
  * Fixed compatibility with SGLang 0.5.10 configuration handling.


Arsene12358:
@coderabbitai pls do a full review of the PR again

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -1160,6 +1160,45 @@ def _normalize_dtype_key(raw: str) -> str:
...
+
+def _resolve_dsa_arch(bucket: Mapping, requested: str, context_key: tuple) -> str:
+    """Return the architecture that has rows in bucket, falling back per
+    _DSA_ARCH_FALLBACKS when the requested architecture has none. Emits a
+    one-time WARNING per (requested_arch, fallback_arch, context_key)."""
+    if bucket.get(requested):
+        return requested
+    fallback = _DSA_ARCH_FALLBACKS.get(requested)
+    if fallback is None or fallback not in bucket or not bucket[fallback]:
+        return requested  # let the caller raise the natural error
```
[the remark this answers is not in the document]
Arsene12358: the fallback only serves as a final safety net in case we are missing dsa data. With the current perf table population, these fallback options are not expected to be active

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -1160,6 +1160,45 @@ def _normalize_dtype_key(raw: str) -> str:
 
 DEFAULT_DSA_ARCHITECTURE = "DeepseekV32ForCausalLM"
 
+# Architecture fallback safety net for DSA perf queries: if the requested
```
tianhaox: why this is needed?

src/aiconfigurator/sdk/perf_database.py
```diff
@@ -6770,7 +6429,18 @@ def get_empirical(
                         f"Context DSA module perf data not loaded for system='{self.system}', "
                         f"backend='{self.backend}', version='{self.version}'."
                     )
-                dsa_dict = dsa_module_data[fmha_quant_mode][kvcache_quant_mode][gemm_quant_mode][architecture]
+                arch_bucket = dsa_module_data[fmha_quant_mode][kvcache_quant_mode][gemm_quant_mode]
+                resolved_arch = _resolve_dsa_arch(
```
tianhaox: is this approach risky?

src/aiconfigurator/sdk/perf_database.py
Arsene12358: As discussed, the DSA_ARCH_Fallback option has been reverted

src/aiconfigurator/sdk/perf_database.py
Arsene12358: This is reverted along with the dsa_arch_fallback

collector/sglang/collect_moe.py
```diff
@@ -349,25 +358,40 @@ def benchmark(
...
+                power_law_alpha=power_law_alpha,
+                workloads=workloads,
+            )
+            return kernel_time, power_stats
+        except TypeError as e:
+            raise RuntimeError(
+                f"nvfp4 CuteDSL kernel does not support this config "
+                f"(num_experts={num_experts}, shard_intermediate_size={shard_intermediate_size}, "
+                f"hidden_size={hidden_size}): {e}"
+            ) from e
```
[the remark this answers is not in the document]
Arsene12358: I think this is low-priority hygiene. The current code is correct for the observed CuteDSL TypeErrors; the narrowing would only matter if a future sglang API change raises a semantically unrelated TypeError from the same call site. Defer to follow-up should bandwidth allow

collector/sglang/collect_wideep_deepep_moe.py
```diff
@@ -32,6 +32,35 @@
...
+    """
+    scale_dim = hidden_size // 128
+    if _is_scale_ue8m0():
+        return torch.ones(
+            num_tokens,
+            _ceil(scale_dim / 4),
+            device=device,
+            dtype=torch.int32,
+        )
+    return torch.ones(num_tokens, scale_dim, device=device, dtype=torch.float32)
```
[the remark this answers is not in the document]
Arsene12358: I think decode produced perf data on Blackwell without the ue8m0 fix. That's why this function was designed to specifically target prefill

collector/sglang/collect_mla_module.py
```diff
@@ -105,14 +105,34 @@ def _resolve_local_model_path(model_id: str) -> str:
...
+    # this, on Blackwell every GLM-5 DSA context probe with max_kv_len
+    # <= 2048 (the default DENSE_ATTN threshold) dispatches to
+    # _forward_standard_mha -> flashinfer.prefill.trtllm_ragged_attention_deepseek
+    # which asserts on GLM-5's v_head_dim=256 and kills the subprocess
+    # after ~14 rows/bucket.
+    if original_arch == "GlmMoeDsaForCausalLM":
+        try:
+            sm = torch.cuda.get_device_capability()[0]
+        except Exception:
+            sm = 0
```
ilyasher: we have a helper method for this

collector/sglang/collect_moe.py
```diff
@@ -487,25 +496,40 @@ def benchmark(
...
+        # re-raising as RuntimeError lets the worker continue to the next task.
+        # Note: DSLCudaRuntimeError (actual kernel crash) is NOT caught here
+        # because the CUDA context is already poisoned by the time Python
+        # receives the exception. That case is handled by collect.py's worker
+        # fatal-error detection.
+        try:
+            kernel_time, power_stats = benchmark_config(
+                None,
+                benchmark_num_tokens,
+                num_experts,
```
ilyasher: Would it make more sense to make sure that if worker catches `TypeError`, it doesn't need to restart?

ilyasher: I don't really understand why we need to catch TypeError and reraise it as RuntimeError. Does RuntimeError get special treatment?

collector/sglang/collect_mla_module.py
Arsene12358: Updated to use the helper method

collector/sglang/collect_moe.py
Arsene12358: Good catches. RuntimeError doesn't get any special treatment. I think this try-except block currently provides no operational values. Removed the block in the latest commit

ilyasher (approved):

PR #882, opened 2026-04-21
chore: Refactor collector to have perf filename in registry.py

ilyasher:
Instead of hardcoding the perf filename in every collector script, factor it out into `registry_types.py`. Also, add the perf_filename to the `OpEntry` class.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated performance output filename configuration into a centralized registry, enabling consistent naming across all benchmark operations and simplifying test case structures.

### DIFF
--- a/collector/registry_types.py
+++ b/collector/registry_types.py
@@ -16,6 +16,13 @@ class PerfFile(str, Enum):
     without ``.value``.
     """
 
+    def __str__(self) -> str:
+        """
+        Override behavior of str(x) and f"{x}" to return
+        the perf filename instead of the enum name like "PerfFile.GEMM".
+        """
+        return self.value
+
     GEMM = "gemm_perf.txt"
     CONTEXT_ATTENTION = "context_attention_perf.txt"
     GENERATION_ATTENTION = "generation_attention_perf.txt"
--- a/collector/registry_types.py
+++ b/collector/registry_types.py
@@ -6,6 +6,38 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
+from enum import Enum
+
+
+class PerfFile(str, Enum):
+    """Canonical output filenames for collector operations.
+
+    Inherits from ``str`` so values pass directly to ``open()`` / ``log_perf()``
+    without ``.value``.
+    """
+
+    GEMM = "gemm_perf.txt"
+    CONTEXT_ATTENTION = "context_attention_perf.txt"
+    GENERATION_ATTENTION = "generation_attention_perf.txt"
+    MOE = "moe_perf.txt"
+    CONTEXT_MLA = "context_mla_perf.txt"
+    GENERATION_MLA = "generation_mla_perf.txt"
+    MLA_BMM = "mla_bmm_perf.txt"
+    GDN = "gdn_perf.txt"
+    MAMBA2 = "mamba2_perf.txt"
+    COMPUTESCALE = "computescale_perf.txt"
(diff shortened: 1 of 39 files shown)


collector/vllm/collect_mla_module_v3.py
```diff
@@ -1071,7 +1070,8 @@ def main():
 
         print(f"Running {len(test_cases)} {args.mode} {attn_type.upper()} module test cases...")
 
-        for i, (s, b, h, kv_dtype, compute, gemm, fname) in enumerate(test_cases):
+        fname = f"{attn_type}_{args.mode}_module_perf.txt"
```
Arsene12358: I think the intent is to use 
```python
            perf_filename = (
                PerfFile.MLA_CONTEXT_MODULE
                if attn_type == "mla" and args.mode == "context"
                else PerfFile.MLA_GENERATION_MODULE
                if attn_type == "mla"
                else PerfFile.DSA_CONTEXT_MODULE
                if args.mode == "context"
                else PerfFile.DSA_GENERATION_MODULE
            )
```
here as well, similar to the usage in v1/v2 collector. Maybe missed in the refactor

Arsene12358 (changes requested):

collector/registry_types.py
```diff
@@ -6,6 +6,38 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
+from enum import Enum
+
+
+class PerfFile(str, Enum):
```
Arsene12358, line 12: Nit — the current `(str, Enum)` pattern has a subtle cross-version inconsistency.

Verified on CPython 3.10 / 3.11 / 3.12:

| Python | `str(PerfFile.GEMM)` | `f"{PerfFile.GEMM}"` |
|---|---|---|
| 3.10 | `'PerfFile.GEMM'` | `'gemm_perf.txt'` |
| 3.11 | `'PerfFile.GEMM'` | `'PerfFile.GEMM'` |
| 3.12 | `'PerfFile.GEMM'` | `'PerfFile.GEMM'` |

Python 3.11 changed `(str, Enum).__format__` to match Enum's repr, so any log or format site that interpolates a `PerfFile` value will render `'PerfFile.GEMM'` on 3.11+ and `'gemm_perf.txt'` on 3.10. Project requires `>=3.10`, so `enum.StrEnum` (3.11+) isn't usable.

What do you think about overriding `__str__` to normalize the behavior?

```python
class PerfFile(str, Enum):
    """Canonical output filenames for collector operations.

    Inherits from ``str`` so values pass directly to ``open()`` / ``log_perf()``
    without ``.value``. ``__str__`` is overridden so f-string / str() formatting
    yields the filename on all supported Python versions (the default behavior
    of ``(str, Enum).__format__`` changed between 3.10 and 3.11).
    """

    def __str__(self) -> str:
        return self.value

    GEMM = "gemm_perf.txt"
    ...
```

Verified on 3.10 / 3.11 / 3.12: with this override, `str()`, f-strings, and `%s` all return the filename consistently. Equality, `open()`, concatenation, and `.endswith()` keep working via the `str` base unchanged.

ilyasher: added, thanks for the suggestion

collector/vllm/collect_mla_module_v3.py
ilyasher: fixed

Arsene12358 (approved):

PR #887, opened 2026-04-21
fix(enumerate): accept total_gpus to cap candidates and size min-fit floor

tedzhouhk:
## Summary

`enumerate_profiling_configs` had no knowledge of the deployment's total GPU budget. Two consequences:

- **64-GPU candidate on a 32-GPU cluster.** For MoE wide-EP, `decode_max_gpus` was `max(num_gpu_per_worker)` = 64, so a 32-GPU system could be handed a decode config that can't possibly schedule.
- **OOM candidates not pruned.** `_calculate_min_tp` was called with `total_gpus=num_gpus_per_node`, capping the memory-fit floor at a single node. For DeepSeek R1 FP8 on 32 H100s, 8-GPU wide-EP candidates passed the floor (floor=8, capped) but OOM'd on deploy — they should have been filtered since R1 needs ≥16 GPUs to hold weights.

## Changes

- `enumerate_profiling_configs` gets a new optional `total_gpus: int | None` parameter. When provided, it caps `prefill_max_gpus` / `decode_max_gpus` by the real budget and feeds the real budget into the memory-fit floor.
- `_calculate_min_tp` gets `allow_multi_node: bool = False`. When enabled (passed from enumerate when `is_moe and enable_wideep`), the floor can span nodes, still bounded by `total_gpus`. Dense and intra-node MoE paths keep the single-node cap.
- 4 new `_calculate_min_tp` unit tests covering R1-sized weights on 8x80 GiB H100 nodes, with and without multi-node.

Backward compatible: callers that don't pass `total_gpus` see identical behavior.

## Test plan

- [x] Existing generator unit tests pass (`pytest tests/unit/generator/`)
- [x] Existing SDK utils/task tests pass (`pytest tests/unit/sdk/test_utils.py tests/unit/sdk/task/`)
- [x] New `TestCalculateMinTp` cases pass (single-node cap, multi-node crosses node, multi-node bounded by budget, small-model base case)
- [x] `pre-commit run` on changed files
- [ ] Downstream Dynamo PR (forthcoming) threads `total_gpus` from `thorough.py` into this call; once both land, the R1-on-32-H100 candidate set drops the 8-GPU OOM and 64-GPU impossible configs

## Related

Paired with a Dynamo-side PR that passes `total_gpus` from `components/src/dynamo/profiler/thorough.py` into this function.

## Summary by CodeRabbit

* **New Features**
  * Added optional total GPU budget parameter to configuration enumeration, allowing the system to cap the search space and generate configurations that fit within specified GPU allocations.

* **Tests**
  * Added comprehensive test coverage for tensor-parallel minimum calculations across single-node and multi-node deployment scenarios.

### DIFF
--- a/src/aiconfigurator/generator/naive.py
+++ b/src/aiconfigurator/generator/naive.py
@@ -218,6 +218,7 @@ def _calculate_min_tp(
     vram_per_gpu: int,
     gpus_per_node: int,
     total_gpus: int,
+    allow_multi_node: bool = False,
 ) -> int:
     """
     Calculate the minimum TP size that fits the model in memory.
@@ -229,9 +230,13 @@ def _calculate_min_tp(
         vram_per_gpu: VRAM per GPU in bytes.
         gpus_per_node: Number of GPUs per node.
         total_gpus: Total GPUs available.
+        allow_multi_node: When True, do not cap the result at ``gpus_per_node``.
+            Use for MoE wide-EP sweeps where an engine can span nodes; the
+            result is still capped at ``total_gpus``.
 
     Returns:
-        Minimum TP size (power of 2, capped at gpus_per_node and total_gpus).
+        Minimum TP size (power of 2). Capped at ``min(gpus_per_node, total_gpus)``
+        by default, or at ``total_gpus`` when ``allow_multi_node=True``.
     """
     # Required VRAM per model copy
     required_vram = model_weight_bytes * _MEMORY_MULTIPLIER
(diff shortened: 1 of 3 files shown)


src/aiconfigurator/generator/naive.py
```diff
@@ -246,17 +251,18 @@ def _calculate_min_tp(
     while tp < min_tp:
         tp *= 2
 
-    # Cap at gpus_per_node (to stay within a single node) and total_gpus
-    max_tp = min(gpus_per_node, total_gpus)
+    # Cap at gpus_per_node (single-node constraint) unless multi-node is allowed.
+    max_tp = total_gpus if allow_multi_node else min(gpus_per_node, total_gpus)
```
simone-chen, line 255: @coderabbitai Why does this solve the problem described in:
> OOM candidates not pruned. _calculate_min_tp was called with total_gpus=num_gpus_per_node, capping the memory-fit floor at a single node. For DeepSeek R1 FP8 on 32 H100s, 8-GPU wide-EP candidates passed the floor (floor=8, capped) but OOM'd on deploy — they should have been filtered since R1 needs ≥16 GPUs to hold weights.

This change increases the cap of possible max_tp, but doesn't it make the `tp > max_tp` condition even harder to be triggered? Can you give me the math steps to filter out DSR1 with 32 H100s?

simone-chen: @coderabbitai 🐰 🐰

simone-chen (approved):

PR #897, opened 2026-04-23
fix: correct mix step latency with chunked prefill

Spycsh:
#### Overview:

When chunked prefill is enabled, ctx_tokens may be smaller than isl. In this case, the multiplier np.ceil(isl / ctx_tokens) remains 1, while the divider may exceed 1, leading to a larger estimation gap.

#### Details:

As above

vllm backend handling

NA

## Summary by CodeRabbit

* **New Features**
  * Added optional chunked prefill mode for improved context token processing efficiency during inference.

### DIFF
--- a/src/aiconfigurator/sdk/backends/vllm_backend.py
+++ b/src/aiconfigurator/sdk/backends/vllm_backend.py
@@ -157,7 +157,7 @@ class VLLMBackend(BaseBackend):
                 if not enable_chunked_prefill or ctx_tokens >= isl:
                     scale_factor = np.ceil(isl / ctx_tokens)
                 else:
-                    scale_factor = 1
+                    scale_factor = 2 / (1 + num_mix_steps)
                 ctx_attention_latency_ms = latency_dict["context_attention"] / scale_factor
                 ctx_attention_energy_wms = energy_wms_dict.get("context_attention", 0.0) / scale_factor
 
--- a/src/aiconfigurator/sdk/backends/vllm_backend.py
+++ b/src/aiconfigurator/sdk/backends/vllm_backend.py
@@ -42,6 +42,7 @@ class VLLMBackend(BaseBackend):
         ctx_seq_imbalance_correction_scale = runtime_config.seq_imbalance_correction_scale
         gen_seq_imbalance_correction_scale = runtime_config.gen_seq_imbalance_correction_scale
         ctx_tokens = kwargs.get("ctx_tokens")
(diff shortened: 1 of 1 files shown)


tianhaox:
Here's an approximation in the modeling. The target is to get the context attention of each step. Assume  ctx_tokens=1024, isl=8192,  
step 0, tokens[0-1023]  
step 1, tokens[1024,2047]  
...  
step 7, tokens[7168, 8191]  
acutally the flops of the attention required by each step is different.  


The approximation I set up here is, we don't compute the exact context attention of each step, but use attn_avg of all the 8 steps to finish one context attention instead.  
That's why there's a scale factor equals np.ceil(isl/ctx_tokens), this is how many steps to finish the context.  
the attn_avg = time_to_compute_full_isl / num_steps_to_finish_the_context = attn(isl) / np.ceil(isl/ctx_tokens)

Spycsh:
> acutally the flops of the attention required by each step is different.

Agree, but attention can be O(N^2), is the current approximation still precise? I wonder whether using isl 8192 attn latency / 8 steps can approximate precisely 1 avg step's latency. From my understanding, latencies for chunked prefill case should be something like:
```

Step 1: 1024*1024
Step 2: 1024*2048
Step 2: 1024*3072
...
Step 8: 1024*8192

```

Generally, avg step latency should be T_ctx_token * (1+num_steps)*num_steps/2 / num_steps = T_ctx_token * (1+num_steps)/2 instead of T_isl / num_steps.

tianhaox:
https://github.com/ai-dynamo/aiconfigurator/blob/main/src/aiconfigurator/sdk/perf_database.py#L4226

'''
                ops = (
                    2 * b * (full_s * full_s - prefix * prefix) * n * h * 2 / 2
                )
'''
the flops of the attenion in each chunk is equivelant to the context attention with previous tokens as cache and the chunksize(1024) as the new tokens.

ops_c0 = C * 1024^2
ops_c1 = C * (2048^2 - 1024^2)
ops_c2 = C * (3072^2 - 2048^2)
...

ops_total = C * 8192^2
If we assume all these attentions are compute-bound, the ideal latency_avg = ops_total / gpu_flops / num_steps_to_finish_the_context.

Practically, I believe there will be overhead of each chunk to store/read the kvcache and lower down the effective flops.

Do you have any observation in real benchmark that doesn't align with this analysis?

Spycsh:
I see. The causal mask let the incremental computation for each chunk as a rectangular of `(full^2-prefix^2)/2`. Then no change from my side. Thanks for explanation.

PR #898, opened 2026-04-23
fix: resolve symlinks in safe_mkdir allowed path prefixes

natoscott:
## Overview

Fix safe_mkdir rejecting valid /tmp and /var/tmp paths on macOS due to unresolved symlinks.

## Details

On macOS, /tmp and /var/tmp are symlinks to /private/tmp and /private/var/tmp. safe_mkdir resolves the user input path but compared it against unresolved allowed prefixes, causing ValueError: Path is outside allowed locations when using --save-dir /tmp/output.

The fix adds .resolve() to the /tmp and /var/tmp entries in the allowed prefixes list so both sides of the comparison use canonical paths.

## Test plan

- [x] Verified aiconfigurator cli default --save-dir /tmp/output succeeds on macOS (previously failed)
- [ ] Verify no regression on Linux (where /tmp is not a symlink)

### DIFF
--- a/src/aiconfigurator/sdk/utils.py
+++ b/src/aiconfigurator/sdk/utils.py
@@ -286,9 +286,9 @@ def safe_mkdir(target_path: str, exist_ok: bool = True) -> Path:
         allowed_prefixes = [
             current_dir,
             Path.home(),
-            Path("/tmp"),
+            Path("/tmp").resolve(),
             Path("/workspace"),
-            Path("/var/tmp"),
+            Path("/var/tmp").resolve(),
             Path(tempfile.gettempdir()).resolve(),
         ]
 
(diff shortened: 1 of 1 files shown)


natoscott:
> Optional improvement: Consider resolving all paths in allowed_prefixes for consistency, including Path.home() and Path("/workspace"), to ensure robustness if these are ever symlinked in certain environments.

I had thought about this previously, and didn't go that route to focus on the minimal fix for macOS here.  However, it may make sense to do this - does anyone else have an opinion or preference?  Thanks.

jasonqinzhou:
> > Optional improvement: Consider resolving all paths in allowed_prefixes for consistency, including Path.home() and Path("/workspace"), to ensure robustness if these are ever symlinked in certain environments.
> 
> I had thought about this previously, and didn't go that route to focus on the minimal fix for macOS here. However, it may make sense to do this - does anyone else have an opinion or preference? Thanks.

Yes I think no harm to resolve all of them. Could you update this part? After that I can approve and merge immediately.

natoscott:
@jasonqinzhou thanks for the review - change made now.

Arsene12358 (approved):

Arsene12358:
Thanks, I think we can merge this

PR #912, opened 2026-04-28
fix: derive MLA KV cache size from config and cover Kimi K2.5

milesial:
The KV cache memory model in trtllm and sglang backends only treated DEEPSEEK / DEEPSEEKV32 as MLA. KIMIK25 (Kimi K2/K2.5) fell through to the generic GQA branch, which overestimates KV per token by ~6.2x for this architecture. On a B200 SXM with isl=8000 osl=1000 the bogus estimate trips the kv_cache budget at b=16, so the agg sweep stops at concurrency=16 even though the model can fit ~95.

Replace the hardcoded 576-element MLA constant with a single get_kvcache_elements_per_token() method on BaseModel that reads kv_lora_rank and qk_rope_head_dim from the parsed HF config (now exposed via extra_params for DeepSeek V3/R1 and Kimi K2.5; V3.2 already exposed them). Both backends call this method, so vLLM picks it up via delegation. Hardcoded 512/64 remain only as a final fallback.

## Summary by CodeRabbit

* **New Features**
  * Added support for KIMIK25 model family with unified memory calculations.
  * Improved KV cache memory estimation consistency across model architectures using model-specific configuration parameters.

* **Bug Fixes**
  * Fixed activation memory estimation for improved accuracy across supported model families.

### DIFF
--- a/src/aiconfigurator/sdk/backends/sglang_backend.py
+++ b/src/aiconfigurator/sdk/backends/sglang_backend.py
@@ -540,7 +540,7 @@ class SGLANGBackend(BaseBackend):
             c_dict = {1: 28, 2: 17, 4: 13, 8: 13}
             activations = 2 * num_tokens * h * c_dict[min(model.config.tp_size, 8)]
             activations = max(activations, 90 * 1024 * 1024)  # Higher minimum for SGLANG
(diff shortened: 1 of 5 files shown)


ilyasher (approved):
Thanks! Could you please add a regression unit test?

milesial:
Added a test thanks

PR #925, opened 2026-04-28
fix: Improve naive config generation robustness for large models (memory-fit awareness)

devivasudevan:
### Summary

This PR enhances the naive configuration path to better handle **large model deployments** by introducing explicit memory-fit detection during tensor-parallel (TP) calculation.

The goal is to avoid silent misconfiguration when a model cannot fit into the available GPU memory, which previously could let invalid deployments continue downstream before failing later, such as in Dynamo RAPID fallback.

---

#### Add memory-fit signal to TP calculation

- Updated `_calculate_min_tp` to return:

  ```python
  (selected_tp: int, fits: bool, required_tp: int)
  ```

- `selected_tp` preserves the existing capped TP behavior.
- `fits=False` indicates the uncapped required TP exceeds the available GPU budget.
- `required_tp` preserves the uncapped TP required by the memory-fit estimate.

#### Surface memory-fit metadata

- Naive generator params now include `ModelConfig.fits_in_memory` and `ModelConfig.required_tp`.
- `enumerate_profiling_configs` now returns `fits` and `required_tp` alongside the prefill/decode candidate lists.

#### Preserve conservative generation behavior

- The selected TP value remains capped to the available GPU budget.
- The change does not automatically switch to an alternative parallelization strategy.
- Logging now includes `fits` and `required_tp` to make infeasible configurations visible earlier.

---

### Enable upstream systems like Dynamo to

- Detect infeasible configurations early.
- Avoid treating capped TP as proof that the model fits.
- Trigger fallback strategies, alternative parallelism, or user warnings.

### Why this is needed

Large models such as:

- Qwen3-235B
- DeepSeek-R1
- LLaMA 70B+

can exceed available GPU memory even at maximum TP. Previously, naive config generation could silently cap TP and allow deployment to continue until a later OOM or runtime failure.

### Example log output

```text
TP calculation: model=470.00GiB, vram=80.00GiB,
required=705.00GiB (1.5x), min_tp=9, selected_tp=8, fit=False, required_tp=16
```

## Summary by CodeRabbit

* **Refactor**
  * Enhanced memory-fit analysis to better report when model configurations require more resources and to preserve detailed tensor-parallelism requirements.
  * Improved internal logging and code formatting for clearer diagnostics.

* **Tests**
  * Updated unit tests to align with the refined configuration-calculation outputs and added assertions for resource-fit reporting.


devivasudevan:
@hhzhang16 @jasonqinzhou PTAL. This would help to fix some of the DGDR issues listed here - https://github.com/ai-dynamo/dynamo/issues/8469

jasonqinzhou:
could u fix coderrabbit comments?

jasonqinzhou:
test failed: [gw1] [ 28%] FAILED tests/unit/generator/test_naive.py::TestCalculateMinTp::test_single_node_caps_at_gpus_per_node

src/aiconfigurator/generator/naive.py
```diff
@@ -233,19 +231,17 @@ def _calculate_min_tp(
         allow_multi_node: When True, do not cap the result at ``gpus_per_node``.
             Use for MoE wide-EP sweeps where an engine can span nodes; the
             result is still capped at ``total_gpus``.
-
     Returns:
-        Minimum TP size (power of 2). Capped at ``min(gpus_per_node, total_gpus)``
-        by default, or at ``total_gpus`` when ``allow_multi_node=True``.
+        selected_tp: selected TP (capped to available GPUs)
+        fits: whether model actually fits in memory
+        tp: true TP required for memory fit (before capping)
```
jasonqinzhou: change tp to required_tp for clearilty.

tests/unit/generator/test_naive.py
jasonqinzhou: Add assertions for the new fits behavior in tests. This is the main value of the PR, so tests should explicitly cover fits is False when capped and fits is True when the required TP is available.

src/aiconfigurator/generator/enumerate.py
```diff
@@ -463,7 +466,7 @@ def enumerate_profiling_configs(
     # across all GPUs), so the memory-fit floor is not capped at a single node.
     # Otherwise (dense, or MoE intra-node) the floor stays within a node.
     effective_total_gpus = total_gpus if total_gpus is not None else num_gpus_per_node
-    min_gpus = _calculate_min_tp(
+    min_gpus, fits, required_tp = _calculate_min_tp(
         model_weight_bytes=model_weight_bytes,
```
jasonqinzhou, line 473: Consider whether enumerate.py should do anything with fits / required_tp. Right now it destructures them and ignores them, so the behavior is unchanged except for avoiding a tuple unpack error. That may be fine, but if the PR’s goal is surfacing memory-fit awareness, enumeration may also need logging or metadata.

src/aiconfigurator/generator/naive.py
```diff
@@ -264,16 +260,15 @@ def _calculate_min_tp(
...
     logger.info(
         f"TP calculation: model={model_weight_bytes / (1024**3):.2f}GiB, "
         f"vram={vram_per_gpu / (1024**3):.2f}GiB, "
         f"required={required_vram / (1024**3):.2f}GiB (1.5x), "
-        f"min_tp={min_tp}, selected_tp={tp}"
+        f"min_tp={min_tp}, selected_tp={selected_tp}, fit={fits}, required_tp={tp}"
     )
-
-    return tp
+    return selected_tp, fits, tp
```
jasonqinzhou, line 271: The implementation still emits configs that may not fit; it only marks them. That is probably the right conservative behavior for now, but the PR description should be clear about that.

jasonqinzhou:
@coderabbitai review

tests/unit/generator/test_naive.py
```diff
@@ -186,44 +186,46 @@ class TestCalculateMinTp:
...
     def test_small_model_fits_on_one_gpu(self):
-        tp = _calculate_min_tp(
+        min_gpus, fits, tp = _calculate_min_tp(
             model_weight_bytes=10 * 1024**3,
             vram_per_gpu=self._H100_VRAM,
             gpus_per_node=8,
             total_gpus=8,
         )
+        assert min_gpus == 1
         assert tp == 1
```
jasonqinzhou, line 239: Could we assert the new `fits` signal directly in this test block? Since that is the main behavior added by the PR, these cases should fail if `fits` regresses while `min_gpus` / `required_tp` still look right.

```suggestion
    def test_single_node_caps_at_gpus_per_node(self):
        """Dense default: min_tp capped at gpus_per_node even if model is larger."""
        min_gpus, fits, tp = _calculate_min_tp(
            model_weight_bytes=self._R1_FP8_WEIGHTS,
            vram_per_gpu=self._H100_VRAM,
            gpus_per_node=8,
            total_gpus=32,
        )
        assert min_gpus == 8  # capped
        assert fits is False
        assert tp == 16  # actual requirement

    def test_multi_node_allows_crossing_node_boundary(self):
        """MoE wide-EP: min_tp can span nodes, floored to power-of-2 fit."""
        min_gpus, fits, tp = _calculate_min_tp(
            model_weight_bytes=self._R1_FP8_WEIGHTS,
            vram_per_gpu=self._H100_VRAM,
            gpus_per_node=8,
            total_gpus=32,
            allow_multi_node=True,
        )
        assert tp == 16
        assert min_gpus == 16
        assert fits is True

    def test_multi_node_capped_by_total_gpus(self):
        min_gpus, fits, tp = _calculate_min_tp(
            model_weight_bytes=self._R1_FP8_WEIGHTS,
            vram_per_gpu=self._H100_VRAM,
            gpus_per_node=8,
            total_gpus=8,
            allow_multi_node=True,
        )
        assert min_gpus == 8  # capped by total GPUs
        assert fits is False
        assert tp == 16  # still required

    def test_small_model_fits_on_one_gpu(self):
        min_gpus, fits, tp = _calculate_min_tp(
            model_weight_bytes=10 * 1024**3,
            vram_per_gpu=self._H100_VRAM,
            gpus_per_node=8,
            total_gpus=8,
        )
        assert min_gpus == 1
        assert fits is True
        assert tp == 1
```

src/aiconfigurator/generator/enumerate.py
```diff
@@ -354,7 +357,7 @@ def enumerate_profiling_configs(
     k8s_pvc_name: str | None = None,
```
jasonqinzhou, line 357: Could we also update the return docstring to match the new return shape? GitHub will not let me attach a clickable suggestion to those unchanged docstring lines, but this is the exact patch:

```diff
     Returns:
-        ``(prefill_candidates, decode_candidates)`` -- two lists of
-        :class:`EnumeratedCandidate` objects, each bundling an aggregated
-        DGD config dict with its parallelization metadata.
+        ``(prefill_candidates, decode_candidates, fits, required_tp)``.
+        The first two values are lists of :class:`EnumeratedCandidate`
+        objects, each bundling an aggregated DGD config dict with its
+        parallelization metadata. ``fits`` indicates whether the model fits
+        in the available GPU budget, and ``required_tp`` is the uncapped TP
+        required for the memory-fit estimate.
```

jasonqinzhou (approved):

PR #954, opened 2026-04-30
fix: llm-d output improvements

natoscott:
Overview:

This PR improves the llm-d output from the generator: fixes whitespace corruption in generated args lists, adds a proper hf_token_secret parameter for K8s auth secret configuration, and adds backend version comments to vLLM container image fields.

Details:

- Fix args list whitespace stripping (sglang/llm-d-values.yaml.j2, vllm/llm-d-values.yaml.j2): The {%- / -%} strip-whitespace Jinja2 modifiers were used on for/endfor tags wrapping the args list entries. This caused all list items to be rendered on a single line without proper YAML indentation, producing invalid K8s manifests. Changed to {% / %} (no whitespace stripping) across all three worker sections (agg, decode, prefill) in both backend templates.
- Add LlmdConfig.hf_token_secret parameter (deployment_config.yaml, vllm/llm-d-values.yaml.j2): The vLLM llm-d template previously referenced service_config.hf_token for the authSecretName field, but that field is a token value, not a K8s secret name. Introduced LlmdConfig.hf_token_secret as a dedicated optional parameter (settable via --generator-set LlmdConfig.hf_token_secret=<secret-name>) and wired it to authSecretName. Updated the inline comment to reflect the new parameter.
- Add backend version comment above image field (vllm/llm-d-values.yaml.j2): Added a comment above each image: field in all three worker sections noting the vLLM version AIC used when predicting the configuration, prompting users to pin the image tag for reproducibility. 

- src/aiconfigurator/generator/config/backend_templates/vllm/llm-d-values.yaml.j2 — contains all
three changes; review the authSecretName block replacement and the for-loop whitespace fix closely
- src/aiconfigurator/generator/config/deployment_config.yaml — the new LlmdConfig.hf_token_secret
schema entry
- src/aiconfigurator/generator/config/backend_templates/sglang/llm-d-values.yaml.j2 — same whitespace
fix applied here

## Summary by CodeRabbit

* **New Features**
  * Introduced a new `hf_token_secret` configuration option, enabling users to securely provision Hugging Face tokens through secret-based methods.

* **Configuration Enhancements**
  * Updated backend configuration to properly source Hugging Face credentials from the newly available secret field, improving flexibility and security in token management across supported backends.

### DIFF
--- a/src/aiconfigurator/generator/config/backend_templates/sglang/llm-d-values.yaml.j2
+++ b/src/aiconfigurator/generator/config/backend_templates/sglang/llm-d-values.yaml.j2
@@ -22,12 +22,8 @@ modelArtifacts:
     llm-d.ai/model: {{ served_name | replace('-', '_') | replace('/', '_') }}
   uri: {{ model_uri }}
   size: {{ llmd_config.model_cache_size | default('100Gi') }}
-  # hf_token should be the name of a K8s Secret containing the HF token
-  {% if service_config.hf_token -%}
-  authSecretName: {{ service_config.hf_token }}
-  {% else -%}
-  authSecretName: ""
-  {% endif -%}
+  # Name of a K8s Secret containing HF_TOKEN. Set via --generator-set LlmdConfig.hf_token_secret=<secret-name>
+  authSecretName: {{ llmd_config.hf_token_secret | default('') }}
   mountPath: {{ llmd_config.model_mount_path | default('/model-cache') }}
 
 # Accelerator configuration
--- a/src/aiconfigurator/generator/config/backend_templates/sglang/llm-d-values.yaml.j2
+++ b/src/aiconfigurator/generator/config/backend_templates/sglang/llm-d-values.yaml.j2
@@ -69,9 +69,9 @@ decode:
         - {{ model_name }}
         - --served-model-name
         - {{ served_name }}
-        {%- for arg in agg_cli_args_list | default([]) %}
+        {% for arg in agg_cli_args_list | default([]) %}
         - {{ arg | tojson }}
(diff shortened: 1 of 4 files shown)


nqzhou2426 (commented):
The new `LlmdConfig.hf_token_secret` is only wired into the vLLM llm-d template, but the SGLang llm-d template still sets `authSecretName` from `ServiceConfig.hf_token`. That leaves SGLang with the same raw-token-vs-secret-name bug this PR is fixing, and users cannot use the documented `--generator-set LlmdConfig.hf_token_secret=<secret-name>` path for SGLang llm-d output. Can we apply the same `authSecretName: {{ llmd_config.hf_token_secret | default('') }}` change to `backend_templates/sglang/llm-d-values.yaml.j2` and add a small test covering both backends?

jasonqinzhou:
> The new `LlmdConfig.hf_token_secret` is only wired into the vLLM llm-d template, but the SGLang llm-d template still sets `authSecretName` from `ServiceConfig.hf_token`. That leaves SGLang with the same raw-token-vs-secret-name bug this PR is fixing, and users cannot use the documented `--generator-set LlmdConfig.hf_token_secret=<secret-name>` path for SGLang llm-d output. Can we apply the same `authSecretName: {{ llmd_config.hf_token_secret | default('') }}` change to `backend_templates/sglang/llm-d-values.yaml.j2` and add a small test covering both backends?

@natoscott could u take a look this comment?

natoscott:
@nqzhou2426 @jasonqinzhou thanks for the review, this should be resolved now.

Ethan-ES (approved):

PR #962, opened 2026-04-30
fix(release/0.8.0): accept bfloat16 dtype in perf data loaders (#53)

Harrilee:
_This is a `release/0.8.0` patch which cannot be directly cherry-picked_

## Reason
1. The new collector is updated with QWEN 3.5 PR, which is out of scope of `release/0.8.0`
2. `bfloat16` is renamed to `float16`, our latest perf data does not match the old naming version

## Solution
1. Copy perf data from ToT `main` to `release/0.8.0`
2. Make an alias for `bfloat16`

## Test Plan
- [x] Rerun support matrix to confirm that the problem is resolved (https://github.com/ai-dynamo/aiconfigurator/actions/runs/25194291000)


jasonqinzhou (approved):

src/aiconfigurator/sdk/common.py
```diff
@@ -647,3 +647,12 @@ class CommQuantMode(Enum):
     half = QuantMapping(2, 0, "half")
     int8 = QuantMapping(1, 0, "int8")
     fp8 = QuantMapping(1, 0, "fp8")
+
+
+# Alias "bfloat16" -> "float16" for the 16-bit float enum members. The perf data
+# on main (post PR #895) uses "bfloat16" as the dtype string. Release/0.8.0 still
+# defines the member as `float16`, so we register the newer name as an alias so
+# the release branch can load main-branch data files without renaming the enum.
+for _enum_cls in (GEMMQuantMode, MoEQuantMode, FMHAQuantMode, KVCacheQuantMode):
```
nqzhou2426, line 656: Could we keep a small regression test for this alias behavior? The fix relies on mutating Enum internals: `GEMMQuantMode["bfloat16"]` works, but `GEMMQuantMode.bfloat16` still does not. A focused test covering `GEMMQuantMode`, `MoEQuantMode`, `FMHAQuantMode`, and `KVCacheQuantMode` would lock in the intended perf-data loader behavior and make future cleanup safer.

jasonqinzhou: FYI: a minor suggestion by agent but P2.

PR #968, opened 2026-05-01
feat: add h800_sxm system support

### LINKED ISSUE #661: [Support Matrix] How to generate system-data for a new machine?
## What coverage are you requesting?

**Hardware / System:**
A800/H800/H20_sxm or maybe other GPUs

## Mode

- [ ] Aggregated (agg)
- [ ] Disaggregated (disagg)
- [*] Both

## Additional context

It seems that the difference between SILICON/HYBRID/EXPERICAL data mode is huge. In order to implement a high accuracy prediction on new machines, how to create and manage these data?

whitelok:
H800 SXM is the China-market variant of H100 SXM, with identical compute (989 TFLOPS BF16 / 1979 TFLOPS FP8), HBM capacity (80 GiB) and HBM bandwidth (3.35 TB/s), but with NVLink bandwidth capped at 400 GB/s bidirectional (vs. 900 GB/s on H100), i.e. 200 GB/s per direction per GPU.

Changes:
- Register `h800_sxm` in `SupportedSystems`
- Add `systems/h800_sxm.yaml` with H800-specific `intra_node_bw=200 GB/s`, reusing the same compute/memory specs as `h100_sxm`
- Add a minimal `systems/data/h800_sxm/trtllm/1.3.0rc10/` op-data tree for compute-bound ops only (gemm/attention/MoE/MLA/...), reusing the corresponding `h100_sxm` LFS pointers (same SM90 silicon)
- **Intentionally omit collective op-data** (`custom_allreduce_perf.txt`, `nccl/<ver>/nccl_perf.txt`) for `h800_sxm` so that `query_custom_allreduce` / `query_nccl` fall back to the bandwidth-based analytical model (which correctly consumes the lower `intra_node_bw`) instead of silently returning H100 latencies in SILICON/HYBRID mode
- Update CLI `--system` help text examples to include `h800_sxm`

#### Overview:

This PR adds first-class support for the **NVIDIA H800 SXM** GPU as a new target system in aiconfigurator. H800 is widely deployed in the China market and is currently the most common Hopper-class SKU in many production clusters there, but it was previously rejected by the CLI with `Backend ... is not supported for system h800_sxm`. After this change, users can run the standard `aiconfigurator cli default/exp/generate/estimate/support` flows with `--system h800_sxm --backend trtllm --database-mode HYBRID` (see "Database mode" note below for SILICON-mode caveats).

The implementation deliberately keeps the diff minimal and additive — no existing system, model or backend behavior is modified. H800 reuses the H100 op-level performance data **for compute-bound ops only** (same SM90 architecture, same compute/memory specs). The NVLink-related collective cost is corrected through the system YAML's `intra_node_bw` value, which the existing communication models already consume; to make sure that this correction actually takes effect, the H100 collective tables are deliberately **not** copied over for H800 (see "Important design decision" below).

#### Important design decision (per review feedback):

An earlier revision of this PR also mirrored the H100 `custom_allreduce_perf.txt` and `nccl_perf.txt` tables under `data/h800_sxm/`. As @jasonqinzhou pointed out, this would silently return H100 collective latencies on H800 in **SILICON / HYBRID** mode for in-node GPU counts, because `query_custom_allreduce` / `query_nccl` read those silicon tables directly when present, and `intra_node_bw` would only affect the SOL / EMPIRICAL / HYBRID-fallback paths.

The fix in this PR is to **omit those two collective files entirely** for `h800_sxm`. The behavior then becomes:

- **SOL / EMPIRICAL**: unchanged — already used `intra_node_bw`, now correctly modeled at 200 GB/s/dir.
- **HYBRID**: collective queries hit `_query_silicon_or_hybrid`, fail to load the missing silicon table, and fall back to the empirical model (which uses `intra_node_bw`). End-to-end runs work and now correctly reflect the halved NVLink.
- **SILICON**: collective queries raise a structured `PerfDataNotAvailableError` with a hint to use HYBRID mode. This is the intended behavior until real H800 silicon profiling data is collected and added.

#### Details:

**1. `src/aiconfigurator/sdk/common.py`**
- Added `"h800_sxm"` to the `SupportedSystems` set so that downstream validation, support-matrix lookup and CLI argument checks all recognize the new system.

**2. `src/aiconfigurator/systems/h800_sxm.yaml`** (new)
- GPU section: identical to `h100_sxm.yaml` for `mem_bw` (3.35 TB/s), `mem_capacity` (80 GiB), `bfloat16/int8/fp8_tc_flops`, `power` (700W) and `sm_version` (90), since H800 and H100 share the same silicon for compute and HBM.
- Node section: **`intra_node_bw: 200000000000`** (200 GB/s per GPU per direction) — this is the only material parameter difference vs. H100. It reflects H800's officially-rated 400 GB/s bidirectional NVLink (down from H100's 900 GB/s bidirectional / 450 GB/s per direction). All NVLink/NCCL-bound operations (AllReduce, AllGather, EP-dispatch, KV transfer cost in disagg, etc.) will now be modeled correctly for H800 in SOL / EMPIRICAL / HYBRID modes.
- `inter_node_bw`, `pcie_bw`, `nccl_mem`, `other_mem` and `nccl_version` follow the H100 SXM defaults — H800 nodes typically share the same CX7 NDR network and PCIe 5.0 fabric. `nccl_version` is kept for schema compatibility even though no `nccl_perf.txt` is shipped (see header comment in the YAML).
- An expanded header comment documents the rationale, in particular: (a) compute-bound op-data is reused from `h100_sxm`, (b) collective op-data is intentionally omitted to force the bandwidth-based fallback, (c) `--database-mode HYBRID` (or SOL / EMPIRICAL) is the recommended mode until real H800 silicon profiling data is collected.

**3. `src/aiconfigurator/systems/data/h800_sxm/`** (new, 15 files)
- Contains a single backend × version: `trtllm/1.3.0rc10/` — `trtllm` is the project `DEFAULT_BACKEND` and `1.3.0rc10` is the latest shipped trtllm version, mirroring `h100_sxm`'s most-recent backend revision.
- All 15 entries are **Git LFS pointers that share the exact same `oid` as the corresponding `h100_sxm/trtllm/1.3.0rc10/` files**, so the patch itself is small and clones don't duplicate any LFS object — the same blobs already pulled for `h100_sxm` are reused for `h800_sxm`.
- The previous broader mirror (`nccl/{2.23,2.27.3}`, `trtllm/1.2.0rc5`, `sglang/{0.5.6.post2,0.5.9}`, `vllm/{0.14.0,0.19.0}`, plus all `custom_allreduce_perf.txt`) was removed in this revision per review feedback to keep the first PR focused on a single backend × version. Additional backends/versions and real H800 silicon profiling data (including the deliberately-omitted collective tables) will be added in follow-up PRs.

**4. `src/aiconfigurator/cli/main.py`**
- Updated 4 `--system` argparse `help=` strings (default / generate / estimate / support sub-commands) to include `h800_sxm` in the example list.

**Validation performed locally:**
```
aiconfigurator cli default \
  --model <Qwen3.5-MoE-FP8> --total-gpus 8 \
  --system h800_sxm --backend trtllm --database-mode HYBRID
```
runs end-to-end and produces a Pareto frontier + top-5 configurations table. Parallel strategies correctly reflect the lower NVLink bandwidth (e.g., favoring lower-TP / higher-DP layouts compared to H100 for the same model), since collective costs now flow through the bandwidth-based fallback that consumes `intra_node_bw=200 GB/s`.

Suggested review order — small files first, then the (small) data tree:

1. **`src/aiconfigurator/systems/h800_sxm.yaml`** — the most important file. Please double-check:
   - `intra_node_bw = 200 GB/s` (per direction, single GPU) matches your reference for H800 NVLink, **or** advise on a different convention if the project elsewhere uses bidirectional figures.
   - The decision to keep `mem_bw / *_tc_flops / mem_capacity / sm_version` identical to H100 SXM.
   - The deliberate omission of `custom_allreduce_perf.txt` and `nccl_perf.txt` (forcing the bandwidth-based fallback) is acceptable as the temporary behavior for the first PR. Real H800 collective profiling will be added in a follow-up.

2. **`src/aiconfigurator/sdk/common.py`** — one-line addition to `SupportedSystems`. Please confirm there is no other registry/whitelist (e.g., support-matrix CSV, autotest config, web UI dropdown) that should also be updated; I left those untouched to keep the PR minimal.

3. **`src/aiconfigurator/cli/main.py`** — purely doc-string updates to 4 `--help` examples; safe to skim.

4. **`src/aiconfigurator/systems/data/h800_sxm/trtllm/1.3.0rc10/`** — 15 new files, all Git LFS pointers re-using existing `h100_sxm/trtllm/1.3.0rc10/` `oid`s. Spot-checking 1–2 files and confirming the `oid` matches the `h100_sxm` counterpart should be enough; no real binary data is added to the repo. Note that `custom_allreduce_perf.txt` is intentionally **not** present here.

#### Follow-ups (out of scope for this PR):

- Collect real H800 silicon profiling data (gemm / attention / moe / collective / nccl) on an 8×H800 SXM NVLink host and replace the LFS-shared compute tables with H800-native data; add the missing `custom_allreduce_perf.txt` and `nccl/<ver>/nccl_perf.txt` so SILICON mode also works for collectives.
- Re-add support for `sglang` / `vllm` backends and additional trtllm versions for `h800_sxm` once their op-data has been validated on H800 hardware.

- Relates to GitHub issue: [[Support Matrix] How to generate system-data for a new machine? #661](https://github.com/ai-dynamo/aiconfigurator/issues/661)

  Note: this PR does not fully resolve #661 (which asks how to *generate* system-data for a new machine). Instead, it provides a pragmatic workaround for H800 specifically by reusing `h100_sxm` compute op-data with corrected NVLink bandwidth (and forcing the bandwidth-based fallback for collectives), until proper H800 silicon profiling data can be collected.

### DIFF
--- a/src/aiconfigurator/systems/h800_sxm.yaml
+++ b/src/aiconfigurator/systems/h800_sxm.yaml
@@ -1,0 +1,37 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# H800 SXM is the China-market variant of H100 SXM.
+# Compute (TFLOPS), HBM capacity, and HBM bandwidth are identical to H100 SXM.
+# The main difference is the NVLink (intra-node) bandwidth, which is approximately
+# halved on H800 (400 GB/s bidirectional vs. 900 GB/s on H100), i.e. 200 GB/s
+# per direction per GPU instead of 450 GB/s.
+# Performance op-data files are reused from h100_sxm (same SM/compute) until
+# H800-specific silicon profiling data is collected.
+data_dir: data/h800_sxm # relative to systems_dir
+gpu:
+  mem_bw: 3350000000000 # 3350GB/s, identical to H100 SXM
+  mem_bw_empirical_scaling_factor: 0.8 # some nonofficial correction based on observations, you should try to modify based on your own observations
+  mem_empirical_constant_latency: 0.000003 # 3us some nonofficial correction based on observations, you should try to modify based on your own observations
+  mem_capacity: 85899345920 # 80GiB, identical to H100 SXM
+  bfloat16_tc_flops: 989000000000000 # 989TFLOPS, identical to H100 SXM
+  int8_tc_flops: 1978000000000000 # 1978TFLOPS, identical to H100 SXM
+  fp8_tc_flops: 1978000000000000 # 1978TFLOPS, identical to H100 SXM
+  power: 700  # Watt
+  sm_version: 90
+
+node:
+  num_gpus_per_node: 8
+  inter_node_bw: 50000000000  # Byte/s per GPU, single direction, 1:1 CX7 NDR 400Gb/s per node
+  intra_node_bw: 200000000000  # Byte/s per gpu, single direction; H800 NVLink is capped at ~400 GB/s bidirectional (200 GB/s per direction)
+  pcie_bw: 64000000000  # Byte/s, single direction, pcie 5.0
+  p2p_latency: 0.00001  # 10us some nonofficial correction based on observations, you should try to modify based on your own observations
+
+misc:
+  nccl_mem: # some nonofficial correction based on observations, you should try to modify based on your own observations
+    1: 0
+    2: 358612992 # 342MB
+    4: 411041792 # 392MB
+    8: 411041792 # 392MB
+  other_mem: 3758096384 # increase from 551MB to 3.5GB for safer deployment, this will cover part of the inaccurate mem calc.
+  nccl_version: '2.27.3'
--- a/src/aiconfigurator/systems/h800_sxm.yaml
+++ b/src/aiconfigurator/systems/h800_sxm.yaml
@@ -6,8 +6,23 @@
 # The main difference is the NVLink (intra-node) bandwidth, which is approximately
 # halved on H800 (400 GB/s bidirectional vs. 900 GB/s on H100), i.e. 200 GB/s
 # per direction per GPU instead of 450 GB/s.
-# Performance op-data files are reused from h100_sxm (same SM/compute) until
-# H800-specific silicon profiling data is collected.
+#
+# Compute-bound op-data (gemm/attention/moe/...) is reused from h100_sxm because
+# H800 shares the same SM90 silicon and HBM, so per-kernel compute latency
+# tracks H100 closely.
+#
+# IMPORTANT: communication-bound op-data (custom_allreduce_perf.txt, nccl_perf.txt)
+# is intentionally NOT shipped here. Reusing the H100 collective tables would
+# silently return H100 latencies in SILICON/HYBRID mode for in-node GPU counts
+# and ignore the halved NVLink bandwidth above. By omitting these files, the
+# database query layer falls back to the bandwidth-based analytical model
+# (see PerfDatabase.query_nccl / query_custom_allreduce) which correctly
+# consumes `intra_node_bw` below.
+#
+# As a result, this initial H800 system definition is meant to be used with
+# `--database-mode HYBRID` (or SOL/EMPIRICAL); pure SILICON mode for collectives
+# will raise PerfDataNotAvailableError until real H800 silicon profiling data is
+# collected and added to data/h800_sxm/.
 data_dir: data/h800_sxm # relative to systems_dir
 gpu:
   mem_bw: 3350000000000 # 3350GB/s, identical to H100 SXM
@@ -34,4 +49,8 @@ misc:
     4: 411041792 # 392MB
     8: 411041792 # 392MB
   other_mem: 3758096384 # increase from 551MB to 3.5GB for safer deployment, this will cover part of the inaccurate mem calc.
+  # nccl_version is kept for schema compatibility; the corresponding
+  # data/h800_sxm/nccl/<version>/nccl_perf.txt is intentionally absent so that
+  # collective queries fall back to the bandwidth-based analytical model that
+  # consumes `intra_node_bw` (see header comment).
   nccl_version: '2.27.3'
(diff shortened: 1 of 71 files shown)


src/aiconfigurator/systems/h800_sxm.yaml
```diff
@@ -0,0 +1,37 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# H800 SXM is the China-market variant of H100 SXM.
+# Compute (TFLOPS), HBM capacity, and HBM bandwidth are identical to H100 SXM.
+# The main difference is the NVLink (intra-node) bandwidth, which is approximately
+# halved on H800 (400 GB/s bidirectional vs. 900 GB/s on H100), i.e. 200 GB/s
+# per direction per GPU instead of 450 GB/s.
+# Performance op-data files are reused from h100_sxm (same SM/compute) until
```
jasonqinzhou: Reusing the H100 op-data means the lower `intra_node_bw` does not actually correct the default SILICON/HYBRID collective costs. `query_custom_allreduce` and `query_nccl` read the silicon tables directly when the H800 files exist, and those new H800 files have the same LFS oids as H100, so H800 returns H100 collective latencies for in-node GPU counts. I verified a sample 8-GPU all-reduce: H100 and H800 both return the same SILICON latency, while SOL/EMPIRICAL diverge because only those paths use `intra_node_bw`. Can we either omit/rescale the copied communication tables or force the H800 communication queries through bandwidth-based fallback until real H800 comm profiling data is available?

jasonqinzhou: Can we collect real perf files for h800?

jasonqinzhou: One framework/framework version is enough as the first PR.

jasonqinzhou:
Many tests are failed due to missing network per files for h800. Could you collect them on h800 and pass the tests?

jasonqinzhou:
@whitelok thanks for the update here. I left a couple of review notes/questions on the PR. Could you take a look and either update the branch or reply with your reasoning? Once those are resolved I can re-review.

jasonqinzhou:
Close this due to no active development and low priority on h800.

PR #971, opened 2026-05-01
fix: derive support matrix target versions

jasonqinzhou:
image

## Summary
- derive support-matrix target versions from the CSV instead of hardcoding backend constants
- store target versions per system so each tab displays the latest tested backend version for that system
- update the static template and checked-in support-matrix page to read the derived targets

## Verification
- python3 -m py_compile tools/support_matrix/generate_static_page.py tests/unit/support_matrix/test_generate_static_page.py
- manual execution of the new unit test functions
- generated /tmp/support-matrix-test.html and confirmed b200_sxm targets are sglang 0.5.10, trtllm 1.3.0rc10, vllm 0.19.0

Note: pytest is not installed in this local environment, so I could not run the pytest command directly here.

## Summary by CodeRabbit

* **New Features**
  * Support matrix target versions are now dynamically computed from test data for each backend, automatically updating matrix table headers to display the latest tested version instead of using predefined static values

* **Tests**
  * Added comprehensive unit tests validating the target version computation logic and matrix generation with dynamically calculated target versions

### DIFF
--- a/docs/support-matrix/index.html
+++ b/docs/support-matrix/index.html
@@ -44,7 +44,6 @@ const { useState, useEffect, useCallback, useMemo, useRef } = React;
 // ── Constants ───────────────────────────────────────────────────────
 const REPO = 'ai-dynamo/aiconfigurator';
 const CSV_PATH = 'src/aiconfigurator/systems/support_matrix.csv';
-const TARGET_VERSIONS = { vllm: '0.14.0', sglang: '0.5.9', trtllm: '1.2.0rc6' };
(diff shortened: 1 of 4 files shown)


jasonqinzhou:
@coderabbitai review

jasonqinzhou:
@coderabbitai resolve

jasonqinzhou:
@Harrilee will have a multiversion support soon that doesn't need this fix anymore.

Harrilee:
Let's continue here https://github.com/ai-dynamo/aiconfigurator/pull/974

PR #974, opened 2026-05-01
feat(support-matrix): regression view + drop unused static generator

Harrilee:
- Color cells by latest-version status: green = latest passes, amber = regression (latest fails but older passes), red = all fail
- Render per-version chips with strikethrough on failing versions; add a "Show all versions" toggle (default: newest only)
- Hover tooltip (instant, no native-title delay) shows full pass / partial / fail version breakdown regardless of toggle
- Multi-version column headers: drop `(target: X)`, list all tested versions when a backend has more than one
- Layout: full-bleed bar backgrounds; content centered at max 1600px
- Remove unused deploy-support-matrix-page.yml workflow plus its Python generator and HTML template (Pages serves docs/ directly)

# Example
https://harrilee.github.io/aiconfigurator/support-matrix

image

## Summary by CodeRabbit

* **New Features**
  * Enhanced support matrix page with version-level filtering, "Show all versions" option, and improved tooltips for detailed test result information across multiple tested versions.

* **Chores**
  * Removed automated deployment workflow and static page generation utilities; support matrix now maintained as direct HTML implementation.

### DIFF
--- a/src/aiconfigurator/sdk/task.py
+++ b/src/aiconfigurator/sdk/task.py
@@ -1152,14 +1152,17 @@ class TaskRunner:
     def _get_database(system: str, backend: str, version: str, database_mode: str | None = None):
         """Fetch a database from the global cache.
 
-        When *database_mode* is set the returned object is a deep copy (because
-        `set_default_database_mode` mutates state).  Otherwise the cached
-        instance is returned directly.
+        When *database_mode* would change the cached database's default mode,
+        return a deep copy first because `set_default_database_mode` mutates
+        query-cache state. If the requested mode already matches, reuse the
+        cached instance directly.
         """
         db = get_database(system=system, backend=backend, version=version)
         if database_mode is not None:
-            db = copy.deepcopy(db)
-            db.set_default_database_mode(common.DatabaseMode[database_mode])
+            mode = common.DatabaseMode[database_mode]
+            if mode != db.get_default_database_mode():
+                db = copy.deepcopy(db)
+                db.set_default_database_mode(mode)
         return db
 
     def run_agg(self, task_config: DefaultMunch) -> dict[str, pd.DataFrame | None]:


docs/support-matrix/index.html
```diff
@@ -198,53 +233,114 @@
...
+        : { version: null, error: null };
+      const aggPass = aggResult.version !== null && aggResult.version !== 'FAIL';
+      const disaggPass = disaggResult.version !== null && disaggResult.version !== 'FAIL';
+      if (aggPass !== disaggPass) {
+        verStatus = 'partial';
+        verError = !aggPass
+          ? `[agg FAIL] ${aggResult.error || 'No error message available'}`
+          : `[disagg FAIL] ${disaggResult.error || 'No error message available'}`;
+      }
+    }
```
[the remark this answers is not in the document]
Harrilee: It is expected. We shall support agg + disagg for every combination. If there's a data missing. We SHOULD mark it as fail.

docs/support-matrix/index.html
```diff
@@ -198,53 +233,114 @@
...
+  //   - latest is strictly pass               -> green
+  //   - latest is fail/partial but any older  -> orange (regression)
+  //   - everything failed                     -> red
+  const sortDesc = (a, b) => compareVersionTuples(versionSortKey(b), versionSortKey(a));
+  const allTested = [...passVersions, ...partialVersions, ...failVersions].sort(sortDesc);
+  const newest = allTested[0];
+  let status;
+  if (passVersions.includes(newest)) status = 'pass';
+  else if (passVersions.length) status = 'partial';
+  else status = 'fail';
```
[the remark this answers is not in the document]
Harrilee: Check my fix

docs/support-matrix/index.html
```diff
@@ -602,6 +722,41 @@ <h1 className="text-lg font-bold leading-tight"><a href="/aiconfigurator/">AI Co
...
+    } else {
+      // Each per-status list is already newest-first (sorted in aggregateBackendCell).
+      // Pick whichever bucket the overall status points at.
+      if (cell.status === 'pass' && cell.passVersions && cell.passVersions.length) {
+        pushPass(cell.passVersions[0]);
+      } else if (cell.status === 'partial') {
+        // Regression: latest is fail/partial. Prefer the latest fail chip if any,
+        // else the latest partial chip — either way that's the version that's
+        // currently broken.
+        if (cell.failVersions && cell.failVersions.length) pushFail(cell.failVersions[0]);
```
jasonqinzhou: The collapsed view can display the wrong version for a partial cell. `aggregateBackendCell` decides the cell status from the newest tested version, but this rendering path always prefers `failVersions[0]` whenever any failed version exists. If the newest version is partial and an older version failed, the cell will show the older struck-through failure instead of the current partial version. Pick the newest entry across the partial/fail buckets here so the collapsed chip matches the status calculation.

```suggestion
      } else if (cell.status === 'partial') {
        const brokenVersions = [
          ...(cell.partialVersions || []).map(v => ({ version: v, kind: 'partial' })),
          ...(cell.failVersions || []).map(v => ({ version: v, kind: 'fail' })),
        ].sort((a, b) => compareVersionTuples(versionSortKey(b.version), versionSortKey(a.version)));
        const latestBroken = brokenVersions[0];
        if (latestBroken?.kind === 'fail') pushFail(latestBroken.version);
        else if (latestBroken) pushPartial(latestBroken.version);
```

Harrilee: Not really. `cell.status` will be `pass`as long as the latest version passed. E.g.
image

Harrilee: cc @coderabbitai  to double check

docs/support-matrix/index.html
```diff
@@ -602,6 +722,48 @@ <h1 className="text-lg font-bold leading-tight"><a href="/aiconfigurator/">AI Co
...
+        // currently broken.
+      } else if (cell.status === 'partial') {
+        const brokenVersions = [
+          ...(cell.partialVersions || []).map(v => ({ version: v, kind: 'partial' })),
+          ...(cell.failVersions || []).map(v => ({ version: v, kind: 'fail' })),
+        ].sort((a, b) => compareVersionTuples(versionSortKey(b.version), versionSortKey(a.version)));
+        const latestBroken = brokenVersions[0];
+        if (latestBroken?.kind === 'fail') pushFail(latestBroken.version);
+        else if (latestBroken) pushPartial(latestBroken.version);
+        else if (cell.partialVersions && cell.partialVersions.length) pushPartial(cell.partialVersions[0]);
```
jasonqinzhou: This attempted to fix the collapsed partial-cell chip, but the empty `else if (cell.status === 'partial')` branch above it now catches every partial cell, so the real `brokenVersions` logic is unreachable. In the default collapsed view, amber cells render with no version chip at all. Removing the empty duplicate branch makes the previous fix apply as intended.

```suggestion
      } else if (cell.status === 'partial') {
        const brokenVersions = [
          ...(cell.partialVersions || []).map(v => ({ version: v, kind: 'partial' })),
          ...(cell.failVersions || []).map(v => ({ version: v, kind: 'fail' })),
        ].sort((a, b) => compareVersionTuples(versionSortKey(b.version), versionSortKey(a.version)));
        const latestBroken = brokenVersions[0];
        if (latestBroken?.kind === 'fail') pushFail(latestBroken.version);
        else if (latestBroken) pushPartial(latestBroken.version);
```

docs/support-matrix/index.html
```diff
@@ -198,53 +233,115 @@
...
+      const disaggRows = verRows.filter(r => r.Mode === 'disagg');
+      const aggResult = aggRows.length
+        ? getLatestSupportedVersion(aggRows, modelId, system, backend)
+        : { version: null, error: null };
+      const disaggResult = disaggRows.length
+        ? getLatestSupportedVersion(disaggRows, modelId, system, backend)
+        : { version: null, error: null };
+      const aggPass = aggResult.version !== null && aggResult.version !== 'FAIL';
+      const disaggPass = disaggResult.version !== null && disaggResult.version !== 'FAIL';
+      if (aggPass !== disaggPass) {
```
jasonqinzhou: This still treats missing mode data as a split failure. If a version only has `agg` rows or only has `disagg` rows, the missing side gets `version: null`, `aggPass !== disaggPass` becomes true, and the cell is marked partial even though there was no actual agg/disagg disagreement. Only compare the two modes when both buckets exist.

```suggestion
      const aggPass = aggResult.version !== null && aggResult.version !== 'FAIL';
      const disaggPass = disaggResult.version !== null && disaggResult.version !== 'FAIL';
      if (aggRows.length > 0 && disaggRows.length > 0 && aggPass !== disaggPass) {
```

docs/support-matrix/index.html
```diff
@@ -33,6 +33,41 @@
...
+    top: 100%;
+    transform: translate(-50%, 6px);
+    z-index: 1000;
+    padding: 0.4rem 0.6rem;
+    background: rgb(15 23 42);
+    color: rgb(241 245 249);
+    font-size: 0.72rem;
+    font-family: ui-monospace, SFMono-Regular, Menlo, Consolas, monospace;
+    line-height: 1.4;
+    white-space: pre;
```
jasonqinzhou: Small follow-up from the earlier review: `white-space: pre` preserves newlines but prevents long error signatures from wrapping, so a 500-character tooltip line can overflow its `max-width`. This keeps newlines while allowing the tooltip to stay inside the intended width.

```suggestion
    white-space: pre-wrap;
    overflow-wrap: anywhere;
```

Harrilee: Why would I have a "500-character tooltip line" when this tool-tip just shows something like
```
sglang
Pass: 0.5.10, 0.5.9
Fail: 0.5.6.post2
```

Is this AI-generated review comment?

docs/support-matrix/index.html
Harrilee: https://github.com/ai-dynamo/aiconfigurator/pull/974#discussion_r3175519530

This is was already addressed by @coderabbitai and resolved.

docs/support-matrix/index.html
Harrilee: True. This bug was introduced in https://github.com/ai-dynamo/aiconfigurator/pull/974#discussion_r3175593119. I shouldn't trust the automatic generated suggections

Harrilee: True. This bug was introduced in https://github.com/ai-dynamo/aiconfigurator/pull/974#discussion_r3175593119. I shouldn't trust the automatic generated suggections.

PR #979, opened 2026-05-05
fix: handle NaN values in Pareto selection

jasonqinzhou:
## Summary
- filter NaN/inf Pareto axis values before frontier computation
- drop non-finite best-config candidates before grouped idxmax selection
- add regression coverage for the all-NaN grouped idxmax crash

Linear: https://linear.app/nvidia/issue/DYN-2796/ga-vdr-fix-aic-profiler-nan-handling-crash-in-pareto-analysis

## Repro
Before the fix, with pandas 2.2.3, all-NaN throughput groups produce NaN idxmax labels and crash:

```text
idxmax result: [nan, nan]
KeyError: "None of [Index([nan, nan], dtype='float64')] are in the [index]"
```

After the fix, the same input returns an empty DataFrame instead of raising.

## Testing
- `/tmp/aic-nan-repro-venv/bin/python - <<'PY' ...` -> `empty: True`
- `PYTHONPATH=src /tmp/aic-nan-repro-venv/bin/python -m pytest -q tests/unit/sdk/picking/test_pareto_sla_filtering.py::TestTpotConstraintFiltering::test_grouped_selection_drops_all_nan_throughput_groups tests/unit/sdk/picking/test_pareto_sla_filtering.py::TestParetoFrontFiltering::test_get_pareto_front_ignores_non_finite_points` -> 2 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced data validation in Pareto analysis to filter out invalid numerical values (infinity, NaN) and verify required input columns exist.
  * Improved constraint-based candidate selection logic to properly exclude invalid data points from results.

* **Tests**
  * Added comprehensive test coverage for edge cases involving invalid data values in Pareto analysis and constraint filtering.

### DIFF
--- a/src/aiconfigurator/sdk/pareto_analysis.py
+++ b/src/aiconfigurator/sdk/pareto_analysis.py
@@ -338,9 +338,19 @@ def get_pareto_front(
         maximize_x: Treat larger values on x axis as better if True, else minimize.
         maximize_y: Treat larger values on y axis as better if True, else minimize.
     """
-    if df is None or df.empty:
+    if df is None:
         return pd.DataFrame()
-    df = df.sort_values(by=x_col)
+    if df.empty:
+        return df.iloc[0:0].copy()
+    if x_col not in df.columns or y_col not in df.columns:
+        return pd.DataFrame(columns=[x_col, y_col])
+
+    working = df[[x_col, y_col]].replace([np.inf, -np.inf], np.nan)
+    valid_mask = working.notna().all(axis=1)
+    if not valid_mask.any():
+        return df.iloc[0:0].copy()
+
+    df = df.loc[valid_mask].sort_values(by=x_col)
 
     def is_pareto(costs: np.ndarray) -> np.ndarray:
         is_better = np.ones(costs.shape[0], dtype=bool)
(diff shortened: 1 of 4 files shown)


jasonqinzhou:
@coderabbitai resolve

src/aiconfigurator/sdk/pareto_analysis.py
```diff
@@ -338,9 +338,19 @@ def get_pareto_front(
...
-    df = df.sort_values(by=x_col)
+    if df.empty:
+        return df.iloc[0:0].copy()
+    if x_col not in df.columns or y_col not in df.columns:
+        return pd.DataFrame(columns=[x_col, y_col])
+
+    working = df[[x_col, y_col]].replace([np.inf, -np.inf], np.nan)
+    valid_mask = working.notna().all(axis=1)
+    if not valid_mask.any():
+        return df.iloc[0:0].copy()
```
ilyasher, line 351: What causes the df to have `inf` or `nan` values in the first place?

jasonqinzhou: Good catch. Root cause is the perf-db interpolation path: when the requested point is not exactly present, AIC uses `scipy.interpolate.griddata(...)`, which can return NaN for degenerate/invalid interpolation cases. Previously `_validate()` only checked `value < 0.0`, and NaN/inf are not rejected by that check, so the non-finite latency could flow into the Pareto DataFrame.

Fixed in `ea356b3`: `_validate()` now rejects NaN/+inf/-inf, with unit coverage for those cases. I also scanned the repo perf CSV/TXT files and did not find existing non-finite numeric cells, so this looks like an interpolation artifact rather than a checked-in bad perf row. The Pareto-layer filter stays as a defensive guard for external real-GPU/AIPerf inputs that already arrive as a DataFrame.

jasonqinzhou: The new _validate() fix catches AIC perf DB / interpolation-produced NaN or inf at the source. That is the right root fix for simulated perf data.

But the Pareto/picking layer still accepts DataFrames from other paths:

- Dynamo real-GPU profiler / AIPerf artifacts
- externally supplied or previously saved perf DataFrames
- future caller code that computes derived columns directly
- tests / user inputs / partial data
- any non-perf-DB source that bypasses PerfDatabase._validate()

So the upper-layer filtering is still useful as a defensive API boundary. It makes get_pareto_front() and config selection robust: invalid candidate rows are ignored instead of crashing plotting/picking.

ilyasher: cool thanks

ilyasher (approved):

PR #988, opened 2026-05-06
fix: preserve experiment prefix

milesial:
`prefix` in the experiment YAML was not being forwarded anywhere and got lost, even when it was documented

```
exp_agg: &exp_agg
  mode: "patch"
  serving_mode: "agg"
  model_path: "nvidia/Kimi-K2.5-NVFP4"
  total_gpus: 8
  system_name: "h200_sxm"
  backend_name: "trtllm"
  database_mode: "HYBRID"
  isl: 4000
  osl: 1000
  prefix: 1000
```

## Release Notes

* **New Features**
  * Per-experiment configuration overrides now support the prefix parameter, enabling per-experiment control of prefix cache length.

* **Tests**
  * Added test coverage verifying that runtime parameters are correctly preserved when applying configuration patches to experiment task configurations.

### DIFF
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -991,7 +991,7 @@ def build_experiment_task_configs(
             task_kwargs["decode_system_name"] = inferred_decode_system or system_name
 
         # Per-experiment overrides for runtime numeric parameters if provided at top level
-        for numeric_key in ("isl", "osl", "ttft", "tpot", "request_latency"):
+        for numeric_key in ("isl", "osl", "prefix", "ttft", "tpot", "request_latency"):
             if numeric_key in exp_config:
                 task_kwargs[numeric_key] = exp_config[numeric_key]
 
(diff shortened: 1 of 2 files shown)


jasonqinzhou:
One small follow-up: since `prefix` is now treated as a top-level experiment override, it should also be added to `_EXPERIMENT_RESERVED_KEYS`.

Right now, YAML experiments that omit an explicit `config:` block still copy `prefix` into the synthesized YAML patch, so the built `TaskConfig` ends up with both `runtime_config.prefix` and a stray top-level `config.prefix`. That is mostly harmless in patch mode, but it can make exported/debug YAML confusing, and it is especially risky for `mode: replace` because the synthesized patch can replace the generated config with only `{"prefix": ...}`.

The fix should be just:

```diff
diff --git a/src/aiconfigurator/cli/main.py b/src/aiconfigurator/cli/main.py
@@
     "isl",
     "osl",
+    "prefix",
     "ttft",
     "tpot",
     "request_latency",
```

It would also be worth adding a small no-`config` test using the example shape from the PR description, so this path is covered too.

jasonqinzhou (approved):

PR #992, opened 2026-05-06
fix: select support matrix versions semantically

jasonqinzhou:
## Summary
- Select support-matrix backend versions with semantic version parsing so `0.5.10` sorts after `0.5.9`.
- Prefer exact model rows before architecture fallback when auto-selecting a support-matrix version.
- Ignore empty/invalid version strings and return `None` when no valid version is available.

## Impact
This improves `aiconfigurator cli support` preflight accuracy when users omit `--backend-version`, especially for branches that contain model-specific newer rows. It does not change `cli default`/`TaskRunner` backend resolution; that path already uses `get_latest_database_version()` from the perf DB layer.

## Tests
- `python3 -m py_compile src/aiconfigurator/cli/main.py tests/unit/cli/test_support_matrix_version.py`
- Manual helper behavior checks for `0.5.10 > 0.5.9` and invalid-version handling

Note: local `pytest`/`ruff` entry points are not installed in this checkout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed support matrix matching to perform case-insensitive comparisons for system and backend values.
  * Improved architecture detection through model-based lookup.

* **New Features**
  * Added automatic version selection from support matrix when no explicit version is provided.

* **Tests**
  * Added comprehensive test coverage for version selection logic and architecture fallback scenarios.

### DIFF
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -11,7 +11,6 @@ from typing import Any
 
 import pandas as pd
 import yaml
-from packaging.version import InvalidVersion, Version
 
 from aiconfigurator import __version__
 from aiconfigurator.cli.report_and_save import log_final_summary, save_results
@@ -30,16 +29,6 @@ from aiconfigurator.sdk.utils import ListFlowDumper, get_model_config_from_model
 logger = logging.getLogger(__name__)
 
 
-def _parse_support_matrix_version(version: str | None) -> Version | None:
-    """Parse backend versions so 0.5.10 sorts after 0.5.9."""
-    if not version:
-        return None
-    try:
-        return Version(version)
-    except InvalidVersion:
-        return None
-
-
 def _latest_support_matrix_version(
     matrix: list[dict[str, str]],
     system: str,
@@ -47,7 +36,15 @@ def _latest_support_matrix_version(
     model: str | None = None,
     architecture: str | None = None,
 ) -> str | None:
-    rows = [row for row in matrix if row["System"] == system and row["Backend"] == backend]
+    """Pick the highest PEP 440 version for the relevant support-matrix rows.
+
+    Matches system and backend case-insensitively. When a model is provided,
+    exact-model rows win, then architecture rows. If neither model nor
+    architecture matches, return None instead of selecting an unrelated row.
+    """
+    rows = [
+        row for row in matrix if row["System"].lower() == system.lower() and row["Backend"].lower() == backend.lower()
+    ]
 
     if model:
         exact_rows = [row for row in rows if row["HuggingFaceID"].lower() == model.lower()]
@@ -57,11 +54,28 @@ def _latest_support_matrix_version(
             architecture_rows = [row for row in rows if row["Architecture"] == architecture]
             if architecture_rows:
                 rows = architecture_rows
+            else:
+                logger.debug(
+                    "No support-matrix rows match model=%s or architecture=%s for system=%s backend=%s",
+                    model,
+                    architecture,
+                    system,
+                    backend,
+                )
+                return None
+        else:
+            logger.debug(
+                "No exact support-matrix rows match model=%s for system=%s backend=%s and no architecture was provided",
+                model,
+                system,
+                backend,
+            )
+            return None
 
     versions = [
         (version, parsed)
         for version in {row["Version"] for row in rows}
-        if (parsed := _parse_support_matrix_version(version))
+        if (parsed := common.parse_support_matrix_version(version))
     ]
     if not versions:
         return None
@@ -1262,14 +1276,14 @@ def _run_support_matrix_mode(args):
     matrix = common.get_support_matrix()
     systems = sorted(common.SupportedSystems) if args.system == "all" else [args.system]
     backends = [b.value for b in common.BackendName] if args.backend == "all" else [args.backend]
-    existing_combos = {(row["System"], row["Backend"]) for row in matrix}
+    existing_combos = {(row["System"].lower(), row["Backend"].lower()) for row in matrix}
 
     results: dict[tuple[str, str], common.SupportResult | None] = {}
     has_inferred = False
 
     for system in systems:
         for be in backends:
-            if (system, be) not in existing_combos:
+            if (system.lower(), be.lower()) not in existing_combos:
                 results[(system, be)] = None
                 continue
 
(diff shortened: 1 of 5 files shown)


src/aiconfigurator/cli/main.py
```diff
@@ -29,6 +30,44 @@
...
+        return None
+
+
+def _latest_support_matrix_version(
+    matrix: list[dict[str, str]],
+    system: str,
+    backend: str,
+    model: str | None = None,
+    architecture: str | None = None,
+) -> str | None:
```
Harrilee, line 38: Case-sensitivity diverges from `sdk/common.check_support`: that function filters rows with `row["System"].lower() == system.lower()` (and `.lower()`-compares `HuggingFaceID`), but here `System` and `Backend` are matched case-sensitively. If a user passes `B200_SXM` or `SGLang`, this helper returns `None`, the matrix loop hits `results[(system, be)] = None` and the cell renders `-`, even though `check_support` would have matched. Worth normalizing here too:

```python
rows = [
    row for row in matrix
    if row["System"].lower() == system.lower()
    and row["Backend"].lower() == backend.lower()
]
```

src/aiconfigurator/cli/main.py
```diff
@@ -29,6 +30,44 @@
...
+    rows = [row for row in matrix if row["System"] == system and row["Backend"] == backend]
+
+    if model:
+        exact_rows = [row for row in rows if row["HuggingFaceID"].lower() == model.lower()]
+        if exact_rows:
+            rows = exact_rows
+        elif architecture:
+            architecture_rows = [row for row in rows if row["Architecture"] == architecture]
+            if architecture_rows:
+                rows = architecture_rows
```
Harrilee, line 56: Silent fall-through hides intent: if `model` is supplied but has no exact rows AND `architecture` is supplied but matches no rows, `rows` falls through to the unfiltered `system+backend` set and we return the latest version across all models. That preserves prior behavior, but the user gets no signal that their model-specific intent was dropped.

Two options:
- Add a `logger.debug("falling back to all %s/%s rows; no exact-model or architecture matches for %s", system, backend, model)` so a surprising version pick is auditable.
- Or be stricter and return `None` here, mirroring the empty-matrix branch — let `check_support` do the architecture-majority-vote inference instead.

Either is fine; the silent fall-through is the part worth fixing.

src/aiconfigurator/cli/main.py
```diff
@@ -29,6 +30,44 @@
 logger = logging.getLogger(__name__)
 
 
+def _parse_support_matrix_version(version: str | None) -> Version | None:
```
Harrilee: Duplicate of `parse_version` in `src/aiconfigurator/webapp/components/support_matrix_tab.py:14`, which already does PEP 440 parsing with `InvalidVersion` fallback. Behaviorally close (webapp returns `Version("0.0.0")` for invalid; this returns `None`), but conceptually the same helper.

Not a blocker for this PR, but a third copy is one too many. Worth pulling into `sdk/common.py` (or a small `sdk/version_utils.py`) as a follow-up so the support-matrix CSV has a single source of truth for version parsing.

src/aiconfigurator/cli/main.py
Harrilee, line 38: Missing docstring on `_latest_support_matrix_version`. The model→architecture→fallthrough policy is non-obvious from the code (especially the silent fall-through noted below). One short paragraph would save the next reader a trace, e.g.:

```
"""Pick the highest PEP 440 version for the given system+backend.

Prefers exact-model rows; falls back to architecture rows if architecture is
provided; otherwise considers all rows for the system+backend combo.
Returns None if no row has a parseable Version."""
```

Harrilee:
@coderabbitai Review

jasonqinzhou:
@coderabbitai review

src/aiconfigurator/cli/main.py
jasonqinzhou: Fixed in d78a1f3. `_latest_support_matrix_version` now matches `System` and `Backend` case-insensitively, and the matrix-mode `existing_combos` check is normalized too. Added a regression test for `B200_SXM` / `SGLang`.

src/aiconfigurator/cli/main.py
jasonqinzhou: Fixed in d78a1f3. The helper no longer silently falls back to unrelated system/backend rows when a model is provided but neither exact-model nor architecture rows match; it logs the miss and returns `None` instead. Added a regression test for that path.

src/aiconfigurator/cli/main.py
jasonqinzhou: Fixed in d78a1f3. I moved the shared PEP 440 parser to `common.parse_support_matrix_version()` and updated the webapp `parse_version()` wrapper to use it while preserving the webapp `Version("0.0.0")` fallback behavior for invalid/empty versions.

src/aiconfigurator/cli/main.py
jasonqinzhou: Fixed in d78a1f3. Added a docstring documenting the PEP 440 selection, case-insensitive system/backend matching, exact-model preference, architecture fallback, and `None` return when no model-relevant rows match.

src/aiconfigurator/cli/main.py
```diff
@@ -29,6 +29,61 @@
...
+
+    versions = [
+        (version, parsed)
+        for version in {row["Version"] for row in rows}
+        if (parsed := common.parse_support_matrix_version(version))
+    ]
+    if not versions:
+        return None
+    return max(versions, key=lambda version: version[1])[0]
+
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in d78a1f3. Ran Ruff formatting on `src/aiconfigurator/cli/main.py`; the GitHub `Lint and Format (Ruff)` job is now passing.

src/aiconfigurator/cli/main.py
```diff
@@ -1312,22 +1363,20 @@ def _run_support_mode(args):
...
+    # If no version specified, find the latest model-relevant version in the support matrix
+    if not version:
+        matrix = common.get_support_matrix()
+        version = _latest_support_matrix_version(matrix, system, backend, model=model, architecture=architecture)
+
+    logger.info("Checking support for model=%s, system=%s, backend=%s, version=%s", model, system, backend, version)
+
     result = common.check_support(
         model=model, system=system, backend=backend, version=version, architecture=architecture
     )
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in d78a1f3. `_run_support_mode` now checks for `version is None` after auto-selection, logs/prints a clear no-valid-version message, and returns before calling `common.check_support`. Added a regression test to ensure `check_support` is not called in that path.

Arsene12358:
@coderabbitai any blocking issues?

Arsene12358:
LGTM

Arsene12358 (approved):

PR #894, opened 2026-04-22
feat: add MXFP4 MoE and attention data for gpt-oss models on H200

natoscott:
## Overview

Add MXFP4 MoE profiling data and attention data for openai/gpt-oss-120b and openai/gpt-oss-20b on H200 SXM with vLLM, enabling full SILICON mode support for these models.

## Details

- Fix vLLM 0.19.x collector compatibility for MXFP4 MoE benchmarking (moe_layer_index reset)
- Add w4a16_mxfp4 MoE profiling data for gpt-oss-120b (128 experts) and gpt-oss-20b (32 experts)
- Add TP=2 distributed MoE profiling data for both models
- Extend attention collector to sweep head_dim and window_size, set sliding_window on hf_text_config
- Add context and generation attention data for head_dim=64 and window_size=128
- Rename float16 to bfloat16 in collected attention data to match main naming convention

Enables aiconfigurator cli default for openai/gpt-oss-120b and openai/gpt-oss-20b on H200+vLLM in full SILICON mode.

Fixes #758

## Test plan

- [x] Unit tests pass
- [x] openai/gpt-oss-120b produces agg + disagg results in SILICON mode
- [x] openai/gpt-oss-20b produces 993 agg + 99 disagg results in SILICON mode
- [x] RedhatAI/gpt-oss-20b produces equivalent results (same architecture)
- [x] Container build and test verified on H200 cluster

PR #1004, opened 2026-05-08
fix: FA3 scheduler_metadata mismatch in vllm attention collector

Arsene12358:
## Tracking

Part of [AIC-1018](https://linear.app/nvidia/issue/AIC-1018/add-missing-data-entries-to-heal-dsv32glm5-support-matrix) — *"Add missing data entries to heal dsv32/GLM5 support matrix"* (Urgent, AIC 0.9.0). This PR fixes the collector bug that has been silently degrading Hopper attention collection (visible only in CI artifact `errors_*.json`, not in the merged perf `.txt`); the data-entry PRs that actually heal the support matrix follow once the in-flight collection pipelines complete on the fix branch.

## Summary

vLLM attention collection on Hopper (H100/H200) was systematically dropping ~10% of `attention_context` and ~23% of `attention_generation` test cases to `RuntimeError: scheduler_metadata must have shape (metadata_size)` from `_vllm_fa3_C.fwd`. The same bug exists on every Hopper run that goes through the FA3 path; both the recent DSV32 collection on `h100_sxm` (vllm 0.19.0, this branch's parent) and the existing `h200_sxm` data added by #894 (vllm 0.19.1) are partial as a result. On Blackwell (B200/B300/GB200/GB300) it's silent because vLLM doesn't pick FA3 there (`aot_schedule = get_flash_attn_version() == 3` is False). Two commits here remove both root causes; an h100_sxm probe with the first commit alone reduced `attention_context` errors 2.5× and `attention_generation` 1.9×, with the entire residual being sliding-window-only cases that the second commit targets.

## Root cause

`collector/vllm/collect_attn.py` builds a minimal `VllmConfig` and a fresh FA3 builder per test case. The FA3 V1 metadata builder makes two implicit assumptions about that config that the existing fixture violates:

1. `ModelConfig.get_num_attention_heads()` / `get_num_kv_heads()` return the actual head counts the kernel will see. The builder uses these to size `scheduler_metadata` via FA3's AOT scheduler (`get_scheduler_metadata`). They route through `model_arch_config`, which is built once at `ModelConfig.__init__` time from the fake model's `config.json` (`16 q-heads / 8 kv-heads`). Mutating `hf_config` afterwards does not propagate — the cached arch values stay stale.

2. `_get_sliding_window_configs(vllm_config)` returns a non-empty set for sliding-window models. It walks registered `Attention` layers via `get_layers_from_vllm_config(...)`; the collector registers none, so the function returns an empty set and the builder defaults `aot_sliding_window` to `(-1, -1)`.

The test sweep iterates arbitrary `(num_heads, num_kv_heads, head_dim, window_size)` tuples, but `create_vllm_config` only forwarded `head_dim`. So the builder pre-computed `scheduler_metadata` for `(16, 8, head_dim, no-sliding)`, then `impl.forward` called FA3 with the test's actual `(num_heads, num_kv_heads, head_dim, (window_size-1, 0))` shape. The kernel computed a different `metadata_size` from the real inputs and rejected the AOT metadata.

## Fix

Two commits, separable:

**`5ec51d6` — collector: pass `num_heads`/`num_kv_heads` through `create_vllm_config`.** Extends `create_vllm_config` with optional `num_heads` / `num_kv_heads` parameters mirroring the existing `head_dim` override. Sets both `hf_config` and `model_arch_config` so the cached arch values match the test's actual shapes. `collect_attn.py:run_attention_torch` forwards the per-case values.

**`57b0712` — collector: align FA3 builder's `aot_sliding_window` with impl.** Pre-populates `builder.aot_sliding_window = (window_size - 1, 0)` when `window_size > 0`, matching what `FlashAttentionImpl.__init__` does for `DECODER` attention (`flash_attn.py:581`).

**`fbb0e43` — collector: extend FA3 `num_heads`/`num_kv_heads` override to MLA collector.** Same root cause was bleeding into `collect_mla_module_v3.py` (vllm ≥ 0.19.0 MLA path). It only mutated `hf_config.{num_attention_heads,num_key_value_heads}` after constructing `VllmConfig` — leaving `model_arch_config` stale, so the FA3 builder pre-computed metadata for the model's natural head count while `impl.forward` ran with the test sweep's `num_heads`. Forwarding `num_heads`/`num_kv_heads` to `create_vllm_config` (uses the same shared override path as commit 1) keeps both attribute paths in sync. For MLA, `get_num_kv_heads()` returns 1 unconditionally (use_mla path), so the kv-heads override is a no-op there but is passed for parity. The pre-existing `hf_config.num_*` post-hoc mutations are kept as-is for minimal diff — they now redundantly set the same values `create_vllm_config` has already set.

All three new attribute writes are guarded by `hasattr` so older vLLM versions (different attribute names, or no `model_arch_config`) silently fall through. The FA3 bug exists from vllm≥0.19, and on older versions FA3 isn't the default backend on Hopper anyway, so the override degrading to a no-op there is acceptable.

## Validation

Numbers below are pulled directly from `collector.log` (`Generated N test cases for X` and `Completed with K errors`):

| Run | Op | Generated | Errors | Error rate |
|---|---|---|---|---|
| OG (job 313277754, vllm 0.19.0, campaign branch) | attention_context | 21,816 | 2,262 | **10.4%** |
| OG | attention_generation | 20,908 | 4,874 | **23.3%** |
| probe#1 (fix branch, commit 1 only) | attention_context | 43,656 | 1,799 | **4.1%** |
| probe#1 | attention_generation | 45,056 | 5,466 | **12.1%** |
| probe#2 (commits 1 + 2) | attention_context | 43,656 | **0** | **0%** |
| probe#2 | attention_generation | 45,056 | **0** | **0%** |

probe#1's residual 1,799 `attention_context` errors are **100% `window_size=128` cases** — exactly what commit 2 patches. After commit 2, probe#2 on the same h100_sxm sweep produces **0 errors** across both attention ops. Static trace through vllm 0.19.0 source confirms the fix path (`flash_attn.py:296`, `:299`, `:236-245`, `:386`).

Concurrent re-runs on `h200_sxm` / `gb200` / `gb300` / `b300_sxm` running this branch (`mla_*+dsa_*` op set) show `mla_context_module: Completed successfully with no errors` on systems where the previous run had several hundred errors, plus zero regression on Blackwell where the fix isn't required.

### Production validation across all 5 DSV32 target systems

The fix branch has now been used end-to-end to collect publishable DSV32 perf data on every Hopper / Blackwell system in scope. 13 data PRs are in flight or merged from this branch — empirical confirmation the fix is correct across the full op sweep, not just attention probes:

| System | Data PR(s) on this branch | mla error rate | dsa error rate | Notes |
|---|---|---|---|---|
| **b300_sxm** | #1007 (full bundle) | 0% | 0% (only OOM/FlashMLA-stride at extremes) | First system to land — cleanest run |
| **h100_sxm** | #1013 (mla), #1009 (dsa_ctx), #1014 (dsa_gen) | 71 / 9.8k = 0.7% (mostly OOM cascade) | 430 dsa_ctx (mostly FlashMLA stride int32 overflow) + 192 dsa_gen OOMs | All FA3 attention errors gone vs OG (which had 71 mla_context errors with ~25% from this same scheduler_metadata bug) |
| **h200_sxm** | #1019 (mla), #1008 (dsa_ctx), #1011 (dsa_gen) | 20 / 9.8k = 0.20% (single-GPU CUDA flake) | 400 dsa_ctx FlashMLA-stride only | Calibration vs Blackwell: h200/b200 ratio 1.42x — expected Hopper progression |
| **gb200** | #1016 (mla), #1012 (dsa_ctx), #1015 (dsa_gen) | 0% | 1200 dsa_ctx FlashMLA-stride only | Calibration: gb200/b200 ratio 0.95x |
| **gb300** | #1010 (mla), #1018 (dsa_ctx), #1017 (dsa_gen) | 0% | 1200 dsa_ctx FlashMLA-stride only | Calibration: b300/gb200 ratio 1.01x — same hardware perf |

All errors observed across 13 data PRs are either (a) configurations that legitimately don't fit (OOM, FlashMLA's known int32 stride overflow on extreme `(b, s)` corners) or (b) rare worker-process flakes. **Zero `scheduler_metadata` errors leaked through** — the bug this PR targets is fully closed across attention, mla_context_module, and mla_generation_module on Hopper. Data files are calibrated consistently across systems (b200/gb200/b300 all within 0.95-1.01x of each other; h200 is 1.4x slower as expected for Hopper vs Blackwell).

**CI outcome on the 13 data PRs**: 12 of 13 are fully green and mergeable as-is. The single exception is #1007 (b300_sxm full bundle), where one cell — Kimi-K2.5 b300_sxm vllm agg — trips the AIC Accuracy Regression gate due to a mode flip (hybrid → silicon) on a cell where hybrid was *coincidentally* very close to ground truth (13% MAPE) and silicon predictions land at the typical vllm-agg silicon error (~30%). It's the same chronic vllm-agg-path miscalibration that's been silently shipping since PR #894 (b200 vllm agg has been at +143% MAPE under silicon mode all along — only invisible to the gate because OLD == NEW there). #1016 (gb200 mla bundle), which we initially expected might also trip, actually *improved* TTFT MAPE on Kimi-K2.5 gb200 vllm disagg — the disagg code path's hybrid-vs-silicon behavior is the inverse of the agg path's. Tracked under [AIC-1033](https://linear.app/nvidia/issue/AIC-1033/vllm-silicon-mode-mla-prediction-ttft-tpot-has-chronic-large-mape-on); not blocking for this PR.

## Cross-checking against the existing H200 attention data

The `h200_sxm/vllm/0.19.0/context_attention_perf.txt` LFS blob currently in `main` is a cumulative file:

| Source | vllm version | Rows | Sweep characteristics |
|---|---|---|---|
| pre-#894 | `0.19.0` | 10,120 | window=0 only, head_dim=128 only — old narrow sweep |
| #894 contribution | `0.19.1` | 39,832 | window=0/128 split 50/50, head_dim=64/128 — the expanded sweep |

The vllm 0.19.0 → 0.19.1 diff has 19 commits, **none touch `flash_attn` or `vllm_flash_attn`**. So the FA3 builder is identical in 0.19.1 — the bug is the same. PR #894's H200 collection added 39,832 rows out of ~43.7k expected cases, an implied ~8.7% error rate. That matches our H100 OG's 10.4% within the expected slop. The H200 data only "looks complete" because the failure stream is `errors_vllm.attention_*.json` in the CI artifacts, never merged to the repo. After this fix, future re-collections (H100, H200, or any new Hopper system) should add ~9% more coverage relative to today's blobs.

## Impact

- Unblocks DSV32 (DeepSeek-V3.2) silicon-mode support for vLLM 0.19 on `h100_sxm` / `h200_sxm` once the missing `mla_*` / `dsa_*` perf data is collected (the in-flight pipelines on this branch will produce that).
- Closes the silent ~10% / ~23% data gap on every existing and future Hopper attention collection (visible only in the artifact `errors_*.json`, not in the perf `.txt`). Re-collecting H200 with this branch should add roughly the missing 3-4k rows that PR #894's run lost.
- Tightens MLA prefill error rates on Hopper (h100 `mla_context_module` had 282 errors in the OG run, a meaningful chunk of which were the same bug bleeding into MLA's prefill path).

## Test plan

- [x] Static analysis traced both fixes through vllm 0.19.0 source
- [x] probe#1 on h100_sxm with `attention_context,attention_generation`: 10.4% → 4.1% / 23.3% → 12.1%, residual 100% sliding-window
- [x] probe#2 on h100_sxm with commits 1 + 2: **0 errors** on both ops (43,656 / 45,056 cases)
- [x] In-flight retries on h200_sxm (mla path): 272 errors confirmed same scheduler_metadata pattern in MLA — motivation for commit 3
- [x] Re-collection on h100_sxm with the full op sweep on the 3-commit branch: PRs #1013 / #1009 / #1014 — zero `scheduler_metadata` errors on attention or MLA paths
- [x] Re-collection on h200_sxm: PRs #1019 / #1008 / #1011 — clean (0.2% rate, single-GPU CUDA flake unrelated to this fix)
- [x] Re-collection on gb200 / gb300 / b300_sxm (Blackwell, fix is no-op there): PRs #1016 / #1012 / #1015 / #1010 / #1018 / #1017 / #1007 — all clean, calibration consistent vs prior data
- [x] Cross-system calibration check: 4-way matched-point latency ratios across Blackwell (b200/gb200/b300) are 0.95-1.01x; h200/b200 = 1.42x as expected for Hopper hardware
- [x] No regression on pre-existing non-DSV32 collections by construction — the change is no-op for callers that don't pass `num_heads`/`num_kv_heads`, and the new attribute writes are `hasattr`-guarded for older vLLM

## Out of scope

While investigating this PR's downstream data PR (#1007), a separate, **pre-existing** calibration issue surfaced: AIC's silicon-mode TTFT/TPOT predictions for vLLM-backend MLA-family models (DSV-family + Kimi-K2.5) miss real silicon ground truth by large MAPE — and the magnitude/direction is path-specific (`run_agg` overshoots, `run_disagg` undershoots in hybrid and is *better* under silicon). This has been silently shipping since PR #894 landed b200's MLA data last week — the regression gate is a delta-detector and only catches *changes* across PRs, not absolute prediction quality once silicon data is in place. **It is NOT caused by this PR**; data files are calibrated consistently across collected systems (verified above). Tracked separately under [AIC-1033](https://linear.app/nvidia/issue/AIC-1033/vllm-silicon-mode-mla-prediction-ttft-tpot-has-chronic-large-mape-on). Of the 13 data PRs in this validation set, **only #1007 trips the regression gate** — same root cause (the mode-flip on a coincidentally-accurate hybrid cell). The other 12 land cleanly because their cells either don't appear in the test fixture or hit the disagg path where silicon mode happens to improve predictions.

### DIFF
--- a/collector/vllm/collect_attn.py
+++ b/collector/vllm/collect_attn.py
@@ -288,6 +288,15 @@ def run_attention_torch(
         builder = builder_cls(kv_cache_spec, layer_names, vllm_config, device)
         if backend_name_str == "FLEX_ATTENTION":
             builder.direct_build = use_direct_block_mask
+        # FA3's metadata builder auto-detects sliding_window by walking registered
+        # Attention layers; the collector doesn't register any, so the auto-detect
+        # finds nothing and aot_sliding_window stays at (-1, -1). Then FA3's
+        # AOT scheduler computes scheduler_metadata for "no sliding window", but
+        # impl.forward later passes the actual (window_size-1, 0) tuple, and FA3's
+        # shape check rejects the mismatched metadata_size. Pre-populate the
+        # builder's aot_sliding_window to match what impl.sliding_window will be.
+        if window_size > 0 and hasattr(builder, "aot_sliding_window"):
+            builder.aot_sliding_window = (window_size - 1, 0)
         attn_metadata = builder.build(
             common_prefix_len=0,
             common_attn_metadata=common_attn_metadata,
--- a/collector/vllm/collect_attn.py
+++ b/collector/vllm/collect_attn.py
@@ -111,6 +111,8 @@ def run_attention_torch(
         use_fp8_kv_cache=use_fp8_kv_cache,
         sliding_window=window_size if window_size > 0 else None,
         head_dim=head_dim,
+        num_heads=num_heads,
+        num_kv_heads=num_kv_heads,
     )
 
     # vLLM >=0.19.0 requires an active config context for backend selection
--- a/collector/vllm/collect_mla_module_v3.py
+++ b/collector/vllm/collect_mla_module_v3.py
@@ -343,6 +343,17 @@ def _create_attention_module(
         max_num_batched_tokens=max(max_batch_size * max_seq_len, 131072) if is_context else max_batch_size,
         use_fp8_kv_cache=use_fp8_kv_cache,
         trust_remote_code=True,
+        # Forward the test sweep's per-case head counts so that
+        # ModelConfig.model_arch_config (built once at __init__ from hf_config) stays in
+        # sync. Without this, the V1 FA3 builder reads the model's natural head count
+        # via get_num_attention_heads(), the AOT scheduler precomputes
+        # scheduler_metadata for that shape, and impl.forward then runs with the test's
+        # actual head count — _vllm_fa3_C.fwd's shape check rejects the mismatch.
+        # MLA collapses KV to 1 head via get_num_kv_heads (use_mla=True), so the
+        # num_kv_heads override is a no-op in that path; we pass it for parity with
+        # the attention collector.
+        num_heads=num_heads,
+        num_kv_heads=num_heads,
     )
 
     # Override quant_config to control linear-layer GEMM precision.
--- a/collector/vllm/utils.py
+++ b/collector/vllm/utils.py
@@ -252,6 +252,8 @@ def create_vllm_config(
     trust_remote_code: bool = False,
     sliding_window: int | None = None,
     head_dim: int | None = None,
+    num_heads: int | None = None,
+    num_kv_heads: int | None = None,
 ) -> VllmConfig:
     """Create a VllmConfig for testing with reasonable defaults."""
 
@@ -320,6 +322,25 @@ def create_vllm_config(
     if head_dim is not None:
         model_config.hf_config.head_dim = head_dim
         model_config.model_arch_config.head_size = head_dim
+    # ModelConfig.model_arch_config is built once from hf_config in __init__,
+    # so mutating hf_config alone leaves the cached arch values stale. Backends
+    # such as the V1 FA3 builder read num_heads/kv_heads via
+    # ModelConfig.get_num_attention_heads / get_num_kv_heads, which look at
+    # model_arch_config — without these overrides the AOT scheduler builds
+    # scheduler_metadata for the fake model's defaults (16 q-heads / 8 kv-heads)
+    # while the kernel call runs with the test's actual head counts, and FA3's
+    # shape check rejects the mismatched scheduler_metadata. The hasattr guards
+    # let this degrade gracefully on older vLLM where the attribute names may
+    # differ (the FA3 bug only exists from vllm>=0.19 anyway).
+    arch_cfg = getattr(model_config, "model_arch_config", None)
+    if num_heads is not None:
+        model_config.hf_config.num_attention_heads = num_heads
+        if arch_cfg is not None and hasattr(arch_cfg, "total_num_attention_heads"):
+            arch_cfg.total_num_attention_heads = num_heads
+    if num_kv_heads is not None:
+        model_config.hf_config.num_key_value_heads = num_kv_heads
+        if arch_cfg is not None and hasattr(arch_cfg, "total_num_kv_heads"):
+            arch_cfg.total_num_kv_heads = num_kv_heads
 
     return VllmConfig(
         model_config=model_config,
--- a/src/aiconfigurator/systems/data/b300_sxm/vllm/0.19.0/dsa_context_module_perf.txt
+++ b/src/aiconfigurator/systems/data/b300_sxm/vllm/0.19.0/dsa_context_module_perf.txt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:8f511ec19dbca6fa508fc973369a25131e882aff6e23a6752cb73d237157363d
+size 4941459
--- a/src/aiconfigurator/systems/data/b300_sxm/vllm/0.19.0/dsa_generation_module_perf.txt
+++ b/src/aiconfigurator/systems/data/b300_sxm/vllm/0.19.0/dsa_generation_module_perf.txt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:a5360bdd5213ccf492fd2d3aee5e9801e4a9dd72f2d89bef697dd24a612e89f8
+size 2644788
--- a/src/aiconfigurator/systems/data/b300_sxm/vllm/0.19.0/mla_context_module_perf.txt
+++ b/src/aiconfigurator/systems/data/b300_sxm/vllm/0.19.0/mla_context_module_perf.txt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:270247cd76e05cf329feee92b0c317e46fb830bbc546a249c1eaf7c008620d2a
+size 2597230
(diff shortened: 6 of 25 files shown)


jasonqinzhou (approved):

jasonqinzhou:
Kimi is using **MLA**, not DSA.

More precisely: Kimi-K2.5 is routed through the DeepSeek-V3/MLA model path. For vLLM it tries the module-level MLA op first:

- context: `context_mla_module`
- generation: `generation_mla_module`

If module data is missing, it falls back to decomposed ops: GEMMs plus vLLM standard attention. That fallback is what the old revision effectively used for B300 vLLM, because `b300_sxm/vllm/0.19.0` did not have `mla_*_module_perf.txt`.

This PR adds:

- `mla_context_module_perf.txt`
- `mla_generation_module_perf.txt`
- also DSA files, but those are not what Kimi uses

So Kimi got worse because the new revision switched B300 vLLM Kimi estimates from the decomposed fallback to the newly added **MLA module-level perf table**. Those new MLA module timings raise TTFT predictions by about 20-30% for many Kimi B300 vLLM aggregate rows.

Example:

```text
kimik2_5_b300_vllm_fp4_agg_8192x1024_c4_cfg813
silicon TTFT: 211.49 ms
old pred:     299.42 ms
new pred:     385.30 ms
```

That pushes the Kimi/B300/vLLM/agg TTFT MAPE from `13.27%` to `29.95%`, so the partition reports `-16.68%` improvement and fails the 10% threshold.

The likely root cause is: the newly collected B300 vLLM **MLA module** data is more pessimistic for Kimi than the old fallback model. It may be “more real” from the collector’s point of view, but relative to the silicon samples in the accuracy set, it predicts too slow.

PR #1021, opened 2026-05-09
chore: Add data from auto-collector pipeline 50773821 (h100_sxm_vllm_0.19.0)

dynamo-ops:
# Error Summary for Auto-Collector Run


Arsene12358:
Closing — collection was cut short by dlcluster orchestration; only ~11% of attention_context cases (~4707/43656) made it to disk and `generation_attention_perf.txt` was never produced. Re-running as split pipelines (one per op) to reduce per-pipeline duration.

PR #1060, opened 2026-05-12
feat: Add "HW_INCOMPATIBLE" result to support matrix

ilyasher:
In the support matrix, all combinations previously resulted in "PASS" or "FAIL". In some cases, the "FAIL" is a AIC bug/shortcoming, but sometimes the model+system combination is not supported together. This PR detects if a model uses a datatype which the system does not support, and marks the support matrix result as "Hardware incompatible" rather than "FAIL". Example: DeepSeek-R1-NVFP4 + a100_sxm.

Example:
notes
1. ignore the empty cells, they're empty because I tested with only a partially-generated support matrix
2. the text spilling from the box was fixed after I took the screenshot
image

## Release Notes

* **New Features**
  * Support matrix now recognizes and displays hardware incompatibility status alongside pass/fail results
  * Added dedicated visual styling (gray) for hardware-incompatible configurations
  * Extended matrix legend and error details to describe hardware incompatibility scenarios

* **Bug Fixes**
  * Improved error message aggregation for incompatible hardware configurations
  * Refined fallback logic to correctly exclude incompatible rows from architecture-based evaluation

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiconfigurator/pull/1060)

### DIFF
--- a/docs/support-matrix/index.html
+++ b/docs/support-matrix/index.html
@@ -57,11 +57,16 @@ tailwind.config = {
     font-size: 0.72rem;
     font-family: ui-monospace, SFMono-Regular, Menlo, Consolas, monospace;
     line-height: 1.4;
-    white-space: pre;
+    white-space: pre-wrap;
+    overflow-wrap: anywhere;
+    word-break: break-word;
     border-radius: 4px;
     box-shadow: 0 2px 8px rgba(0,0,0,.25);
     pointer-events: none;
-    max-width: 480px;
+    width: max-content;
+    max-width: min(480px, calc(100vw - 2rem));
+    box-sizing: border-box;
+    text-align: left;
   }
   html.dark td[data-tip]:hover::after {
     background: rgb(30 41 59);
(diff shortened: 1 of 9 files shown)


docs/support-matrix/index.html
```diff
@@ -79,6 +84,7 @@
 // ── Constants ───────────────────────────────────────────────────────
 const REPO = 'ai-dynamo/aiconfigurator';
 const CSV_PATH = 'src/aiconfigurator/systems/support_matrix.csv';
+const STATUS_HW_INCOMPATIBLE = 'HW_INCOMPATIBLE';
```
Harrilee, line 87: Let's follow up on the discussion here to see if we want to proceed with the partial solution.

https://nvidia.slack.com/archives/C096945HE8M/p1778626839451589

Harrilee:
Starting a test run: https://github.com/ai-dynamo/aiconfigurator/actions/runs/25892173157/job/76097525930

Harrilee (approved):
Let

PR #1063, opened 2026-05-13
feat(sdk): warn when FallbackOp silently swallows the primary op failure

simone-chen:
#### Overview:

Make `FallbackOp` visible: log a WARNING when it silently picks the fallback because the primary raised, and tag the result with `source="fallback"` so per-ops attribution shows it.

#### Before / after on Kimi K2.5 NVFP4 on B200 (the config that motivated this fix):

`cli_estimate(model="nvidia/Kimi-K2.5-NVFP4", system="b200_sxm", backend="trtllm", ...)` — the MLA module table is present but has no NVFP4 GEMM-quant rows, so `FallbackOp.primary` (MLAModule) raises `AssertionError: values is None or empty` and the fallback chain (GEMM + MLABmm + GenerationMLA + ...) is summed instead.

**Before:** silent. `TTFT=135.06ms TPOT=22.21ms` with no signal that the MLA attention component was modelled by the fallback path.

**After:** one WARNING per `FallbackOp` instance per exception type (dedup'd across the 1024-iter generation loop):

```
WARNING aiconfigurator.sdk.operations | FallbackOp 'context_mla_block':
  primary op 'context_mla_module' failed (AssertionError: values is None
  or empty ... gemm_quant_mode=nvfp4 ...); returning sum of fallback ops
  [context_downscale_gemm, context_q_b_proj_gemm, context_kv_b_proj_gemm,
  context_attention, context_proj_gemm]. Predictions for this op will be
  tagged source='fallback'.

WARNING aiconfigurator.sdk.operations | FallbackOp 'generation_mla_block':
  primary op 'generation_mla_module' failed (AssertionError: values is
  None or empty ... gemm_quant_mode=nvfp4 ...); returning sum of fallback
  ops [generation_downscale_gemm, generation_q_b_proj_gemm,
  generation_bmm_pre, generation_attention, generation_bmm_post,
  generation_proj_gemm]. Predictions for this op will be tagged
  source='fallback'.
```

#### Details:

- **WARNING on swallow**: includes op name, primary name, exception type+message, and the fallback op list.
- **Dedup**: per-instance `set[exception_type]` — first occurrence is WARNING, subsequent are DEBUG, so sweeps don't spam.
- **Source tag**: fallback `PerformanceResult` gets `source="fallback"`; propagates through existing `_merge_source` into `per_ops_source.json`.

`src/aiconfigurator/sdk/operations.py` (`FallbackOp.query`).

#### Related Issues:

- Closes AIC-1059
- Would have caught AIC-1057 immediately

### DIFF
--- a/src/aiconfigurator/sdk/inference_summary.py
+++ b/src/aiconfigurator/sdk/inference_summary.py
@@ -59,8 +59,8 @@ class InferenceSummary:
         self._generation_latency_dict = {}  # ms
         self._context_energy_wms_dict = {}  # RENAMED from _context_power_dict, W·ms
         self._generation_energy_wms_dict = {}  # RENAMED from _generation_power_dict, W·ms
-        # Per-op data source ("silicon", "empirical", or "mixed") populated by
-        # base_backend phase helpers from PerformanceResult.source.
+        # Per-op data source ("silicon", "empirical", "fallback", or "mixed")
+        # populated by base_backend phase helpers from PerformanceResult.source.
         self._context_source_dict: dict[str, str] = {}
         self._generation_source_dict: dict[str, str] = {}
         self._is_oom = None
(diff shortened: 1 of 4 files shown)


simone-chen:
Closed as DS is the only model that uses Fallback Ops. Not worth the refactor

PR #1068, opened 2026-05-13
fix(interpolation): preserve energy in interp_1d and grid extrapolation

kaim-eng:
#### Overview:

Fix the PR153 power-estimation feature so that `energy` (W·ms) is preserved through all interpolation paths. Two paths in `src/aiconfigurator/sdk/interpolation.py` only operated on `latency` and `power` when the leaf was a dict and silently dropped `energy`, which caused downstream `KeyError: 'energy'` failures or — more insidiously — power values of `0 W` for any GEMM/attention shape that landed on an extrapolated grid point.

#### Details:

PR153 (per-GPU average-power estimation) added an `energy` key alongside `latency` and `power` in the dict leaves of the perf database. Two places didn't get the memo:

1. `interp_1d()` — for dict leaves it hard-coded extracting `latency` and `power` and returned a dict containing only those two keys. Any caller that later looked up `energy` either crashed with `KeyError` or, after recent defensive cleanups, fell back to `energy = 0.0`.
2. `_extrapolate_data_grid()` `sqrt_y_value` branch — applied the sqrt/square transform to a hard-coded subset of keys, so newly materialized extrapolated grid cells lacked an `energy` entry.

This PR rewrites both paths to be generic over numeric keys in dict leaves: every `(int|float)` value present in both endpoint dicts is interpolated/transformed independently, so `energy` is preserved and any future metric (e.g., `tgp_w`, `sm_clock_mhz`) is handled automatically.

Two regression tests are added in `tests/unit/sdk/database/test_interpolation.py`:

- `test_interp_1d_preserves_energy_dict_leaves` — asserts `interp_1d` returns a dict containing `energy` for midpoint, endpoint, and extrapolated values, and that the interpolated energy matches the expected linear interpolation of the two endpoint energies.
- `test_extrapolate_data_grid_preserves_energy_dict_leaves` — builds a sparse 2D grid with `energy` leaves and asserts the newly materialized cells produced by extrapolation contain a non-zero `energy` value.

Both tests reproduce the original symptom on a worktree where the interpolation fix is reverted (`KeyError: 'energy'` from `interp_1d`; missing-key assertion from the grid path) and pass with this change.

End-to-end validation reproduced the documented PR153 power-estimation sweeps locally on the H200:

- vLLM 0.19.1 bfloat16 overlay (§10.8.8) — non-zero `power_w` restored.
- TRT-LLM 1.3.0rc2 FP8 overlay (§10.8.2) — non-zero `power_w` restored.
- Stock latency-only sweeps (§10.8.7, §10.12.3) — unchanged, as expected.

- `src/aiconfigurator/sdk/interpolation.py` — the two changed functions, `interp_1d` and `_extrapolate_data_grid`. Focus on:
  - The new `_interp_scalar` helper closure inside `interp_1d` and the `for key in y0.keys() & y1.keys():` loop that replaces the hard-coded latency/power extraction.
  - The generalization of the `sqrt_y_value` branch in `_extrapolate_data_grid` to iterate over numeric keys in dict leaves.
- `tests/unit/sdk/database/test_interpolation.py` — the two new regression tests, which document the invariants this PR enforces.

#### Verification:

- `pytest tests/unit/sdk/database/test_interpolation.py` — 21 / 21 passed.
- `pytest tests/unit/sdk/database` (excluding `test_moe_dispatch.py` which needs `torch`) — 297 / 297 passed.

#### Related Issues:

- Relates to PR153 (power-estimation feature).

## Summary by CodeRabbit

* **Bug Fixes**
  * Interpolation and extrapolation now consistently preserve all numeric metric fields (e.g., energy) for both scalar and dict-formatted data, and retain prior sign-crossing protection to avoid spurious sign flips.

* **Tests**
  * Added regression tests that verify metric preservation and correct interpolation/extrapolation behavior across 1D and multi-dimensional grid scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiconfigurator/pull/1068?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/aiconfigurator/sdk/interpolation.py
+++ b/src/aiconfigurator/sdk/interpolation.py
@@ -158,35 +158,35 @@ def extract_latency_and_energy_3d(data: dict) -> tuple[dict, dict]:
 
 
 def interp_1d(x: list[int], y: list, value: int):
-    """Linear interpolation in 1-D. Handles both float and dict leaf values."""
+    """Linear interpolation in 1-D. Handles both float and dict leaf values.
+
+    When the leaves are dicts, every numeric metric present in BOTH endpoints
+    is interpolated (e.g. ``latency``, ``power``, ``energy``). Dropping
+    ``energy`` here used to zero out every extrapolated grid point and made
+    the SDK power model return 0 W for any GEMM shape that wasn't an exact
+    collection point.
+    """
     x0, x1 = x
     y0, y1 = y
 
+    def _interp_scalar(v0, v1):
+        a, b = v0, v1
+        if (x0 - x1) * (a - b) < 0 and (value - x0) * (value - x1) > 0:
+            b = a
+        if a == b:
+            return a
+        return a + (b - a) / (x1 - x0) * (value - x0)
+
     if isinstance(y0, dict) and isinstance(y1, dict):
-        lat0, lat1 = y0["latency"], y1["latency"]
-        pow0, pow1 = y0["power"], y1["power"]
+        result = {}
+        for key in y0.keys() & y1.keys():
+            v0 = y0[key]
+            v1 = y1[key]
+            if isinstance(v0, (int, float)) and isinstance(v1, (int, float)):
+                result[key] = _interp_scalar(v0, v1)
+        return result
 
-        if (x0 - x1) * (lat0 - lat1) < 0 and (value - x0) * (value - x1) > 0:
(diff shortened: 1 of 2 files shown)


jasonqinzhou (approved):
Please fix the coderabbit comments.

PR #1074, opened 2026-05-13
fix: Infer moe_tp and moe_ep when only 1 is provided

### LINKED ISSUE #9398: mocker: MiniMax-M2 is marked PASS in AIC, but cannot start with --aic-perf-model
# MiniMax-M2 is marked PASS in AIC, but Dynamo Mocker cannot start with `--aic-perf-model`

## Bottom Line

The AIC support matrix marks several MiniMax-M2 cells as `PASS`, but Dynamo Mocker cannot start those same models with `--aic-perf-model`.

The failure is deterministic on Linux and happens before the mock engine finishes startup:

```text
Exception: Failed to create AIC callback (--aic-perf-model was requested):
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```

This is a contract bug between AIC and Dynamo Mocker. The AIC CLI can still find a valid configuration during a coarse sweep, but that does not prove the per-op/per-shape runtime path used by Mocker is usable. Today the support matrix reads as broader than what is actually validated.

## Confirmed Scope

Validated on 2026-05-13 in a Linux CPU pod on the NScale cluster.

| Model | System | Backend | AIC / data version | Mocker result |
|---|---:|---:|---:|---|
| `nvidia/MiniMax-M2.5-NVFP4` | `b200_sxm` | `vllm` | `0.19.0`, AIC current main `9b13af67` | Fails with `NoneType * int` |
| `nvidia/MiniMax-M2.7-NVFP4` | `b200_sxm` | `vllm` | `0.19.0`, AIC current main `9b13af67` | Fails with `NoneType * int` |
| `MiniMaxAI/MiniMax-M2.5` | `b200_sxm` | `vllm` | `0.14.0`, AIC `0.8.0` | Fails with `NoneType * int` |

Current AIC main no longer ships the `b200_sxm/vllm/0.14.0` database, so the `MiniMaxAI/MiniMax-M2.5` row should be treated as a released-stack `aiconfigurator==0.8.0` repro. The NVFP4 rows still reproduce with current AIC main.

The current support matrix still lists the NVFP4 MiniMax rows as `PASS`, for example:

```csv
nvidia/MiniMax-M2.5-NVFP4,b200_sxm,vllm,0.19.0,agg,PASS
nvidia/MiniMax-M2.5-NVFP4,b200_sxm,vllm,0.19.0,disagg,PASS
nvidia/MiniMax-M2.7-NVFP4,b200_sxm,vllm,0.19.0,agg,PASS
nvidia/MiniMax-M2.7-NVFP4,b200_sxm,vllm,0.19.0,disagg,PASS
```

## Reproduction

```bash
python -m dynamo.mocker \
  --model-path nvidia/MiniMax-M2.5-NVFP4 \
  --num-workers 1 \
  --speedup-ratio 1.0 \
  --aic-perf-model \
  --aic-system b200_sxm \
  --aic-backend vllm \
  --aic-backend-version 0.19.0 \
  --aic-tp-size 1 \
  --aic-moe-ep-size 1 \
  --discovery-backend file \
  --request-plane tcp
```

Observed tail:

```text
WARN interpolation.extrapolate_data_grid:
  Skipping interpolation for z=51200 as it does not exist in both x_left=16384 and x_right=32768 for y=262144
...
INFO helpers._apply_model_quant_defaults:
  Resolved quant modes for MiniMaxM2ForCausalLM: gemm=nvfp4 moe=nvfp4 kvcache=fp8 fmha=bfloat16 comm=half
  File ".../dynamo/mocker/main.py", line 210, in launch_workers
    engine_config = await make_engine(runtime, entrypoint_args)
Exception: Failed to create AIC callback (--aic-perf-model was requested):
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```

The same command pattern fails for `nvidia/MiniMax-M2.7-NVFP4`. With `aiconfigurator==0.8.0`, the same command pattern also fails for `MiniMaxAI/MiniMax-M2.5` using `--aic-backend-version 0.14.0`.

## What Still Works

`aiconfigurator cli default` succeeds for the NVFP4 MiniMax cases. It emits the same interpolation-gap warnings but still returns a top configuration.

Representative direct AIC callback probes also succeed:

```python
from dynamo._internal.aic import create_session

s = create_session(
    "vllm",
    "b200_sxm",
    "nvidia/MiniMax-M2.5-NVFP4",
    1,
    "0.19.0",
    1,
    1,
    1,
)

s.predict_prefill(80, 4000, 0)    # succeeds
s.predict_decode(80, 4000, 2)     # succeeds
s.predict_prefill(1, 51200, 0)    # succeeds
s.predict_decode(80, 196608, 2)   # succeeds
```

So the problem is not simply "AIC cannot load this model" or "AIC cannot answer any query for this model." The failing path is the full Dynamo Mocker startup path, which appears to exercise an additional AIC model-construction, startup-estimation, or per-op query that the CLI and simple direct probes do not cover.

## Likely Root Cause

MiniMax-M2 is a large MoE shape: 256 experts, top-8 routing, hidden size 3072, MoE intermediate size 1536, very long context.

During Mocker startup, AIC emits many interpolation-gap warnings around grid coordinates such as `z=51200`, `z=65537`, and `z=131073`. One downstream path then attempts to multiply a missing numeric value:

```text
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```

The grid-gap warnings are not sufficient proof by themselves, because some direct AIC calls emit those warnings and still succeed. But the full Mocker path clearly reaches a missing-value case that is not handled as a typed AIC error.

The current error message is also too lossy. Dynamo reports only "failed to create AIC callback", which hides the op, shape, and AIC stack frame that produced the `None`.

## Product Impact

This breaks the promised "Dynamo Mocker + AIC as a CPU twin of Dynamo" workflow for MiniMax-M2.

The failure mode is especially bad because users see a green AIC support-matrix cell, then the Mocker path crashes before it can run. That makes `PASS` ambiguous: it currently means "the AIC CLI can produce at least one config", not "the model is usable through the Mocker runtime integration."

## Fix Requests

1. AIC should never return or internally propagate `None` for perf data in a numeric path.

   If a grid point is unsupported, raise a typed exception with enough context to debug it: model, system, backend, backend version, op name, quant mode, TP/EP, batch size, sequence length, prefix, and the missing grid coordinate.

2. Dynamo should preserve the underlying AIC error context.

   The Mocker startup error should say which AIC op/query failed. The Rust/Python callback bridge should avoid collapsing this into only "failed to create AIC callback."

3. The support matrix should distinguish CLI support from Mocker support.

   Either split the matrix into separate columns, for example `CLI pareto PASS` and `Mocker callback PASS`, or gate `PASS` on a minimal Linux Mocker startup test when the cell is meant to imply runtime integration support.

4. Add a regression test for these exact models.

   Minimal test: run `python -m dynamo.mocker --aic-perf-model` far enough to create the AIC callback for:

   - `nvidia/MiniMax-M2.5-NVFP4`, `b200_sxm`, `vllm`, `0.19.0`
   - `nvidia/MiniMax-M2.7-NVFP4`, `b200_sxm`, `vllm`, `0.19.0`
   - `MiniMaxAI/MiniMax-M2.5`, `b200_sxm`, `vllm`, `0.14.0` if preserving the AIC 0.8.0 support claim

5. Fill or explicitly mark the missing perf-grid coverage.

   If MiniMax-M2 needs shapes around `z=51200` or related long-context MoE coordinates, collect those data points or make the extrapolation policy explicit. If those shapes are intentionally unsupported, the matrix should say so.

## Validation Notes

Environment used for validation:

```text
Cluster: NScale CPU pod, Linux, Python 3.12
ai-dynamo: 1.2.0
ai-dynamo-runtime: 1.2.0
AIC released stack: aiconfigurator 0.8.0
AIC current main tested: 9b13af67
Dynamo Python current main overlaid for spot check: 78146fe2
```

The NVFP4 crash reproduces with:

1. published `ai-dynamo[mocker]==1.2.0` plus its resolved `aiconfigurator==0.8.0`, and
2. published Dynamo runtime with current AIC main and current Dynamo Python sources overlaid.

Direct AIC session probes succeeded for representative shapes, so the next debugging step should instrument the exact AIC query invoked during Mocker startup and print the op/shape before each `op.query` or startup-estimation call.

Related internal writeup: https://gitlab-master.nvidia.com/-/snippets/14598

ilyasher:
Fixes https://github.com/ai-dynamo/dynamo/issues/9398

When `moe_ep_size` is provided but `moe_tp_size` is not, `moe_tp_size` defaulted to `None` which caused a crash. This PR sets `moe_tp_size` and `moe_ep_size` to an int for MoE models.

## Summary by CodeRabbit

* **Improvements**
  * MoE model parallelism configuration validation now occurs earlier in the initialization process with clearer error messages for invalid setups.
  * Missing MoE parallelism dimensions are automatically inferred when possible with improved consistency.

* **Tests**
  * Added comprehensive test coverage for MoE parallelism resolution and validation scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiconfigurator/pull/1074)

### DIFF
--- a/src/aiconfigurator/cli/api.py
+++ b/src/aiconfigurator/cli/api.py
@@ -23,6 +23,7 @@ from aiconfigurator.cli.main import (
     build_experiment_task_configs,
 )
 from aiconfigurator.cli.report_and_save import save_results
+from aiconfigurator.sdk.config import ModelConfig
 from aiconfigurator.sdk.models import check_is_moe
 from aiconfigurator.sdk.task import (
     DEFAULT_DECODE_LATENCY_CORRECTION_SCALE,
@@ -557,26 +558,19 @@ def _resolve_moe_parallelism(
 ) -> tuple[int, int]:
     """Resolve and validate MoE parallelism widths, returning (moe_tp_size, moe_ep_size).
 
-    For dense (non-MoE) models the width constraint is not enforced because
-    MoE parallelism has no effect on the computation.
+    For dense (non-MoE) models, MoE parallelism has no effect on the
+    computation, so leave the fields as provided.
     """
-    if moe_tp_size is None and moe_ep_size is None:
-        moe_tp_size = tp_size
-        moe_ep_size = attention_dp_size
-    elif moe_tp_size is None:
-        moe_tp_size = tp_size * attention_dp_size // moe_ep_size
-    elif moe_ep_size is None:
-        moe_ep_size = tp_size * attention_dp_size // moe_tp_size
+    if model_path is not None and not check_is_moe(model_path):
+        return moe_tp_size, moe_ep_size
 
-    attn_width = tp_size * attention_dp_size
-    moe_width = moe_tp_size * moe_ep_size
-    if attn_width != moe_width and (model_path is None or check_is_moe(model_path)):
-        raise ValueError(
-            f"Parallelism width mismatch: tp_size({tp_size}) * attention_dp_size({attention_dp_size}) = {attn_width}, "
-            f"but moe_tp_size({moe_tp_size}) * moe_ep_size({moe_ep_size}) = {moe_width}. "
-            f"These must be equal."
-        )
-    return moe_tp_size, moe_ep_size
+    cfg = ModelConfig(
+        tp_size=tp_size,
+        attention_dp_size=attention_dp_size,
+        moe_tp_size=moe_tp_size,
+        moe_ep_size=moe_ep_size,
+    )
+    return cfg.resolve_moe_parallelism()
 
 
 def _build_model_config(
--- a/src/aiconfigurator/cli/api.py
+++ b/src/aiconfigurator/cli/api.py
@@ -677,8 +677,12 @@ def cli_estimate(
             prefill/decode TP when their specific args are omitted.
         pp_size: Pipeline parallelism size. Default is 1.
         attention_dp_size: Attention data parallelism size. Default is 1.
-        moe_tp_size: MoE tensor parallelism size. Default is None (auto).
-        moe_ep_size: MoE expert parallelism size. Default is None (auto).
+        moe_tp_size: MoE tensor parallelism size. At least one of ``moe_tp_size``
+            or ``moe_ep_size`` is required for MoE models; the missing dimension
+            is inferred when possible.
+        moe_ep_size: MoE expert parallelism size. At least one of ``moe_tp_size``
+            or ``moe_ep_size`` is required for MoE models; the missing dimension
+            is inferred when possible.
         gemm_quant_mode: GEMM quantization mode. Default is None (auto-inferred).
         kvcache_quant_mode: KV cache quantization mode. Default is None (auto-inferred).
         fmha_quant_mode: FMHA quantization mode. Default is None (auto-inferred).
--- a/src/aiconfigurator/sdk/config.py
+++ b/src/aiconfigurator/sdk/config.py
@@ -42,8 +42,8 @@ class ModelConfig:
         For MoE models, the attention width must match the expert width:
         ``tp_size * attention_dp_size == moe_tp_size * moe_ep_size``. If one
         MoE dimension is missing, infer it from the other. If both are missing,
-        keep the common default shape of ``moe_tp_size=tp_size`` and
-        ``moe_ep_size=attention_dp_size``.
+        raise an error so callers do not silently get an MoE layout they did
+        not request.
         """
 
         def _validate_positive(name: str, value: int) -> None:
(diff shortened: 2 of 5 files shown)


src/aiconfigurator/sdk/config.py
```diff
@@ -36,6 +36,64 @@ class ModelConfig:
...
+
+        _validate_positive("tp_size", self.tp_size)
+        _validate_positive("attention_dp_size", self.attention_dp_size)
+
+        attn_width = self.tp_size * self.attention_dp_size
+        moe_tp_size = self.moe_tp_size
+        moe_ep_size = self.moe_ep_size
+        if moe_tp_size is None and moe_ep_size is None:
+            moe_tp_size = self.tp_size
+            moe_ep_size = self.attention_dp_size
```
simone-chen: @coderabbitai I think it holds for vllm and trtllm, as they only allow `tp == moe_tp`.
What about sglang? It may allow `tp != moe_tp`

simone-chen: @ilyasher Might worth drop a TODO here to modify with wideep path.

simone-chen (approved):

src/aiconfigurator/cli/api.py
TrojanXu, line 552: a quick question, why we allow default value for moe_tp_size and moe_ep_size to be None when it's a moe model, should we just raise an error. We'd better not set default values for the user

ilyasher: good point. changed it so that at least one of moe_tp_size and moe_tp_size is required for MoE

PR #1114, opened 2026-05-16
fix: improve B200 support matrix coverage and replay tooling

jasonqinzhou:
This PR improves B200 SXM support-matrix coverage across the target SGLang, TRT-LLM, and vLLM versions, refreshes the recorded B200 performance data used by the matrix, and hardens the replay/tooling paths needed to keep the published support rows reproducible.

## What changed

- Added and expanded B200 collector cases and runtime collectors for attention, DSA/DSV4, MLA, MoE, GEMM, and FP8 static overhead data across the target backends.
- Refreshed B200 perf data and `b200_sxm` support-matrix rows with replayable AIC CLI commands for verified PASS cells.
- Updated support-matrix comparison/generation, docs/web rendering, and tests so row classifications and replay evidence are easier to audit.
- Stabilized replay for sparse B200 context-attention interpolation slices while keeping generic sparse GEMM misses strict.
- Completed final classification of deterministic non-PASS cells: remaining gaps are recorded as framework-incompatible or hardware/version-incompatible with row-level evidence, and no raw FAIL rows remain in the healed matrix.

## Coverage summary

Before counts normalize legacy `FAIL` rows to `FRAMEWORK_INCOMPATIBLE`, matching the dashboard snapshot.

| Scope | Rows | PASS | FRAMEWORK_INCOMPATIBLE | HW_INCOMPATIBLE | Pass rate |
| --- | ---: | ---: | ---: | ---: | ---: |
| Before, target branch | 470 | 306 | 164 | 0 | 65.1% |
| After, healed matrix | 546 | 410 | 48 | 88 | 75.1% |

Net PASS delta: `+104`.

Latest target versions only (`sglang 0.5.10`, `trtllm 1.3.0rc10`, `vllm 0.19.0`):

| Scope | Rows | PASS | FRAMEWORK_INCOMPATIBLE | HW_INCOMPATIBLE | Pass rate |
| --- | ---: | ---: | ---: | ---: | ---: |
| Before, target branch | 282 | 188 | 94 | 0 | 66.7% |
| After, healed matrix | 330 | 274 | 38 | 18 | 83.0% |

Latest-version net PASS delta: `+86`.

## Verification/status

- Delta replay verification passed: `205/205` replay commands passed for `410` rows changed to PASS.
- Focused tests passed: interpolation regressions, FP8 static support behavior, task validation, support-matrix versioning/framework-incompatibility tests, and CSV loading.
- Full-list before/after support-matrix screenshots were captured and uploaded.
- This PR remains a draft for final review.

### Before

B200 support matrix before

### After

B200 support matrix after


jasonqinzhou:
Added the B200 SGLang 0.5.10 heal work here instead of using a separate PR. This update includes the collector fix for head_dim=192/window_size logging plus the healed B200 SGLang 0.5.10 perf data. Verified the six previously failing SGLang 0.5.10 model families now pass both agg and disagg via SupportMatrix.run_single_test locally after fetching required LFS data, and the same pass result was verified on the live B200 debug node.

jasonqinzhou:
B200 SGLang MXFP4 healing update:

- Added SGLang 0.5.10 collector support for GPT-OSS `w4a16_mxfp4` and DeepSeek-V4 `w4a8_mxfp4_mxfp8` MoE paths on SM100/SM103.
- Recollected targeted B200 rows on `umbriel-b200-039` for GPT-OSS MoE plus missing context/generation attention shapes.
- Updated `b200_sxm.csv`: `openai/gpt-oss-20b`, `openai/gpt-oss-120b`, and `deepseek-ai/DeepSeek-V4-Flash` now PASS for SGLang 0.5.10 agg/disagg.
- Reclassified `deepseek-ai/DeepSeek-V4-Pro` SGLang 0.5.10 from stale MXFP4 HW_INCOMPATIBLE to FAIL with the current remaining blocker: missing mHC silicon data for hidden_size=7168. The stock `lmsysorg/sglang:v0.5.10-cu130` image rejects DeepSeek-V4 `deepseek_ref`, so Pro mHC recollection needs a DeepSeek-V4-capable SGLang image.

Verification:

- `PYTHONPATH=src:. SupportMatrix.run_single_test(...)` PASS for `openai/gpt-oss-20b`, `openai/gpt-oss-120b`, and `deepseek-ai/DeepSeek-V4-Flash` on `b200_sxm / sglang / 0.5.10`.
- `deepseek-ai/DeepSeek-V4-Pro` still FAILs with missing mHC data, no longer unsupported MXFP4.
- `uvx ruff check collector/sglang/collect_moe.py collector/sglang/collect_attn.py`
- `uvx ruff format collector/sglang/collect_moe.py collector/sglang/collect_attn.py --check`
- `git diff --check`

B200 debug allocation was released after collection.

jasonqinzhou:
B200 DeepSeek-V4-Pro-FP8 / SGLang 0.5.10 update:

- Recollected missing DeepSeek-V4 mHC rows for hidden_size=7168, hc_mult=4 on a B200 debug node.
- Added targeted fp8_block MoE rows for the Pro-FP8 shape, including power_law_1.01 8-way splits.
- Updated collect_mhc_module.py so the collector can use DeepSeek-V4 SGLang source/runtime paths: model_type=deepseek_v4, dsv4 attention backend, TileLang fallback for HC prenorm, and the newer hc_pre return tuple.
- Re-ran SupportMatrix.run_single_test('sgl-project/DeepSeek-V4-Pro-FP8', 'b200_sxm', 'sglang', '0.5.10'). The original missing mHC failure is gone, but the cell still fails on the remaining SGLang planner limitation: non-WideEP SGLang asserts TP>1 and DP>1 attention are not enabled simultaneously.
- Updated b200_sxm.csv for these two rows from stale HW_INCOMPATIBLE/missing-mHC wording to FAIL with the current GPU-verified reason.

Local checks: uvx ruff check collector/sglang/collect_mhc_module.py; uvx ruff format collector/sglang/collect_mhc_module.py --check; git diff --check.

jasonqinzhou:
B200 DeepSeek-V4-Pro SGLang 0.5.10 follow-up:

- Allocated one-GPU B200 debug node `6u1g-0017` under Slurm job `2202629`.
- Recollected 270 native `w4a8_mxfp4_mxfp8` MoE perf rows for `deepseek-ai/DeepSeek-V4-Pro` shape `hidden_size=7168/inter_size=3072/topk=6/num_experts=384` across one-node TP/EP combinations.
- Updated `b200_sxm.csv`: `deepseek-ai/DeepSeek-V4-Pro / sglang 0.5.10` is now `PASS` for both agg and disagg.
- Verification: `SupportMatrix.run_single_test(..., compare_engine_step_backends=False)` returns `{'agg': 'PASS', 'disagg': 'PASS'}`. The Rust comparison path on my Mac is separately blocked by missing `cargo`, not by missing support-matrix data.

src/aiconfigurator/sdk/operations/dsa.py
```diff
@@ -864,7 +864,76 @@ def missing_generation_dsa_error() -> PerfDataNotAvailableError:
                 )
             try:
                 dsa_dict = dsa_module_data[kv_cache_dtype][gemm_quant_mode][architecture]
-                result = interpolation.interp_3d(num_heads, b, s, dsa_dict, "cubic", database._extracted_metrics_cache)
+                # Generation callers pass ``s`` as total decode length, while
```
jasonqinzhou: We need it because generation DSA has an off-by-one convention mismatch.

The caller passes `s` as total decode length, but some collected DSA generation tables store the same workload by past-KV length. During decode, those differ by one: total length `150` corresponds to past-KV length `149`.

Without this block, the code can:

- miss an exact measured row keyed by `s - 1`
- interpolate at `s` even after discovering the table uses `s - 1`
- do cross-batch interpolation using mixed/wrong sequence coordinates
- fall back to `interp_3d(..., s, ...)`, still one token shifted

The new `normalized_s` logic says: prefer exact `s` if the table has it; otherwise, if this `num_heads` slice has `s - 1`, use that consistently for exact lookup, sequence interpolation, batch interpolation, and 3D fallback.

So it is a correctness fix for silicon lookup, not just a defensive fallback. It keeps total-length tables unchanged, but makes past-KV-keyed collector data return the measured/interpolated point for the actual decode workload.

src/aiconfigurator/sdk/operations/gemm.py
```diff
@@ -644,8 +656,10 @@ def query(self, database: PerfDatabase, **kwargs) -> PerformanceResult:
         energy = result.energy
         source = getattr(result, "source", "silicon")
 
-        # Adjust for fp8_static: subtract compute_scale overhead, only fix for trtllm now
-        if is_fp8_static:
+        # TRT-LLM's static-FP8 GEMM timing includes quantization overhead that
```
jasonqinzhou: I would keep it TRT-LLM-only for now.

The adjustment should apply only when the `fp8_static` GEMM row already includes quantization/scale overhead that AIC also models separately. Current collector behavior points that way:

- TRT-LLM: `fp8_static` uses `force_dynamic_quantization=True`, so the measured GEMM path includes compute-scale work. Subtracting avoids double counting.
- SGLang: `fp8_static` pre-quantizes activations/weights before the timed kernel and times `fp8_scaled_mm` directly. There is no dynamic quant overhead in that GEMM row to subtract.
- vLLM: the collector uses static activation FP8 path, so its GEMM row should already be the direct static runtime measurement.

So doing this for all frameworks would likely undercount SGLang/vLLM static-FP8 latency and energy.

The right rule is: subtract only for backends whose `fp8_static` GEMM collector measures overhead-inclusive rows. Today that appears to be TRT-LLM only.

Arsene12358 (approved):

collector/trtllm/collect_moe.py
```diff
@@ -487,9 +488,10 @@ def run_moe_torch(
     # the proper shuffle/permutation (torch.ops.trtllm.shuffle_matrix) is applied,
     # which the kernel expects for correct memory access patterns.
     if moe_type in ("w4a16_mxfp4", "w4a8_mxfp4_mxfp8"):
-        w1_bias = torch.randn((num_experts, inter_size), dtype=dtype, device=device)
-        w2_bias = torch.randn((num_experts, hidden_size), dtype=dtype, device=device)
-        w3_bias = torch.randn((num_experts, inter_size), dtype=dtype, device=device)
+        local_num_experts = num_experts // moe_ep_size
```
tianhaox, line 491: why using local experts, need @YijiaZhao to comment here whether the full expert should be used.

src/aiconfigurator/sdk/operations/gemm.py
tianhaox: base = query_gemm(m, n, k, fp8_static)
     = 实际查 gemm_perf.txt 里的 fp8 表, 这里实际上是dynamic quant

如果 low_precision_input=False:
  latency = base - compute_scale(m, k)

如果 low_precision_input=True:
  latency = base - compute_scale(m, k) - scale_matrix(m, k)

在一个fp8_static 的模型中，ffn2的输入是low_precision，其余地方的都是bf16的输入，所以上面的low_precision_input false 和 true都需要。我们可以选择采集bf16 input Static + fp8 input static 来cover两种精读，也可以选择复用fp8 dyanmic，但是额外采集计算和apply scale的两个算子，这样采集时间能少很多，这就是trtllm里面的设计。
对应vllm，sglang，如果要完整支持fp8 static，也需要沿用这种设计，否则ffn1的static那部分，多算了一个compute_scale，在blackwell上可能有5-10%的端到端影响。

src/aiconfigurator/cli/main.py
```diff
@@ -329,6 +330,13 @@ def _add_default_mode_arguments(parser):
         "When set, MoE models use EP-only parallelism with deepep_moe backend. "
         "Applies to both DeepSeek and Qwen3-235B on SGLang.",
     )
+    parser.add_argument(
```
tianhaox, line 332: should we add complexity to default mode with additional yaml config? why it's required

src/aiconfigurator/sdk/operations/dsa.py
tianhaox: what will happen if we still go with the interp for s-1? another option is that we may need to fix the mismatch instead of adding 70 lines to handle this special case

collector/trtllm/collect_moe.py
jasonqinzhou: I think local experts are intended here. `num_experts` remains the global expert count for routing and for the perf row, but this block is loading synthetic weights into the rank-local TRT-LLM MoE module after `mapping.moe_ep_size` is set. Under EP, each rank owns `num_experts // moe_ep_size` experts, so the manually loaded MXFP4 weight tensors should be local-sized. Using the full expert count would over-allocate and would not match the EP-sharded load path.

src/aiconfigurator/sdk/operations/dsa.py
jasonqinzhou: Agree. I moved this to the collector/data convention instead of keeping the SDK special case. Generation DSA rows now log `isl=1` and `step=past_kv` (`s-1`), so the loader’s `isl + step` key matches the total decode length the SDK queries. I also normalized the touched B200 TRT-LLM DSA generation parquet and removed the query-side fallback.

collector/trtllm/collect_moe.py
jasonqinzhou: Good question. My understanding is that `num_experts` should remain the full/global expert count for the collected row and for routing, but the tensors passed into this manual `load_weights()` path should be rank-local.

This block is specific to MXFP4 (`w4a16_mxfp4`, `w4a8_mxfp4_mxfp8`). For these types, the collector cannot just rely on the random weights created by `create_moe()`: it explicitly fabricates MXFP4 weights and calls `moe.load_weights()` so TRT-LLM applies the required MXFP4 shuffle/permutation before benchmarking. Since `mapping.moe_ep_size` has already been set, the MoE module is EP-sharded, so the load path should receive only the experts owned by this rank, i.e. `num_experts // moe_ep_size`.

The global `num_experts` is still used elsewhere:
- router logits are still shaped over the full expert set
- the perf row still logs `num_experts` as the model/global expert count
- `moe_ep_size` is logged separately, so DB lookup still represents the original global shape

So this is not changing the schema or pretending the model has fewer experts. It only changes the synthetic weight tensors passed into the rank-local MXFP4 load path.

This is mainly a correctness fix for the manual MXFP4 weight-loading path, with a side benefit of reducing memory/collection pressure. Other quant modes do not need this exact change because they do not manually construct and load full per-expert weight dictionaries in the same way; `create_moe()`/the backend initialization handles their EP-local tensors internally. GPT-OSS SwiGLU params already use `num_experts // moe_ep_size` as well.

Yijia, please confirm, but based on the EP-sharded `load_weights()` contract, local experts look like the intended shape here.

src/aiconfigurator/sdk/operations/dsa.py
jasonqinzhou: Small correction after checking the collector worker semantics and updating the branch: the module collectors' generation `seq_len` is already the past-KV/KV-cache length. So the collector convention should stay `isl=1, step=seq_len`, and the loader indexes that as `isl + step` = total decode length.

No DSA recollection is needed. I removed the SDK-side `s`/`s-1` fallback, added a loader test for the `isl + step` indexing, restored the B200 TRT-LLM DSA generation parquet to the original values, and left only a SGLang comment clarification on the collector side.

src/aiconfigurator/cli/main.py
jasonqinzhou: Good point, we don’t need to add default-mode YAML config complexity here. I removed the CLI/default YAML path and dropped the `main.py` diff from the PR; the large B200 MoE worker shape is now handled in the TaskConfig defaults instead, so support-matrix replay commands stay plain.

src/aiconfigurator/sdk/operations/gemm.py
jasonqinzhou: Agree, thanks for the detailed breakdown. I updated this to follow the TRT-LLM decomposition for SGLang/vLLM as well:

- Treat `fp8_static` GEMM rows as the dynamic FP8 base measurement.
- For BF16-input static FP8, subtract `compute_scale(m, k)`.
- For low-precision-input static FP8, subtract both `compute_scale(m, k)` and `scale_matrix(m, k)`, which covers the ffn2/down-proj case.
- Added SGLang/vLLM `compute_scale` collectors that also emit `scale_matrix` rows, and gate `fp8_static` support on having both overhead tables so we do not advertise support before the data exists.

So yes, we still need to collect the SGLang/vLLM `compute_scale` and `scale_matrix` datapoints. This keeps the collection/modeling approach consistent with TRT-LLM and avoids charging ffn1/static-BF16-input paths for an extra compute-scale cost.

YijiaZhao:
I think we should validate this change with before/after perf runs on existing MoE cases across TRT-LLM, SGLang, and vLLM.

For the local-expert related changes, the main question is whether the measured latency remains consistent with the previous collector behavior for cases that already existed before this PR. The PR changes how rank-local expert weights/modules are constructed in some paths, but the routing inputs and
logged perf keys still generally use global `num_tokens` and global `num_experts`.

src/aiconfigurator/sdk/operations/gemm.py
jasonqinzhou: Correction after re-checking the intended contract: TRT-LLM's SDK modeling here was already correct. The update is to apply the same modeling convention to SGLang/vLLM, not to fix TRT-LLM specifically.

The current branch now treats `fp8_static` as a derived GEMM mode:

- base GEMM lookup uses the dynamic `fp8` GEMM table
- BF16-input static FP8 = `fp8` base - `compute_scale(m, k)`
- low-precision-input static FP8 = `fp8` base - `compute_scale(m, k)` - `scale_matrix(m, k)`

So we do **not** need to collect separate SGLang/vLLM `fp8_static` GEMM datapoints. I removed `fp8_static` from their GEMM collector sweeps in `520e6d24`; the only additional datapoints needed for SGLang/vLLM static-FP8 support are the `compute_scale` and `scale_matrix` overhead rows. The support matrix now only advertises `fp8_static` when the dynamic `fp8` GEMM base and both overhead tables are present.

jasonqinzhou:
> I think we should validate this change with before/after perf runs on existing MoE cases across TRT-LLM, SGLang, and vLLM.
> 
> For the local-expert related changes, the main question is whether the measured latency remains consistent with the previous collector behavior for cases that already existed before this PR. The PR changes how rank-local expert weights/modules are constructed in some paths, but the routing inputs and logged perf keys still generally use global `num_tokens` and global `num_experts`.

We don't have perf run cases of gpt-oss btw. Can we do that analysis later?

collector/trtllm/collect_moe.py
jasonqinzhou: Comment on line R491,   do u know that without this local_num_experts change, will this break some support matrix?


Yes, but with a nuance.

It won’t break support-matrix generation directly, because support matrix reads existing parquet data and does not run collector/trtllm/collect_moe.py.

But it can break the data collection needed for those matrix rows. Without local_num_experts = num_experts // moe_ep_size, the TRT-LLM MXFP4 manual load_weights() path would pass global expert tensors into an EP-sharded MoE module. For moe_ep_size > 1, that can cause shape/key mismatch or much higher memory pressure. Then MXFP4 MoE rows may fail to collect or be missing.

So the affected surface is:

only TRT-LLM MoE collector
only w4a16_mxfp4 / w4a8_mxfp4_mxfp8
only EP cases where moe_ep_size > 1
relevant to current GPT-OSS TRT-LLM 1.3.0rc10 support rows on b200_sxm, b300_sxm, gb200, gb300
I would not say “reverting this line immediately changes the support matrix,” but I would say “without it, recollecting the MXFP4 MoE perf data that those support-matrix PASS rows depend on may fail or collect invalid shapes.”

PR #1121, opened 2026-05-17
refactor: task v2

tianhaox:
#### Overview:
Replaces the nested-Munch `TaskConfig` (and the `find_best_*` / `agg_pareto` / `disagg_pareto` search paths it dispatched through) with a flat `Task` dataclass + a new standalone `sweep` module. Output is bit-identical to V1, locked by integration parity tests. V1 is preserved under `sdk/task_v1.py` so the CLI / webapp / generator paths that still consume the old API keep working during the migration window.
#### Details:
- **Flat `Task` dataclass** (`sdk/task.py`): replaces nested `DefaultMunch` + `deep_merge` + override-spread with a single flat dataclass; `__post_init__` resolves model identity, backend version, quant modes, and search space to concrete values. Strict prefix discipline — in disagg mode, setting any top-level worker-spec field raises `ValueError`. Entry points: `.run()` (sweep), `.run_single_agg()` / `.run_single_disagg()` (subsume `cli_estimate`), `.to_dict()` / `.to_yaml()`. Disagg rate-match degradation factors and autoscale TTFT correction are now Task fields (defaults V1-identical, retunable per deployment).
- **New `sweep` module** (`sdk/sweep.py`): `sweep_agg` / `sweep_disagg` own the full search loop and replace the legacy `find_best_*` paths. Output schemas unchanged so downstream picking works as-is. Includes an ~80× speedup for tpot sweeps (build backend + model once per parallel, not per (parallel, tpot)).
- **Pluggable `Predictor` strategy** (`sdk/predictor.py` + `predict.py`): Protocol with `predict_agg_worker` / `predict_disagg_phase`; default `AnalyticPredictor` wraps `backend.run_agg` / `backend.run_static` bit-identically. Stub for future Mocker / dynamic-traffic predictors.
- **V1 preserved**: original `task.py` moved to `task_v1.py`; CLI / webapp / support-matrix continue to import from V1 unchanged.
- **New error**: `UnsupportedWideepConfigError(ValueError)` for WideEP-specific quant-mode mismatches.
- **Standalone runner**: `tools/run_sweep_new.py` — `--yaml example_new.yaml` → emits Pareto CSV.
- **Tests**: integration parity (old `TaskRunner` vs new `Task.run` byte-equal on agg/disagg matrix) + unit tests for sweep / predict / predictor / Task. 1071 unit + 5 parity + 6 picking-integration tests pass.
1. `src/aiconfigurator/cli/exps/example_new.yaml` — new flat YAML schema, anchors the field naming.
2. `src/aiconfigurator/sdk/task.py` — the new `Task` dataclass; focus on `__post_init__` resolution chain, `validate()`, and the `run()` / `run_single_*` entry points.
3. `src/aiconfigurator/sdk/sweep.py` — `sweep_agg` / `sweep_disagg`; read fresh, parity tests guard correctness.
4. `src/aiconfigurator/sdk/predict.py` + `predictor.py` — small; confirm the strategy injection layer.
5. `tests/integration/test_old_vs_new_parity.py` — the parity contract for the whole refactor.
`sdk/task_v1.py` is the original `task.py` renamed — only worth diffing if you want to confirm V1 is untouched.

#### Case 1: New Task Sweep Workflow

This branch moves sweep execution out of the legacy `TaskRunner -> pareto_analysis -> InferenceSession` stack and into explicit SDK-level sweep functions.

Old path:

```text
TaskConfig / TaskRunner
  -> pareto_analysis.agg_pareto() / disagg_pareto()
    -> InferenceSession / DisaggInferenceSession
      -> find_best_* helpers
        -> backend.run_agg() / backend.run_static()
```

New path:

```text
Task
  -> Task.run()
    -> sweep_agg() / sweep_disagg()
      -> predict_agg_worker() / predict_disagg_phase()
        -> Predictor
          -> backend.run_agg() / backend.run_static()
```

The code-level responsibilities are now cleaner:

- `Task` owns user-facing configuration, config resolution, database loading, and dispatch.
- `sweep_agg()` owns the aggregated search loop: parallelism, batch size, `ctx_tokens`, SLA filtering, and result sorting.
- `sweep_disagg()` owns the disaggregated search loop: prefill candidates, decode candidates, worker-count matching, rate matching, and SLA filtering.
- `predict.py` owns per-point prediction wrappers.
- `Predictor` is an injectable strategy interface. The default `AnalyticPredictor` still calls the existing backend methods, so the new path remains parity-compatible with current behavior.
- Sweep output keeps the existing `ColumnsAgg` / `ColumnsDisagg` DataFrame schemas, so downstream picking / Pareto visualization can continue to consume the same shape.

One important semantic cleanup is that Pareto helpers no longer drive the actual search. The new sweep functions return the SLA-feasible candidate set directly. Pareto frontier computation becomes a downstream view instead of being part of the core search path.

#### Case 2: New Task Build Flow From YAML / CLI

The new `Task` config format is intentionally flat. Every YAML key maps directly to one field on the `Task` dataclass.

The legacy nested YAML shape is not supported in the new path:

```yaml
# Legacy shape: no longer used by new Task
worker_config:
  ...
replica_config:
  ...
advanced_tuning_config:
  ...
```

The new shape is flat:

```yaml
agg_example:
  serving_mode: agg
  model_path: deepseek-ai/DeepSeek-V3
  system_name: h200_sxm
  total_gpus: 8
  isl: 4000
  osl: 1000
  ttft: 1000.0
  tpot: 40.0
  backend_name: trtllm
  backend_version: 1.2.0rc5
```

For aggregated serving, worker fields stay at the top level:

```yaml
agg_full:
  serving_mode: agg

  # Workload / SLA
  isl: 4000
  osl: 1000
  prefix: 0
  ttft: 1000.0
  tpot: 40.0
  total_gpus: 8
  database_mode: SILICON

  # Worker spec
  model_path: deepseek-ai/DeepSeek-V3
  system_name: h200_sxm
  backend_name: trtllm
  backend_version: 1.2.0rc5
  enable_wideep: false
  enable_chunked_prefill: false
  enable_eplb: false
  moe_backend: null

  # Quantization
  quant_preset: null
  gemm_quant_mode: fp8_block
  moe_quant_mode: fp8_block
  kvcache_quant_mode: bfloat16
  fmha_quant_mode: bfloat16
  comm_quant_mode: half

  # Search space
  agg_num_gpu_candidates: [4, 8]
  agg_tp_candidates: [1, 2, 4, 8]
  agg_pp_candidates: [1]
  agg_dp_candidates: [1, 2, 4, 8]
  agg_moe_tp_candidates: [1]
  agg_moe_ep_candidates: [1, 2, 4, 8]
```

For disaggregated serving, role-specific fields must be explicit. The new path does not silently reuse top-level `model_path`, `system_name`, `backend_name`, or quant fields for both roles.

```yaml
disagg_full:
  serving_mode: disagg

  # Shared workload / SLA
  isl: 4000
  osl: 1000
  ttft: 1000.0
  tpot: 40.0
  total_gpus: 32
  moe_backend: null

  # Prefill worker spec
  prefill_model_path: deepseek-ai/DeepSeek-V3
  prefill_system_name: h200_sxm
  prefill_backend_name: trtllm
  prefill_backend_version: 1.2.0rc5
  prefill_enable_wideep: false
  prefill_enable_chunked_prefill: false
  prefill_enable_eplb: false

  # Prefill quantization
  prefill_quant_preset: null
  prefill_gemm_quant_mode: fp8_block
  prefill_moe_quant_mode: fp8_block
  prefill_kvcache_quant_mode: bfloat16
  prefill_fmha_quant_mode: bfloat16
  prefill_comm_quant_mode: half

  # Prefill search space
  prefill_num_gpu_candidates: [4, 8]
  prefill_tp_candidates: [1, 2, 4, 8]
  prefill_pp_candidates: [1]
  prefill_dp_candidates: [1]
  prefill_moe_tp_candidates: [1]
  prefill_moe_ep_candidates: [1, 2, 4, 8]

  # Decode worker spec
  decode_model_path: deepseek-ai/DeepSeek-V3
  decode_system_name: h200_sxm
  decode_backend_name: trtllm
  decode_backend_version: 1.2.0rc5
  decode_enable_wideep: false
  decode_enable_eplb: false

  # Decode quantization
  decode_quant_preset: null
  decode_gemm_quant_mode: fp8_block
  decode_moe_quant_mode: fp8_block
  decode_kvcache_quant_mode: bfloat16
  decode_fmha_quant_mode: bfloat16
  decode_comm_quant_mode: half

  # Decode search space
  decode_num_gpu_candidates: [4, 8]
  decode_tp_candidates: [1, 2, 4, 8]
  decode_pp_candidates: [1]
  decode_dp_candidates: [1, 2, 4, 8]
  decode_moe_tp_candidates: [1]
  decode_moe_ep_candidates: [1, 2, 4, 8]

  # Disagg orchestration
  num_gpu_per_replica: [8, 16, 24, 32, 40, 48, 56, 64]
  max_gpu_per_replica: 128
  max_prefill_workers: 32
  max_decode_workers: 32
  prefill_max_batch_size: 1
  decode_max_batch_size: 512

  # Disagg calibration knobs
  prefill_latency_correction: 1.1
  decode_latency_correction: 1.08
  rate_match_prefill_degradation: 0.9
  rate_match_decode_degradation: 0.92
  autoscale_ttft_correction_factor: 1.8
```

YAML and CLI construction share the same field surface:

```python
task = Task.from_yaml(yaml_data, **cli_overrides)
```

`Task.from_yaml()` intentionally stays thin:

1. Accept keys that match `Task` fields.
2. Convert string quant modes into SDK enums.
3. Warn and ignore unknown keys.
4. Apply CLI overrides last, so CLI values win over YAML values.

Then `Task.__post_init__()` resolves the effective task:

1. Enforces disagg prefix discipline.
2. Resolves model identity, MoE status, and model family.
3. Resolves backend version if omitted.
4. Resolves quant modes with this priority:

```text
explicit field > quant_preset > HF config inference > fallback
```

5. Fills default search spaces if candidate lists are omitted.

After construction, callers use the same execution entry point:

```python
df = task.run()
```

`Task.run()` loads the required database internally and dispatches based on `serving_mode`:

```text
serving_mode=agg
  -> get_database(system_name, backend_name, backend_version)
  -> sweep_agg(...)

serving_mode=disagg
  -> get_database(prefill_system_name, prefill_backend_name, prefill_backend_version)
  -> get_database(decode_system_name, decode_backend_name, decode_backend_version)
  -> sweep_disagg(...)
```


tianhaox:
@coderabbitai resolve

tianhaox:
@coderabbitai full review

tianhaox:
@coderabbitai resolve

jasonqinzhou:
One migration concern: this changes the public `aiconfigurator.sdk.task` import surface. Internal callers were updated to `task_v1`, but external SDK users may still have code like:

```python
from aiconfigurator.sdk.task import TaskConfig, TaskRunner
```

After this PR, that import path now points at the new flat `Task` module and no longer exposes the legacy `TaskConfig` / `TaskRunner`. If `sdk.task` is considered public SDK surface, can we preserve a compatibility path, e.g. re-export the legacy symbols from `task.py` during the migration window, or add a clear deprecation/migration note?

The staged V1/V2 split is good; I just want to avoid turning an internal refactor into an accidental breaking SDK change.

jasonqinzhou (approved):

src/aiconfigurator/sdk/predictor.py
```diff
@@ -0,0 +1,149 @@
...
+        backend: BaseBackend,
+        database: PerfDatabase,
+        runtime_config: config.RuntimeConfig,
+        ctx_tokens: int,
+        **backend_kwargs: Any,
+    ) -> InferenceSummary:
+        """Predict perf for an aggregated IFB worker at a single (parallel, batch, ctx_tokens) point."""
+        ...
+
+    def predict_disagg_phase(
```
simone-chen: Would naming it `predict_disagg_worker` easier to read and pair with `predict_agg_worker`?

src/aiconfigurator/sdk/predictor.py
```diff
@@ -0,0 +1,149 @@
...
+
+Future implementations may swap in different prediction strategies:
+
+- ``DynamicPredictor`` -- models dynamic-traffic effects (queueing under
+  concurrency) beyond the steady-state assumption baked into the analytic
+  predictor.
+- ``MockerPredictor`` -- delegates to Dynamo Mocker for request-level
+  discrete-event simulation; useful when the user has a real Mooncake-
+  style trace and wants per-request metrics rather than aggregate
+  latency / throughput numbers.
```
simone-chen: I'm confused here. It feels like the `MockerPredictor` needs to import Dynamo Mocker's API to implement the protocol functions, but it is the reversed case as mocker will import AIC and invoke its `run_static` and `run_agg`.

tests/integration/test_task_v1_v2_parity.py
simone-chen, line 1: I think naming it `test_task_v1_v2_parity.py` might be easier to understand.

src/aiconfigurator/sdk/predict.py
```diff
@@ -0,0 +1,134 @@
...
+def predict_disagg_phase(
+    *,
+    model: BaseModel,
+    backend: BaseBackend,
+    database: PerfDatabase,
+    runtime_config: config.RuntimeConfig,
+    role: DisaggRole,
+    latency_correction: float = 1.0,
+    stride: int = 32,
+    predictor: Predictor | None = None,
```
simone-chen, line 54: We can just do `predictor: Predictor | None = DEFAULT_PREDICTOR` in replace of the `(predictor or DEFAULT_PREDICTOR).predict_disagg_phase(` below

src/aiconfigurator/sdk/predictor.py
```diff
@@ -0,0 +1,149 @@
...
+        role: DisaggRole,
+        latency_correction: float = 1.0,
+        stride: int = 32,
+    ) -> InferenceSummary:
+        """Predict perf for one phase of a disagg worker (prefill-only or decode-only)."""
+        ...
+
+
+class AnalyticPredictor:
+    """Default predictor: steady-state analytic predictions (zero-queue).
```
simone-chen, line 87: It doesn't directly relate to the purpose of the PR, but should `AnalyticPredictor` own a function validating if the engine will queue under the given bs?

src/aiconfigurator/sdk/predictor.py
tianhaox: Done — renamed `predict_disagg_phase` → `predict_disagg_worker` across the Predictor protocol, `AnalyticPredictor`, `predict.py`, `sweep.py`, `task_v2.py` and the tests, to pair with `predict_agg_worker`. Thanks for the suggestion!

src/aiconfigurator/sdk/predictor.py
tianhaox: Good catch — fixed the docstring. It now notes the integration is inverted: Dynamo Mocker imports AIC and drives `run_static`/`run_agg`, rather than AIC delegating to Mocker.

tests/integration/test_task_v1_v2_parity.py
tianhaox: Done — renamed to `test_task_v1_v2_parity.py` and updated the references in `test_predict.py` / `test_sweep.py`.

src/aiconfigurator/sdk/predictor.py
tianhaox: sounds good, we can add one and modify the docstring

PR #1126, opened 2026-05-18
feat: Add support for Google Gemma4 MoE

Arsene12358:
## Summary

Adds first-class aiconfigurator support for `google/gemma-4-26B-A4B` — a 25.2B-total / 3.8B-active MoE model with hybrid SWA/global attention and a shared dense MLP running in parallel with a routed top-8-of-128 MoE on every layer. The model is selectable in the webapp dropdown, its MoE controls render correctly, and the CLI's `default` and `generate` subcommands work end-to-end.

Tracking issue: [AIC-1098](https://linear.app/nvidia/issue/AIC-1098/aiconfigurator-add-support-for-googlegemma-4-26b-a4b-moe) · Project: [Model: Google Gemma4](https://linear.app/nvidia/project/model-google-gemma4-fdab453ba5c3) · Design doc: [linked in AIC-1098](https://linear.app/nvidia/document/design-aiconfigurator-support-for-gemma-4-26b-a4b-moe-0aaa442db701) · Model card: https://huggingface.co/google/gemma-4-26B-A4B

This PR is the **Phase 1 + Phase 2** ship: SDK model class + webapp wiring. The model is fully usable from this PR onwards via `--database-mode HYBRID` (SOL + empirical estimates). **Perf-DB collection (Phase 3) and validation benchmark (Phase 4) will follow in separate PRs** — see below.

## What's in this PR

**SDK (`d1d0a79f`, `f686c272`)**

- New `Gemma4MoEConfig` dataclass in `sdk/common.py` capturing per-layer-type attention dims (SWA: 8 KV heads × head_dim 256; global: 2 KV heads × head_dim 512), `sliding_window_size`, and the global `attention_k_eq_v` flag.
- `"GEMMA4MOE"` registered in `ModelFamily`, `ARCHITECTURE_TO_MODEL_FAMILY` (mapping `Gemma4ForConditionalGeneration`), `MULTIMODAL_TEXT_CONFIG_KEY` (flatten nested `text_config`), and `_MOE_MODEL_FAMILIES`.
- New `sdk/models/gemma4_moe.py` with `Gemma4MoEModel` (sibling of `HybridMoEModel`, not an extension — design doc §4 documents the four reasons). Two layer-type recipes (SWA + global), each emitting attention norm → QKV GEMM → ContextAttention/GenerationAttention → projection GEMM → FFN norm → shared dense MLP (gate-up + act + down) → router GEMM (when `num_experts ≥ 128`) → MoE dispatch + experts + dispatch. Overrides `get_kvcache_elements_per_token` and `get_kvcache_bytes_per_sequence` (DeepSeekV4/V32 precedent) so the SWA window cap is applied — this is the architectural payoff that keeps 256K context in ~5.2 GiB at TP=1, bf16.
- `sdk/utils.py`: new `Gemma4ForConditionalGeneration` parse branch in `_parse_hf_config_json`. Generic widening: `topk` parse falls back to `top_k_experts` since Gemma 4 uses that key instead of HF's conventional `num_experts_per_tok`.
- 11 new unit tests in `tests/unit/sdk/test_utils.py` covering parse-time field mapping (including the 5:1 SWA:global layer pattern), per-layer-type QKV-out widths with and without `attention_k_eq_v`, KV-cache window cap at multiple seq_len, TP sharding, and validation errors.

**Webapp (`850d8618`)**

- Add `"google/gemma-4-26B-A4B"` to `DefaultHFModels` so the webapp dropdown lists it (`get_default_models()` is the union of support-matrix CSVs and this set).
- Cache the HF config at `model_configs/google--gemma-4-26B-A4B_config.json` so `get_model_family()` resolves offline, matching the convention every other Default model follows.
- New `tests/unit/webapp/test_event_fn.py` regression test asserting `update_model_related_components` returns `visible=True` for all 7 controls when given a GEMMA4MOE model. Uses `pytest.importorskip("gradio")` so it runs in webapp-capable envs and skips cleanly in minimal lint envs.

**Style (`8cd2e016`)**

- Ruff lint pass (RUF002 ambiguous unicode in docstrings, RUF022 isort `__all__`, RUF100 unused noqa) and `ruff format` to match repo conventions.

**Merge with main (`d83608de`)**

- Pulled in #1108 (kimik2.5 default quant config in webapp). That PR refactored the family-keyed MoE-control gate in `update_model_related_components` from a hand-maintained tuple to `check_is_moe(model_path)`. Since Phase 1 already added `GEMMA4MOE` to `_MOE_MODEL_FAMILIES`, the family is now auto-detected, and our earlier explicit one-token tuple addition became redundant. Conflict resolved by taking main's version. **Net effect on `event_fn.py` in this branch vs current main: zero change.**

## Verification

End-to-end CLI smoke (no fixes needed across three variants):

| Invocation | Result |
| --- | --- |
| `cli default --model google/gemma-4-26B-A4B --total-gpus 8 --system b200_sxm --backend trtllm --database-mode HYBRID` | Full Pareto, 23 s, best agg 9332 tokens/s/gpu, both agg and disagg recommendations |
| `cli default ... --database-mode SILICON` (default) | Clean failure with explicit "Consider using HYBRID mode. Missing silicon data" — designed UX, no tracebacks |
| `cli default ... --system h200_sxm --backend auto --database-mode HYBRID` | Full Pareto across all three backends, 58 s |
| `cli generate --model google/gemma-4-26B-A4B --total-gpus 8 --system b200_sxm --backend trtllm` | 7 deployment artifacts emitted (agg_config.yaml, bench_run.sh, k8s_deploy.yaml, run_0.sh, k8s_bench.yaml, generator_config.yaml, sflow.yaml). agg_config has correct MoE fields; run_0.sh uses `--served-model-name "$SERVED_MODEL_NAME"`. |

Test sweep:
- `tests/unit/sdk/test_utils.py`: 56/56 pass (was 45; +11 new Gemma 4 cases).
- `tests/unit/sdk/`: 655/655 pass (1 unrelated torch-dependency skip). No regressions in MiMo-V2-Flash / Llama 4 / Qwen3.5 / DeepSeek paths.
- `tests/unit/webapp/test_event_fn.py`: skipped in the local gradio-less env, runs in CI.
- Mocked-gradio sanity check of `EventFn.update_model_related_components("google/gemma-4-26B-A4B")`: returns 7 `gr.update` objects with `visible: True`.
- `ruff check`, `ruff format --check`: clean.
- **Upstream CI on this PR: 10/10 checks ✅ SUCCESS** (AIC Accuracy Regression, Cargo Deny, copyright-checks, Ruff Lint, ESLint, Build and Test unit, Build and Test e2e, CodeRabbit, DCO, PR title validation).

## Follow-up PRs (out of scope here)

This PR ships Phase 1+2 only — the model is usable today via `--database-mode HYBRID`. The rest of the project lands in separate PRs:

- **Phase 3 — Perf-DB collection** (next, separate PR). Fire `aiconfigurator-ci` over 6 systems (`h100_sxm`, `h200_sxm`, `b200_sxm`, `b300_sxm`, `gb200`, `gb300`) × 3 backends × 2 backend-versions. Dry-run first. Will commit the resulting `*_perf.txt` files under `src/aiconfigurator/systems/data/<system>/<backend>/<version>/` per the standard layout.
- **Phase 4 — Validation benchmark** (separate PR). One config on `b200_sxm × trtllm` at TP=4, EP=8. Acceptance target ±15% on throughput + TTFT.
- **Phase 5 — Add to `test_aiconfigurator_moe_models` parametrization + release notes** (separate PR after Phase 3+4 land).

## What's not in this PR

- **Perf-DB silicon data** — handled in Phase 3 follow-up. `--database-mode HYBRID` works today (SOL + empirical estimates).
- **Measured-vs-predicted validation** — handled in Phase 4 follow-up.
- **Vision encoder / audio tower**: text-only LLM scope. Multimodal estimation is a separate epic.
- **Adjacent KIMIK25 fix in the (now-obsolete) family tuple** ([DYN-3044](https://linear.app/nvidia/issue/DYN-3044)): structurally fixed upstream by #1108's `check_is_moe()` refactor.

## Test plan

- [x] `tests/unit/sdk/test_utils.py` 56 passing locally
- [x] `tests/unit/sdk` 655 passing locally (1 unrelated torch-skip)
- [x] `cli default` and `cli generate` smoke-tested on b200_sxm × trtllm
- [x] `cli default --backend auto` smoke-tested on h200_sxm
- [x] `ruff check`, `ruff format --check` clean
- [x] All 10 upstream CI checks ✅ green
- [x] Mocked-gradio run of `update_model_related_components("google/gemma-4-26B-A4B")` returns 7 visible updates


src/aiconfigurator/sdk/utils.py
```diff
@@ -540,7 +540,7 @@ def _parse_hf_config_json(config: dict) -> dict:
     d = config.get("head_dim") or config.get("attention_head_dim") or (hidden_size // n if n > 0 else 0)
 
     # MoE parameters
-    topk = config.get("num_experts_per_tok", 0)
+    topk = config.get("num_experts_per_tok") or config.get("top_k_experts") or 0
```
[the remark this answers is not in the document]
Arsene12358: Applied per the suggested fix in commit 9553ecd1 — switched to explicit `None` checks so `num_experts_per_tok=0` is preserved instead of falling through to `top_k_experts`. Comment kept inline to document the rationale.

Existing `test_parse_llama_config` already asserts `topk == 0` for the absent-key path, so the change is regression-tested. No real-world config has `num_experts_per_tok=0` combined with a nonzero `top_k_experts`, so the fix is defensive rather than load-bearing — but the reviewer's reading is correct and the original code's semantic is now preserved.

src/aiconfigurator/sdk/models/gemma4_moe.py
```diff
@@ -0,0 +1,389 @@
...
+                moe_tp,
+                moe_ep,
+                moe_q,
+                wl_dist,
+                attn_dp,
+            ),
+            ops.MoEDispatch(
+                f"{prefix}_moe_post_dispatch", count, h, self._topk, self._num_experts, moe_tp, moe_ep, attn_dp, False
+            ),
+        ]
```
jasonqinzhou, line 205: `MoEDispatch` also needs the MoE quant mode. Right now `_moe_ops()` passes `moe_q` into `ops.MoE`, but both dispatch ops are constructed without `quant_mode`, so `MoEDispatch.query()` sees `None`. On TRT-LLM SM100/NVL72 all-to-all paths that raises `ValueError("MoEDispatch requires quant_mode...")`, and on other paths it can size communication as an unknown/default format.

This keeps the change local and preserves the existing operation structure:

```suggestion
        return router_ops + [
            ops.MoEDispatch(
                f"{prefix}_moe_pre_dispatch",
                count,
                h,
                self._topk,
                self._num_experts,
                moe_tp,
                moe_ep,
                attn_dp,
                True,
                quant_mode=moe_q,
            ),
            ops.MoE(
                f"{prefix}_moe",
                count,
                h,
                self._moe_inter_size,
                self._topk,
                self._num_experts,
                moe_tp,
                moe_ep,
                moe_q,
                wl_dist,
                attn_dp,
            ),
            ops.MoEDispatch(
                f"{prefix}_moe_post_dispatch",
                count,
                h,
                self._topk,
                self._num_experts,
                moe_tp,
                moe_ep,
                attn_dp,
                False,
                quant_mode=moe_q,
            ),
        ]
```

jasonqinzhou (commented):
Two integration comments for Gemma4MoE support.

src/aiconfigurator/sdk/models/helpers.py
```diff
@@ -23,7 +23,7 @@
 
 logger = logging.getLogger(__name__)
 
-_MOE_MODEL_FAMILIES = {"MOE", "DEEPSEEK", "DEEPSEEKV32", "DEEPSEEKV4", "KIMIK25", "HYBRIDMOE"}
+_MOE_MODEL_FAMILIES = {"MOE", "DEEPSEEK", "DEEPSEEKV32", "DEEPSEEKV4", "KIMIK25", "HYBRIDMOE", "GEMMA4MOE"}
```
jasonqinzhou, line 26: Since this registers `GEMMA4MOE` as an MoE family, the backend memory estimators also need to treat it as routed MoE. Today both TRT-LLM and SGLang only special-case `"MOE"`/DeepSeek-like families, so Gemma4MoE falls through to the dense fallback and can undercount activation/workspace memory for fit checks.

The backend files are outside this PR diff, so GitHub cannot apply this as a suggestion block directly, but this patch should be copy/apply-ready:

```diff
diff --git a/src/aiconfigurator/sdk/backends/trtllm_backend.py b/src/aiconfigurator/sdk/backends/trtllm_backend.py
--- a/src/aiconfigurator/sdk/backends/trtllm_backend.py
+++ b/src/aiconfigurator/sdk/backends/trtllm_backend.py
@@
-        h = model._num_heads * model._head_size
+        h = model._num_heads * model._head_size
+        moe_workspace_h = getattr(model, "_hidden_size", h)
@@
-        elif model.model_family == "MOE":
+        elif model.model_family in ("MOE", "GEMMA4MOE"):
             c_dict = {1: 22, 2: 13, 4: 10, 8: 10}
             activations = 2 * num_tokens * h * c_dict[min(model.config.tp_size, 8)]
+            if model.model_family == "GEMMA4MOE":
+                activations += (
+                    num_tokens
+                    * moe_workspace_h
+                    * model.config.attention_dp_size
+                    * model._num_experts
+                    * model._topk
+                    / model.config.moe_ep_size
+                    / 128
+                    * 4
+                )
             activations = max(activations, 70 * 1024 * 1024)  # minimum act
diff --git a/src/aiconfigurator/sdk/backends/sglang_backend.py b/src/aiconfigurator/sdk/backends/sglang_backend.py
--- a/src/aiconfigurator/sdk/backends/sglang_backend.py
+++ b/src/aiconfigurator/sdk/backends/sglang_backend.py
@@
-        elif model.model_family == "MOE":
+        elif model.model_family in ("MOE", "GEMMA4MOE"):
             c_dict = {1: 28, 2: 17, 4: 13, 8: 13}
             activations = 2 * num_tokens * h * c_dict[min(model.config.tp_size, 8)]
+            if model.model_family == "GEMMA4MOE":
+                activations += (
+                    num_tokens
+                    * moe_workspace_h
+                    * model.config.attention_dp_size
+                    * model._num_experts
+                    * model._topk
+                    / model.config.moe_ep_size
+                    / 128
+                    * 4
+                )
             activations = max(activations, 90 * 1024 * 1024)  # Higher minimum for SGLANG
```

jasonqinzhou (approved):

src/aiconfigurator/sdk/models/helpers.py
Arsene12358: Applied your patch verbatim in commit 402ca5a2 — `GEMMA4MOE` now joins the `MOE` branch in both `trtllm_backend.py` and `sglang_backend.py` and conditionally picks up the MoE dispatch workspace term, mirroring the DEEPSEEK family's accounting. TRT-LLM backend gained the `moe_workspace_h` fallback (SGLang already had it).

Good catch — without this, the dense-fallback path was undercounting activation memory by ~2.2x (c_dict 10 vs 22 at TP=1) and skipping the workspace term entirely, so memory-fit checks would have green-lit configs that OOM at deploy time. CLI smoke unchanged (best agg 9332.53 tokens/s/gpu on b200_sxm × trtllm × HYBRID), which confirms the new accounting doesn't disqualify top configs — just adds correct headroom on tighter ones.

For the record: `HYBRIDMOE` (Llama 4 Scout/Maverick, MiMo-V2-Flash) has the same falls-through-to-dense bug today. Pre-existing and outside this PR's scope, but worth filing separately.

PR #1129, opened 2026-05-19
feat(afd): add AFD estimate mode

gehujun:
#### Overview:

Adds **Attention-FFN Disaggregated (AFD)** estimate mode to aiconfigurator. AFD is orthogonal to P/D-disaggregation: it splits each layer's compute into an A-Worker pool (embedding / attention / boundary ops) and an F-Worker pool (FFN / MoE) running on separate GPU groups with explicit cross-pool communication, and models the resulting ping-pong pipeline. This PR delivers the full estimate path — new ops, partitioning module, session/CLI/task wiring, memory accounting, and unit tests — and is reachable via `aiconfigurator cli estimate --estimate-mode afd ...`.

#### Details:

The change is structured as 5 commits, each independently reviewable:

1. **`feat(afd): add AFDTransfer op for AFD cross-pool / intra-pool comm modeling`** (`1a07f051`)
   New `AFDTransfer` operation in `sdk/operations.py`. Returns a per-layer latency dict keyed by `t_a2f` / `t_f2a` (one-direction cross-pool) and `t_a` / `t_f` (intra-pool AllGather / ReduceScatter / combine added to each pool's compute). Token-volume billed, so prefill (≈ `isl` tokens per request) and decode (1 token per step) are both modeled correctly. Picks the right BW tier based on whether the transfer is intra-node or cross-rack (consults `system_spec`).

2. **`feat(afd): add op partitioning module to split a model into A/F pools`** (`f187ae33`)
   New module `sdk/afd_partition.py` exposing `AFDOpsPartition`, `classify_op_side`, and `build_afd_ops_partition`. Classification is name-keyed (allow-list per pool + skip-set for already-modeled comm ops). Boundary ops (`add_norm_2`, `logits_gemm`) are routed via a single `boundary_on_attn` flag so users can flip them between pools for sensitivity studies without touching the model definition. `OverlapOp` groups are atomic — `AFDPartitionError` is raised if an overlap spans the A/F boundary instead of silently stretching it. Also adds an `AFD Operation Partitioning Compatibility` section to `docs/add_a_new_model.md` documenting the maintenance contract.

3. **`feat(afd): wire AFD estimate end-to-end (CLI, task, session, config)`** (`de1e0370`)
   Wires AFD into the estimate pipeline:
   - `sdk/config.py`: new `AFDConfig` with two enforced invariants — `gpus_per_node` is **derived from `system_spec['node']['num_gpus_per_node']` as the single source of truth** (silent mismatch with the perf DB topology used to be possible); `tp_f` is **derived from `n_f_workers` under the Phase-1 `F-DP=1` assumption** (was redundant and could contradict `n_f_nodes × gpus_per_node`). Explicit `ValueError` on either invariant mismatch.
   - `sdk/inference_session.py`: new `AFDInferenceSession` with `run_afd(phase=prefill|decode|both)`, per-phase `_simulate_phase`, and `_pipeline_tcycle` for the ping-pong overlap model. Decode is **integrated along `osl` with `stride=32` sampling** (`s_i = isl + i + 1`) and the pipeline cycle is evaluated **per step before summing** — `sum_i max(t_a_i, t_f_i, t_c)`, not `max(sum_i t_a_i, sum_i t_f_i, N·t_c)` — to keep the bottleneck judgement correct in `osl ~ isl` and `osl ≫ isl` regimes. Single-point would under-count A-side latency by ~33% for typical chat and several-fold for long generation, silently flipping AFD sizing.
   - `cli/main.py` + `cli/api.py`: new `--estimate-mode afd` plus AFD-specific args (`--n-a-nodes`, `--n-f-nodes`, `--a-tp-size`, `--a-batch-size`, `--f-moe-ep-size`, `--num-microbatches`, `--pipeline-model`, `--afd-phase`, `--afd-combined-with-pd`, `--boundary-on-ffn`, ...). Removes redundant args that conflicted with the new invariants (`--gpus-per-node`, `--f-tp-size`).
   - `sdk/task.py`: AFD path for the YAML task layer, with `_finalize_afd` cross-checking any explicit `gpus_per_node` override against the system spec.

4. **`test(afd): add unit tests for AFDTransfer and afd_partition`** (`9ae45df6`)
   23 unit tests covering: AFDTransfer message-size math, intra- vs cross-rack BW tier selection, partition allow/skip-lists, boundary op routing, OverlapOp atomicity invariant, and `AFDPartitionError` paths.

5. **`feat(afd): align memory checks with backend summaries`** (`877c98af`)
   Replaces the hand-rolled AFD pool memory estimate with the existing backend memory model so A-Worker / F-Worker OOM checks use the same weights / activation / KV cache / NCCL / other-memory accounting as the agg/disagg summaries. CLI and task entry points now thread `free_gpu_memory_fraction` and `max_seq_len` into AFD. TRT-LLM provides its backend-specific KV-cache defaults; other backends inherit `BaseBackend` defaults.

The fastest way to internalize the design is to read in this order:

- **`src/aiconfigurator/sdk/config.py`** — `AFDConfig` and its two derivation invariants. Sets the mental model for what is user-input vs system-derived.
- **`src/aiconfigurator/sdk/afd_partition.py`** — small (~300 lines) and self-contained. The classification rules are the contract every future model author needs to satisfy.
- **`src/aiconfigurator/sdk/operations.py`** — `AFDTransfer` class. Look at the docstring + `query()` return shape; the per-layer breakdown dict is the interface used by `_simulate_phase`.
- **`src/aiconfigurator/sdk/inference_session.py`** — `AFDInferenceSession._simulate_phase`, `_integrate_decode_phase`, and `_pipeline_tcycle`. This is where the per-step ping-pong overlap math lives.
- **`src/aiconfigurator/cli/api.py`** — `_run_afd_estimate` shows the full happy path (config → session → summary → CLI render).
- **`docs/add_a_new_model.md`** — the new `AFD Operation Partitioning Compatibility` section is the maintenance contract for adding new model families.
- **`tests/unit/sdk/test_afd_partition.py` / `test_afd_transfer.py`** — runnable spec of the partitioning contract and the comm op.

A working invocation that exercises the full path:

```bash
aiconfigurator cli estimate \
  --model deepseek-ai/DeepSeek-V3 \
  --system h200_sxm --backend trtllm \
  --estimate-mode afd \
  --isl 4096 --osl 1024 \
  --n-a-nodes 1 --n-f-nodes 1 \
  --a-tp-size 4 --a-batch-size 64 \
  --f-moe-ep-size 8 \
  --afd-phase decode

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added Attention‑FFN Disaggregated (AFD) serving: CLI estimate mode, enriched CLI output, A/F estimation path, and end‑to‑end A/F inference simulation with per‑phase latency, throughput, memory, and per‑op breakdowns.

* **Documentation**
  * New AFD operation partitioning guidance describing op classification, overlap rules, and handling for unsupported ops.

* **SDK / API**
  * New AFD config, partitioning, inference session, A/F communication operations, and AFD-aware summary columns; backend memory/kv-cache helpers added.

* **Tests**
  * Comprehensive unit tests for partitioning, transfers, CLI estimation, task config, and AFD end‑to‑end behavior.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

File: src/aiconfigurator/sdk/inference_session.py [hunk 1]
```diff
@@ -828,3 +829,867 @@ def _find_best_result_under_constraints(
...
+        is_context: bool,
+    ):
+        """Sum the query() latencies for a list of ops, returning (total, per-op dict).
+
+        For prefill (``is_context=True``) we pass ``seq_imbalance_correction_scale``;
+        for decode we pass ``gen_seq_imbalance_correction_scale``.  Tokens
+        processed per call = ``batch_size`` for decode, ``batch_size*seq_len``
+        for prefill (one token per sequence vs. full sequence).
+        """
+        x = batch_size * seq_len if is_context else batch_size
```
liz-badada, line 923: For prefill, use `batch_size * (isl - runtime_config.prefix)` instead for more reasonable TTFT?

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -828,3 +829,867 @@ def _find_best_result_under_constraints(
...
+        b_total = cfg.n_a_workers * cfg.a_batch_size
+
+        # AFD cross-pool comm (AFDTransfer, F-node AllGather/ReduceScatter,
+        # A-side combine) bills by *token* volume per step, not request
+        # count.  In prefill each request contributes ``isl`` tokens per
+        # layer; in decode it contributes 1 token per step.  Pre-compute
+        # the per-step token totals to feed into the AFD ops so the
+        # per-link byte volume is correctly scaled by sequence length
+        # (without this, prefill comm latency is under-estimated by ``isl``).
+        tokens_per_req = isl if phase == "prefill" else 1
```
liz-badada: For prefill, use (isl - runtime_config.prefix) instead?

src/aiconfigurator/cli/api.py
```diff
@@ -1346,6 +1402,155 @@ def _run_disagg_estimate(
...
+        f_moe_ep_size=f_moe_ep_size,
+        a_batch_size=a_batch_size,
+        num_microbatches=num_microbatches,
+        pipeline_model=pipeline_model,
+        comm_overhead_factor=comm_overhead_factor,
+        phase=afd_phase,
+        combined_with_pd=bool(afd_combined_with_pd),
+        boundary_on_attn=bool(afd_boundary_on_attn),
+    )
+    runtime_config = RuntimeConfig(isl=isl, osl=osl, batch_size=afd_config.n_a_workers * a_batch_size)
```
liz-badada: Add prefix info?

src/aiconfigurator/cli/api.py
```diff
@@ -1346,6 +1402,155 @@ def _run_disagg_estimate(
...
+        gemm_quant_mode, kvcache_quant_mode, fmha_quant_mode,
+        moe_quant_mode, comm_quant_mode,
+    )
+    f_model_config = _build_model_config(
+        f_tp_size, 1, 1,
+        f_moe_tp_size, f_moe_ep_size,
+        gemm_quant_mode, kvcache_quant_mode, fmha_quant_mode,
+        moe_quant_mode, comm_quant_mode,
+    )
+
```
liz-badada, line 1698: Will consider `_apply_nextn` in this version?

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -828,3 +829,867 @@ def _find_best_result_under_constraints(
...
+        # ``scale_factor=num_layers`` (per-layer ops such as qkv_gemm) or
+        # ``scale_factor=1`` (once-per-step ops such as embedding /
+        # logits_gemm).  ``_sum_latency`` therefore returns the *full
+        # per-step* contribution of each pool, not a single layer.  The
+        # AFD pipeline model is layer-granular, so amortize across layers
+        # to recover the per-layer cycle ingredients before pipelining.
+        # Once-per-step ops (embedding/logits_gemm) get folded into the
+        # per-layer average; their absolute cost is small relative to
+        # ``num_layers`` per-layer compute and AFD does not currently
+        # model them as separate stages.
```
liz-badada, line 1413: So this will be a future todo or not?

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -828,3 +829,867 @@ def _find_best_result_under_constraints(
...
+            free_gpu_memory_fraction=free_gpu_memory_fraction,
+            kv_cache_reserved_fraction=reserved_fraction,
+            kv_cache_tolerance=tolerance,
+        )
+        return summary
+
+    # Stride for sampling KV-cache length ``s`` along the decode trace.
+    # Mirrors ``base_backend._run_generation_phase`` so the AFD path
+    # uses the same numerical integration grid as agg/disagg.
+    _AFD_DECODE_STRIDE = 32
```
liz-badada, line 1197: Why fixed to magic number here? Is it based on any testing or empirical validation?

src/aiconfigurator/sdk/afd_partition.py
```diff
@@ -0,0 +1,311 @@
...
+            return _unknown_or_default(op, allow_unknown_ops=allow_unknown_ops, unknown_side=unknown_side)
+
+        raise AFDPartitionError(f"OverlapOp '{name}' spans A/F boundaries and cannot be kept atomic.")
+
+    if _is_skipped_model_internal_op(op, name):
+        return "skip"
+    _raise_if_unclassifiable_layer_family(op)
+    if _is_boundary_op(name):
+        return "boundary"
+    if _is_ffn_side_op(name):
```
yunyang1999: But` _classify_op` and `_classify_inner_overlap_op` are attn → ffn. Substring-based markers can overlap (proj_gemm is the canonical example), so the order matters. Even if no current model triggers it, an inconsistency like this is a future-bug factory.

Fix: unify all three on attn → ffn, or factor out a _classify_by_name(name) helper used by all three.

src/aiconfigurator/sdk/afd_partition.py
```diff
@@ -0,0 +1,311 @@
...
+            "add_norm_2",
+            "_ffn_norm",
+            "_moe_norm",
+            "_mlp_norm",
+            "_dense_ffn_norm",
+        )
+    ) or name.endswith("_combine")
+
+
+def _is_attention_side_op(name: str) -> bool:
```
yunyang1999, line 273: `_is_attention_side_op` matches the substring `proj_gemm` (intended for the attention output projection), but FFN ops in some models are also named `down_proj_gemm` / `up_proj_gemm` / `gate_proj_gemm`. Since `_is_attention_side_op` runs before `_is_ffn_side_op`, those would all be misclassified as attention-side.

Please verify that every model under models/ actually avoids proj_gemm in FFN op names (Qwen3.5 / GPT-OSS MoE in particular). If any model does use it, the partition silently misroutes weights and per-pool latency without raising.

Or we can just tighten the marker to `attn_proj_gemm` / `o_proj_gemm`, or have `_is_attention_side_op` short-circuit to False when the name also contains down_/up_/gate_proj_gemm.

src/aiconfigurator/sdk/afd_partition.py
```diff
@@ -0,0 +1,311 @@
...
+    if _is_boundary_op(name):
+        return "boundary"
+    if _is_ffn_side_op(name):
+        return "ffn"
+    if _is_attention_side_op(name):
+        return "attn"
+    return _unknown_or_default(op, allow_unknown_ops=allow_unknown_ops, unknown_side=unknown_side)
+
+
+def _classify_inner_overlap_op(
```
yunyang1999, line 162: `_classify_op `/ `_classify_overlap_op` (no-inner-ops branch) / `_classify_inner_overlap_op `are ~90% the same logic. Can refactor:
```
def _classify_by_markers(op, name, *, allow_unknown_ops, unknown_side):
    if _is_skipped_model_internal_op(op, name): return "skip"
    _raise_if_unclassifiable_layer_family(op)
    if _is_boundary_op(name): return "boundary"
    if _is_attention_side_op(name): return "attn"
    if _is_ffn_side_op(name): return "ffn"
    return _unknown_or_default(op, ...)
```

All three callsites use this single helper. That alone would avoid attn ffn order issue.

src/aiconfigurator/sdk/operations.py
```diff
@@ -2108,3 +2108,275 @@ def get_weights(self, **kwargs):
...
+
+    Consolidates four conceptual ops behind one ``query()``:
+
+      * A→F cross-pool dispatch (one-direction per-layer DMA)
+      * F→A cross-pool return (symmetric: same per-link payload)
+      * F-node intra-node AllGather + ReduceScatter (only under
+        ``rank_mapping == "one_to_one"`` and ``tp_f > 1``)
+      * A-side cross-EP combine reduce (only when ``f_moe_ep_size > 1``)
+
+    ``query()`` returns a dict ::
```
yunyang1999: `AFDTransfer.query() `returns a dict, breaking the `Operation.query()` → `PerformanceResult` contract

Consequences:

- `AFDInferenceSession._simulate_phase` has to special-case unpack the four sub-dicts; you cannot drop AFDTransfer into the generic per-op latency accumulator.
- AFDTransfer cannot be appended to `partition.attn_ops` / `ffn_ops` and processed uniformly with other ops.
- Phase 2 work (shared-expert replication traffic, MoE EP all-to-all sharing the NIC fabric with AFDTransfer) will either keep tacking new keys onto this nested dict (silent schema drift) or force a full refactor. The longer this lives, the more callers depend on the current shape.

Minimal change can be: return a dataclass `AFDTransferBreakdown(t_a2f, t_f2a, t_a, t_f) `so the schema is at least explicit and type-checked.

Or split into four distinct Operation subclasses — `AFDCrossPoolTransfer` `AFDIntraFAllGather`, `AFDIntraFReduceScatter`, `AFDCombine`. Each returns PerformanceResult. `AFDInferenceSession` holds the four references and composes them in the pipeline formula. The `Operation.query() `contract stays intact, and Phase 2 becomes an additive operation rather than schema surgery.

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -828,3 +829,867 @@ def _find_best_result_under_constraints(
...
+            "ttft": round(ttft, 3),
+            "tpot": round(tpot, 3),
+            "request_latency": round(request_latency, 3),
+            "b_total": b_total,
+            "tokens/s": round(tokens_per_s, 2),
+            "tokens/s/gpu": round(tokens_per_s_per_gpu, 2),
+            "tokens/s/user": round(tokens_per_s_per_user, 2),
+            "seq/s": round(seq_per_s, 3),
+            "concurrency": b_total,
+            "tpuc": round(tokens_per_s_per_gpu, 2),
```
yunyang1999: Same value as tokens/s/gpu on line 1627?

tpuc stands for "Tokens Per User Concurrency"?

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -828,3 +829,867 @@ def _find_best_result_under_constraints(
...
+            "t_step": round(t_step, 3),
+            "ttft": round(ttft, 3),
+            "tpot": round(tpot, 3),
+            "request_latency": round(request_latency, 3),
+            "b_total": b_total,
+            "tokens/s": round(tokens_per_s, 2),
+            "tokens/s/gpu": round(tokens_per_s_per_gpu, 2),
+            "tokens/s/user": round(tokens_per_s_per_user, 2),
+            "seq/s": round(seq_per_s, 3),
+            "concurrency": b_total,
```
yunyang1999, line 1826: In the K=3 ping-pong pipeline, true in-flight requests are roughly `num_microbatches * b_total (each microbatch carries its own activation set)`. concurrency typically denotes total in-flight — so when comparing AFD vs disagg concurrency columns, this discrepancy is amplified.

Fix: either "concurrency": cfg.num_microbatches * b_total, or drop the concurrency column and keep only b_total to avoid ambiguity.

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -828,3 +829,867 @@ def _find_best_result_under_constraints(
...
+        isl = runtime_config.isl
+        osl = runtime_config.osl or 1
+        b_total = cfg.n_a_workers * cfg.a_batch_size
+
+        # Pick the metrics that drive latency/throughput headline numbers.
+        drive = decode_metrics if decode_metrics is not None else prefill_metrics
+        assert drive is not None, "At least one phase must be simulated"
+
+        # Per-phase scalar metrics (zero-fill the ones we didn't run).
+        t_a_layer = drive["t_a_layer"]
```
yunyang1999: In phase="both", headline scalars come only from `decode_metrics`; prefill metrics with the same keys are dropped

request_latency = ttft + tpot * (osl - 1) does combine both phases, but the column reader has no way to know t_a_layer reflects only decode. Prefill t_a_layer / t_f_layer / balance_ratio can be very different, and silently showing only one is misleading.

Can add prefill-prefixed paired columns or in phase="both" mode, refuse to populate the un-prefixed scalars

src/aiconfigurator/sdk/task.py
```diff
@@ -93,7 +121,7 @@ def resolve(self, ctx: TaskContext) -> dict:
 
 @dataclass
 class TaskContext:
-    serving_mode: Literal["agg", "disagg"]
+    serving_mode: Literal["agg", "disagg", "afd"]
```
yunyang1999, line 182: gpt-oss Blackwell quant promotion else branch never adapts to AFD
around line R402-423:
```
            if ctx.serving_mode == "agg":
                if ctx.system_name in _blackwell_systems:
                    _deep_merge(config_dict, {"worker_config": quant_override})
                    applied_layers.append("gptoss-blackwell-mxfp8")
            else:
                prefill_system = ctx.system_name
                decode_system = ctx.decode_system_name or ctx.system_name
                promoted = {}
                if prefill_system in _blackwell_systems:
                    promoted["prefill_worker_config"] = quant_override
                if decode_system in _blackwell_systems:
                    promoted["decode_worker_config"] = quant_override
                if promoted:
                    _deep_merge(config_dict, promoted)
                    applied_layers.append("gptoss-blackwell-mxfp8")
```
AFD reads only worker_config (same shape as agg) and never prefill_/decode_worker_config. So gpt-oss-120b + Blackwell + --estimate-mode afd ends up with worker_config.moe_quant_mode un-promoted; the quantization optimization is silently lost.

Can fix as:
```
if ctx.serving_mode == "agg":
    ...
elif ctx.serving_mode == "disagg":
    ...
elif ctx.serving_mode == "afd":
    if ctx.system_name in _blackwell_systems:
        _deep_merge(config_dict, {"worker_config": quant_override})
```

jiangkuaixue123:
Thanks for the detailed AFD implementation. Do we already have any experimental results or validation runs for the AFD simulation path that could be shared with reviewers?

It would be helpful to see a few representative outputs, for example:
- one or two end-to-end `aiconfigurator cli estimate --estimate-mode afd ...` runs,
- the model/system/backend/version used,
- how the estimated A/F balance, communication-hidden flag, TPOT/TTFT, and memory/OOM checks look,
- and, if available, any comparison against a baseline agg/disagg estimate or prototype measurement.

This would make it easier to sanity-check the new pipeline model and understand the expected magnitude of benefit from AFD.

src/aiconfigurator/sdk/afd_partition.py
xutizhou, line 273: Could we make this A/F placement policy more flexible instead of hard-coding it entirely in the classifier? Today all attention-side ops (for example QKV/MLA/proj GEMMs) and FFN/MoE ops are fixed by name rules, with only boundary ops configurable via `boundary_on_attn`. For AFD design-space exploration it would be useful to let users/configs decide selected op placement, e.g. a partition policy with defaults plus overrides for specific op names or op groups. That would also make experiments like moving shared experts, router, or projection GEMMs between A and F possible without editing this classifier.

src/aiconfigurator/sdk/inference_session.py
jieliu1997: Fixed in the latest push. `_simulate_phase` now computes `effective_prefill_len = max(isl - runtime_config.prefix, 1)` and uses it both for the per-pool latency query and for `AFDTransfer` token volume. TTFT now matches the prefix-aware form you suggested.

src/aiconfigurator/sdk/inference_session.py
jieliu1997: Same fix as L1: `tokens_per_req = effective_prefill_len if phase == "prefill" else 1`.

src/aiconfigurator/cli/api.py
jieliu1997: Fixed. `_run_afd_estimate` now accepts `prefix` (along with `nextn` and`nextn_accept_rates`) and threads it into `RuntimeConfig`, mirroring `_run_agg_estimate`. The `--prefix` CLI flag is now honored for AFD as well; covered by a unit test.

src/aiconfigurator/sdk/inference_session.py
jieliu1997: Yes, this is a Phase-2 TODO.

src/aiconfigurator/cli/api.py
jieliu1997: Plumbing only in this PR: `_run_afd_estimate` now threads `nextn` /`nextn_accept_rates` through `_apply_nextn(a_model_config, ...)` / `_apply_nextn(f_model_config, ...)`, so the CLI flag isn't dropped and activation-memory picks up the `(nextn+1)` scaling via `backend.get_partition_memory_usage`. The AFD pipeline-latency path itself does not yet consume `nextn` — proper MTP modeling (asymmetric A-side draft-layer scaling, acceptance-rate TPOT division, AFDTransfer volume recalibration) is tracked as a follow-up TODO.

src/aiconfigurator/sdk/inference_session.py
jieliu1997: Not arbitrary — it mirrors the stride used by `base_backend._run_generation_phase`.

src/aiconfigurator/sdk/afd_partition.py
jieliu1997: Fixed. All three classify paths (`_classify_op`, `_classify_overlap_op`, `_classify_inner_overlap_op`) now delegate to a single `_classify_by_markers(name, ...)` helper, so the `boundary → attn → ffn` order lives in exactly one place.

src/aiconfigurator/sdk/afd_partition.py
jieliu1997: Handled in the code — `_is_attention_side_op` now short-circuits to False when the name contains any of `down_proj_gemm / up_proj_gemm / gate_proj_gemm`.

src/aiconfigurator/sdk/afd_partition.py
jieliu1997: Same fix as above: `_classify_by_markers` is the helper you suggested. The three call sites collapse to "OverlapOp recursion + delegate".

src/aiconfigurator/sdk/inference_session.py
jieliu1997: Same value — removed the duplicate field.

src/aiconfigurator/sdk/inference_session.py
jieliu1997: Discussed and reconciled — concurrency now reports total in-flight as `num_microbatches * b_total`, consistent with the agg/disagg column semantics. The per-microbatch view is exposed separately as `b_micro_total / micro_bs` so the pipeline structure stays inspectable.

src/aiconfigurator/sdk/inference_session.py
jieliu1997: Fixed. Per-phase layer scalars now exist in *paired* `prefill_*` / `decode_*` columns; in `phase="both"` the un-prefixed headline form is explicitly set to NaN/None so prefill metrics can no longer be silently shadowed. The CLI prints a dual-block layout for `phase="both"` and a `Composition` row whenever an AFD phase was merged with a static estimate, making the AFD-vs-static split visible at a glance.`ColumnsAFD` extended accordingly.

src/aiconfigurator/sdk/task.py
jieliu1997: Fixed. The promotion now dispatches on `serving_mode` explicitly with an `elif ctx.serving_mode == "afd"` branch that writes to `worker_config` (AFD shape), matching the agg path. The previous two-branch fallthrough wrote to `prefill_/decode_worker_config` keys that AFD never reads — exactly the silent loss of `w4a8_mxfp4_mxfp8` you flagged. Regression test added covering AFD + Blackwell.

Arsene12358:
@gehujun thanks for contributing, could you do a rebase and resolve the CI failing checks first? Also Could you help address the remaining review comments from coderabbitai ? Thanks

src/aiconfigurator/sdk/operations.py
```diff
@@ -2133,3 +2133,305 @@ def get_weights(self, **kwargs):
...
             weights += op.get_weights(**kwargs)
         return weights
+
+
+def _afd_send_prob(num_experts: int, topk: int, num_f_nodes: int) -> float:
+    """Probability that a token must be dispatched to a given F-node.
+
+    For MoE with expert parallelism (num_experts > 0, topk > 0,
+    num_f_nodes > 1): uses the combinatorial formula
+    ``P_send = 1 - C(E - E/Nf, k) / C(E, k)`` -- the probability that
```
liyuanzhe1991: Does this mean that when computing with the power-law distribution, a different distribution is adopted for communication?

src/aiconfigurator/sdk/operations.py
```diff
@@ -2133,3 +2133,305 @@ def get_weights(self, **kwargs):
...
+        x = int(kwargs.get("x", 1) or 1)
+        nf = self.num_f_nodes
+        p_send = _afd_send_prob(self._num_experts, self._topk, nf)
+        bpe = self._comm_quant_mode.value.memory
+        # x = tokens held by a single A-rank; per_link_bytes is
+        # the volume on one A-rank → F-rank P2P connection.
+        per_link_bytes = int(p_send * x * self._hidden_size * bpe)
+        if per_link_bytes <= 0:
+            return PerformanceResult(0.0, 0.0, source="empirical")
+        result = database.query_p2p(per_link_bytes)
```
liyuanzhe1991: AFD Transfer uses a single expected per-link payload with query_p2p(), but the real A/F exchange is a many-to-many transfer across all A workers and F nodes, so this likely under-models fan-in/fan-out and make it less accuracy.

src/aiconfigurator/sdk/afd_partition.py
```diff
@@ -0,0 +1,343 @@
...
+    # ``_pipeline_tcycle``. That hides the MoE EP all-to-all from the AFD
+    # pipeline overlap math, so contention between the MoE all-to-all and
+    # the cross-pool A<->F transfer on the same NIC fabric is silently
+    # dropped. Surface ``MoEDispatch`` cost as an additional contribution
+    # to ``t_c_layer`` (or split the OverlapOp into compute + dispatch
+    # sub-stages) when AFD is active, instead of folding it into ``t_f``.
+    if isinstance(op, (operations.CustomAllReduce, operations.P2P, operations.NCCL)):
+        return True
+
+    return any(
```
liyuanzhe1991, line 211: If communication operations are placed inside overlap_op, they will still be factored into latency calculations and cannot be skipped as expected.

gehujun:
> @gehujun thanks for contributing, could you do a rebase and resolve the CI failing checks first? Also Could you help address the remaining review comments from coderabbitai ? Thanks

Okay, I will solve all the problems with coderabbitai.

src/aiconfigurator/cli/api.py
```diff
@@ -664,6 +664,18 @@ def cli_estimate(
     nextn: int = 0,
     nextn_accept_rates: list[float] | None = None,
     stride: int = 32,
+    # AFD-specific parameters (ignored when mode != 'afd')
+    n_a_nodes: int | None = None,
+    n_f_nodes: int | None = None,
+    a_tp_size: int = 1,
+    a_batch_size: int = 128,
+    f_moe_ep_size: int = 1,
+    num_microbatches: int = 3,
```
tianhaox, line 673: can you update the docstring of these new args?

src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -616,6 +617,71 @@ def _run_generation(bs: int, bw: int, effective_isl: int, eff_osl: int, strd: in
...
+            **kwargs,
+        )
+        memory = dict(memory)
+        memory["weights"] = sum(op.get_weights() for op in partition_ops) / (1 << 30)
+        if include_kvcache:
+            memory["kvcache"] = memory.get("kvcache", 0.0) * max(kvcache_multiplier, 1)
+        else:
+            memory["kvcache"] = 0.0
+
+        memory.setdefault("activations", 0.0)
```
tianhaox, line 688: why we set act, nccl and othres to 0.0?

tianhaox:
please resolve lint and DCO. https://github.com/ai-dynamo/aiconfigurator/blob/main/DEVELOPMENT.md
I'm okay with the overall design. We have an ongoing refactor, https://github.com/ai-dynamo/aiconfigurator/pull/1121, will merge AFD first and merge task v2 (2 phases, this pr is phase1) to unify the task entrypoint.

src/aiconfigurator/sdk/afd_partition.py
zhengqianyi: Good suggestion. We'll add a side_overrides mechanism afterward, threaded through the full CLI + SDK path，so users can override placement of any op by name without editing the classifier.

  CLI:

  aiconfigurator estimate --mode afd \
    --afd-side-override generation_router_gemm=attn \
    --afd-side-override generation_proj_gemm=ffn

  SDK:

  AFDConfig(
      ...,
      side_overrides={
          "generation_router_gemm": "attn",
          "generation_proj_gemm": "ffn",
      },
  )

src/aiconfigurator/sdk/operations.py
gehujun: Fixed. We went with your second suggestion — split into four distinct Operation subclasses:

  - AFDTransfer — cross-pool A↔F P2P transfer
  - AFDFAllGather — intra-F-node AllGather (token dimension)
  - AFDFReduceScatter — intra-F-node ReduceScatter (post-FFN)
  - AFDCombine — EP combine when f_moe_ep_size > 1

  Each returns PerformanceResult via its own query(), so the Operation.query() contract is preserved. AFDInferenceSession holds the four references and composes them in the pipeline formula — no special-case dict unpacking, no schema
  drift. Phase 2 additions (shared-expert traffic, MoE EP contention) become additive operations rather than surgery on a nested dict.

  See src/aiconfigurator/sdk/operations/afd_transfer.py.

src/aiconfigurator/sdk/afd_partition.py
```diff
@@ -0,0 +1,343 @@
...
+    # skip-list only excludes it from the overlap *classification vote* --
+    # its cost is still folded into the OverlapOp's F-pool latency via
+    # ``OverlapOp.query()`` and stays invisible to the AFD comm ops /
+    # ``_pipeline_tcycle``. That hides the MoE EP all-to-all from the AFD
+    # pipeline overlap math, so contention between the MoE all-to-all and
+    # the cross-pool A<->F transfer on the same NIC fabric is silently
+    # dropped. Surface ``MoEDispatch`` cost as an additional contribution
+    # to ``t_c_layer`` (or split the OverlapOp into compute + dispatch
+    # sub-stages) when AFD is active, instead of folding it into ``t_f``.
+    if isinstance(op, (operations.CustomAllReduce, operations.P2P, operations.NCCL)):
```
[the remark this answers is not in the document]
gehujun: Acknowledged — tracked as the Phase-2 TODO at lines 203-213.

  In the Phase-1 pipeline model, MoE EP all-to-all dispatch latency is folded into t_f (the F-pool compute stage), while cross-pool A↔F transfers live in the separate t_c stage. Under the K=3 optimistic formula t_cycle = max(t_a,
  t_f, t_c), these occupy different pipeline stages — when the F-pool is running FFN + EP dispatch for microbatch i, the NIC is carrying A↔F transfers for microbatch i-1. They are temporally staggered by design, not superimposed.

  Regarding NIC bandwidth contention: Phase-1 assumes uniform token distribution across F-nodes (_afd_send_prob uses combinatorial probability), so both EP dispatch and A↔F transfer loads are evenly spread across links. Severe
  contention would require both collectives to saturate the same physical link in the same time window — but the pipeline's purpose is precisely to prevent that overlap. When the pipeline cannot stagger them (insufficient
  microbatches), the model already falls back to conservative mode max(t_a + t_a2f, t_f + t_f2a), where EP dispatch latency inside t_f is serialized with t_f2a, implicitly and conservatively accounting for the shared fabric.

  Phase-2 will explicitly model NIC contention between EP dispatch and cross-pool transfers (particularly at multi-node EP scales where the pipeline depth sits near the K=2/K=3 boundary), but for Phase-1's typical topologies and
  pipeline depths, the estimation error is within acceptable bounds

src/aiconfigurator/sdk/config.py
```diff
@@ -119,3 +119,149 @@ class RuntimeConfig:
...
+        # invariant; otherwise it silently implies F-DP > 1.
+        derived_tp_f = self.n_f_workers
+        if self.tp_f and self.tp_f != derived_tp_f:
+            raise ValueError(
+                f"tp_f ({self.tp_f}) is derived under the Phase 1 F-DP=1 "
+                f"assumption as n_f_nodes * gpus_per_node = {derived_tp_f}; "
+                "an explicit different value would silently imply F-DP > 1. "
+                "Remove the override, or change n_f_nodes / gpus_per_node."
+            )
+        self.tp_f = derived_tp_f
```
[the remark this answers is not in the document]
gehujun: Fixed. Added fail-fast validation at the top of AFDConfig.__post_init__ covering all enum and positivity constraints:

  - phase must be one of {"prefill", "decode", "both"}
  - pipeline_model must be one of {"optimistic", "conservative"}
  - n_a_nodes / n_f_nodes >= 1
  - a_batch_size >= 1
  - num_microbatches >= 1
  - comm_overhead_factor > 0
  - f_moe_ep_size >= 1

src/aiconfigurator/cli/api.py
```diff
@@ -1346,6 +1466,272 @@ def _run_disagg_estimate(
...
+        pp_size=afd_result.pp_size,
+        model_path=afd_result.model_path,
+        system_name=afd_result.system_name,
+        backend_name=afd_result.backend_name,
+        backend_version=afd_result.backend_version,
+        raw=result_dict,
+        mode="afd",
+        per_ops_data=per_ops_data,
+        kv_cache_warning=static_result.kv_cache_warning,
+    )
```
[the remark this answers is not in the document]
gehujun: Fixed. Both AFD EstimateResult constructors now carry per_ops_source:

  1. _combine_afd_static_estimate_results (AFD + static merge path): added per_ops_source=dict(afd_result.per_ops_source or {})
  2. _run_afd_estimate direct return path: added per_ops_source=summary.get_per_ops_source()

  Consistent with the disagg path (line 1465). The summary object already exposes get_per_ops_source() — the AFD paths simply omitted the call. --detail source now correctly displays per-op provenance for AFD mode.

src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -616,6 +617,71 @@ def _run_generation(bs: int, bw: int, effective_isl: int, eff_osl: int, strd: in
...
+        memory.setdefault("nccl", 0.0)
+        memory.setdefault("others", 0.0)
+        memory["total"] = (
+            memory["weights"]
+            + memory["activations"]
+            + memory["kvcache"]
+            + memory["nccl"]
+            + memory["others"]
+        )
```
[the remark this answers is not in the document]
gehujun: Fixed. get_partition_memory_usage now divides partition weights by max(model.config.pp_size, 1), consistent with the convention in _get_memory_usage (line 1364).

  In AFD Phase-1, pp_size is always 1 (hardcoded when _run_afd_estimate builds A/F model configs), so this is a defensive fix — no behavioral change today, but it ensures that if Phase-2 introduces pipeline parallelism within AFD
  pools, partition weights won't be over-counted and trigger false OOM

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -839,3 +840,1025 @@ def _find_best_result_under_constraints(
...
+        if prefill_metrics is not None:
+            ttft = prefill_metrics["t_step"]
+        else:
+            ttft = 0.0
+
+        if phase == "prefill":
+            request_latency = ttft
+        elif phase == "decode":
+            request_latency = tpot * max(osl - 1, 1) if tpot > 0 else tpot
+        else:  # both
```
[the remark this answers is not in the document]
gehujun: Fixed. Changed request_latency = tpot * max(osl - 1, 1) to tpot * max(osl - 1, 0) in the phase="decode" branch, consistent with the phase="both" branch (line 1706).

  The original max(..., 1) forced at least one decode step worth of latency when osl=1, but osl=1 means zero decode steps (the output is just the first token produced by prefill). decode_metrics itself remains valid — it's the
  per-step latency that's correct, just multiplied by the actual number of steps.

src/aiconfigurator/sdk/operations.py
```diff
@@ -2133,3 +2133,305 @@ def get_weights(self, **kwargs):
...
+    def query(self, database: PerfDatabase, **kwargs) -> PerformanceResult:
+        x = int(kwargs.get("x", 1) or 1)
+        nf = self.num_f_nodes
+        p_send = _afd_send_prob(self._num_experts, self._topk, nf)
+        bpe = self._comm_quant_mode.value.memory
+        # x = tokens held by a single A-rank; per_link_bytes is
+        # the volume on one A-rank → F-rank P2P connection.
+        per_link_bytes = int(p_send * x * self._hidden_size * bpe)
+        if per_link_bytes <= 0:
+            return PerformanceResult(0.0, 0.0, source="empirical")
```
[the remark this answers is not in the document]
gehujun: Fixed. All three query() methods (AFDTransfer, AFDFAllGather, AFDFReduceScatter) now short-circuit on x <= 0:

  x = int(kwargs.get("x", 0))
  if x <= 0:
      return PerformanceResult(0.0, 0.0, source="empirical")

  In the current call path x is always >= 1 (a_batch_size * tokens_per_req, both >= 1), and the downstream per_link_bytes <= 0 / per_rank_elements <= 0 guards would have caught it anyway. This change is primarily semantic alignment
  — "0 tokens = 0 communication" is more explicit than relying on or 1 followed by downstream guards, and prevents phantom latency if a future caller passes x=0 to model an empty/cached phase.

src/aiconfigurator/sdk/operations.py
```diff
@@ -2133,3 +2133,305 @@ def get_weights(self, **kwargs):
...
+
+    def query(self, database: PerfDatabase, **kwargs) -> PerformanceResult:
+        if self._f_moe_ep_size <= 1:  # no expert parallelism
+            return PerformanceResult(0.0, 0.0, source="empirical")
+        x = int(kwargs.get("x", 1) or 1)
+        tokens_per_a_rank = max(1, x // self._tp_a)
+        bpe = self._comm_quant_mode.value.memory
+        total_bytes = int((self._f_moe_ep_size + 1) * tokens_per_a_rank * self._hidden_size * bpe)
+        if total_bytes <= 0:
+            return PerformanceResult(0.0, 0.0, source="empirical")
```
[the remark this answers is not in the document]
gehujun: Fixed with two changes:

  1. x=0 short-circuit: consistent with AFDTransfer / AFDFAllGather / AFDFReduceScatter — returns zero latency when x <= 0.
  2. Floor → ceil sharding: tokens_per_a_rank = max(1, x // tp_a) changed to (x + tp_a - 1) // tp_a. The combine stage is paced by the busiest A-rank, so ceil-division correctly models the hot-rank data volume when token counts
  don't divide evenly.

  Practical impact is minimal — the caller passes x = a_batch_size * tokens_per_req, which divides evenly by tp_a in most configurations. When it doesn't, the difference is one token's worth of HBM access (nanosecond-scale). But
  semantically ceil is correct: combine latency is bounded by the rank with the most data.

src/aiconfigurator/sdk/task.py
```diff
@@ -667,6 +717,108 @@ def _finalize_disagg(cls, config: DefaultMunch, ctx: TaskContext) -> None:
...
+                return
+            total_requested = (n_a_nodes + n_f_nodes) * spec_gpus_per_node
+            if total_requested > ctx.total_gpus:
+                logger.warning(
+                    "AFD config requests %d GPUs "
+                    "(n_a_nodes=%d + n_f_nodes=%d) * gpus_per_node=%d "
+                    "but total_gpus=%d. Consider adjusting.",
+                    total_requested, n_a_nodes, n_f_nodes, spec_gpus_per_node,
+                    ctx.total_gpus,
+                )
```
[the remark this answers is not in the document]
gehujun: Fixed. Changed from logger.warning to raise ValueError when total_requested > ctx.total_gpus.

  This check only triggers when the user explicitly passes --total-gpus (ctx.total_gpus is not None and ctx.total_gpus > 0). If no GPU budget is set, total_gpus is None and the entire validation block is skipped — unconstrained runs
  are unaffected.

  When the user provides an explicit GPU budget but the AFD topology (n_a_nodes + n_f_nodes) * gpus_per_node exceeds it, continuing would return estimates for a larger cluster than budgeted, silently misleading downstream sizing
  decisions. Fail-fast ensures the user immediately knows to adjust n_a_nodes/n_f_nodes or relax the total_gpus constraint. Consistent with how agg/disagg paths enforce total_gpus as a hard constraint.

src/aiconfigurator/sdk/task.py
```diff
@@ -1590,6 +1747,110 @@ def _wc_get(wc: object, key: str, fallback):
...
+            afd_config=afd_config,
+        )
+        summary = session.run_afd(
+            runtime_config,
+            phase=afd_config.phase,
+            free_gpu_memory_fraction=getattr(task_config, "free_gpu_memory_fraction", None),
+            max_seq_len=getattr(task_config, "max_seq_len", None),
+        )
+        result_df = summary.get_summary_df()
+        return {"pareto_df": result_df}
```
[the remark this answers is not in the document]
gehujun: Acknowledged. prefix, nextn, and the combined_with_pd static merge are all handled correctly in the default CLI path (cli_estimate → _run_afd_estimate). The Task path (run_afd) is a simplified entry point for automated search that
  currently only does single-point AFD estimation.

  These will be aligned when the task v2 refactor (PR #1121) unifies the task entrypoint.

docs/add_a_new_model.md
```diff
@@ -111,6 +111,21 @@ Steps required:
...
+2. **F-worker ops**: operations that belong to the FFN / MoE side, such as router GEMMs, dense `ffn` / `mlp` ops, routed expert ops, shared expert ops, and activation / gate / up / down GEMMs.
+3. **Boundary ops**: operations that naturally sit at the A/F boundary, such as `add_norm_2`, FFN/MoE/MLP norms, `logits_gemm`, `reduce_add`, and `*_combine`. These default to the A-worker and can be moved to the F-worker with the AFD boundary placement option.
+4. **Skipped model-internal communication ops**: communication or dispatch operations already represented by the AFD communication model, such as `CustomAllReduce`, `P2P`, `NCCL`, TP all-gather / reduce-scatter, and MoE dispatch ops. These should not be counted again in either compute pool because AFD adds its own cross-pool and intra-pool communication through `AFDTransfer`.
+5. **Overlap ops**: an `OverlapOp` can stay atomic only when every non-skipped inner op belongs to the same side. If an overlap group spans the A/F boundary, the partitioner must fail or be extended to split / rebuild that overlap explicitly.
```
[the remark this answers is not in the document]
gehujun: Fixed. The docs now enumerate all four AFD communication primitives: AFDTransfer (cross-pool P2P), AFDFAllGather (intra-F-node AllGather), AFDFReduceScatter (intra-F-node ReduceScatter), and AFDCombine (A-side EP combine).

src/aiconfigurator/sdk/afd_partition.py
```diff
@@ -0,0 +1,343 @@
...
+    # ``proj_gemm`` is the canonical attention output projection (``W_O``)
+    # name used across deepseek / llama / gpt / moe / hybrid_moe / qwen35
+    # / nemotron, but the bare substring also matches HF-style FFN names
+    # (see ``_FFN_PROJ_GEMM_MARKERS`` above). Allow ``proj_gemm`` only
+    # when none of the FFN-style forms is present; the matching FFN
+    # markers added to ``_is_ffn_side_op`` below pick up the excluded
+    # names on the second-pass attn -> ffn check.
+    if "proj_gemm" in name and not any(ffn in name for ffn in _FFN_PROJ_GEMM_MARKERS):
+        return True
+    return False
```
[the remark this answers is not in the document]
gehujun: Fixed. Returning the boolean condition directly to satisfy Ruff SIM103.

tests/unit/cli/test_afd_phase_completion.py
```diff
@@ -0,0 +1,753 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for completing single-phase AFD estimates with regular static phases."""
+
+import math
+from types import SimpleNamespace
+
+import pytest
```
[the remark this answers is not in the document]
gehujun: Fixed. Added from typing import ClassVar import.

tests/unit/cli/test_afd_phase_completion.py
```diff
@@ -0,0 +1,753 @@
...
+
+
+def test_run_afd_estimate_passes_prefix_and_nextn(monkeypatch):
+    captured = {}
+
+    class FakeDatabase:
+        system_spec = {
+            "node": {"num_gpus_per_node": 8},
+            "gpu": {"mem_capacity": 80 * (1 << 30)},
+        }
```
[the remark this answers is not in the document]
gehujun: Fixed. Same as above — annotated second system_spec with ClassVar[dict]

tests/unit/cli/test_afd_phase_completion.py
```diff
@@ -0,0 +1,753 @@
...
+        "_build_models",
+        lambda self: (SimpleNamespace(_num_layers=4), SimpleNamespace(_num_layers=4)),
+    )
+    monkeypatch.setattr(AFDInferenceSession, "_simulate_phase", fake_simulate_phase)
+
+    class FakeDatabase:
+        version = "test-version"
+        system = "test-system"
+        system_spec = {"gpu": {"mem_capacity": 80 * (1 << 30)}}
+
```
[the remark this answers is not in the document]
gehujun: Fixed. Annotated system_spec with ClassVar[dict] (Ruff RUF012)

tests/unit/cli/test_afd_phase_completion.py
```diff
@@ -0,0 +1,753 @@
...
+    # Static-prefill stand-in: a minimal EstimateResult that the combiner
+    # treats as the "other phase". Only the fields the combiner reads
+    # (mode/raw/throughput numbers) need to be plausible.
+    static_raw = {
+        "model": afd_raw["model"] if "model" in afd_raw else "test-model",
+        "(p)tp": 1, "(p)pp": 1, "(p)dp": 1, "(p)bs": 1, "(p)gpus": 1,
+        "(p)impl": "trtllm",
+        "ttft": 25.0, "seq/s": 5.0, "tokens/s/user": 100.0,
+    }
```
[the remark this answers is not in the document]
gehujun: Fixed. Simplified to afd_raw.get("model", "test-model") (Ruff SIM401).

tests/unit/sdk/test_task_blackwell_gptoss.py
```diff
@@ -0,0 +1,155 @@
...
+    """Non-Blackwell + AFD must NOT promote; no Blackwell layer should be recorded."""
+    ctx = _make_ctx("afd", "h200_sxm")
+    config, applied = TaskConfigFactory.create(ctx)
+
+    assert config["worker_config"].get("moe_quant_mode") != "w4a8_mxfp4_mxfp8"
+    assert not any(label.startswith("gptoss-blackwell-mxfp8") for label in applied)
+
+
+def test_agg_gptoss_blackwell_back_compat():
+    """The agg branch keeps writing ``worker_config.moe_quant_mode`` and emitting the bare ``gptoss-blackwell-mxfp8`` label."""
```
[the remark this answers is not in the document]
gehujun: Fixed. Wrapped the docstring to stay within 120 chars (Ruff E501)

src/aiconfigurator/sdk/backends/base_backend.py
jieliu1997: This code does not set it to 0: `setdefault("activations", 0.0)` only fills the key if the backend does not return it, and keeps backend-provided values. For TRTLLM/SGLang, `_get_memory_usage()` already returns `activations`, `nccl`, and `others`. The only field intentionally set to 0 for F-worker is `kvcache`, because F runs FFN/MoE and does not own attention KV cache.

src/aiconfigurator/sdk/operations.py
yuxingtyx: Currently, AFDTransfer's P_send uses a combinatorial formula that assumes uniform expert routing — an intentional Phase 1 simplification providing an analytical estimate independent of the actual expert frequency distribution. We plan to calibrate it against real AFD communication profiles.

src/aiconfigurator/sdk/operations.py
yuxingtyx: The current model assumes a 1-to-1 rank mapping between A-ranks and F-ranks, so each link is modeled as an independent P2P transfer via query_p2p(per_link_bytes). Fan-in/fan-out contention in many-to-many scenarios is not yet captured — an intentional Phase 1 simplification.

src/aiconfigurator/cli/api.py
Arsene12358: @gehujun could you update the docstring as requested, thanks

gehujun: ok

gehujun: Done — I've updated the docstring of cli_estimate() in src/aiconfigurator/cli/api.py to document all the new AFD-specific args (n_a_nodes, n_f_nodes, a_tp_size, a_batch_size, f_moe_ep_size, num_microbatches, pipeline_model, comm_overhead_factor, afd_phase, afd_combined_with_pd, afd_boundary_on_attn), and also mentioned the 'afd' mode in the mode description. Pushed in a new commit. Thanks for the review!

Arsene12358:
@gehujun All comments are resolved, thanks again for contributing @gehujun and look forward to future PRs

Arsene12358 (approved):

PR #1147, opened 2026-05-27
fix: correct mix-step decode token count for b >= osl

### LINKED ISSUE #1146: TPOT predictions are identical for all batch sizes b >= osl in vLLM aggregated mode
## Observation

When running `cli_estimate` or `cli_default` in aggregated vLLM mode with a workload where the batch size equals or exceeds the output sequence length (b >= osl), AIC produces identical TPOT predictions regardless of batch size.

Measured at ISL=9000, OSL=30, Qwen3-8B, H200 SXM, vLLM 0.18.0 using single-replica guidellm sweeps:

| b | AIC TPOT | Measured ITL |
|---|----------|--------------|
| 16 | 117.87 ms | 93.63 ms |
| 32 | 238.58 ms | 236.8 ms |
| 64 | **238.58 ms** (identical to b=32) | 308.8 ms |

b=32 and b=64 produce the same AIC prediction despite the measured ITL continuing to grow. The prediction error at b=64 is +29%.

## Analysis

In `run_agg`, when `steps_to_finish_ctx >= osl`, `num_mix_gen_tokens` is computed as:

```python
num_mix_gen_tokens = max(1, b // (steps_to_finish_ctx / osl))
```

Since `steps_to_finish_ctx = ceil(isl * b / ctx_tokens)` with `ctx_tokens` defaulting to `isl`, this simplifies to `max(1, b // (b/osl)) ≈ osl` — constant for any b. The mix step latency (and therefore TPOT) then no longer scales with b.

## Scheduler context

vLLM v1's scheduler sets `max_num_partial_prefills=1` by default, meaning exactly one request prefills per forward pass while the remaining `b-1` requests decode. The number of decode tokens per mix step therefore grows linearly with b, which the current formula does not reflect.

The `steps_to_finish_ctx < osl` branch already uses `b - ceil(ctx_tokens / isl)` for `num_mix_gen_tokens`, which is consistent with this scheduler behaviour. The two branches are therefore inconsistent with each other.

natoscott:
Fixes #1146.

## Summary

When `steps_to_finish_ctx >= osl`, `num_mix_gen_tokens` was computed as `max(1, b // (steps_to_finish_ctx / osl)) ≈ osl` for any b, causing identical TPOT predictions for all batch sizes where `b >= osl`. At ISL=9000, OSL=30, b=32 and b=64 both predicted 238.58 ms while measured ITL was 236.8 ms and 308.8 ms respectively.

Introduces `BaseBackend._mix_step_gen_tokens(b, ctx_tokens, isl, osl)`, covering both scheduling regimes. `VLLMBackend` overrides with `b - ceil(ctx_tokens / isl)`, derived from vLLM v1's `max_num_partial_prefills=1` scheduler default (confirmed from source).

## Validation

ISL=9000, OSL=30, Qwen3-8B, H200 SXM, vLLM 0.18.0:

| b | Before | After | Measured |
|---|--------|-------|----------|
| 32 | 238.58 ms (+0.8%) | 238.90 ms (+1%) | 236.8 ms |
| 64 | 238.58 ms (+29%) | 248.76 ms (-19%) | 308.8 ms |

b=32 and b=64 now produce distinct predictions.

### DIFF
--- a/src/aiconfigurator/sdk/backends/base_backend.py
+++ b/src/aiconfigurator/sdk/backends/base_backend.py
@@ -83,6 +83,21 @@ class BaseBackend:
         """
         return getattr(model, "_hidden_size", h)
 
+    def _mix_step_gen_tokens(self, b: int, ctx_tokens: int, isl: int, osl: int) -> int:
+        """Return the number of decode tokens per mix step for a batch of b requests.
+
+        A mix step is a forward pass that contains both prefill tokens (for requests
+        still completing their context phase) and decode tokens (for requests already
+        generating). This method encodes the engine's scheduling policy for how many
+        decode-phase requests participate alongside the prefilling request(s).
+
+        Subclasses should override to match their engine's scheduling behaviour.
+        """
+        steps_to_finish_ctx = np.ceil(isl * b / ctx_tokens)
+        if steps_to_finish_ctx >= osl:
+            return max(1, int(b // (steps_to_finish_ctx / osl)))
+        return max(1, b - int(np.ceil(ctx_tokens / isl)))
+
     def _resolve_agg_kwargs(self, kwargs: dict, isl: int, osl: int) -> dict:
         """Resolve backend-specific run_agg kwargs to defaults.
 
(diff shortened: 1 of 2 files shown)


jasonqinzhou (approved):

jasonqinzhou:
Thanks a lot for this fix!

PR #1151, opened 2026-05-27
fix: correct mix-step latency over-estimation in run_agg

### LINKED ISSUE #1150: mix-step latency is systematically over-estimated, causing TPOT prediction errors at small b and large ISL
## Observation

AIC systematically over-estimates the latency of mixed forward passes (steps that process both prefill and decode tokens). The over-estimation grows with ISL and ranges from 1.5× to 3× across configurations measured on H200 SXM with vLLM 0.18.0.

Mix step and genonly step latencies were isolated by measuring ITL at two OSL values for the same (ISL, b) and solving the two-equation TPOT system. Across 12 configurations (ISL: 512–4096, b: 4–16), the actual mix step latency is consistently 33–69% of AIC's prediction. Genonly step latency is accurate.

| ISL | b | mix_lat (actual) | mix_lat (AIC) | ratio |
|-----|---|-----------------|---------------|-------|
| 512 | 4 | 7.41 ms | 13.49 ms | 0.55 |
| 512 | 16 | 9.56 ms | 13.79 ms | 0.69 |
| 2048 | 4 | 15.09 ms | 45.64 ms | 0.33 |
| 4096 | 4 | 31.35 ms | 91.55 ms | 0.34 |
| 4096 | 16 | 52.81 ms | 93.58 ms | 0.56 |

The existing `num_mix_steps - 3` pipeline-bubble correction in `run_agg` partially compensates for this effect but is the wrong functional form: the over-estimation is ISL-dependent, which a fixed offset cannot capture.

## Root cause theory

Both vLLM and TRT-LLM process all tokens in a mixed forward pass through a single unified GEMM call (confirmed from source). Per-op silicon data measures each operation in isolation. The data is consistent with a GPU batching efficiency effect: when prefill and decode tokens share a forward pass, the weight matrices may be loaded once from HBM for the combined batch rather than once per token type, making the marginal cost of the prefill tokens substantially lower than isolated measurements predict. The effect is larger at higher ISL where the prefill GEMM is more weight-read-dominated.

natoscott:
Fixes #1150.

## Summary

Introduces `BaseBackend._mix_step_efficiency(ctx_tokens, gen_tokens)` and removes the existing `num_mix_steps - 3` pipeline-bubble correction, which was an empirical approximation for the same underlying effect.

## Model derivation

Mix step and genonly step latencies were isolated by measuring ITL at two OSL values (30 and 128) for 12 (ISL, b) configurations and solving the two-equation TPOT system. A further 4-point OSL sweep at fixed (ISL=1024, b=8) validated the triangulation (std = 0.22 ms across OSL=10/20/30/60).

Power-law model fitted by minimising mean absolute TPOT error (Nelder-Mead, tolerance 1e-7):
```
efficiency = 1.238 × (gen_tokens / (ctx_tokens + gen_tokens)) ^ 0.157
```

| Correction | MAE across 12 training points |
|------------|-------------------------------|
| None | 25.0% |
| -3 correction only (prior) | 25.0% |
| Efficiency model alone | 8.0% |
| Efficiency model + -3 | 16.6% |
| Efficiency model (optimal K, alpha) | 3.72% |

## Accuracy (ISL=9000, OSL=30, Qwen3-8B, H200 SXM, vLLM 0.18.0)

| b | Before | After | Measured |
|---|--------|-------|----------|
| 32 | 238.58 ms | 238.90 ms | 236.8 ms |
| 64 | 238.58 ms | 248.76 ms | 308.8 ms |

## Summary by CodeRabbit

* **Improvements**
  * Mixed context/generation latency and energy calculations are now scaled using per-backend mixed-step efficiency, keeping per-step breakdowns consistent with updated totals.
  * Adjusted mix-step counting for more accurate throughput-related estimates (with pipeline-drain corrections applied only where appropriate).

* **New Features**
  * Backends can override mixed-step efficiency and mix-step counting behavior via new overridable hooks; one backend retains the default efficiency behavior.

* **Tests**
  * Added cross-backend unit tests validating default efficiency, backend-specific step adjustments, and correct clamping behavior.

### DIFF
--- a/src/aiconfigurator/sdk/backends/base_backend.py
+++ b/src/aiconfigurator/sdk/backends/base_backend.py
@@ -100,6 +100,28 @@ class BaseBackend:
             return max(1, int(b // (steps_to_finish_ctx / osl)))
         return max(1, b - int(np.ceil(ctx_tokens / isl)))
 
+    def _mix_step_efficiency(self, ctx_tokens: int, gen_tokens: int) -> float:
+        """GPU batching efficiency factor for a mixed prefill/decode forward pass.
+
+        Per-op silicon data measures each operation in isolation, overstating the
+        marginal cost of prefill tokens when they share a forward pass with decode
+        tokens. Weight matrices are loaded once from HBM for the combined batch.
+        Default: 1.0 (no correction — preserves existing behaviour for backends
+        without empirical efficiency data).
+        """
+        return 1.0
+
+    def _tpot_mix_steps(self, num_mix_steps: int) -> int:
+        """Return the effective mix-step count for TPOT calculation.
+
+        Engines with pipeline-drain latency at the context/decode boundary
+        (requests cannot be immediately enqueued after prefill finishes) may
+        reduce the effective step count to account for that bubble. Default:
+        use the full mix step count. Subclasses should override with an
+        empirically calibrated correction.
+        """
+        return num_mix_steps
+
     def _resolve_agg_kwargs(self, kwargs: dict, isl: int, osl: int) -> dict:
         """Resolve backend-specific run_agg kwargs to defaults.
 
@@ -1133,12 +1155,10 @@ class BaseBackend:
                 num_genonly_tokens = 0
                 num_mix_steps_for_tpot_calc = num_mix_steps
             else:
-                # 3-step is an empirical correction for pipelining requests where new requests
-                # cannot be enqueued immediately after last request's exit
                 num_mix_steps = steps_to_finish_ctx
                 num_genonly_steps = osl - num_mix_steps
                 num_genonly_tokens = b
-                num_mix_steps_for_tpot_calc = max(1, num_mix_steps - 3)
+                num_mix_steps_for_tpot_calc = self._tpot_mix_steps(num_mix_steps)
         elif b == 1:
             # special case for b=1
             num_mix_steps = 1
@@ -1155,6 +1175,11 @@ class BaseBackend:
         mix_step_latency_ms, mix_step_energy_wms, mix_per_ops, mix_per_ops_src = self._get_mix_step_latency(
             model, database, runtime_config, num_mix_ctx_tokens, num_mix_gen_tokens, isl, osl, prefix
         )
+        mix_efficiency = self._mix_step_efficiency(num_mix_ctx_tokens, num_mix_gen_tokens)
+        mix_step_latency_ms *= mix_efficiency
+        mix_step_energy_wms *= mix_efficiency
+        if mix_efficiency != 1.0:
+            mix_per_ops = {op: v * mix_efficiency for op, v in mix_per_ops.items()}
         per_ops_data["mix_step"] = mix_per_ops
         per_ops_source["mix_step"] = mix_per_ops_src
 
--- a/src/aiconfigurator/sdk/backends/sglang_backend.py
+++ b/src/aiconfigurator/sdk/backends/sglang_backend.py
@@ -37,3 +37,8 @@ class SGLANGBackend(BaseBackend):
     def __init__(self):
         super().__init__()
         self.name = common.BackendName.sglang
+
+    def _tpot_mix_steps(self, num_mix_steps: int) -> int:
+        # Same pipeline-drain correction as TRT-LLM: ~3 steps elapse before
+        # new requests can be enqueued after the last prefill finishes.
+        return max(1, num_mix_steps - 3)
--- a/src/aiconfigurator/sdk/backends/trtllm_backend.py
+++ b/src/aiconfigurator/sdk/backends/trtllm_backend.py
@@ -65,6 +65,12 @@ class TRTLLMBackend(BaseBackend):
             return getattr(model, "_hidden_size", h)
         return h
 
+    def _tpot_mix_steps(self, num_mix_steps: int) -> int:
+        # TRT-LLM in-flight batching has a pipeline-drain latency at the
+        # context/decode boundary: ~3 steps elapse before new requests can be
+        # enqueued after the last prefill finishes. Empirical correction.
+        return max(1, num_mix_steps - 3)
+
     def get_default_free_gpu_memory_fraction(self) -> float | None:
         return TRTLLM_DEFAULT_FREE_GPU_MEMORY_FRACTION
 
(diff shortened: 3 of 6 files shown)


src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -83,6 +83,26 @@ def _moe_workspace_width(self, model: BaseModel, model_family: str, h: int) -> i
         """
         return getattr(model, "_hidden_size", h)
 
+    def _mix_step_efficiency(self, ctx_tokens: int, gen_tokens: int) -> float:
```
jasonqinzhou, line 103: This lands the fitted correction in BaseBackend, so it changes vLLM, TRT-LLM, and SGLang aggregate estimates. The evidence in the PR is vLLM/H200-oriented, and the coefficients look empirical enough that I would prefer either moving this into a VLLMBackend override first or adding validation showing the same coefficients hold for TRT-LLM and SGLang. I do not have a safe one-click suggestion here because the narrower fix would require moving the hook implementation into the backend-specific file after rebasing onto #1147.

src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -83,6 +83,26 @@ def _moe_workspace_width(self, model: BaseModel, model_family: str, h: int) -> i
...
+        empirical (ISL, b) measurements, where gen_frac = gen_tokens /
+        (ctx_tokens + gen_tokens). Mix step and genonly step latencies were
+        isolated by solving the TPOT equation at two OSL values. K=1.238,
+        alpha=0.157 minimise mean absolute TPOT error (3.72%) across the
+        measurement grid. See GitHub issue #1150.
+        """
+        total = ctx_tokens + gen_tokens
+        if total == 0 or gen_tokens == 0:
+            return 1.0
+        return 1.238 * ((gen_tokens / total) ** 0.157)
```
jasonqinzhou: The fitted curve exceeds 1.0 once gen_tokens / (ctx_tokens + gen_tokens) is roughly above 0.26, which means the correction can increase mixed-step latency even though the linked issue describes systematic over-estimation. If values above 1 are intentional, please document and validate that regime; otherwise this should be bounded to keep the correction from turning into a penalty outside the measured long-prefill cases.

```suggestion
        return min(1.0, 1.238 * ((gen_tokens / total) ** 0.157))
```

src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -965,6 +985,9 @@ def run_agg(
         mix_step_latency_ms, mix_step_energy_wms, mix_per_ops, mix_per_ops_src = self._get_mix_step_latency(
             model, database, runtime_config, num_mix_ctx_tokens, num_mix_gen_tokens, isl, osl, prefix
         )
+        mix_efficiency = self._mix_step_efficiency(num_mix_ctx_tokens, num_mix_gen_tokens)
+        mix_step_latency_ms *= mix_efficiency
+        mix_step_energy_wms *= mix_efficiency
         per_ops_data["mix_step"] = mix_per_ops
```
jasonqinzhou, line 1183: After scaling mix_step_latency_ms, per_ops_data["mix_step"] still stores the unscaled per-op breakdown. The CLI detail report sums that dict directly, so users can see a Mix Step total that does not match the latency used for TTFT/TPOT. Scaling the breakdown keeps the report reconciled with the adjusted latency.

```suggestion
        if mix_efficiency != 1.0:
            mix_per_ops = {op: latency * mix_efficiency for op, latency in mix_per_ops.items()}
        per_ops_data["mix_step"] = mix_per_ops
```

jasonqinzhou (commented):
Thanks for digging into the mixed-step modeling issue. I think the direction is plausible, but I would not merge this branch as-is yet. It is currently dirty against main because #1147 added _mix_step_gen_tokens in the same hook area, so please rebase and make sure the final code includes both the corrected mix-step token count and this efficiency model. I would also like to see a small regression test that computes expected TPOT with a fake backend, so the efficiency factor and the removal of the old num_mix_steps - 3 correction are pinned independently of the accuracy-regression job.

src/aiconfigurator/sdk/backends/base_backend.py
natoscott: Done; _mix_step_efficiency now lives in VLLMBackend with the fitted formula and is capped at 1.0.  BaseBackend defines the hook with a default of 1.0 (no correction), so TRT-LLM and SGLang are unaffected.

src/aiconfigurator/sdk/backends/base_backend.py
natoscott: Agreed - great point; added min(1.0, ...) to the VLLMBackend override.

src/aiconfigurator/sdk/backends/base_backend.py
natoscott: Done.  I made it unconditional for simplicity (multiply by 1 is a no-op), but no problem adding that back if there's a preference that way?

natoscott:
The PR has been updated to add a companion hook to the previously introduced `_mix_step_efficiency`: `_tpot_mix_steps(num_mix_steps)`, used in the mixed-step branch of `run_agg` for the TPOT calculation.

**Background:** when `_mix_step_efficiency` was introduced and the hard-coded `max(1, num_mix_steps - 3)` offset was removed, that offset was an empirical pipeline-drain correction for engines (TRT-LLM, SGLang) where ~3 steps elapse at the context/decode boundary before new requests can be enqueued after the last prefill finishes. Removing it globally lost that correction for both TRT-LLM and SGLang. The new hook restores it per-engine:

- `BaseBackend`: returns `num_mix_steps` (no correction — default)
- `TRTLLMBackend`: returns `max(1, num_mix_steps - 3)`
- `SGLANGBackend`: returns `max(1, num_mix_steps - 3)` (same pipeline behaviour)
- `VLLMBackend`: inherits default (no correction needed for
  `max_num_partial_prefills=1` scheduling)

Tests in `test_agg_hooks.py` are updated to cover the new hook and reflect that all backends now return 1.0 from `_mix_step_efficiency` (VLLMBackend explicitly, others by inheritance).

**Full-corpus gate results (this PR standalone):**

| Partition  | TTFT     | TPOT     | Gate |
|------------|----------|----------|------|
| all        | −1.49%   | −0.42%   | ✓✓   |
| vllm-agg   | −2.97%   | −0.83%   | ✓✓   |
| Kimi-K2.5  | +52.05%  | +2.12%   | ✓✓   |

`all` and `agg` pass both thresholds. The Kimi-K2.5 TTFT improvement (+52%) is from the `_mix_step_gen_tokens` change in #1147 base interacting with this PR's TPOT correction.

tianhaox:
closing. #1166 has been merged.

PR #1152, opened 2026-05-27
feat: add FPM forward pass perf model with online tuning

tedzhouhk:
## Summary
- add `ForwardPassPerfModel` as an additive Rust API for native AIC estimates, fallback regression, and online FPM tuning/correction
- expose the new model through the C ABI and Python `RustForwardPassPerfModel` ctypes wrapper
- add nested `tune_with_fpms` support, diagnostics, min/max/avg correction factor getters, and fixed correction-grid bounds
- keep existing `EngineStepEstimator` and legacy Python ctypes loading backward compatible

## Linear Coverage
- Fully resolves [AIC-1122](https://linear.app/nvidia/issue/AIC-1122/simple-online-fine-tuning-solution-from-planner-to-aic): this PR moves the simple online FPM fine-tuning/correction path from Planner-style logic into AIC's Rust API surface.
- Does not fully resolve [AIC-1123](https://linear.app/nvidia/issue/AIC-1123/move-self-benchmark-mode-from-planner-to-aic): this PR can consume benchmark/observed FPMs through `tune_with_fpms`, but it does not move Planner self-benchmark generation/orchestration mode into AIC.

## FPM Workload Semantics
- `estimate_forward_pass_time_ms(metrics_by_rank)` estimates one forward-pass iteration; the input is one FPM per attention-DP rank
- estimation treats FPM as a workload descriptor: scheduled request fields are used, while `wall_time` and queued request fields are ignored
- `tune_with_fpms(iterations)` accepts multiple observed iterations; outer list is iterations, inner list is per-attention-DP-rank FPMs for that iteration
- tuning treats FPM as observed telemetry: scheduled request fields are workload features and positive `wall_time` is the latency target
- workload kind is inferred from scheduled requests only:
  - prefill: `[sum_prefill_tokens]`
  - decode: `[num_decode_requests, sum_decode_kv_tokens]`
  - mixed/agg: `[sum_prefill_tokens, sum_decode_kv_tokens]`
- multi-rank tuning records one observation using the max-load rank feature vector and max positive `wall_time`

## Native Correction Semantics
- native AIC estimates are available immediately; online tuning applies a multiplicative correction factor on top of the native estimate
- correction grids use fixed constructor-time ranges from `ForwardPassPerfOptions`:
  - `sum_prefill_tokens`: `0..max_num_tokens`, default `8192`
  - `num_decode_requests`: `0..max_batch_size`, default `512`
  - `sum_decode_kv_tokens`: `0..max_kv_tokens`, default `2000000`
- `min_observations` is checked per inferred workload kind, matching planner regression behavior, not per correction region
- correction regions default to factor `1.0`
- once a workload kind has enough total retained observations, populated in-range regions use their own median `observed_ms / native_ms` correction factor
- estimates outside configured correction bounds use factor `1.0`; native-correction tuning observations outside configured bounds are ignored

## Fallback Regression Semantics
- `best_available(config, options)` uses native AIC when supported and falls back to regression only for unsupported native model/data cases
- regression-only mode returns `None` for non-empty estimates until the inferred workload kind has enough retained observations
- bucketed storage is used for sample retirement and coverage, while regression fitting uses all retained samples for the workload kind

## Testing
- `cargo test --manifest-path rust/aiconfigurator-core/Cargo.toml`
- `cargo build --release --manifest-path rust/aiconfigurator-core/Cargo.toml`
- `uv run --extra dev pytest tests/unit/sdk/test_rust_engine_step.py`
- `uv run --extra dev ruff format --check src/aiconfigurator/sdk/rust_engine_step.py tests/unit/sdk/test_rust_engine_step.py`
- `uv run --extra dev ruff check src/aiconfigurator/sdk/rust_engine_step.py tests/unit/sdk/test_rust_engine_step.py`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added a forward-pass performance model (native, best-available native→regression fallback, and regression-only), with optional ms estimates, tuning from flexible iteration inputs, diagnostics JSON, min/max/avg correction-factor queries, and explicit model lifecycle handling in the SDK.
* **Tests**
  * Added unit and integration tests covering fallback behavior, tuning across workloads/ranks, diagnostics, correction-factor reporting, SDK wrapper wiring, and compatibility with older native libraries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiconfigurator/pull/1152?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


rust/aiconfigurator-core/src/fpm_perf.rs
```diff
@@ -0,0 +1,848 @@
...
+                    f64::from(max_decode_requests),
+                    f64::from(max_decode_kv_tokens),
+                ],
+            },
+            (true, true) => Self {
+                shape: FeatureShape::Mixed,
+                x: vec![
+                    f64::from(max_prefill_tokens),
+                    f64::from(max_decode_kv_tokens),
+                ],
```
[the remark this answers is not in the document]
tedzhouhk: Good catch as a modeling limitation. This PR intentionally keeps mixed/agg features planner-compatible as `[sum_prefill_tokens, sum_decode_kv_tokens]`, matching Dynamo current `AggRegressionModel`. Extending mixed to include `num_decode_requests` would change the learned feature contract from 2D to 3D, affect bucket sizing/retirement semantics, tests, Python docs, and downstream expectations. I am leaving this as a follow-up modeling decision rather than changing it in this compatibility-focused PR.

rust/aiconfigurator-core/src/fpm_perf.rs
```diff
@@ -0,0 +1,1092 @@
...
+///   `[sum_prefill_tokens, sum_decode_kv_tokens]`
+/// - empty: no scheduled prefill or decode work, estimates `0.0` and is not
+///   used for tuning
+///
+/// Queued request fields are accepted for FPM schema parity but ignored by this
+/// forward-pass-level model. For attention-DP configurations, the input for one
+/// iteration is one FPM per attention-DP rank; tuning merges that list into one
+/// observation by taking max-rank load features and max nonzero `wall_time`.
+#[derive(Clone, Debug)]
+pub struct ForwardPassPerfModel {
```
jasonqinzhou, line 146: Shall we define PerfModel?

jasonqinzhou:
I like the direction here: the layering is clean, and the Rust core / FFI / Python wrapper split is easy to follow. The main thing that still feels hard as a reader is the vocabulary around what this model is serving.

A few suggestions that would make the PR easier to understand:

- `FeatureShape` feels overloaded. Since this is really prefill/decode/mixed workload classification rather than tensor shape, I think `WorkloadKind`, `ForwardPassKind`, or `RequestMix` would read better.
- `ShapeStores` might be clearer as `WorkloadStores` if the enum above is renamed.
- `Native` / `Regression` are a little muddy. The first is really “existing estimator plus optional online correction”; the second is “fit directly from observed FPM wall-time samples.” Names like `EstimatorCorrected` / `ObservedFit`, or `CalibratedEstimator` / `EmpiricalFit`, would make the mental model easier.
- `auto()` is a constructor/factory, not really a mode. It tries the estimator first and falls back to observed-fit only for unsupported model/data cases. `with_fallback()` or `best_available()` may communicate that better.
- The docs should more explicitly separate “FPM as workload descriptor” from “FPM as observed telemetry.” `ForwardPassMetrics` is observed telemetry in Dynamo, but `estimate_forward_pass_time_ms()` uses only the scheduled workload fields and ignores `wall_time`; `tune_with_fpms()` uses `wall_time` as the observation target.

One behavioral thing I think is worth tightening: `retire_from_fattest_bucket()` removes samples but does not recompute `axis_min` / `axis_max` or rebuild buckets after eviction. If the retired sample was a min/max point, later correction lookup can still use stale lifetime bounds, which makes the “out-of-range queries default to 1.0” semantics less exact over time. I’d either recompute bounds from retained samples after eviction or document that bounds are lifetime bounds rather than retained-sample bounds.

rust/aiconfigurator-core/src/fpm_perf.rs
jasonqinzhou: ok I think FPPF is a fixed concept. don't need parent class.

tedzhouhk:
thanks for the review! 

> * `FeatureShape` feels overloaded. Since this is really prefill/decode/mixed workload classification rather than tensor shape, I think `WorkloadKind`, `ForwardPassKind`, or `RequestMix` would read better.

updated 

> * `ShapeStores` might be clearer as `WorkloadStores` if the enum above is renamed.

updated

> * `Native` / `Regression` are a little muddy. The first is really “existing estimator plus optional online correction”; the second is “fit directly from observed FPM wall-time samples.” Names like `EstimatorCorrected` / `ObservedFit`, or `CalibratedEstimator` / `EmpiricalFit`, would make the mental model easier.

I feel correction is a shared layer on top of native and regression model so I would prefer the current naming more. 

> * `auto()` is a constructor/factory, not really a mode. It tries the estimator first and falls back to observed-fit only for unsupported model/data cases. `with_fallback()` or `best_available()` may communicate that better.

updated

> * The docs should more explicitly separate “FPM as workload descriptor” from “FPM as observed telemetry.” `ForwardPassMetrics` is observed telemetry in Dynamo, but `estimate_forward_pass_time_ms()` uses only the scheduled workload fields and ignores `wall_time`; `tune_with_fpms()` uses `wall_time` as the observation target.

updated

> One behavioral thing I think is worth tightening: `retire_from_fattest_bucket()` removes samples but does not recompute `axis_min` / `axis_max` or rebuild buckets after eviction. If the retired sample was a min/max point, later correction lookup can still use stale lifetime bounds, which makes the “out-of-range queries default to 1.0” semantics less exact over time. I’d either recompute bounds from retained samples after eviction or document that bounds are lifetime bounds rather than retained-sample bounds.

Good catch. The bounds should be fixed (by bs, mnt, and kv_tokens). Updated.

jasonqinzhou (approved):

PR #1164, opened 2026-05-29
feat: add decode overhead and throughput cap hooks to agg pipeline

natoscott:
## Summary

Introduces two overridable hooks to `BaseBackend` that let backend subclasses correct agg predictions for real-world engine behaviour not captured in silicon benchmarks:

- **`_decode_overhead_factor(b)`** — multiplicative correction on genonly step latency. Silicon data measures isolated kernel time; in practice engines carry scheduling, KV cache management and CUDA graph dispatch overhead that grows with batch size. Default: 1.0. `VLLMBackend` overrides with `1 + (b-1)^0.662 * 0.084`, calibrated on H200 SXM vLLM 0.18.0 at ISL=9000, Qwen3-8B, b=1..62.

- **`_throughput_cap(step_throughput, ttft, tpot, b, osl)`** — clamps the step-latency-derived throughput to a Little's Law upper bound. Prevents recommending operating points that cannot be sustained in steady state. Default: no cap. `VLLMBackend` overrides to apply the cap.

TRT-LLM and SGLang have the same hook paths available.

## Dependencies

Assumes [PR #1151](https://github.com/ai-dynamo/aiconfigurator/pull/1151) (`fix/mix-step-efficiency`) is merged first, as the decode overhead constant was calibrated on the corrected mix-step efficiency model.

## Test plan

- [ ] Unit tests pass (887 passing, 1 pre-existing failure in test_moe_dispatch unrelated to this change)
- [ ] `ruff format` and `ruff check` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added per-backend controls for mixed-step efficiency, TPOT mix-step adjustment, and throughput capping.
  * Throughput planning now applies an engine-specific cap based on steady-state request latency.

* **Improvements**
  * Mixed-step totals and per-operation latency breakdowns are scaled by configurable efficiency.
  * Pipeline-drain TPOT/mix-step corrections are applied for select backends for more realistic modeling.
  * Output throughput estimation was refactored for improved accuracy.

* **Tests**
  * Added unit coverage for default hook behavior and cross-backend override differences.

PR #1128, opened 2026-05-19
perf: TTFT queuing correction

Spycsh:
#### Overview:

Draft the TTFT correction with queuing modeling instead of the original implementation. It reduces some outliers when calculating MAPE in our experiments. However, pls keep this PR as a draft for now because it still have some gap when some empirical correction is a must. Distribution of requests in real scenarios is also another tricky issue to handle.

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)

- closes GitHub issue: #xxx

## Release Notes

* **New Features**
  * Added CLI options `--ttft-queue-model` and `--ttft-request-interval-ms` to configure time-to-first-token predictions
  * Support for multiple queueing models (default, M/D/1, D/D/1) to improve TTFT estimation accuracy

* **Tests**
  * Added unit tests for time-to-first-token queueing behavior across different model configurations

PR #1165, opened 2026-05-29
fix: replace previous TTFT correction with queuing model hook

natoscott:
## Summary

Replaces the previous TTFT heuristic (`min(2 + (steps-3)/2/10, 4)` capped at 4×) with an overridable `_ttft_queuing_factor(b, steps_to_finish_ctx)` hook in `BaseBackend`. The default preserves the previous formula so all existing backends are unaffected.

`VLLMBackend` overrides with the `(b+1)/2` simultaneous-arrival model: vLLM v1 processes one request's context phase at a time via `max_num_partial_prefills=1`. With `b` requests arriving simultaneously, request `k` waits for `k-1` prefills to complete. Average TTFT = `(b+1)/2 * per-request prefill time`.

The pure prefill step latency is used (efficiency reduction from `_mix_step_efficiency` undone before TTFT calculation), since decode tokens sharing a forward pass do not reduce the prefill cost.

**Validated on H200 SXM, ISL=9000, OSL=30, Qwen3-8B across b=1..62:**
- TTFT error b=1..8 (SLA-relevant range): +5% to +11%
- Throughput error b=1..32: within 2%
- Above b=16 TTFT diverges as the (b+1)/2 model over-predicts near saturation, but at those operating points TTFT SLAs are already violated so the prediction is not actionable.

## Dependencies

- [PR #1151](https://github.com/ai-dynamo/aiconfigurator/pull/1151) (`fix/mix-step-efficiency`) — mix-step efficiency model required for correct prefill step latency
- [PR #1164](https://github.com/ai-dynamo/aiconfigurator/pull/1164) (`fix/agg-throughput-hooks`) — decode overhead and throughput cap hooks

## Test plan

- [ ] Unit tests pass (887 passing)
- [ ] `ruff format` and `ruff check` clean

## Summary by CodeRabbit

* **New Features**
  * Added backend-specific tuning for mixed-step efficiency, TPOT mix-step adjustment, TTFT queuing scaling, and throughput capping.

* **Improvements**
  * Refined mixed-step latency/energy modeling using the new tuning hooks.
  * Updated TTFT and throughput estimates with backend-specific queuing and Little’s-Law-based caps to improve downstream rate/latency calculations.

* **Tests**
  * Added unit tests validating hook defaults and per-backend behavior for mixed-step efficiency, TPOT adjustments, TTFT queuing, and throughput cap enforcement.

### DIFF
--- a/src/aiconfigurator/sdk/backends/base_backend.py
+++ b/src/aiconfigurator/sdk/backends/base_backend.py
@@ -122,6 +122,16 @@ class BaseBackend:
         """
         return num_mix_steps
 
+    def _throughput_cap(self, step_throughput: float, ttft: float, tpot: float, b: int, osl: int) -> float:
+        """Return the effective output throughput after any engine-specific cap.
+
+        Default: returns step_throughput unchanged. Subclasses may override to
+        apply a tighter constraint — e.g. a Little's Law cap that prevents the
+        model from recommending operating points that cannot be sustained in
+        steady state given the predicted request latency.
+        """
+        return step_throughput
+
     def _resolve_agg_kwargs(self, kwargs: dict, isl: int, osl: int) -> dict:
         """Resolve backend-specific run_agg kwargs to defaults.
 
@@ -1213,9 +1223,10 @@ class BaseBackend:
         _total_step_latency_ms = (
             encoder_latency_ms + num_mix_steps * mix_step_latency_ms + num_genonly_steps * genonly_step_latency_ms
         )
-        output_throughput = (
+        _step_throughput = (
             (1000 / _total_step_latency_ms * b * (osl - 1)) if (osl > 1 and _total_step_latency_ms > 0) else 0.0
         )
+        output_throughput = self._throughput_cap(_step_throughput, ttft, tpot, b, osl)
         logger.debug(
             f"ctx_tokens: {ctx_tokens}, b: {b}, osl: {osl}, isl: {isl}, "
             f"num_mix_steps: {num_mix_steps}, num_genonly_steps: {num_genonly_steps}, "
--- a/src/aiconfigurator/sdk/backends/base_backend.py
+++ b/src/aiconfigurator/sdk/backends/base_backend.py
@@ -100,6 +100,28 @@ class BaseBackend:
             return max(1, int(b // (steps_to_finish_ctx / osl)))
         return max(1, b - int(np.ceil(ctx_tokens / isl)))
 
+    def _mix_step_efficiency(self, ctx_tokens: int, gen_tokens: int) -> float:
+        """GPU batching efficiency factor for a mixed prefill/decode forward pass.
+
+        Per-op silicon data measures each operation in isolation, overstating the
+        marginal cost of prefill tokens when they share a forward pass with decode
+        tokens. Weight matrices are loaded once from HBM for the combined batch.
+        Default: 1.0 (no correction — preserves existing behaviour for backends
+        without empirical efficiency data).
+        """
+        return 1.0
+
+    def _tpot_mix_steps(self, num_mix_steps: int) -> int:
+        """Return the effective mix-step count for TPOT calculation.
+
+        Engines with pipeline-drain latency at the context/decode boundary
+        (requests cannot be immediately enqueued after prefill finishes) may
+        reduce the effective step count to account for that bubble. Default:
+        use the full mix step count. Subclasses should override with an
+        empirically calibrated correction.
+        """
+        return num_mix_steps
+
     def _resolve_agg_kwargs(self, kwargs: dict, isl: int, osl: int) -> dict:
         """Resolve backend-specific run_agg kwargs to defaults.
 
@@ -1133,12 +1155,10 @@ class BaseBackend:
                 num_genonly_tokens = 0
                 num_mix_steps_for_tpot_calc = num_mix_steps
             else:
-                # 3-step is an empirical correction for pipelining requests where new requests
-                # cannot be enqueued immediately after last request's exit
                 num_mix_steps = steps_to_finish_ctx
                 num_genonly_steps = osl - num_mix_steps
                 num_genonly_tokens = b
-                num_mix_steps_for_tpot_calc = max(1, num_mix_steps - 3)
+                num_mix_steps_for_tpot_calc = self._tpot_mix_steps(num_mix_steps)
         elif b == 1:
             # special case for b=1
             num_mix_steps = 1
@@ -1155,6 +1175,11 @@ class BaseBackend:
         mix_step_latency_ms, mix_step_energy_wms, mix_per_ops, mix_per_ops_src = self._get_mix_step_latency(
             model, database, runtime_config, num_mix_ctx_tokens, num_mix_gen_tokens, isl, osl, prefix
         )
+        mix_efficiency = self._mix_step_efficiency(num_mix_ctx_tokens, num_mix_gen_tokens)
+        mix_step_latency_ms *= mix_efficiency
+        mix_step_energy_wms *= mix_efficiency
+        if mix_efficiency != 1.0:
+            mix_per_ops = {op: v * mix_efficiency for op, v in mix_per_ops.items()}
         per_ops_data["mix_step"] = mix_per_ops
         per_ops_source["mix_step"] = mix_per_ops_src
 
--- a/src/aiconfigurator/sdk/backends/base_backend.py
+++ b/src/aiconfigurator/sdk/backends/base_backend.py
@@ -122,6 +122,17 @@ class BaseBackend:
         """
         return num_mix_steps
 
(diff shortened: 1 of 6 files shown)


jasonqinzhou (commented):
Thanks for working through the TTFT correction. I don't think this is ready to merge yet: the accuracy workflow is green only because the regression gate is allow-to-fail. I downloaded the posted artifact and ran the threshold test against it; it fails with `all` TTFT regression `192.128148%`, `all` TPOT regression `20.164597%`, `h200_sxm` TTFT regression `313.223050%`, and `vllm-agg` TTFT regression `329.471744%`. Also, the PR is currently dirty/not rebaseable against `main` now that #1147 has landed, so please rebase on current `main`, scope the heuristic to the validated regime, and rerun the accuracy artifact before we evaluate the numbers again.

src/aiconfigurator/sdk/backends/vllm_backend.py
```diff
@@ -35,3 +35,32 @@ class VLLMBackend(BaseBackend):
...
         super().__init__()
         self.name = common.BackendName.vllm
+
+    def _ttft_queuing_factor(self, b: int, steps_to_finish_ctx: float) -> float:
+        # vLLM v1 processes one request's context phase at a time
+        # (max_num_partial_prefills=1). With b requests arriving simultaneously,
+        # request k waits for k-1 prefills to complete before its first token.
+        # Average TTFT across all b requests = (b+1)/2 * per-request prefill time.
+        # Validated on H200 SXM ISL=9000: 2% error at b=4, 4% error at b=8.
+        return (b + 1) / 2
```
jasonqinzhou: Can we scope this heuristic to the regime it was validated for instead of applying it to every vLLM agg prediction? The `(b + 1) / 2` simultaneous-arrival model makes sense when a batch arrives together and vLLM processes one partial prefill at a time, but the accuracy corpus includes H200 vLLM agg rows with larger `b` plus different models/ISL/OSL combinations. Those rows get very large TTFT multipliers and dominate the regression artifact. Since we plan to plug Dynamo Mocker into AIC for a more faithful scheduler/TTFT model, I think this should be gated/clamped to the validated/SLA-relevant slice, or made opt-in through scenario config, instead of becoming the global vLLM TTFT correction.

src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -965,6 +1016,9 @@ def run_agg(
         mix_step_latency_ms, mix_step_energy_wms, mix_per_ops, mix_per_ops_src = self._get_mix_step_latency(
             model, database, runtime_config, num_mix_ctx_tokens, num_mix_gen_tokens, isl, osl, prefix
         )
+        mix_efficiency = self._mix_step_efficiency(num_mix_ctx_tokens, num_mix_gen_tokens)
+        mix_step_latency_ms *= mix_efficiency
+        mix_step_energy_wms *= mix_efficiency
         per_ops_data["mix_step"] = mix_per_ops
```
jasonqinzhou, line 1204: Can we scale the per-op breakdowns when applying these step-level corrections? `mix_step_latency_ms` and `genonly_step_latency_ms` are now multiplied by `mix_efficiency` / `decode_overhead`, but `per_ops_data["mix_step"]` and `per_ops_data["genonly_step"]` are stored unscaled, so the detail report can disagree with the top-level TTFT/TPOT and scheduling latencies. A small fix would be to multiply the per-op dict values by the same factor before storing them, or add an explicit correction row/factor so the totals reconcile.

src/aiconfigurator/sdk/backends/vllm_backend.py
natoscott: I've reworked this (see full commentary below).

src/aiconfigurator/sdk/backends/base_backend.py
natoscott: Good points - resolved now, thanks.

natoscott:
@jasonqinzhou Thank you for review and assessments - these made it much easier to pinpoint the root causes.

**Rebase:** Done on current main via the updated #1164 base.

**On the per-op breakdown inconsistency (your inline suggestion on base_backend.py line 1022):**
  
Fixed in #1164 (same approach). 

**Why `(b+1)/2` caused the 192–329% TTFT regression:**

`(b+1)/2` is the exact formula for a single burst of b simultaneous arrivals.  The corpus measures steady-state TTFT. In steady state growth is sub-linear - the corpus median TTFT/T_prefill at b=64 is 2.09×, not 32.5×.

**On "scope the heuristic to the validated regime":**

The validated regime is vLLM v1 with `max_num_partial_prefills=1`. Verified against the full corpus: once entries are normalised within matching (model, system, ISL, OSL, tp_size) groups the vLLM data is fully consistent with serial-prefill queuing theory. TRT-LLM's profile (0.17× T_prefill at b=4, from in-flight batching) is modelled via the ctx_tokens sweep and is unaffected.

**The new formula: `min(1 + log₂(b)/8, 2.0)`:**

    b=1: 1.00×   b=8: 1.375×   b=32: 1.625×   b=128: 1.875×
    b=4: 1.25×   b=16: 1.50×   b=64: 1.75×    b=256: 2.00× (cap = 2T)

MAPE on tp_size-matched vLLM corpus:

  Old code (TTFT = T_prefill):  26.4%
  (b+1)/2 - original formula:  199.8%
  1 + log₂(b)/8 - new formula:  18.0%  (improvement over baseline)

A principled M/D/1 treatment requiring T_decode as a hook input is the natural follow-on (also see draft PR #1128 from @Spycsh which explores this direction) and is noted in the code comment.

**Standalone gate results:**

| Partition     | TTFT     | TPOT     | Gate |
|---------------|----------|----------|------|
| all           | −2.77%   | +0.06%   | ✓✓   |
| vllm-agg      | −5.50%   | +0.12%   | ✓✓   |
| Kimi-K2.5     | +54.47%  | −2.74%   | ✓✓   |
| MiniMax|h100  | +70.09%  | −1.16%  | ✓✓   |
| Llama-70B|h100| +27.65%  | +2.08%  | ✓✓   |
| h200_sxm      | −11.35%  | +0.29%   | ✗✓   |

`all` passes both thresholds. Remaining TTFT leaf failures (MiniMax −15.92%, h200_sxm −11.35%) are in ISL=1024, long-OSL configurations where T_prefill is short relative to T_decode - the system is lightly loaded and TTFT barely grows with b. The formula adds queuing overhead not present in those entries; this is the ISL/OSL-dependence a full M/D/1 model would handle.

Note: in the full stack including #1166 dispatch overhead, nearly all leaf failures resolve (see #1166 for combined numbers).

natoscott:
> scope the heuristic to the validated regime

@jasonqinzhou @tianhaox while I have the data handy, I wanted to share detailed results of comparisons across various TTFT formula that have been discussed, which led to selecting the log₂ formula currently in this PR.

## TTFT queuing formula comparison

Analysis against the full accuracy corpus (tp_size-matched vLLM agg entries, 99 configurations with `b=1` anchors for T_prefill normalisation).

### Overall MAPE

| Formula | Overall MAPE | vs no-queuing baseline |
|---|---|---|
| Old code — `TTFT = T_prefill`, no queuing | 26.4% | baseline |
| `(b+1)/2` — original formula in this PR | 199.8% | +173 pp |
| `min((b+1)/2, 2.0)` — cap at 2T | 46.1% | +20 pp |
| **`1 + log₂(b)/8`** — updated formula | **18.0%** | **−8 pp ✓** |

`(b+1)/2` is the exact formula for a single burst of simultaneous arrivals and is correct in that regime, but the corpus measures steady-state TTFT where growth is sub-linear. At b=64 the formula predicts 32.5× T_prefill; the corpus median is 2.09×.

Capping at 2T fixes the large-b over-prediction but introduces a sharp discontinuity at b=3 and still regresses vs the no-queuing baseline overall.

`1 + log₂(b)/8` (= log₂₅₆(b) + 1, saturating at 2T at b=256) is the only formula that **improves** on the original no-queuing baseline across all batch sizes, achieving 18.0% MAPE vs 26.4%.

### Per batch size

| b | N | old (no queuing) | `(b+1)/2` | cap 2T | `log₂(b)/8` |
|---|---|---|---|---|---|
| 1 | 9 | 0.0% | 0.0% | 0.0% | 0.0% |
| 2 | 9 | 6.4% | 40.4% | 40.4% | 6.3% |
| 4 | 20 | 19.1% | 104.5% | 63.6% | 24.5% |
| 8 | 20 | 19.5% | 267.1% | 64.1% | 22.2% |
| 16 | 20 | 42.7% | 421.9% | 44.2% | 23.0% |
| 32 | 11 | 38.0% | 923.6% | 33.3% | 27.5% |
| 64 | 10 | 52.1% | 1458.3% | 36.2% | 24.9% |

At small b (b=2, b=8) `log₂(b)/8` stays close to the old code.
At large b (b=16+) it substantially reduces error where the old code was most wrong and `(b+1)/2` diverges most severely.

tianhaox:
closing. #1166 has been merged.

PR #1166, opened 2026-05-29
fix: add prefill dispatch overhead hook to T_prefill

natoscott:
## Summary

Silicon benchmarks measure isolated GPU kernel time. In practice vLLM adds CPU-side Python dispatch cost per prefill request that scales with transformer layer count and is not captured in per-kernel measurements.

Adds `_prefill_dispatch_overhead_ms(model)` hook to `BaseBackend` (default 0.0, preserving existing behaviour for all backends). The model object is provided so subclasses can factor in architecture properties beyond layer count. `VLLMBackend` overrides with `model._num_layers * 1.4ms`. TRT-LLM and SGLang have the same hook available but no default change.

The overhead is added to T_prefill before the queuing factor so that per-request service time (and hence queue wait) reflects the true cost.

**Calibrated at ~1.4ms/layer on H200 SXM vLLM 0.18.0 at ISL=9000, b~1:**

| Model | Layers | Gap | After correction |
|---|---|---|---|
| Qwen3-0.6B | 28 | +41ms | −2% |
| Qwen3-8B | 32 | +45ms | +2% |
| Qwen3-14B | 40 | +52ms | +1% |

## Dependencies

- [PR #1165](https://github.com/ai-dynamo/aiconfigurator/pull/1165) (`fix/ttft-queuing-model`) — TTFT queuing hook required for the `_ttft_per_request` call site this PR extends

## Test plan

- [ ] Unit tests pass (887 passing)
- [ ] `ruff format` and `ruff check` clean

## Summary by CodeRabbit

* **New Features**
  * Improved latency/energy estimation using efficiency-aware mixed-step scaling.
  * Refined time-to-first-token modeling with queuing factors and added per-request prefill dispatch overhead.
  * More accurate effective mix-step counting with engine-specific pipeline-drain adjustments.
  * Added an output throughput cap to better reflect real-world limits; updated related debug/metrics.

* **Tests**
  * Expanded cross-engine unit coverage for the new estimation and cap/limit behaviors, including edge cases and default values.

### DIFF
--- a/src/aiconfigurator/sdk/backends/vllm_backend.py
+++ b/src/aiconfigurator/sdk/backends/vllm_backend.py
@@ -38,6 +38,17 @@ class VLLMBackend(BaseBackend):
         super().__init__()
         self.name = common.BackendName.vllm
 
+    def _mix_step_efficiency(self, ctx_tokens: int, gen_tokens: int) -> float:
+        # vLLM v1 serialises prefill (max_num_partial_prefills=1): each mix step
+        # processes one request's full ISL alongside a handful of decode tokens
+        # from other requests. With gen_frac = (b-1)/ISL ≈ 0.001 at typical
+        # operating points, the base-class power-law formula extrapolates to
+        # ~0.19 — an 80% reduction with no physical basis. Full-corpus analysis
+        # (1928 vLLM agg entries) shows median implied efficiency of 1.115,
+        # confirming the base-class formula is inapplicable to this regime.
+        # Return 1.0: no correction applied for this backend.
+        return 1.0
+
     def _mix_step_gen_tokens(self, b: int, ctx_tokens: int, isl: int, osl: int) -> int:
         # vLLM v1 scheduler sets max_num_partial_prefills=1 by default, meaning
         # exactly one request is in partial-prefill state per forward pass.
--- a/src/aiconfigurator/sdk/backends/vllm_backend.py
+++ b/src/aiconfigurator/sdk/backends/vllm_backend.py
@@ -58,6 +58,18 @@ class VLLMBackend(BaseBackend):
         # Source: vllm/v1/core/sched/scheduler.py, SchedulerConfig.max_num_partial_prefills
         return max(1, b - int(np.ceil(ctx_tokens / isl)))
 
+    def _ttft_queuing_factor(self, b: int, steps_to_finish_ctx: float) -> float:
+        # vLLM v1 serialises prefill (max_num_partial_prefills=1): requests queue
+        # behind the active prefill, so TTFT grows with concurrency. In steady
+        # state, growth is sub-linear — calibrated to the silicon corpus
+        # (tp_size-matched vLLM agg entries, b=1..64) as log_256(b), which
+        # improves MAPE from 26.4% (no correction) to 18.0% overall.
+        # Formula: 1 + log2(b)/8, capped at 2xT_prefill (saturates at b=256).
+        # A principled M/D/1 treatment (requiring T_decode input) is a follow-on.
+        if b <= 1:
+            return 1.0
+        return float(min(1.0 + np.log2(b) / 8.0, 2.0))
+
     def _throughput_cap(self, step_throughput: float, ttft: float, tpot: float, b: int, osl: int) -> float:
         # Cap throughput at the Little's Law limit: b concurrent requests each
         # taking (ttft + tpot*(osl-1)) ms cannot sustain more than
--- a/src/aiconfigurator/sdk/backends/vllm_backend.py
+++ b/src/aiconfigurator/sdk/backends/vllm_backend.py
@@ -57,3 +57,13 @@ class VLLMBackend(BaseBackend):
         # giving a consistent formula across both scheduling regimes.
         # Source: vllm/v1/core/sched/scheduler.py, SchedulerConfig.max_num_partial_prefills
         return max(1, b - int(np.ceil(ctx_tokens / isl)))
+
+    def _throughput_cap(self, step_throughput: float, ttft: float, tpot: float, b: int, osl: int) -> float:
+        # Cap throughput at the Little's Law limit: b concurrent requests each
+        # taking (ttft + tpot*(osl-1)) ms cannot sustain more than
+        # b*(osl-1)*1000 / request_latency_ms output tokens/s in steady state.
+        request_latency_ms = ttft + tpot * max(osl - 1, 0)
+        if request_latency_ms <= 0:
+            return step_throughput
+        ll_throughput = b * max(osl - 1, 0) * 1000.0 / request_latency_ms
+        return min(step_throughput, ll_throughput)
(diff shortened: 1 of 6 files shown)


src/aiconfigurator/sdk/backends/vllm_backend.py
```diff
@@ -35,3 +36,37 @@ class VLLMBackend(BaseBackend):
...
+
+    def _ttft_queuing_factor(self, b: int, steps_to_finish_ctx: float) -> float:
+        # vLLM v1 processes one request's context phase at a time
+        # (max_num_partial_prefills=1). With b requests arriving simultaneously,
+        # request k waits for k-1 prefills to complete before its first token.
+        # Average TTFT across all b requests = (b+1)/2 * per-request prefill time.
+        # Validated on H200 SXM ISL=9000: 2% error at b=4, 4% error at b=8.
+        return (b + 1) / 2
+
+    def _decode_overhead_factor(self, b: int) -> float:
```
tianhaox: this is approaching 3 when bs is large enough, how to interpret this behavior? for cuda-graph enabled case, the overhead will approach 0 when bs is large enough

src/aiconfigurator/sdk/backends/vllm_backend.py
```diff
@@ -35,3 +36,37 @@ class VLLMBackend(BaseBackend):
...
+        # steps carry additional cost (Python scheduling, KV cache block
+        # management, CUDA graph dispatch) that grows with batch size.
+        # correction(b) = 1 + (b-1)^0.662 * 0.084
+        # Calibrated on H200 SXM, vLLM 0.18.0, ISL=9000, Qwen3-8B, b=1..62.
+        # Needs re-validation across model sizes and hardware.
+        if b <= 1:
+            return 1.0
+        return 1.0 + ((b - 1) ** 0.662) * 0.084
+
+    def _throughput_cap(self, step_throughput: float, ttft: float, tpot: float, b: int, osl: int) -> float:
```
tianhaox, line 80: how this is impacting the final throughput, do we see capped results and how much?

src/aiconfigurator/sdk/backends/vllm_backend.py
```diff
@@ -35,3 +36,37 @@ class VLLMBackend(BaseBackend):
     def __init__(self):
         super().__init__()
         self.name = common.BackendName.vllm
+
+    def _prefill_dispatch_overhead_ms(self, model: BaseModel) -> float:
+        # CPU-side dispatch overhead scales with layer count and is not captured
+        # in silicon benchmarks. Calibrated at ~1.4ms/layer across dense models.
+        return model._num_layers * 1.4
+
+    def _ttft_queuing_factor(self, b: int, steps_to_finish_ctx: float) -> float:
```
tianhaox, line 68: the previsou correction is trying to correct the queueing time of the first batch when benchmark with concurrency based requests. i will share the doc about how we derive this. it's actually coupled with how many requests are sent.

tianhaox: [TTFT metric investigation.pdf](https://github.com/user-attachments/files/28470113/TTFT.metric.investigation.pdf)

natoscott: @tianhaox thanks for the review comments and pdf - I have a couple of unrelated deadlines to meet this week, but I will read through these and get back to you as soon as I'm able to.

tianhaox: Thanks, i'm available for teams meeting if needed

tianhaox: Looks like there're severals PRs incrementally and this is the last one, shall we just discuss over this thread?

src/aiconfigurator/sdk/backends/vllm_backend.py
natoscott: I've significantly reworked this aspect, please see full details below.

src/aiconfigurator/sdk/backends/vllm_backend.py
natoscott: I've reworked this as well - please see results summary below.

src/aiconfigurator/sdk/backends/vllm_backend.py
natoscott: Sure thing - the summary table below covers the full stack of work.

natoscott:
@tianhaox Thank you for the detailed review comments and especially for sharing the TTFT metric investigation pdf - it was super helpful for understanding the steady-state queuing behaviour under load.

The branch has been rebased on the updated stack.

**On `_decode_overhead_factor` approaching 3x at large batch sizes, and your point about CUDA graph behaviour:**

You are right on both counts. A growing overhead with b is physically wrong with CUDA graph capture, where dispatch cost is fixed per graph replay. Through this work I also found that the implied TPOT overhead varies widely by model family, so a single constant cannot improve accuracy across the full corpus without hurting specific families. I have therefore dropped `_decode_overhead_factor` from the stack entirely.

**On the throughput cap impact:**

The `_throughput_cap` applies a Little's Law upper bound:

    cap = b × (osl−1) × 1000 / (ttft + tpot × (osl−1))  [tokens/s]

For a representative case (Qwen3-8B, ISL=9000, OSL=30, H200, b=32): using corrected TTFT ≈ 800 ms and TPOT ≈ 6 ms, the cap evaluates to approximately 946 tok/s versus the raw step-throughput estimate. The cap activates at higher concurrencies where request latency grows and step-throughput becomes over-optimistic; for moderate b it leaves predictions unchanged.

**Full combined stack gate results (PRs #1151 + #1164 + #1165 + #1166):**

| Partition     | TTFT     | TPOT     | Gate |
|---------------|----------|----------|------|
| all           | +2.76%   | +0.06%   | ✓✓   |
| vllm-agg      | +5.49%   | +0.12%   | ✓✓   |
| sglang-agg    |  0.00%   | −1.08%   | ✓✓   |
| trtllm-agg    |  0.00%   | −1.39%   | ✓✓   |
| h200_sxm      | −1.16%   | +0.29%   | ✓✓   |
| h100_sxm      | +1.77%   | +0.37%   | ✓✓   |
| b200_sxm      | +7.50%   | −0.39%   | ✓✓   |
| b300_sxm      | −4.78%   | +0.13%   | ✓✓   |
| Kimi-K2.5     | +48.80%  | −2.74%   | ✓✓   |
| MiniMax-M2.5  | −0.25%   | +0.60%   | ✓✓   |
| Llama-3.1-70B | +0.44%   | −0.19%   | ✓✓   |

The full stack passes every partition on both TTFT and TPOT. `all` TTFT is +2.76% — a net improvement over the baseline. The dispatch overhead (0.8ms/layer, recalibrated against the full silicon corpus across hardware platforms and model families) compensates for the queuing formula's over-predictions at low utilisation, producing a cleaner combined result than either PR delivers in isolation.

tianhaox (approved):

PR #1199, opened 2026-06-04
ci: accelerate accuracy regression testing

jasonqinzhou:
## Summary
This PR speeds up the AIC accuracy regression workflow by avoiding repeated estimator work for identical silicon sample configurations.

## Why
The accuracy workflow runs `predict_silicon_sample.py` over the full silicon sample for both the incoming and base revisions. Many rows resolve to the same estimator arguments or the same validation/fallback outcome, so the script was repeatedly paying for duplicate `cli_estimate` calls.

## Changes
- memoize prediction results by exact estimator arguments
- memoize repeated estimator exceptions so duplicate invalid configurations are cheap
- preflight incomplete disaggregated rows before calling `cli_estimate`
- cache missing database-version fallback decisions before retrying with AIC defaults
- emit `prediction cache: <hits> hits, <misses> misses` to stderr for visibility

## Benchmark
Full-sample local benchmark on `src/aiconfigurator/systems/silicon_sample.csv` (5,221 rows), comparing the `origin/main` script against this PR's script on the same machine:

| Script | Time |
| --- | ---: |
| `origin/main` | 446.53s |
| this PR | 382.65s |

That is 63.88s saved per prediction pass, a 14.3% wall-clock reduction (1.17x faster). The accuracy workflow runs both new and old prediction passes, so the expected workflow-level savings is roughly 2m08s per run on this machine.

The PR script reported `prediction cache: 1816 hits, 3405 misses`, reducing `cli_estimate` calls by 34.8% per pass. The benchmark CSV output matched the `origin/main` output.

## Validation
- `uv run --extra dev ruff check tools/accuracy_regression_testing/predict_silicon_sample.py`
- `uv run --extra dev ruff format --check tools/accuracy_regression_testing/predict_silicon_sample.py`
- `uv run python -m py_compile tools/accuracy_regression_testing/predict_silicon_sample.py`
- GitHub Actions are green on this PR, including AIC Accuracy Regression Testing

### DIFF
--- a/tools/accuracy_regression_testing/predict_silicon_sample.py
+++ b/tools/accuracy_regression_testing/predict_silicon_sample.py
@@ -15,11 +15,18 @@ AGG_FIELDS = ("batch_size", *PARALLEL_FIELDS)
 DISAGG_FIELDS = tuple(
     f"{stage}_{field}" for stage in ("prefill", "decode") for field in (*PARALLEL_FIELDS, "batch_size", "num_workers")
 )
+REQUIRED_DISAGG_FIELDS = ("prefill_batch_size", "prefill_num_workers", "decode_batch_size", "decode_num_workers")
+EstimateArgs = dict[str, str | int | None]
+EstimateCacheKey = tuple[tuple[str, str | int | None], ...]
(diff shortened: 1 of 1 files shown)


jasonqinzhou:
@coderabbitai review

Harrilee (approved):

Issue #1159, opened 2026-05-28
Proposal: AIC Rust KV cache capacity estimation API


tedzhouhk:
# Proposal: AIC Rust KV Cache Capacity Estimation API

Status: Draft
Area: AIC / planner / mocker / runtime capacity

## Summary

Add an AIC Rust API that estimates rank-local KV cache capacity for a model, GPU, backend, parallelization mapping, and engine memory configuration.

The API should support two paths:

1. Native AIC estimation for model/backend/system combinations that AIC can parse and model.
2. A conservative naive fallback for unsupported models, unsupported backends, or incomplete AIC performance data.

The primary output is an estimated KV cache budget, expressed in bytes and tokens.

## Motivation

Dynamo already has consumers that need KV capacity before starting a real engine:

- mocker and replay need `num_gpu_blocks` when users do not provide an explicit override.
- planner and future router/runtime components need a fast estimate of how much KV capacity is available for admission, placement, and scaling.
- Python-only AIC wrappers are not enough for Rust-side consumers.

PR `ai-dynamo/dynamo#9598` added a Python wrapper in `dynamo._internal.aic.estimate_num_gpu_blocks()` for mocker/replay. That path directly uses AIC internals, but it is Python-only and returns only block count. This proposal moves the core capability into AIC Rust and returns a richer, typed capacity result.

## Goals

- Provide a Rust API with stable input/output structs.
- Estimate available KV cache for AIC-supported and AIC-unsupported cases.
- Return enough detail for consumers to convert capacity into `num_gpu_blocks` or max KV tokens.
- Keep fallback estimates conservative when native AIC memory breakdown is unavailable.
- Keep backend-specific memory knobs explicit, especially `gpu_memory_utilization`, `mem_fraction_static`, and TRT-LLM-style `free_gpu_memory_fraction`.

## Non-Goals

- Exact reproduction of every backend allocator detail.
- Full queueing or SLA modeling.
- Replacing forward-pass latency estimation.
- Guaranteeing model fit for every arbitrary HF model without enough configuration metadata.

### Request

```rust
pub struct KvCacheEstimateRequest {
    pub gpu_sku: String,
    pub hf_model_id: String,
    pub backend: Backend,
    pub engine: EngineMemoryConfig,
    pub parallel: ParallelMapping,
    pub quantization: AicQuantizationConfig,
    pub kv_fraction: f64,
    pub options: KvCacheEstimateOptions,
}

pub enum Backend {
[...]

PR #1201, opened 2026-06-04
feat(rust): estimate_kv_cache KV-cache memory API

simone-chen:
#### Overview:

Top-level KV-cache memory (capacity) estimation in the core, separate from latency prediction. Called once per session (e.g. by the Dynamo Mocker) to size `num_gpu_blocks`. The complete estimate lives in Python (`sdk/memory.py`, the single source of truth); the Rust entry point is a **pure** PyO3 forwarder with no math of its own.

> **Stacked on #1200** — review/merge that PR first; this PR's diff is only the capacity commit on top.

#### Details (file map):

```text
src/aiconfigurator/sdk/memory.py        NEW — single source of truth. Two estimators, each built then
                                          asked to .estimate(): KVCacheEstimator (native: AIC model +
                                          backend + perf DB, OfFree/OfTotal budget) and NaiveKVCacheEstimator
                                          (naive: HF-config-only, reserves a fraction of post-weight memory,
                                          standard K/V vs MLA split, fail-closed). estimate_kv_cache
                                          validates + routes native↔naive + applies the tolerance margin;
                                          estimate_num_gpu_blocks (floor helper); shared kv_cache_budget_bytes.
                                          bytes->tokens follows the model's true KV curve
                                          (get_kvcache_max_tokens), so hybrid SWA / compressed models aren't
                                          extrapolated from the seq_len=1 rate; nextn/MTP applied before the
                                          model build. engine.py untouched.

src/aiconfigurator/sdk/models/          Model.get_kvcache_max_tokens — capacity-sizing inverse of
  base.py / gemma4.py /                   get_kvcache_bytes_per_sequence. Base is exact floor-division for
  deepseek_v4.py                          linear KV (MHA/GQA/MLA); Gemma 4 (SWA) and DeepSeek-V4
                                          (window-capped + compressed) override it to invert their
                                          piecewise curve via a monotonic binary search.

rust/aiconfigurator-core/
├── src/memory.rs                       PURE PyO3 forwarder to sdk.memory; owns KvCacheEstimate* result
│                                         types. No tolerance math (moved to Python): forwards
│                                         tolerance_fraction, parses tolerance_adjusted back.
├── src/py.rs                           drops the estimate_kv_cache #[pyfunction] (redundant Python surface)
├── src/lib.rs                          re-export the memory API at the crate root
├── tests/memory_round_trip.rs          Rust round-trip: forwarder end-to-end (tolerance fwd + parse)
└── docs/phase-1.5-*.md                 design / follow-up notes (capacity-followup, execution-plan)

src/aiconfigurator_core/__init__.py     drops the estimate_kv_cache export
tests/unit/sdk/test_memory_estimation.py     pure-Python unit tests (budget formulas; standard/MLA/
                                          fail-closed per-token; naive fallback incl. unsupported arch +
                                          hybrid sliding-window; nextn applied to the model config)
tests/unit/sdk/models/test_model_config.py   get_kvcache_max_tokens: linear floor-division + DeepSeek-V4 curve
tests/unit/sdk/test_utils.py                  get_kvcache_max_tokens: Gemma 4 window-capped curve
tests/integration/test_memory_estimation.py  native path end-to-end vs a real perf DB
```

- **compat**: drops the `aiconfigurator_core.estimate_kv_cache` `#[pyfunction]` and its `__init__.py` export — in-repo Python callers use `sdk.memory` directly; the Rust fn is for embedded callers (the Mocker).

#### Rust caller usage (the Mocker):

`aiconfigurator_core::estimate_kv_cache(req)` is a plain `pub fn` (not a `#[pyfunction]`), like `build_aic_engine`. The crate links libpython; the call acquires the GIL internally and imports `aiconfigurator.sdk.memory`, so the caller just needs `aiconfigurator` importable in the embedded interpreter (no GIL handling on the caller side).

```rust
use aiconfigurator_core::{
    estimate_kv_cache, BackendKind, EngineConfig, KvCacheEstimateOptions,
    KvCacheEstimateRequest, KvCacheMemoryFraction, ParallelMapping, QuantizationConfig,
    ENGINE_CONFIG_SCHEMA_VERSION,
};

let est = estimate_kv_cache(KvCacheEstimateRequest {
    engine: EngineConfig {
        schema_version: ENGINE_CONFIG_SCHEMA_VERSION,
        model_name: "Qwen/Qwen3-32B".into(),
        system_name: "h200_sxm".into(),
        systems_path: None,                  // None -> bundled systems data
        backend: BackendKind::Trtllm,
        backend_version: Some("1.3.0rc10".into()),
        kv_block_size: None,
        parallel: ParallelMapping {
            tp_size: 2, pp_size: 1, attention_dp_size: Some(1),
            moe_tp_size: None, moe_ep_size: None,
        },
        quantization: QuantizationConfig {   // None -> inferred from the HF config
            weight_dtype: None, moe_dtype: None,
            activation_dtype: None, kv_cache_dtype: None,
        },
        speculative: None,
        extra: Default::default(),
    },
    max_num_tokens: 8192,
    max_batch_size: 256,
    // TRT-LLM -> OfFree(free_gpu_memory_fraction); vLLM/SGLang -> OfTotal.
    kv_cache_memory_fraction: KvCacheMemoryFraction::OfFree(0.9),
    gpu_memory_capacity_bytes_override: None,  // Some(bytes) for unknown SKUs / required by the naive fallback
    tolerance_fraction: Some(0.05),            // None -> raw estimate only
    options: KvCacheEstimateOptions {
        allow_naive_fallback: true,
        allow_hf_config_download: false,
        naive_kv_reservation: 0.80,            // post-weight reserve on the naive path
    },
})?;

// Result is rank-local. Convert to scheduler blocks with the scheduler's own block size,
// preferring the tolerance-adjusted token count when a tolerance was requested:
let tokens = est.tolerance_adjusted.as_ref()
    .map(|a| a.total_kv_size_tokens)
    .unwrap_or(est.total_kv_size_tokens);
let num_gpu_blocks_per_rank = tokens / scheduler_block_size;
// est.source: Native | NaiveFallback;  est.memory_breakdown: Some(..) native, None on fallback.
```

`src/aiconfigurator/sdk/memory.py` (the logic) → `rust/aiconfigurator-core/src/memory.rs` (the forwarder)

#### Related Issues:

- Relates to #1159
- Follow-up (deferred OOM-check consolidation onto `kv_cache_budget_bytes`): #1208

## Summary by CodeRabbit

* **New Features**
  * KV-cache GPU capacity estimation with native and fallback strategies
  * Tolerance-adjusted capacity and conversion to scheduler block counts
  * Detailed memory breakdown and max-sequence-length computation

* **Documentation**
  * Phase 1.5 execution plan and capacity API design docs updated

* **Tests**
  * New unit and integration tests validating estimation math and end-to-end parity

### DIFF
--- a/rust/aiconfigurator-core/src/lib.rs
+++ b/rust/aiconfigurator-core/src/lib.rs
@@ -23,6 +23,7 @@ mod common;
 mod config;
 pub mod engine;
 mod fpm;
+pub mod memory;
 mod operators;
 mod perf_database;
 mod session;
@@ -43,6 +44,13 @@ pub use fpm::{
 // `engine/runtime.rs`) keep resolving after the types moved into `fpm`.
 pub(crate) use fpm::validate_forward_pass_metrics;
 pub use fpm::{ForwardPassMetrics, QueuedRequestMetrics, ScheduledRequestMetrics, FPM_VERSION};
+// KV-cache memory API. Top-level surface, not a method on
+// `AicEngine`: estimation runs once at startup, separate from the latency path.
+pub use memory::{
+    estimate_kv_cache, EstimateSource, KvCacheEstimate, KvCacheEstimateAdjusted,
+    KvCacheEstimateError, KvCacheEstimateOptions, KvCacheEstimateRequest, KvCacheMemoryFraction,
+    MemoryBreakdown,
+};
 // PyO3 bindings. `AicEngine` is the Python -> Rust hot-path pyclass;
 // `build_aic_engine` is the Rust -> Python -> Rust embedded build entry point
 // for callers in OTHER crates (the Dynamo Mocker,
--- a/rust/aiconfigurator-core/src/memory.rs
+++ b/rust/aiconfigurator-core/src/memory.rs
@@ -1,0 +1,436 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! KV-cache memory estimation.
+//!
+//! [`estimate_kv_cache`] is a top-level crate function (NOT a method on
+//! `AicEngine`): estimation runs once at startup, uses overlapping but not
+//! identical inputs to `build_aic_engine`, and is a separate concern from
+//! latency prediction. The Dynamo Mocker is the primary external consumer; it
+//! calls this once and derives `num_gpu_blocks_per_rank` from
+//! `total_kv_size_tokens`.
+//!
+//! ## Rust is a pure forwarder; the estimate is computed in Python
+//!
+//! This mirrors how `build_aic_engine` forwards to the Python `compile_engine`:
+//! ALL of the work -- fraction + tolerance validation, HF-config parsing, the
+//! AIC backend memory model, the OfFree/OfTotal budget math, the naive heuristic
+//! fallback, AND the tolerance margin -- lives in
+//! `aiconfigurator.sdk.memory.estimate_kv_cache`. The Rust side:
+//!
+//! 1. crosses into Python once (`with_gil → import → call estimate_kv_cache →
+//!    extract dict`), forwarding `tolerance_fraction` through;
+//! 2. rebuilds a [`KvCacheEstimate`] from that dict (including
+//!    `tolerance_adjusted`), with no math of its own.
+//!
+//! The two budget formulas (TRT-LLM free-fraction vs vLLM/SGLang total-fraction),
+//! the naive fallback, and the tolerance margin all live on the Python side; see
+//! the docstring of `aiconfigurator.sdk.memory.estimate_kv_cache`. The
+//! [`KvCacheMemoryFraction`] enum still encodes the backend↔fraction XOR so the
+//! request shape is unambiguous; the variant is validated against
+//! `engine.backend` in Python.
+
+use serde::{Deserialize, Serialize};
+
+use crate::{BackendKind, EngineConfig};
+
+/// KV-cache memory request. `engine` reuses the modularised [`EngineConfig`];
+/// the remaining fields describe the runtime sizing budget.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct KvCacheEstimateRequest {
+    pub engine: EngineConfig,
+    pub max_num_tokens: u32,
+    pub max_batch_size: u32,
+    pub kv_cache_memory_fraction: KvCacheMemoryFraction,
+    /// Override for unknown SKUs; when `Some`, it wins over the SystemSpec
+    /// capacity reported by the native path.
+    pub gpu_memory_capacity_bytes_override: Option<u64>,
+    /// `None` = raw estimate only; `Some(0.05)` = 5% safety margin.
+    pub tolerance_fraction: Option<f64>,
+    pub options: KvCacheEstimateOptions,
+}
+
+/// Backend-tagged memory fraction. The variant encodes the XOR between
+/// TRT-LLM's free-fraction and vLLM/SGLang's total-fraction semantics; the
+/// Python `estimate_kv_cache` validates it against `engine.backend` and returns
+/// an error (mapped to [`KvCacheEstimateError::IncompatibleMemoryFraction`]) if
+/// mismatched.
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
+pub enum KvCacheMemoryFraction {
+    /// Fraction of TOTAL GPU memory. Compatible with vLLM
+    /// (`gpu_memory_utilization`) and SGLang (`mem_fraction_static`).
+    OfTotal(f64),
+    /// Fraction of FREE (post-non-KV) GPU memory. Compatible with TRT-LLM
+    /// (`free_gpu_memory_fraction`).
+    OfFree(f64),
+}
+
+impl KvCacheMemoryFraction {
+    /// `(kind, value)` pair for the flat Python call. `kind` is the wire string
+    /// the Python `estimate_kv_cache` expects (`"of_total"` / `"of_free"`).
+    fn to_wire(self) -> (&'static str, f64) {
+        match self {
+            Self::OfTotal(f) => ("of_total", f),
+            Self::OfFree(f) => ("of_free", f),
+        }
+    }
+}
+
+/// Default for [`KvCacheEstimateOptions::naive_kv_reservation`] so a JSON request
+/// from an embedded caller (the Mocker) that omits this newer field still
+/// deserializes (to the same 0.80 the Python side defaults to).
+fn default_naive_kv_reservation() -> f64 {
+    0.80
+}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
+pub struct KvCacheEstimateOptions {
+    pub allow_naive_fallback: bool,
+    pub allow_hf_config_download: bool,
+    /// Fraction of post-weight memory the naive fallback reserves for KV
+    /// (default `0.80`). Ignored on the native path. Exposed so the Mocker can
+    /// tune the crude fallback budget.
+    #[serde(default = "default_naive_kv_reservation")]
+    pub naive_kv_reservation: f64,
+}
+
+/// KV-cache memory estimate.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct KvCacheEstimate {
+    pub total_gpu_capacity_bytes: u64,
+    pub total_kv_size_bytes: u64,
+    pub kv_size_per_token_bytes: u64,
+    pub total_kv_size_tokens: u64,
+    pub source: EstimateSource,
+    /// `Some` on the native path; `None` on the naive fallback.
+    pub memory_breakdown: Option<MemoryBreakdown>,
+    /// `Some` iff `tolerance_fraction` set; `None` for the raw estimate.
+    pub tolerance_adjusted: Option<KvCacheEstimateAdjusted>,
+}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub enum EstimateSource {
+    /// AIC's full backend memory model used.
+    Native,
+    /// Post-weight reservation heuristic (`naive_kv_reservation`, default 80%).
+    NaiveFallback,
+}
+
+/// Non-KV memory components, in bytes. Maps AIC's `_get_memory_usage` dict:
+/// `weights → weights`, `activations → activations`, `others →
+/// runtime_overhead`, `nccl → comm_overhead`.
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct MemoryBreakdown {
+    pub weights_bytes: u64,
+    pub activations_bytes: u64,
+    pub runtime_overhead_bytes: u64,
+    pub comm_overhead_bytes: u64,
+}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
+pub struct KvCacheEstimateAdjusted {
+    pub tolerance_fraction: f64,
+    pub total_kv_size_bytes: u64,
+    pub total_kv_size_tokens: u64,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub enum KvCacheEstimateError {
+    Unsupported {
+        model: String,
+        backend: BackendKind,
+        gpu_sku: String,
+        reason: String,
+    },
+    InsufficientModelMetadata {
+        missing_fields: Vec<String>,
+    },
+    NoKvBudget {
+        total_gpu_capacity_bytes: u64,
+        non_kv_bytes: u64,
+    },
+    IncompatibleMemoryFraction {
+        backend: BackendKind,
+        variant_kind: &'static str,
+    },
+    BadConfig {
+        field: String,
+        reason: String,
+    },
+    HfConfigFetchFailed {
+        hf_id: String,
+        source: String,
+    },
+}
+
+impl std::fmt::Display for KvCacheEstimateError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Unsupported {
+                model,
+                backend,
+                gpu_sku,
+                reason,
+            } => write!(
+                f,
+                "unsupported model/backend/GPU for KV-cache estimation: model={model}, \
+                 backend={backend:?}, gpu_sku={gpu_sku}: {reason}"
+            ),
+            Self::InsufficientModelMetadata { missing_fields } => {
+                write!(
+                    f,
+                    "insufficient model metadata; missing: {missing_fields:?}"
+                )
+            }
+            Self::NoKvBudget {
+                total_gpu_capacity_bytes,
+                non_kv_bytes,
+            } => write!(
+                f,
+                "no KV budget: non-KV memory ({non_kv_bytes} bytes) meets/exceeds the \
+                 KV-cache memory limit (capacity={total_gpu_capacity_bytes} bytes)"
+            ),
+            Self::IncompatibleMemoryFraction {
+                backend,
+                variant_kind,
+            } => write!(
+                f,
+                "incompatible memory fraction: backend {backend:?} does not accept \
+                 KvCacheMemoryFraction::{variant_kind}"
+            ),
+            Self::BadConfig { field, reason } => {
+                write!(f, "bad memory config field {field:?}: {reason}")
+            }
+            Self::HfConfigFetchFailed { hf_id, source } => {
+                write!(f, "HF config fetch failed for {hf_id:?}: {source}")
+            }
+        }
+    }
+}
+
+impl std::error::Error for KvCacheEstimateError {}
+
+/// Estimate KV-cache memory (raw estimate + optional tolerance margin).
+///
+/// Pure forwarder: crosses into Python once to compute the COMPLETE estimate. The
+/// Python `aiconfigurator.sdk.memory.estimate_kv_cache` does the backend↔fraction
+/// and tolerance validation, the native AIC memory breakdown + budget math, the
+/// naive heuristic fallback, AND the tolerance margin (`tolerance_adjusted`); the
+/// Rust side rebuilds a [`KvCacheEstimate`] from the returned dict with no math
+/// of its own.
+///
+/// Errors from the Python side (unsupported model/backend, incompatible memory
+/// fraction, out-of-range tolerance, no KV budget, HF config fetch failure) cross
+/// the PyO3 boundary as a `ValueError` whose message carries the failure detail
+/// and are surfaced here as [`KvCacheEstimateError::Unsupported`] with that
+/// message.
+pub fn estimate_kv_cache(
+    req: KvCacheEstimateRequest,
+) -> Result<KvCacheEstimate, KvCacheEstimateError> {
+    fetch_python_estimate(&req)
+}
+
+/// Cross into Python once to compute the complete estimate.
+///
+/// Mirrors the `build_aic_engine` → `compile_engine` forwarder shape: `with_gil
+/// → import aiconfigurator.sdk.memory → call estimate_kv_cache(...) → extract
+/// the returned dict`. `tolerance_fraction` is forwarded; the Python fn applies
+/// the tolerance and returns `tolerance_adjusted` in the dict.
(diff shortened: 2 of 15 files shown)


rust/aiconfigurator-core/src/config.rs
```diff
@@ -121,6 +121,21 @@ impl BackendKind {
             Self::Vllm => "vllm",
         }
     }
+
+    /// Parse a wire backend name (`"trtllm" | "sglang" | "vllm"`) into a
+    /// [`BackendKind`]. Used by the `estimate_kv_cache` `#[pyfunction]` to build
+    /// an [`EngineConfig`] from flat scalar args. Returns the offending string in
+    /// the `Err` so the caller can surface a precise message.
+    pub(crate) fn parse(s: &str) -> Result<Self, String> {
```
simone-chen: Why don't we do case-insensitive match?

rust/aiconfigurator-core/src/lib.rs
```diff
@@ -43,6 +44,13 @@ pub use fpm::{
 // `engine/runtime.rs`) keep resolving after the types moved into `fpm`.
 pub(crate) use fpm::validate_forward_pass_metrics;
 pub use fpm::{ForwardPassMetrics, QueuedRequestMetrics, ScheduledRequestMetrics, FPM_VERSION};
+// KV-cache memory API (Issue #1159). Top-level surface, not a method on
```
simone-chen: Let's avoid mentioning about Issue #1159 in this PR. Same applies to the rest of code changes.

rust/aiconfigurator-core/src/config.rs
```diff
@@ -151,6 +166,18 @@ pub enum DataType {
     W4a8Mxfp4Mxfp8,
 }
 
+impl DataType {
+    /// Parse a wire dtype/quant-mode name (the serde `rename` strings, e.g.
+    /// `"bfloat16"`, `"fp8_block"`) into a [`DataType`]. Used by the
+    /// `estimate_kv_cache` `#[pyfunction]` to build a [`QuantizationConfig`]
+    /// from flat scalar args. Uses serde so the accepted names stay in lockstep
+    /// with the `#[serde(rename = ...)]` attributes above.
+    pub(crate) fn parse(s: &str) -> Result<Self, String> {
```
simone-chen: What's it mapping to? Seems it takes a string and just maps to a string. Where's the list of known types defined?

src/aiconfigurator_core/__init__.py
```diff
@@ -10,17 +10,20 @@
...
 
-``build_aic_engine`` is intentionally NOT re-exported: it is a Rust-only entry
-point for embedded callers (e.g. the Dynamo Mocker), not part of the Python
-surface.
+KV-cache capacity estimation (Issue #1159) is intentionally NOT re-exported: the
+estimate is computed entirely in Python
+(``aiconfigurator.sdk.memory.estimate_kv_cache``); the Rust
+``aiconfigurator_core::memory::estimate_kv_cache`` is a Rust-only forwarder for
+embedded callers (e.g. the Dynamo Mocker), like ``build_aic_engine``, and is not
+part of the Python surface.
```
simone-chen: Might be in a follow up PR when we do https://github.com/ai-dynamo/aiconfigurator/issues/1208?

tests/integration/test_memory_estimation.py
simone-chen, line 1: Should we unify the name `tests/integration/test_memory_estimation.py` and `tests/unit/sdk/test_kv_cache_estimate.py`?

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,795 @@
...
+- **Naive fallback** (:func:`_naive_fallback`): for models/backends AIC cannot
+  build natively, parse the HF ``config.json`` (reusing the SDK's existing
+  loaders + :func:`~aiconfigurator.sdk.utils._parse_hf_config_json`) and reserve
+  a fraction (default 80%) of post-weight memory for KV. Per #1159 the fallback
+  handles standard K/V attention and MLA/compressed-latent attention separately,
+  and fails closed to a conservative constant when the geometry is unclear.
+
+The budget math is factored into :func:`kv_cache_budget_bytes` so the (deferred)
+consolidation of ``InferenceSummary._check_and_set_kv_cache_oom`` onto a single
+formula is a small follow-up. See ``docs/phase-1.5-capacity-followup.md``.
```
simone-chen: Can we add a TODO: https://github.com/ai-dynamo/aiconfigurator/issues/1208?

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,795 @@
...
+    _load_pre_downloaded_hf_config,
+    _parse_hf_config_json,
+)
+
+_ONE_GIB = 1 << 30
+
+# Last-resort weight estimate (bytes) when the HF config lacks the fields for
+# even a rough param count. ~70 GiB: a deliberately recognisable, model-sized
+# placeholder for models AIC cannot model at all (not a calibrated number).
+_DEFAULT_FALLBACK_WEIGHT_BYTES = 70 * _ONE_GIB
```
simone-chen: This is bad and rootless. We should raise an error as proposed in https://github.com/ai-dynamo/aiconfigurator/issues/1159 if the case basic metadata isn't available.

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,795 @@
...
+
+# Last-resort per-token KV size (bytes) when the HF config lacks the attention
+# geometry to compute it (or MLA latent width is unclear). 128 KiB/token; a
+# recognisable round placeholder, conservatively large (-> fewer tokens).
+_DEFAULT_FALLBACK_KV_BYTES_PER_TOKEN = 131_072
```
simone-chen: same as above

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,795 @@
...
+
+    Thin convenience over :func:`estimate_kv_cache` (called directly, in-process
+    -- no Rust hop). Computes the per-rank KV-cache token capacity (native backend
+    memory model, with an optional naive fallback) and converts it to the number
+    of scheduler blocks the Dynamo Mocker would allocate. When
+    ``tolerance_fraction`` is set, the tolerance-adjusted token count is used (the
+    raw count otherwise).
+
+    This is the in-repo reference for the conversion the Dynamo shim
+    (``dynamo._internal.aic.estimate_num_gpu_blocks``) mirrors.
```
simone-chen: Please be more concise and try not to wire with Dynamo in the AIC repo. The feature is requested by dynamo but AIC should live standalone.

The middle section alone should be enough to convey what this function is doing.

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,795 @@
...
+            parsed = None
+        dtype_bytes = _dtype_bytes(raw)
+        geom = _naive_geometry(raw, parsed)
+        kv_per_token = _naive_kv_bytes_per_token(geom, dtype_bytes, tp_size)
+        weight_bytes = _naive_weight_bytes(geom, dtype_bytes, tp_size, pp_size)
+    else:
+        kv_per_token = None
+        weight_bytes = None
+    kv_per_token = float(kv_per_token if kv_per_token is not None else _DEFAULT_FALLBACK_KV_BYTES_PER_TOKEN)
+    weight_bytes = float(weight_bytes if weight_bytes is not None else _DEFAULT_FALLBACK_WEIGHT_BYTES)
```
simone-chen: no fallbacks for these basic info, as mentioned above.

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,795 @@
...
+    gpu_memory_capacity_bytes_override: int | None,
+    naive_kv_reservation: float,
+    allow_hf_config_download: bool,
+) -> dict[str, Any]:
+    """Conservative naive fallback: reserve ``naive_kv_reservation`` of post-weight memory.
+
+    Requires ``gpu_memory_capacity_bytes_override`` (the native path failed, so
+    the SystemSpec capacity is unavailable here). Per-rank bytes throughout,
+    matching the native path and the Mocker consumer. Does NOT honor the requested
+    ``memory_fraction_value`` -- the post-weight reservation is the crude budget.
```
simone-chen: I'd prefer restructure it into the cases in bullet points when we need naive fallback:
* hardware unknown to AIC-> thus we need gpu_memory_capacity_bytes_override
* model family unsupported -> thus we need to compute naive_weight_bytes

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,795 @@
...
+    model. Attention-DP replicates rather than shards, so it does NOT divide here.
+    """
+    hidden = geom.get("hidden")
+    layers = geom.get("layers")
+    if not hidden or not layers:
+        return None
+    hidden = int(hidden)
+    layers = int(layers)
+    vocab = int(geom.get("vocab") or 0)
+    inter = int(geom.get("inter") or (4 * hidden))
```
simone-chen: hidden, layer, vocab, inter should all be compulsory metadata and will incur return None when unavailable and eventually cause the naive fallback to fail.

Isn't it funny to fallback to 0 for vocab lol

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,795 @@
...
+
+# --------------------------------------------------------------------------- #
+# Naive fallback: parse the HF config, reserve a fraction of post-weight memory.
+#
+# Reuses the SDK's existing config loaders + `_parse_hf_config_json`, which
+# REJECTS architectures AIC does not support --
+# exactly the fallback's target -- so a supported arch (perf DB missing) uses the
+# normalized parse, while an unsupported arch reads the raw config keys directly.
+# Both feed one KV formula and one weight formula via `_naive_geometry`.
+# --------------------------------------------------------------------------- #
```
simone-chen, line 441: There are so many _naive* functions below. Would it make sense to create a MemoryEstimator abstract class to define the API shape, implement an AIC native subclass and a Naive fallback class, grouping functions inside of classes?

rust/aiconfigurator-core/src/config.rs
simone-chen: `BackendKind::parse` had no callers after the `#[pyfunction]` removal, so I deleted it (and `DataType::parse`) rather than make dead code case-insensitive. The Rust API takes `BackendKind` directly; the forwarder uses `.as_str()`.

rust/aiconfigurator-core/src/lib.rs
simone-chen: Removed all `#1159` mentions from code (`lib.rs`, `memory.rs`, `memory.py`, `__init__.py`). Kept it only in the design docs + PR Related-Issues for traceability.

rust/aiconfigurator-core/src/config.rs
simone-chen: It mapped a serde wire-string (e.g. `"fp8_block"`) to the `DataType` enum (the variants right above it) — but it was only used by the removed `#[pyfunction]`, so I deleted it along with `BackendKind::parse`.

src/aiconfigurator_core/__init__.py
simone-chen: #1208 is the deferred OOM-check consolidation — it doesn't touch the Python surface, so it won't re-export this. Trimmed the note and dropped the `#1159` ref.

tests/integration/test_memory_estimation.py
simone-chen: Renamed `tests/unit/sdk/test_kv_cache_estimate.py` -> `tests/unit/sdk/test_memory_estimation.py`, so the unit and integration tests share the name (matching `sdk/memory.py`).

src/aiconfigurator/sdk/memory.py
simone-chen: Added the `#1208` reference to the module docstring (there's also a `TODO(#1208)` at the `kv_cache_budget_bytes` helper).

src/aiconfigurator/sdk/memory.py
simone-chen: Agreed — removed `_DEFAULT_FALLBACK_WEIGHT_BYTES`; the fallback now raises `insufficient model metadata` when the weight-estimate fields are missing, instead of guessing.

src/aiconfigurator/sdk/memory.py
simone-chen: Same — removed `_DEFAULT_FALLBACK_KV_BYTES_PER_TOKEN`; raises when the per-token KV geometry can't be computed.

src/aiconfigurator/sdk/memory.py
simone-chen: Trimmed and removed the Dynamo wiring — kept the concise middle.

src/aiconfigurator/sdk/memory.py
simone-chen: Done — no placeholder fallbacks; raises when kv-per-token or weights can't be computed from the config.

src/aiconfigurator/sdk/memory.py
simone-chen: Restructured the docstring into the two bulleted cases (hardware unknown -> needs `gpu_memory_capacity_bytes_override`; model unsupported -> estimate weights/KV from the HF config).

src/aiconfigurator/sdk/memory.py
simone-chen: Done — `hidden` / `layers` / `vocab` / `inter` are all required now (dropped the `or 0` / `or 4*hidden`); a missing one returns None and the fallback raises. (vocab->0 is gone.)

src/aiconfigurator/sdk/memory.py
simone-chen: After dropping the default-constant machinery the naive surface is small and cohesive, so I kept it functional rather than add an abstract class for two dict-returning paths. Happy to do the `MemoryEstimator` refactor as a follow-up if you'd prefer — reopen this if so.

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,805 @@
...
+        (= AIC ``others``) / ``comm_overhead_bytes`` (= AIC ``nccl``)
+      - ``non_kv_bytes`` = sum of the four above (== ``total - kvcache``)
+      - ``kv_size_per_token_bytes`` = ``model.get_kvcache_bytes_per_sequence(1)``
+        (MLA-aware: latent ``kv_lora_rank + qk_rope_head_dim`` for compressed
+        attention, else ``2 * kv_heads_per_gpu * head_size`` summed over layers)
+      - ``gpu_memory_capacity_bytes`` = ``system_spec["gpu"]["mem_capacity"]``
+    """
+    resolved_moe_tp = moe_tp_size if moe_tp_size is not None else 1
+    resolved_moe_ep = moe_ep_size if moe_ep_size is not None else 1
+    model_config = _build_model_config(
```
jasonqinzhou: [P1] Apply `nextn` before building the native memory model

`estimate_kv_cache` accepts `nextn` / `nextn_accept_rates` and Rust forwards them, but `kv_cache_memory_breakdown` never applies them to the `ModelConfig` before calling `get_model`. `_get_memory_usage` later reads `model.config.nextn` to scale activation memory for MTP, so speculative/MTP requests currently estimate non-KV memory as if `nextn=0`. That can overstate available KV budget and size `num_gpu_blocks` too high for the Mocker. Please mirror the existing `_apply_nextn(model_config, nextn, nextn_accept_rates)` path before `get_model`.

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,805 @@
...
+        max_seq_len=1,
+    )
+
+    weights_bytes = float(memory["weights"]) * _ONE_GIB
+    activations_bytes = float(memory["activations"]) * _ONE_GIB
+    runtime_overhead_bytes = float(memory["others"]) * _ONE_GIB
+    comm_overhead_bytes = float(memory["nccl"]) * _ONE_GIB
+    non_kv_bytes = weights_bytes + activations_bytes + runtime_overhead_bytes + comm_overhead_bytes
+
+    kv_size_per_token_bytes = float(model.get_kvcache_bytes_per_sequence(1))
```
tedzhouhk: This assumes KV cache growth is linear by using `get_kvcache_bytes_per_sequence(1)` as a per-token constant. That is not true for hybrid/sliding-window attention models, where SWA layers stop growing after the window while global layers continue. For those models this can significantly under-estimate token capacity / num_gpu_blocks. Consider using a model-level inverse capacity helper, or at least detecting non-linear KV growth instead of deriving capacity from seq_len=1.

src/aiconfigurator/sdk/memory.py
```diff
@@ -0,0 +1,805 @@
...
+        "qk_rope_head_dim": raw.get("qk_rope_head_dim"),
+        "vocab": raw.get("vocab_size"),
+        "hidden": hidden,
+        "inter": raw.get("intermediate_size"),
+        "num_experts": raw.get("n_routed_experts") or raw.get("num_local_experts") or 0,
+        "moe_inter": raw.get("moe_intermediate_size") or raw.get("intermediate_size") or 0,
+    }
+
+
+def _naive_kv_bytes_per_token(geom: dict, dtype_bytes: int, tp_size: int) -> int | None:
```
tedzhouhk: Maybe also add some simple logic to cover hybrid model's fallback? No need to be super accurate.

src/aiconfigurator/sdk/memory.py
simone-chen: Fixed in 5d44e913. The native path no longer treats `get_kvcache_bytes_per_sequence(1)` as a constant per-token rate. I added a model-level inverse `Model.get_kvcache_max_tokens(budget_bytes)` (ff46e490): the base model is exact floor-division for linear KV growth, while Gemma 4 (sliding-window) and DeepSeek-V4 (window-capped + compressed) override it to invert their true piecewise `get_kvcache_bytes_per_sequence` curve via a monotonic binary search. `KVCacheEstimator` sizes both the raw and the tolerance-adjusted token counts through that inverse, so SWA layers cap at the window while global layers keep growing — no more `seq_len=1` extrapolation.

src/aiconfigurator/sdk/memory.py
simone-chen: Done in 5d44e913. `NaiveKVCacheEstimator` now detects a sliding-window layout from the HF config (`layer_types`, or `sliding_window` + `sliding_window_pattern`) and, for standard-attention hybrids, inverts a simple piecewise curve that caps the SWA layers at the window while the global layers keep growing. It falls back to the linear per-token estimate when the layout can't be determined or for MLA hybrids — kept deliberately coarse, per your note.

src/aiconfigurator/sdk/memory.py
simone-chen: Fixed in 5d44e913. `nextn` / `nextn_accept_rates` are now applied to the `ModelConfig` via `_apply_nextn` before `get_model` (in `KVCacheEstimator.from_request`, mirroring the agg/disagg/static estimate paths), so `_get_memory_usage` scales activation memory for MTP instead of sizing non-KV memory as `nextn=0`. Added a unit test asserting the config reaching `get_model` carries the nextn + default accept rates.

rust/aiconfigurator-core/src/memory.rs
```diff
@@ -0,0 +1,436 @@
...
+    // PyErr → KvCacheEstimateError inline (keeps the error enum self-contained).
+    // The Python side raises a ValueError whose message already carries the
+    // specific failure detail; surface it as Unsupported so the (deferred)
+    // Mocker fallback decision still keys on the same variant it did before.
+    .map_err(|e| KvCacheEstimateError::Unsupported {
+        model: engine.model_name.clone(),
+        backend: engine.backend.clone(),
+        gpu_sku: engine.system_name.clone(),
+        reason: format!("estimate_kv_cache: {e}"),
+    })
```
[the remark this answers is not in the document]
simone-chen: Intentional, declining for now. `sdk.memory.estimate_kv_cache` is the single source of truth for the failure detail — it raises a `ValueError` whose message already carries the specific cause, and the Rust side surfaces that verbatim in the `Unsupported { reason }` payload. The only current consumer is the (deferred) Dynamo Mocker fallback, which keys on the `Unsupported` variant exactly as before. Introducing a stable error-code/class contract and mapping it to typed Rust variants (`BadConfig`/`IncompatibleMemoryFraction`/`NoKvBudget`/…) is a coordinated Rust+Python change that belongs with that consumer — tracked alongside the Mocker integration in #1208. Mapping on Python message strings now would be the fragile half-measure, so I am leaving the documented single-variant behavior in place.

jasonqinzhou (approved):

tedzhouhk (approved):

PR #1203, opened 2026-06-05
fix: skip sparse L40S vLLM 0.22.0 database

jasonqinzhou:
## Summary

Temporary fix for the CI failures introduced by #1158. The L40S vLLM 0.22.0 data added there is too sparse — the support-matrix healer is purely model-driven, so some interpolation-required generation-attention points were never collected, and downstream PRs fail when the sanity checks hit them.

- Mark `l40s/vllm/0.22.0` as incomplete via `INCOMPLETE.txt` so `get_supported_databases` skips it during database discovery. CI sanity checks keep using L40S vLLM 0.14.0 until 0.22.0 is recollected with sufficient generation-attention coverage.
- Add unit coverage (`test_get_supported_databases_skips_incomplete_versions`) confirming incomplete versions are excluded from discovery and from `get_latest_database_version`.

This PR is intentionally minimal to unblock CI quickly. The `validate_database.ipynb` coverage preflight (`sanity_coverage.py` + its tests) that was previously on this branch is deferred to a follow-up PR.

## Verification

- `uv run --extra dev pytest tests/unit/sdk/database/test_database_helpers.py -q` → 21 passed

### DIFF
--- a/src/aiconfigurator/systems/data/l40s/vllm/0.22.0/INCOMPLETE.txt
+++ b/src/aiconfigurator/systems/data/l40s/vllm/0.22.0/INCOMPLETE.txt
@@ -1,0 +1,4 @@
+L40S vLLM 0.22.0 data is intentionally excluded from supported database discovery.
+The available generation attention data is too sparse for the sanity-check workload
+and can fail interpolation, so keep using the older L40S vLLM database until this
+version is recollected with sufficient coverage.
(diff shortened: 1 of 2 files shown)


simone-chen:
Thanks for the fix! One concern with the temporary fix: AIC CLI default mode will still detect latest as 0.22.0 with this fix. Should we also modify code to avoid using version as latest if a INCOMPLETE.txt is present? @coderabbitai Can you confirm?

simone-chen (approved):

PR #1211, opened 2026-06-08
fix(operations): correct kv_cache_bytes formula in DSv4 attention SOL

yinan1995:
DeepSeek-V4 uses compressed attention (SWA/CSA/HCA) with MQA-equivalent
    KV cache storage: each token stores a single head_dim-sized vector
    shared by all query heads, confirmed by
    DeepSeekV4Model.get_kvcache_bytes_per_sequence which derives
    cache_entry_bytes = head_dim * kvcache_quant_mode.value.memory
    (without a num_heads multiplier).
    
    The SOL bandwidth formula in _deepseek_v4_attention_sol computed:
    
        kv_cache_bytes = attention_pairs * num_heads * head_dim * memory
    
    which over-counted KV-cache HBM traffic by a factor of num_heads
    (~128x for DSv4-Pro). The inflated sol_mem term made the analytic
    SOL latency exceed silicon-measured latency at non-trivial decode
    batch sizes, violating the SOL <= silicon roofline invariant.
    
    Fix: simply remove the num_heads multiplier:
    
        kv_cache_bytes = attention_pairs * head_dim * memory
    
    This is a minimal one-factor fix. attention_pairs already correctly
    models the KV entries accessed per batch (window_pairs + compressed_pairs
    = b * entries_per_request for generation), matching the storage layout
    in get_kvcache_bytes_per_sequence.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected memory calculation for DSv4 models to accurately reflect compressed KV-cache storage, providing more precise resource usage estimates.

### DIFF
--- a/src/aiconfigurator/sdk/operations/dsv4.py
+++ b/src/aiconfigurator/sdk/operations/dsv4.py
@@ -302,7 +302,8 @@ def _deepseek_v4_attention_sol(
         * (hidden_size + q_lora_rank + num_heads * head_dim + head_dim + local_groups * o_lora_rank)
         * gemm_quant_mode.value.memory
     )
-    kv_cache_bytes = attention_pairs * num_heads * head_dim * kvcache_quant_mode.value.memory
+    # DSv4 compressed attention stores one head_dim KV vector per token, shared by all query heads.
+    kv_cache_bytes = attention_pairs * head_dim * kvcache_quant_mode.value.memory
     rope_bytes = tokens * num_heads * rope_head_dim * fmha_quant_mode.value.memory
 
     sol_math = (


src/aiconfigurator/sdk/operations/dsv4.py
```diff
@@ -391,7 +307,20 @@ def _tc_flops(quant_mode):
...
+    # reused for every query token (Flash Attention tiling).  With MQA,
+    # the KV cache has ``num_kv_heads`` heads (1 for DeepSeek-V4) shared by
+    # all query heads.  The previous formula ``attention_pairs * num_heads * ...``
+    # overcounted by treating each Q-K pair as a separate cache line, inflating
+    # the memory-bound SOL by orders of magnitude for compressed-attention layers.
+    if compress_ratio:
+        actual_kv_entries = kv_len // compress_ratio + window_size
+    else:
+        actual_kv_entries = kv_len
+    kv_cache_bytes = b * actual_kv_entries * num_kv_heads * head_dim * 2 * kvcache_quant_mode.value.memory
```
tianhaox: why there's a 2 here?

tianhaox: btw, maybe we don't need num_kv_heads and just use 1

yinan1995: > why there's a 2 here?

Good catch — the * 2 was my AI co-author being overly enthusiastic 😅. My previous reply was wrong, let me correct it: it saw the 2 * n_kv * s * h pattern in attention.py (standard MHA/GQA with separate K and V caches) and thought "aha, same thing here!" — except DeepSeek-V4's compressed attention (SWA/CSA/HCA) stores a single head_dim-sized vector per token, shared across all query heads, similar to an MQA storage layout. Our own get_kvcache_bytes_per_sequence confirms: cache_entry_bytes = head_dim * quant_memory, no * 2 anywhere. Classic case of doing more work than necessary and being confidently wrong about it. I'll drop it.

yinan1995: > btw, maybe we don't need num_kv_heads and just use 1

Same goes for num_kv_heads — you're right that we don't need it. The whole fix is really just removing num_heads from the original formula. Everything else was over-engineering.

tianhaox:
closing this PR as I just found another PR #1216  fixed this bug. thanks for the contribution.

PR #1213, opened 2026-06-10
fix: enable b60 support in generator

mandy-li:
Fixed hard-coded CUDA_VISIBLE_DEVICES in the generated artifacts by passing system name into the template. If system is b60, use ONEAPI_DEVICE_SELECTOR. Otherwise, unchanged.

## New Features
* Added automatic GPU device binding that detects and selects the appropriate environment configuration based on system hardware type (Intel oneAPI or CUDA).


jasonqinzhou (approved):
could u take a look the comments of Coderabbit?

mandy-li:
> could u take a look the comments of Coderabbit?

Thanks for review. Coderabbit comments don't make sense for this case. b60 or intel HW currently only support vllm and system name is case-sensitive.

PR #1215, opened 2026-06-10
refactor: integrate v2 Task into CLI - phase 2

tianhaox:
The flat **v2 `Task`** (`sdk/task_v2.py`) is now the only task config; the legacy v1 `TaskConfig` / `TaskRunner` (`sdk/task.py`) is deleted. CLI, `generator/module_bridge`, and tooling all build and run the v2 `Task`. Legacy v1 YAML still loads through a **deprecated, limited compatibility shim**, and all bundled YAML + docs are migrated to the flat format. On top of the core swap this PR also lands the v1→v2 parity fixes, a disagg-sweep performance fix, stricter YAML validation, and two restored config knobs.

#### Overview:

Phase 2 of the task-config v2 work. Replaces the nested v1 task schema with the flat v2 `Task` (one key ↔ one field) across the whole stack, removes v1, and keeps existing v1 YAML working via a one-way auto-conversion shim that is explicitly deprecated. Also migrates every bundled experiment/example YAML and the docs to the flat format so users no longer see the legacy shape as the norm.

#### Details:

- **Core swap**: v1 `TaskConfig`/`TaskRunner` (`sdk/task.py`) deleted. CLI (`cli/main` `build_default_tasks` / `build_experiment_tasks` / `_execute_tasks`, `cli/api`, `cli/utils`, `report_and_save`), `generator/module_bridge`, `enumerate` / `support_matrix` / `run_sweep_new` all build and run the v2 `Task`. `build_disagg_parallel_lists` promoted to a public `sdk.task_v2` function.
- **v1 YAML compat (deprecated)**: `sdk/task_v1_compat.py` — `Task.from_yaml` auto-detects the legacy nested format (`mode`/`config`/`worker_config`/`replica_config`/`profiles`) and converts it with a `DeprecationWarning`. V1 fields with **no v2 equivalent are rejected, not silently dropped**.
- **Ported v1-only behavior into v2**: gpt-oss Blackwell MoE-quant promotion, DeepSeek-V4 native-FP4-on-Hopper rejection, and multimodal image inputs; plus default-path parity fixes (quant/nextn/search-space, WideEP+validate timing, `num_gpu_per_replica` state, fp8 FMHA fail-fast, `database_mode` actually taking effect).
- **Stricter YAML**: unknown/misspelled keys and YAML-unexpressible strategy fields now raise instead of being logged-and-ignored (the warning was invisible in long logs). `quant_preset` removed — v1 `profiles` expand to explicit `*_quant_mode` fields.
- **Restored config knobs**: `attention_backend` (flashinfer/fa3) and `wideep_num_slots` are exposed again as flat `Task` fields (v1 had them, the initial v2 schema dropped them); both flow into `ModelConfig`.
- **Performance**: disagg sweep's `_match_workers` `lru_cache` is now shared across all (ttft, tpot) constraint points instead of being rebuilt per point — removes a ~22% wall-clock regression vs v1; output is byte-identical.
- **Naming cleanup**: `task_configs`→`tasks`, `_execute_task_configs`→`_execute_tasks`, etc.; dropped the leftover `*_task_config` local names.
- **YAML + docs migration**: all `cli/exps/*.yaml` and `cli/example.yaml` rewritten to flat v2 (each verified to construct a byte-identical `Task`); `example_new.yaml` removed; new `example_v1_deprecated.yaml` kept as a marked migration reference; `cli_user_guide.md` / `advanced_tuning.md` walkthroughs rewritten to v2 and note that v1 is deprecated with only limited compatibility.
- **Tests**: new `test_v1_compat` + `test_blackwell_gptoss`; field-effect / no-orphan guards against accepted-but-ignored fields; `build_experiment_tasks` regression test for flat-v2 disagg; cli/model/picking tests updated to the flat schema; v1-only tests removed.

- `src/aiconfigurator/sdk/task_v2.py` — the flat `Task`, `from_yaml` / `build_model_config` / `validate`.
- `src/aiconfigurator/sdk/task_v1_compat.py` — the deprecated v1→v2 conversion (and what it rejects).
- `src/aiconfigurator/cli/main.py` — `build_default_tasks` / `build_experiment_tasks` (incl. the flat-v2 disagg pre-flight fix) / `_execute_tasks`.
- `src/aiconfigurator/sdk/sweep.py` — `sweep_disagg` / `_match_workers` cache scope (perf fix).

## Summary by CodeRabbit

* **Documentation**
  * Updated CLI and advanced tuning guidance to the canonical flat V2 experiment YAML schema; legacy V1 nested YAML is still accepted with a DeprecationWarning.
  * Refreshed example and experiment configuration templates to match the new flat format.

* **Improvements**
  * CLI, reporting, saving, support-matrix, and generators now operate on the V2 Task model and expose clearer primary model/system/backend identity fields.
  * Stricter V2 YAML validation and improved disaggregation/wide-EP task handling; shared caching speeds up disagg sweep evaluation.

* **New Features**
  * Added multimodal image workload knobs (height, width, images-per-request) to Task.

* **Tests**
  * Updated CLI/unit tests for the V2 Task interface; removed legacy parity coverage.


tianhaox:
@coderabbitai resolve

tianhaox:
@coderabbitai full review

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -450,16 +540,95 @@ def from_cli(cls, **kwargs: Any) -> Task:
...
+        if self.moe_backend is not None:
+            return
+        wideep = (
+            self.enable_wideep
+            if self.serving_mode == "agg"
+            else (self.prefill_enable_wideep or self.decode_enable_wideep)
+        )
+        if wideep:
+            self.moe_backend = "deepep_moe"
+
```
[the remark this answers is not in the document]
tianhaox: Agreed this is a real concern for SGLang mixed-WideEP disagg (one role WideEP, the other not): `self.moe_backend` is shared and `build_model_config` feeds it to every role, so the non-WideEP role still gets modeled/validated as `deepep_moe`. Note this is inherited from v1 (`_normalize_wideep_moe_backend` mirrors the v1 `__init__` derivation), so it's a pre-existing latent issue rather than a regression introduced by this PR. As flagged (heavy lift), a safe fix needs a per-role effective moe_backend that model construction, DB validation, and the disagg search all route through. Deferring from this PR to keep scope bounded; tracking as a follow-up.

src/aiconfigurator/generator/enumerate.py
```diff
@@ -491,10 +491,12 @@ def enumerate_profiling_configs(
...
         backend_name=backend,
+        is_moe=is_moe,
         prefill_system=system,
         decode_system=system,
-        is_moe=is_moe,
-        enable_wideep=enable_wideep,
+        prefill_enable_wideep=enable_wideep,
+        decode_enable_wideep=enable_wideep,
+        moe_backend=None,
     )
```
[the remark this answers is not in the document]
tianhaox: Confirmed: `enumerate.py` hardcodes `moe_backend=None` at both `build_disagg_parallel_lists` call sites, and `moe_backend` only affects the SGLang non-WideEP MoE parallel ladder, so SGLang MoE without WideEP skips the megamoe/deepep search space here. This is pre-existing (not introduced by this PR) and lives in the generator enumeration path, so deferring from this PR. Follow-up: resolve the intended `moe_backend` from the Task/model config before enumeration and thread it through, after confirming the generator enum path is meant to cover megamoe/deepep.

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -1018,10 +1347,28 @@ def sweep_disagg_kwargs(self, *, prefill_database, decode_database) -> dict[str,
...
+        # list is filtered by the cap; a None list (WideEP) becomes range(1, cap+1) so the
+        # replica size stays bounded (v2 sweep gates by this list, not a max ceiling).
+        if self.num_gpu_per_replica:
+            num_gpu_list = self.num_gpu_per_replica
+            if self.max_gpu_per_replica is not None:
+                num_gpu_list = [n for n in num_gpu_list if n <= self.max_gpu_per_replica]
+        elif self.max_gpu_per_replica is not None:
+            num_gpu_list = list(range(1, self.max_gpu_per_replica + 1))
+        else:
+            num_gpu_list = None
```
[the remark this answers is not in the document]
tianhaox: The empty-list edge is worth noting, but the stated trigger isn't reachable: `_apply_total_gpus_budget()` intentionally does NOT filter `num_gpu_per_replica` (only the per-worker `*_num_gpu_candidates`; see the comment there mirroring v1's get_working_list), so it never narrows this to `[]`. That leaves only a caller passing `num_gpu_per_replica: []` explicitly, where the current truthiness treats empty-as-unset and widens — a degenerate input rather than a realistic constraint. Leaving as-is for now; happy to switch to `is not None` if we decide an explicit `[]` should mean “no feasible replica size”.

src/aiconfigurator/cli/example_v1_deprecated.yaml
```diff
@@ -0,0 +1,135 @@
...
+  osl: 1000 # output sequence length, optional, default to 1000
+  prefix: 0 # prefix cache len, default to 0
+  ttft: 1000.0  # Target TTFT in ms, optional, default to 1000.0
+  tpot: 40.0   # Target TPOT in ms, optional, default to 40.0
+  enable_wideep: false # enable wide ep for prefill/decode, optional, default to false
+  profiles: [] # some inherit presets for easier patch, optional
+  config: # all optional, used to patch default values
+    nextn: 1 # mtp 1
+    nextn_accept_rates: [0.85,0,0,0,0] # each position maps to the accept rate of the ith draft token, nextn 1 will only use the first draft token accept rate.
+    moe_backend: None # moe backend for SGLang wide ep, optional, default to None
```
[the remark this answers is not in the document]
tianhaox: Technically correct — `None` is parsed as a YAML string here, not null. This file is a verbatim copy of the legacy example, kept only as a deprecated migration reference of the old nested shape, so I'd rather not edit its content; the canonical, correct template is the flat `example.yaml`. Can switch it to `null` if we'd prefer the deprecated reference to also be strictly valid YAML.

src/aiconfigurator/cli/report_and_save.py
```diff
@@ -760,31 +756,32 @@ def get_safe_model_name(path: str) -> str:
...
+                        row_task_key = f"{exp_name}_{row_backend}"
+                        row_task = tasks[row_task_key]
                         row_backend_version = generated_backend_versions.get(
-                            row_backend, row_task_config.backend_version
+                            row_backend, row_task.primary_backend_version
                         )
                     else:
-                        row_task_config = exp_task_config
+                        row_task = exp_task
                         row_backend_version = effective_generated_version
```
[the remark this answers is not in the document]
tianhaox: Agreed it's defensive. The `else` branch is reachable only when `backend == "auto"` AND a merged row lacks a `backend` column; `_merge_into_top_n` always sets `df["backend"]` for auto merges, so in practice that sub-case doesn't occur and `exp_task` / `effective_generated_version` are always assigned (non-auto path) when the `else` runs. So no live NameError today — leaving it, though an explicit guard would be reasonable hardening if the merge invariant ever changes.

src/aiconfigurator/sdk/task_v1_compat.py
```diff
@@ -0,0 +1,290 @@
...
+        extra ``profiles``) are collected and the conversion raises -- they are
+        never silently dropped.  The ``mode`` selector is V1 structure and is
+        dropped harmlessly.
+    """
+    out: dict = {}
+    unmapped: list[str] = []
+    serving_mode = v1.get("serving_mode", "agg")
+    config = v1.get("config") or {}
+    profiles = v1.get("profiles") or []
+
```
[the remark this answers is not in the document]
tianhaox: The robustness point is fair, but the cited example is off: `config: []` hits `[] or {}` → `{}`, so it's handled. A non-empty non-dict (e.g. `config: [foo]`, or a non-list `profiles`) is what would reach `.get(...)` and raise `AttributeError` instead of a clear conversion error. Low priority (malformed legacy YAML), but consistent with the fail-loud direction — deferring an explicit type check for now.

docs/advanced_tuning.md
simone-chen, line 1: Out of curiosity, what was our intention to use flat fields in `Task` that drives this refactor?

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -120,12 +96,60 @@ def _resolve_quant_str(key: str, value: Any) -> Any:
...
+        return False
+    spec = load_system_spec(system_name)
+    return int(spec.get("gpu", {}).get("sm_version", -1)) >= 100
+
+
+def _is_hopper_system(system_name: str | None) -> bool:
+    """True for Hopper-class systems (h100 / h200 / gh200)."""
+    if not system_name:
+        return False
+    return system_name.startswith(("h100", "h200", "gh200"))
```
simone-chen:
```suggestion
    return system_name.lower().startswith(("h100", "h200", "gh200"))
```

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -120,12 +96,60 @@ def _resolve_quant_str(key: str, value: Any) -> Any:
...
+
+# Native FP4 routed-expert DeepSeek-V4 checkpoints and their FP8 replacements.
+# The native FP4 weights are unsupported on Hopper.
+_DEEPSEEK_V4_NATIVE_FP4_TO_FP8_MODEL = {
+    "deepseek-ai/DeepSeek-V4-Flash": "sgl-project/DeepSeek-V4-Flash-FP8",
+    "deepseek-ai/DeepSeek-V4-Pro": "sgl-project/DeepSeek-V4-Pro-FP8",
+}
+
+
+def _is_blackwell_system(system_name: str | None) -> bool:
```
jasonqinzhou: shall we put _is_blackwell_system and _is_hopper_system in system folder, or in system_spec.py?

src/aiconfigurator/sdk/task_v2.py
jasonqinzhou: can we use sm versions?

def _is_hopper_system(system_name: str | None) -> bool:
    """True for Hopper-class systems (SM 90, e.g. h100 / h200 / gh200)."""
    if not system_name:
        return False
    spec = load_system_spec(system_name)
    return int(spec.get("gpu", {}).get("sm_version", -1)) == 90

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -51,43 +51,19 @@
...
-    },
-}
+# Families that natively ship MTP (nextn=1) -- used only as a fallback when the
+# checkpoint's HF config does NOT declare ``num_nextn_predict_layers`` at all.
+# A config that explicitly states the field (including 0, e.g. Kimi-K2.5) wins.
+_MTP_DEFAULT_FAMILIES = {"DEEPSEEK", "DEEPSEEKV32", "DEEPSEEKV4", "KIMIK25", "QWEN35"}
+
+# Legacy V1 TaskRunner swept TPOT over this fixed grid to build the latency/throughput
+# Pareto frontier. Used when ``pareto_sweep=True`` (the default) so v2 matches v1.
+_LEGACY_TPOT_SWEEP: list[int] = list(range(1, 20, 1)) + list(range(20, 300, 5))
```
jasonqinzhou, line 66: shall we generate pareto curve around users' TPOT target?

simone-chen (approved):
LGTM!

jasonqinzhou (approved):

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -4,10 +4,9 @@
 """
 Task — flat user-facing config for sweep_agg / sweep_disagg.
```
jasonqinzhou, line 5: shall we refactor this file to make the sweep config generation process cleaner with less if-else conditions?

src/aiconfigurator/sdk/task_v2.py
tianhaox: Done in 2daf9f15 — moved both into `perf_database.py` (next to `load_system_spec`) as public `is_blackwell_system` / `is_hopper_system`, since they're generic system classifiers rather than task-specific helpers.

src/aiconfigurator/sdk/task_v2.py
tianhaox: Done in 2daf9f15 — `is_hopper_system` now keys off `sm_version == 90` (consistent with `is_blackwell_system`'s `sm_version >= 100`) instead of the `h100`/`h200`/`gh200` name-prefix match. Verified: h100/h200/gh200 → hopper, b200/b300/gb200/gb300 → blackwell.

src/aiconfigurator/sdk/task_v2.py
tianhaox: Good idea. Generating the Pareto curve around the user's TPOT target is a sweep-semantics change rather than part of this v1→v2 integration, so I'd prefer to keep it out of this PR and track it as a follow-up enhancement.

src/aiconfigurator/sdk/task_v2.py
tianhaox: Agreed the file has grown and the mode/role branching could be cleaner. That's a sizable refactor of its own, so I'd rather keep it out of this (already large) integration PR and do it separately as a follow-up.

docs/advanced_tuning.md
tianhaox: the flas fileds is only part of the refactor. it creates a 1:1 mapping of the arg class. the previous yaml+cli patch design is confusing and difficult to add new filed, that the main reason of the refactor. meanwhile, the sweep logic is moved out from the inference session and make the definitions clear.

src/aiconfigurator/sdk/task_v2.py
tianhaox: Tracked as a follow-up: #1226

src/aiconfigurator/sdk/task_v2.py
tianhaox: Tracked as a follow-up: #1227

PR #1221, opened 2026-06-15
fix: fail accurately on insufficient data instead of silent handling

tianhaox:
#### Overview:

#1114/#1158 made the interpolation engine silently return wrong values (or crash uncaught) on insufficient perf-data slices. This makes it fail accurately instead — degenerate/missing slices raise `PerfDataNotAvailableError`, which sweep (skip) and HYBRID (empirical) already handle. No new recovery logic.

#### Details:

- Remove `allow_singleton_axes` (ContextAttention): it returned the `batch=1` latency for any batch. Now fails loudly like other ops.
- Cubic `QhullError` (previously escaped uncaught) → `ValueError`.
- Classify interpolation `ValueError` as missing silicon data so it surfaces as `PerfDataNotAvailableError`.

4 files, +12/−69. Doesn't touch the extrapolation warnings (ragged data → re-collection) or `fp8_static` (separate PR).

`perf_database.py` (`_MISSING_SILICON_DATA_EXCEPTIONS`) and `interpolation.py`.

#### Related Issues:

Relates to #1114, #1158

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling for degenerate or missing interpolation data by introducing a dedicated `InterpolationDataNotAvailableError` (subclass of `ValueError`) for consistent reporting.
  * Performance-data lookups in silicon/hybrid modes now treat interpolation-data unavailability as missing data, matching existing fallback behavior.

* **Breaking Changes**
  * Stricter interpolation validation now enforces required axis coverage and rejects singleton-axis cases (the previous relaxed behavior is removed).

### DIFF
--- a/src/aiconfigurator/sdk/perf_database.py
+++ b/src/aiconfigurator/sdk/perf_database.py
@@ -24,7 +24,7 @@ databases_cache = defaultdict(lambda: defaultdict(lambda: defaultdict()))
 logger = logging.getLogger(__name__)
 
 _SYSTEMS_PATHS: list[str] = [os.fspath(pkg_resources.files("aiconfigurator") / "systems")]
-_MISSING_SILICON_DATA_EXCEPTIONS = (KeyError, IndexError)
+_MISSING_SILICON_DATA_EXCEPTIONS = (KeyError, IndexError, ValueError)
 
 
 def _normalize_systems_paths(raw_paths: str | Iterable[str] | None) -> list[str]:
--- a/src/aiconfigurator/sdk/perf_database.py
+++ b/src/aiconfigurator/sdk/perf_database.py
@@ -17,6 +17,7 @@ import yaml
 
 from aiconfigurator.sdk import common
 from aiconfigurator.sdk.common import PerfDataFilename, parse_support_matrix_version
+from aiconfigurator.sdk.interpolation import InterpolationDataNotAvailableError
 from aiconfigurator.sdk.performance_result import PerformanceResult
 from aiconfigurator.sdk.system_spec import SystemSpec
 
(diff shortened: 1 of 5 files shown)


src/aiconfigurator/sdk/perf_database.py
```diff
@@ -24,7 +24,7 @@
 logger = logging.getLogger(__name__)
 
 _SYSTEMS_PATHS: list[str] = [os.fspath(pkg_resources.files("aiconfigurator") / "systems")]
-_MISSING_SILICON_DATA_EXCEPTIONS = (KeyError, IndexError)
+_MISSING_SILICON_DATA_EXCEPTIONS = (KeyError, IndexError, ValueError)
```
Arsene12358: I feel ValueError might be too broad here. Would it be better if we define a specific error in **interpolation.py** like
```python
class InterpolationDataNotAvailableError(ValueError):
    """Raised when interpolation cannot produce a real value from available data."""
```
and raise it when interpolation data is actually missing

Arsene12358 (dismissed):

src/aiconfigurator/sdk/perf_database.py
tianhaox: Good call — I've taken this approach in 89b5a939. New class `InterpolationDataNotAvailableError(ValueError)` lives in `interpolation.py`, and `_MISSING_SILICON_DATA_EXCEPTIONS` is narrowed to it. Every `raise ValueError` in the interpolation module that signals insufficient data was migrated (axis coverage, out-of-range/singleton bracket, constant-dim target mismatch, non-finite result, degenerate cubic slice from `QhullError`). It keeps subclassing `ValueError` so existing `except ValueError` callers still work, but a plain `ValueError` from a genuine bug deeper in the silicon-query stack now surfaces with a real traceback instead of being misclassified as missing data. Added a contract test in `test_perf_database_errors.py` to lock that down.

tianhaox:
@coderabbitai re: the "broadening `_MISSING_SILICON_DATA_EXCEPTIONS` to include `ValueError`" note — addressed in 89b5a939. Introduced `InterpolationDataNotAvailableError(ValueError)` and narrowed the exception tuple to it, so a real bug raising plain `ValueError` in the silicon-query call stack no longer gets misclassified as missing data. All five interpolation raise-sites (`_require_3d_axis_coverage`, `_reduced_griddata` constant-dim mismatch, `validate_interpolation_result` non-finite, `nearest_1d_point_helper` out-of-range/singleton, `interp_2d_1d` `QhullError → ValueError`) now use the new class. Subclassing `ValueError` keeps existing `except ValueError` callers working. Locked the contract down with a test in `test_perf_database_errors.py`.

jasonqinzhou (approved):

PR #1228, opened 2026-06-17
fix(sdk): stop double-counting MTP activation in KV-cache capacity estimate

Arsene12358:
## Summary

Fixes **AIC-1110**: on the Dynamo Mocker, the Kimi-K2.5 + GB200 DEP4 **prefill** worker failed to start with a *negative* KV-cache budget once Eagle (speculative decoding) was enabled.

## Root cause

The KV-cache capacity path (`sdk/memory.py`) sizes non-KV memory through `BaseBackend._get_memory_usage`, which unconditionally scaled activation memory by `(nextn + 1)` as an MTP correction. On that path `num_tokens` is the engine's `max_num_tokens` budget, which already caps the total tokens processed per forward pass (draft tokens included). Multiplying again by `(nextn + 1)` therefore double-counts: for Kimi-K2.5 + GB200 DEP4 prefill it pushed non-KV memory past GPU capacity, so `estimate_kv_cache` raised `no KV budget` and the prefill worker could not start. `nextn` changed *only* the activation term in this breakdown (weights, KV-per-token, and overhead are all constant), so the multiplier was its sole, spurious effect here.

Reproduction (Kimi-K2.5, GB200, attn_dp=4 / moe_ep=4, nvfp4, 16k prefill chunk), prefill KV budget:

| nextn | before | after |
|------:|-------:|------:|
| 0 | 17.1 GiB | 17.1 GiB |
| 1 | 3.0 GiB | 17.1 GiB |
| 2 | **negative → raises** | 17.1 GiB |
| 3 | **negative → raises** | 17.1 GiB |

## Fix

- `_get_memory_usage` gains an `mtp_activation_scaling: bool = True` flag; the `(nextn + 1)` multiplier is gated on it. The default preserves the latency sweep (static / agg / disagg) exactly.
- `KVCacheEstimator.from_request` passes `mtp_activation_scaling=False`. The capacity breakdown is now `nextn`-independent, consistent with the engine's per-iteration token cap. The draft module's marginal weight/KV cost is not separately modeled here — a small, conservative approximation versus the gross prior over-count.

## Testing

- New integration regression test (`test_eagle_prefill_kv_budget_nonnegative_kimi_gb200`) asserting a non-negative prefill KV budget across feasible Eagle draft lengths; it failed with `no KV budget` before the fix and passes after.
- 136 tests pass across the affected suites: `test_memory_estimation` (unit + integration), the dedicated `test_trtllm_kv_cache_capacity` (79), `test_base_backend`, and the rust↔python `test_engine_step_parity`.
- `ruff check` and `ruff format --check` clean on the changed files.

## Not in scope (potential follow-ups)

- The latency sweep's `static_ctx` prefill activation also applies `(nextn + 1)` to pure context tokens (arguably over-counts), but it is pre-existing, not what AIC-1110 reported, and changing it would shift sweep goldens — left for a separate change.
- Modeling the draft module's own weights in the capacity breakdown.

## Related

Sibling issue **AIC-1111** (Eagle raises predicted ITL) is a *separate* latency-fidelity investigation — the `(nextn + 1)` decode-batch scaling there is correct, and the high-batch inversion is largely physical. Details captured in that issue's thread.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected KV-cache memory estimation for speculative decoding to prevent double-counting activation scaling, which could previously lead to negative KV-cache capacity estimates for longer draft sequences.

* **Tests**
  * Added an integration regression test that verifies KV-cache estimation remains non-negative across multiple `nextn` values and that activation capacity (non-KV) remains consistent with the `nextn=0` baseline.

### DIFF
--- a/src/aiconfigurator/sdk/backends/base_backend.py
+++ b/src/aiconfigurator/sdk/backends/base_backend.py
@@ -1535,6 +1535,7 @@ class BaseBackend:
         prefix: int = 0,
         max_seq_len: int | None = None,
         encoder_memory: dict[str, float] | None = None,
+        mtp_activation_scaling: bool = True,
     ) -> dict[str, float]:
         """
         Get the memory usage of the backend.
@@ -1545,6 +1546,13 @@ class BaseBackend:
             max_seq_len: per-slot KV cache pre-allocation budget. Defaults to
                 ``isl + beam_width * osl`` when not supplied.
             encoder_memory: optional colocated encoder component to add to this worker.
+            mtp_activation_scaling: whether to scale activation by ``(nextn + 1)`` for
+                speculative decoding (see the MTP correction below). True for the
+                latency sweep, where ``num_tokens`` is the per-step token count that the
+                multiplier turns into the verified ``nextn + 1`` tokens. False for the
+                KV-cache capacity path, where ``num_tokens`` is the engine's
+                ``max_num_tokens`` budget that already caps total per-forward tokens
+                (draft tokens included), so re-multiplying would double-count.
         """
         weights = 0.0
         for op in model.context_ops:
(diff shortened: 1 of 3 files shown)


Arsene12358:
@coderabbitai review the PR

jasonqinzhou (approved):

PR #1229, opened 2026-06-17
fix: guard run_agg tokens_s_user against tpot==0 for osl=1

panpan0000:
When I tried to run Prefill-only simulation with `--osl 1 `:

```
python3 -m uv run aiconfigurator cli default   --system b300_sxm --model nvidia/GLM-5-NVFP4  --total-gpus 2   --backend vllm --isl 8000 --osl 1 
```

It fails:

```
16:20:43 [aiconfigurator] [main.py:1496] [E] No successful experiment runs to compare.

16:20:43 [aiconfigurator] [main.py:1498] [E] -> Experiment agg failed: No results found for any parallel configuration. Showing last exception: float division by zero

16:20:43 [aiconfigurator] [main.py:1498] [E] -> Experiment disagg returned no results. Possible causes: (1) TTFT/TPOT constraints are too tight — try relaxing --ttft or --tpot; (2) the model does not fit on the available GPUs — try increasing --total-gpus; (3) no perf data in the database for this configuration. 
```

#### Overview:

Fix a `ZeroDivisionError` in `BaseBackend.run_agg` when `osl=1`. With `osl=1` there is no decode step, so `tpot` becomes 0 and `tokens_s_user = 1000 / tpot` crashed the agg sweep. The same fallback (`0.0 if tpot == 0.0 else 1000.0 / tpot`) was already used in `run_static` further down in the same file, so the fix keeps the two paths consistent.

The bug also broke `aiconfigurator cli default` end-to-end: the `ZeroDivisionError` was caught in `pareto_analysis.agg_pareto` and re-raised as `RuntimeError: No results found for any parallel configuration. Showing last exception: float division by zero`, so the agg experiment returned no pareto and the whole default run reported `No successful experiment runs to compare.` Users hitting this could not run the default sweep with `osl=1`, which is a valid setup (no-decode / prefill-only benchmarking).

#### Details:

- **`src/aiconfigurator/sdk/backends/base_backend.py`** — In `run_agg`, guard `tokens_s_user` against `tpot == 0` so `osl=1` does not raise. The expression is now `tokens_s_user = 0.0 if tpot == 0.0 else 1000.0 / tpot`, matching the existing pattern in `run_static` (line ~550).
- **`tests/unit/sdk/backends/test_base_backend.py`** — New regression test `test_run_agg_with_osl_one_does_not_divide_by_zero` that runs `run_agg` with `osl=1`, `b=1` and asserts no `ZeroDivisionError` is raised and that `tokens/s/user` is `0.0`. Verified to fail on the unpatched code and pass after the fix. Also dropped the redundant `run_agg` override in the local `_TestBackend` helper so the test exercises the real `BaseBackend.run_agg` path.

1. **`src/aiconfigurator/sdk/backends/base_backend.py`** around the `run_agg` summary assembly (the single-line change near `tokens_s_user`).
2. **`tests/unit/sdk/backends/test_base_backend.py`** — the new regression test and the small `_TestBackend` cleanup that lets the test hit `BaseBackend.run_agg` directly.

- (Add a specific issue number if applicable, e.g. `Fixes #xxx`.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a throughput/user-throughput calculation that could produce errors or infinite values when decoding wasn’t occurring (e.g., when timing values were zero).
  * Ensured user throughput reports `0.0` for cases like single-step runs where the metric isn’t valid to compute.
* **Tests**
  * Added a regression test covering the single-step scenario to prevent division-by-zero behavior.


panpan0000:
@jasonqinzhou  @tianhaox  Can you please take a look? thank you

src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -1248,7 +1076,7 @@ def run_agg(
         moe_tp = model.config.moe_tp_size
         moe_ep = model.config.moe_ep_size
         tokens_s_gpu = output_throughput / pp / tp / dp
-        tokens_s_user = 1000 / tpot
+        tokens_s_user = 0.0 if tpot == 0.0 else 1000.0 / tpot
```
tianhaox: can we use osl to guard the tokens_sec_per_user?, osl > 1, 1000/tpot, else, 0.0

it seems like there's a bug in agg modeling. when bs > 1, osl==1, it will have positive tpot value. but it's out of the scope of this fix. then, better to use osl > 1 to guard.

panpan0000:
@tianhaox Thanks for the note (https://github.com/ai-dynamo/aiconfigurator/pull/1229#discussion_r3457947852). Agree: for osl=1 (prefill-only) there are no decode tokens, so tokens/s/user should be 0 even if the current agg tpot calc ends up >0 when bs>1. I updated tokens/s/user to be guarded by osl>1 (and kept a tpot>0 guard), and adjusted the unit test to cover bs>1 + osl=1 where tpot>0.

panpan0000:
Merged latest main into this branch to resolve the merge conflict. No functional change beyond conflict resolution.

tianhaox (approved):

PR #1238, opened 2026-06-22
feat(sdk): data-calibrated hybrid (util-space empirical) + MiniMax-M3/MSA support

tianhaox:
# Data-calibrated HYBRID mode: util-space empirical + fine-grained transfer policy

## Summary

This PR makes the **HYBRID / EMPIRICAL** prediction path *data-calibrated* instead of
relying on fixed per-op `scale_factor` constants, and turns the ways HYBRID fills a
coverage gap into **first-class, individually controllable "transfer kinds"**. It also
adds the first model that depends end-to-end on this path (**MiniMax-M3**, all-MoE +
sparse attention, no own silicon data) and surfaces, per support-matrix cell, whether a
prediction is pure silicon or hybrid-rescued and at which tier.

## Core idea: util-space empirical

Each op's empirical estimate is now `latency = SOL(query) / util`, where
`util = SOL / measured > 0` is read best-effort from real collected data. `util` is an
effective calibration factor, not a bounded physical efficiency: hardware/kernel effects
missing from the analytic SOL can make it exceed 1, and large values remain visible as
model/data-quality signals instead of being silently clamped. One-dimensional curves use
bracketing inverse-distance interpolation in log space; generic multi-dimensional grids
use normalised log-space nearest neighbours, and out-of-range queries freeze boundary util.
SOL is analytic and quant-aware, so it absorbs structural differences while `util` carries
kernel efficiency. This removes the catastrophic tail of the old fixed-constant fallback
(GEMM LOO mean MAPE 64–341% → 7–15%; util reconstruction is largely invariant to wrong
SOL constants because the ratio cancels).

When **no** calibration data exists for a slice (no own-shape, no transfer reference),
the path now raises `EmpiricalNotImplementedError` instead of fabricating a
`SOL / constant`. Coverage gaps surface honestly. Genuinely table-less ops
(mem / p2p / element-wise) keep their analytic formulas and never call `estimate()`.

## Transfer ladder + fine-grained policy

When an op's own slice has no data, HYBRID may borrow utilisation from a neighbour. The
borrow kinds are an enum (`common.TransferKind`), each independently gateable by a
**transfer policy**:

| kind | what it borrows | typical MAPE |
|---|---|---|
| `xshape` | nearest collected shape, same quant (MoE shape; attention head_size) | ~28% (MoE LOO) |
| `xquant` | a sibling quant in the **same** (memory, compute) profile | ~13–17% |
| `xprofile` | a quant in a **different** profile, rescaled by `e(query)/e(ref)` | ~24% (LOO) |
| `xop` | a related op's util (MSA ← DSA), via a manual level-alignment scale | model-dependent |

`transfer_policy` has `off` / `conservative` / `balanced` / `aggressive` presets and is
read at query time, so a cached DB can be retuned. Surfaces:

- **CLI**: `--transfer-policy` on the `default` / `estimate` subcommands.
- **YAML / API**: a flat `transfer_policy` field on the v2 `Task`.
- **support matrix**: `AIC_SM_TRANSFERS` env.

Sibling-version/framework shared-layer reuse is not a transfer kind. After #1219 it is
part of the collected-data SILICON path and is enabled for default/SILICON/HYBRID,
independently of empirical transfer policy.

### Why cross-profile cross-quant uses a single per-quant level (not a compute/mem SOL split)

A per-component (compute / mem) SOL split was implemented and then **reverted**: it adds
no value because (a) NN lookup naturally returns same-binding-regime samples, and (b) a
quant's compute and memory factors are correlated, so the roofline knee barely moves
across quants — the binding bound essentially never flips (0 / 84992 GEMM shapes for
bf16↔fp8). Cross-profile error is instead ~pure systematic *kernel-efficiency* bias,
which a single per-quant level `e(q)` corrects (58% → 24% LOO). `e(q)` ratios are
~stack-stable (w4a16/fp8 = 0.17 on b200 vs 0.18 on h100).

## Support matrix: silicon-first with hybrid rescue (default)

The support matrix runs each cell in **SILICON first**, including manifest-declared
shared collected rows. It retries only structured performance-data misses and explicitly
classified data/framework coverage gaps in HYBRID; arbitrary schema/programming errors and
hardware incompatibilities are not retried. `PASS` is reserved for collected SILICON data,
and a successful empirical rescue is reported separately as `HYBRID_PASS`:

| silicon | hybrid | `Status` | `Source` |
|---|---|---|---|
| PASS | — | `PASS` | `silicon` |
| structured/known gap | PASS, empirical tier fired | `HYBRID_PASS` | that tier (`xshape`/`xquant`/`xprofile`/`xop`) |
| structured/known gap | PASS, no finer tier | `HYBRID_PASS` | `empirical` |
| structured/known gap | FAIL | original SILICON status | empty |
| programming error / HW incompatible | — | original status | empty |

Shared-layer hits remain `silicon`: they are real collected rows merged before lookup,
and the loader does not retain per-row origin after the merge. Replay commands use the
mode that produced the row. `AIC_SM_ALLOW_HYBRID=0` disables rescue.

## MiniMax-M3 / MSA

Adds the `MSA` op (GQA sparse attention: per-block FP8 indexer + top-k sparse attn) and
the `MiniMax-M3` text model (all-MoE + MSA). It has no collected silicon data, so it runs
in HYBRID via cross-op transfer from DSA. Validated against the sglang B200 cookbook:
HYBRID TPOT 23.5 ms vs 24.1 ms measured (within 2%).

## Models unblocked by HYBRID (support-matrix both-fail rescues)

On the `b200/trtllm/1.3.0rc10` support matrix these models failed in **both** SILICON
and HYBRID; this PR fixes the modeling/validation gaps so they pass in HYBRID:

- **Nemotron-3 (MTP)** — Nemotron-3 ships `num_nextn_predict_layers=1`, which tripped a
  hard `assert self._nextn == 0` in the NemotronH model. MTP is now modeled with the
  DeepSeek-pattern generation-side scale factor `1/(1+E[accept]) · (nextn+L)/L`, applied
  to the generation path only (`nextn=0` is an exact 1.0 no-op, so non-MTP NemotronH is
  unchanged). All four Nemotron-3 variants (Super-120B-NVFP4, Ultra-550B BF16/FP8/NVFP4)
  go FAIL → HYBRID PASS.
- **Transfer-reachable MoE quant in validation** — `supported_quant_mode` is a
  *data-presence* list, not a backend-capability list. In SILICON it equals what we can
  model (stays strict), but in HYBRID/EMPIRICAL the MoE `xquant` path can synthesize a
  quant from a collected same-(memory,compute)-profile quant. Validation now admits a
  data-absent MoE quant when a supported quant shares its profile (MoE only — gemm and
  attention have no cross-quant path, so they stay strict and keep failing early rather
  than crashing late in the sweep). Kimi-K2.5 (`int4_wo`, profile shared with collected
  `w4a16_mxfp4`) goes FRAMEWORK_INCOMPATIBLE → HYBRID PASS while SILICON stays strict.

## Windowed (sliding-window) attention

Windowed attention prefers its real collected slice; when that slice is absent or too
sparse to interpolate, it derives from the `window=0` (full-attention) measurement scaled
by the window-aware SOL ratio (a bounded fallback, validated at ~25–50% vs measured). The
physical invariant *windowed ≤ full* holds by construction. This fixed a 683546 ms → 183 ms
TTFT blow-up for MiMo-V2-Flash that came from cubic extrapolation on a ragged borrowed
grid (now also guarded upstream by fail-accurately interpolation).

## Scope of calibration

util-space empirical is wired for: GEMM, context/generation attention (+ cross-head_size,
windowed), MoE (regular + WideEP + low-latency kernel selection), MLA, DSA, DSV4 (MHC /
context / generation), MSA, NCCL + custom_allreduce. The sibling-version shared layer is
a separate collected-data source used by default/SILICON/HYBRID and disabled in
formula-only EMPIRICAL/SOL.

## Boundary with SILICON

**SILICON stays collected-data-only**: it combines active-version rows with
manifest-declared shared rows from sibling versions/frameworks, then applies same-slice
interpolation and explicitly supported numeric boundary handling. It never uses SOL/util
transfer.

**HYBRID / util_empirical owns gap-filling**: cross-shape, windowed, cross-quant,
cross-profile, and cross-op transfer. A failed collected-data lookup is handed to HYBRID;
the request raises only if that path also has no coverage.

## Validation

The following offline study ran the same model and workload independently in explicit
SILICON and EMPIRICAL modes. Results use strict own-data EMPIRICAL
(`--transfer-policy off`), so no HYBRID or transfer fallback is included.
SILICON follows its production policy and includes manifest-declared shared collected rows;
formula-only EMPIRICAL loads only the active version.
APE is `abs(empirical - silicon) / silicon`; each cell is **mean / p90 / max APE (%)** with
the number of comparable points in parentheses. Unless marked, a row contains ten
deterministic irregular off-grid points per phase on B200/SGLang 0.5.10.

Across every phase and sample kind in the primary matrix, 518/546 silicon-eligible pairs
are comparable: **1.69% mean, 0.98% median, 4.69% p90, 8.68% max, and 1.13% WAPE**.

| model / configuration | type | quantization | prefill | decode | encoder |
|---|---|---|---:|---:|---:|
| **Primary off-grid aggregate** | mixed | mixed | **1.16 / 2.73 / 7.04 (n=185)** | **2.58 / 5.87 / 8.68 (n=170)** | — |
| Qwen3-32B | dense | BF16 | 0.72 / 1.34 / 1.53 (n=10) | 1.03 / 1.52 / 3.59 (n=10) | — |
| Qwen3-32B | dense | FP8 | 0.98 / 1.30 / 4.29 (n=10) | 2.28 / 4.70 / 6.82 (n=10) | — |
| Qwen3-32B | dense | FP8-static | 0.56 / 1.19 / 1.37 (n=10) | 2.23 / 2.65 / 3.38 (n=10) | — |
| Llama-3.1-70B | dense | FP8-static | 0.71 / 1.01 / 1.10 (n=10) | 3.38 / 3.73 / 3.87 (n=10) | — |
| Qwen3.5-27B⁴ | dense/GDN | BF16 | 4.68 / 5.33 / 5.46 (n=10) | 6.25 / 7.75 / 7.91 (n=10) | — |
| Qwen3-235B-A22B | MoE | BF16 | 0.53 / 0.40 / 2.71 (n=10) | 2.42 / 3.38 / 3.47 (n=10) | — |
| Qwen3-235B-A22B | MoE | FP8 | 0.59 / 0.83 / 2.25 (n=10) | 2.90 / 3.82 / 3.85 (n=10) | — |
| Qwen3-235B-A22B | MoE | NVFP4 | 0.26 / 0.57 / 0.67 (n=10) | 4.15 / 5.91 / 5.96 (n=10) | — |
| MiniMax-M2.5 | MoE | FP8-block | 0.24 / 0.30 / 1.07 (n=10) | 2.60 / 3.93 / 4.24 (n=10) | — |
| MiniMax-M2.5 | MoE | NVFP4 | 0.53 / 1.43 / 2.34 (n=10) | 3.67 / 6.52 / 7.04 (n=10) | — |
| Nemotron-Nano | hybrid/MoE | BF16 | 0.17 / 0.34 / 0.34 (n=10) | 3.43 / 5.53 / 7.28 (n=10) | — |
| Kimi-K2.5 | MLA/MoE | BF16 + INT4-WO | — | — | — |
| DeepSeek-V3.2 (TP2/PP2) | DSA | FP8-block | 2.62 / 5.91 / 5.91 (n=5) | 0.51 / 0.82 / 0.98 (n=10) | — |
| DeepSeek-V3.2 (TP8) | DSA | FP8-block | 2.00 / 6.40 / 7.04 (n=10) | 0.74 / 1.36 / 2.33 (n=14)¹ | — |
| GLM-5 (PP2) | DSA | BF16 | 1.30 / 2.20 / 2.25 (n=10) | 0.68 / 1.59 / 1.69 (n=10) | — |
| GLM-5 | DSA | FP8 | 1.34 / 2.00 / 2.23 (n=10) | 1.12 / 1.93 / 2.52 (n=10) | — |
| GLM-5 | DSA | NVFP4 | 2.10 / 3.05 / 3.63 (n=10) | 1.31 / 2.29 / 2.66 (n=10) | — |
| DeepSeek-V4-Pro (PP2) | DSV4 | FP8-block | 0.89 / 1.30 / 1.42 (n=10) | 3.73 / 6.26 / 8.68 (n=10) | — |
| DeepSeek-V4-Pro | DSV4 | FP8 + MXFP4/MXFP8 | 1.12 / 1.32 / 1.98 (n=10) | 2.20 / 4.93 / 5.36 (n=10) | — |
| DeepSeek-V4-Flash | DSV4 | FP8 + MXFP4/MXFP8 | 1.45 / 2.49 / 2.73 (n=10) | — | — |
| Qwen3-VL-32B image | dense/VL | BF16 | 0.58 (n=1)² | 0.29 (n=1)² | 5.06 (n=1)² |
| Qwen3-VL-30B-A3B image | MoE/VL | BF16 | 0.34 (n=1)² | 2.45 (n=1)² | 5.23 (n=1)² |
| **Cross-stack off-grid aggregate**³ | mixed | mixed | **0.89 / 1.50 / 9.87 (n=64)** | **1.07 / 1.83 / 3.25 (n=64)** | — |

1. The TP8 decode row uses all 14 boundary-util extrapolation probes; its ten irregular
   probes have 0.84% mean APE with the same 1.36% p90 and 2.33% max.
2. VL image values are one-point smoke checks, not distributions.
3. The separate stack sweep covers B200/H100/B300 and SGLang/TRT-LLM/vLLM. Its 9.87%
   maximum is the DSA `index_topk` transition; the B300/TRT-LLM MoE case has no SILICON
   reference and is excluded.
4. Qwen3.5's SILICON path reuses sibling-version collected GDN kernels; strict EMPIRICAL
   has no active-version GDN table and uses its analytic fallback.

The primary off-grid coverage is 185/195 (94.9%) for prefill and 170/180 (94.4%) for
decode. Five DeepSeek-V3.2 TP2/PP2 prefill probes are OOM. Kimi-K2.5 is a strict
EMPIRICAL own-data gap, while DeepSeek-V4-Flash decode has no SILICON reference; neither
is included in MAPE.

## Summary by CodeRabbit

* **New Features**
  * Added configurable transfer-policy controls (CLI + task/database support) to govern hybrid/empirical filling.
  * Added MiniMax-M3 model support, including a MiniMax Sparse Attention (MSA) performance modeling path.

* **Bug Fixes**
  * Replaced constant fallback estimate behavior with calibration-based, data-driven empirical/util estimation, and standardized “missing-data” failures.

* **Documentation**
  * Added a new design document covering hybrid/empirical prediction and transfer behavior.

* **Tests**
  * Expanded regression coverage for empirical fidelity, provenance/transfer gating, and cache/error handling.


tianhaox:
@coderabbitai review

tianhaox:
@coderabbitai full review

tianhaox:
@coderabbitai full review

tianhaox:
@coderabbitai full review

tianhaox:
It's a super long PR. Difficult to split the design.

src/aiconfigurator/sdk/errors.py
```diff
@@ -13,3 +13,20 @@ class UnsupportedWideepConfigError(ValueError):
...
+    """Raised when the empirical (SOL/util) path has no basis to estimate an op.
+
+    Distinct from ``PerfDataNotAvailableError`` (SILICON has no exact bracket but
+    HYBRID may still estimate): this is the terminal signal that even the
+    empirical fallback found nothing to calibrate from — no own-shape util, no
+    cross-shape/sibling transfer reference. We raise instead of returning a
+    placeholder ``SOL / constant`` so missing coverage surfaces honestly rather
+    than as a fabricated number. Genuinely table-less ops (mem / p2p /
+    element-wise) keep their own analytic formulas and never reach here.
+    """
```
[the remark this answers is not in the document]
tianhaox: Verified as already addressed in ffc14f3f. sdk.errors is now the single definition site for both PerfDataNotAvailableError and EmpiricalNotImplementedError. perf_database and all operation modules import or re-export those canonical classes, so every path uses the same exception identity.

tests/unit/cli/test_cli_api.py
```diff
@@ -114,6 +141,67 @@ def fake_run_disagg_estimate(**kwargs):
...
+
+        assert hybrid_off is not cached_db
+        assert hybrid_off.mode is common.DatabaseMode.HYBRID
+        assert hybrid_off.transfer_policy == frozenset()
+        assert silicon_default is not cached_db
+        assert silicon_default.mode is common.DatabaseMode.SILICON
+        assert silicon_default.transfer_policy == common.ALL_TRANSFERS
+        assert hybrid_default is not cached_db
+        assert hybrid_default.mode is common.DatabaseMode.HYBRID
+        assert hybrid_default.transfer_policy == common.ALL_TRANSFERS
```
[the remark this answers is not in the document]
tianhaox: Covered by the database-view tests. The CLI test now stubs get_database_view and no longer owns a shared cached database. test_database_view_configuration_is_isolated_and_same_key_is_reused explicitly creates configured views and then asserts that the root template remains SILICON with ALL_TRANSFERS. Repeating that assertion in the CLI stub test would only test the stub.

Arsene12358:
### Design clarification: should empirical calibration include synthesized extrapolation points?

Could you clarify the intended semantics for the empirical utilization grids here?

The design doc says utilization is read from the operation's “calibration rows” and that query axes are clamped to the “measured range.” However, several loaders currently extrapolate their canonical table in place before the empirical grid is built. For example:

- `extrapolate_data_grid()` [adds target coordinates in place](https://github.com/ai-dynamo/aiconfigurator/blob/9207a8b2c5a4d51f5b7df865ffdf04bd02454fcc/src/aiconfigurator/sdk/interpolation.py#L638-L646).
- GEMM [extrapolates `gemm_loaded`](https://github.com/ai-dynamo/aiconfigurator/blob/9207a8b2c5a4d51f5b7df865ffdf04bd02454fcc/src/aiconfigurator/sdk/operations/gemm.py#L208-L234), then its empirical path [builds from that same table](https://github.com/ai-dynamo/aiconfigurator/blob/9207a8b2c5a4d51f5b7df865ffdf04bd02454fcc/src/aiconfigurator/sdk/operations/gemm.py#L472-L490). Context/encoder attention and several MLA paths follow the same pattern.
- `build_samples()` [flattens every leaf without measured/synthesized provenance](https://github.com/ai-dynamo/aiconfigurator/blob/9207a8b2c5a4d51f5b7df865ffdf04bd02454fcc/src/aiconfigurator/sdk/operations/util_empirical.py#L113-L131), so synthesized rows participate as ordinary calibration samples.

A small example illustrates the behavioral difference. With measured points `x={1,2}`, latency `1`, and `SOL(x)=x`, extrapolating a row at `x=100` causes an empirical query at `100` to return utilization `100` and latency `1`. Building from measured rows only clamps to `x=2`, giving utilization `2` and latency `50`.

Is the intended design:

1. **Measured-only calibration:** empirical utilization is derived only from collected rows, while the extrapolated table remains available for SILICON queries; or
2. **Expanded-table calibration:** latency-space extrapolation is intentionally allowed to create additional empirical utilization samples?

There are examples of both patterns in this PR: GenerationAttention's own empirical grid and DSA preserve/use a pre-extrapolation raw view, while GEMM, ContextAttention, EncoderAttention, and the MLA paths generally use the extrapolated view. That makes the intended contract unclear.

If option 1 is intended, the existing DSA/GenerationAttention raw-cache pattern looks reusable. If option 2 is intended, could the design doc clarify that “calibration rows” and “measured range” include synthesized points, along with the rationale for deriving empirical utilization from extrapolated latency?

src/aiconfigurator/sdk/operations/communication.py
```diff
@@ -132,9 +135,33 @@ def get_sol(quant_mode: common.CommQuantMode, tp_size: int, size: int) -> tuple[
...
+            # Data-calibrated: util = SOL/measured read from the collected
+            # custom_allreduce curve (smooth in message size), reconstructed with
+            # this query's SOL. Falls back to the legacy 1/0.8 constant when the
+            # slice has no data. SOL uses the real tp_size; the util grid is built
+            # from the effective (node-capped) tp slice, so SOL ratio carries any
+            # multi-node bandwidth scaling.
+            sol_q = get_sol(quant_mode, tp_size, size)[0]
+            if tp_size <= 1 or sol_q <= 0:
+                return sol_q
+            eff = min(tp_size, database.system_spec["node"]["num_gpus_per_node"])
```
Arsene12358, line 148: **Blocker: CustomAllReduce bypasses the XSHAPE policy.** When `tp_size` exceeds the per-node count, `eff` selects another TP slice. Lines 161–163 explicitly classify that as `xshape`, but no policy check occurs, so `transfer_policy="off"` still returns a transferred estimate. Please reject the borrow before building the grid:

```python
if eff != tp_size and common.TransferKind.XSHAPE not in database.transfer_policy:
    raise EmpiricalNotImplementedError(
        f"No measured custom-allreduce calibration for tp_size={tp_size}; "
        "node-capped XSHAPE transfer is disabled by the transfer policy."
    )
```

Keep the existing `xshape` provenance when the transfer is enabled. Please also test `tp_size=16` with collected maximum `8` under both `off` and an XSHAPE-enabled policy.

src/aiconfigurator/sdk/operations/communication.py
```diff
@@ -350,9 +386,47 @@ def get_sol(
...
+            if not src.loaded:
+                raise EmpiricalNotImplementedError(
+                    f"No NCCL data to estimate {operation} ({dtype.value.name}, num_gpus={num_gpus})."
+                )
+            try:
+                by_op = require_nccl_bucket(
+                    util_empirical.require_data_slice(src, dtype, operation),
+                    f"dtype={dtype.value.name}, operation={operation!r}",
+                )
+                eff = min(num_gpus, max(by_op.keys()))
```
Arsene12358, line 419: **Blocker: NCCL has the same policy bypass.** For a 16-GPU query with data only through 8 GPUs, this sets `eff=8`, builds utilization from the 8-GPU curve, applies it to the 16-GPU SOL, and records `xshape`—even when the policy is `off`. After resolving `eff` and before `grid_for`, require XSHAPE permission:

```python
if eff != num_gpus and common.TransferKind.XSHAPE not in database.transfer_policy:
    raise EmpiricalNotImplementedError(
        f"No measured NCCL calibration for num_gpus={num_gpus}; "
        "GPU-count XSHAPE transfer is disabled by the transfer policy."
    )
```

Please cover the disabled and enabled cases, and retain `xshape` provenance only for the enabled overflow borrow.

Arsene12358 (changes requested):
Requesting changes because `transfer_policy="off"` does not currently disable communication XSHAPE transfers. I reproduced this for both CustomAllReduce and NCCL with collected counts `{1, 2, 4, 8}` and a requested count of `16`: each API returned an empirical value and recorded `xshape` even though the effective policy was an empty set.

The design contract says transfer kinds are independently enabled and a disabled kind is skipped or raises. Please gate both effective-count borrows on `TransferKind.XSHAPE` and add regression coverage for: exact count + `off` succeeds as own-data empirical; overflow count + `off` raises `EmpiricalNotImplementedError`; overflow count + XSHAPE enabled succeeds and records `xshape`.

This request applies to the EMPIRICAL transfer path. It does not require disabling the separate SILICON boundary-correction behavior. If rank overflow is instead intended to be non-transfer boundary handling, please reclassify it, remove the `xshape` provenance, and document that exception; the current state both calls it XSHAPE and bypasses the XSHAPE policy.

src/aiconfigurator/sdk/operations/moe.py
```diff
@@ -1813,8 +2210,50 @@ def get_empirical_from_sol(
                 quant_mode,
                 node_num,
             )[0]
-            scale_factor = 0.5
-            return latency / scale_factor
+            ks = cls._select_alltoall_kernel(database, quant_mode, moe_ep_size, topk, moe_backend=moe_backend)
+            tqm = cls._normalize_quant_mode_for_table(quant_mode)
+
+            def _slice():
+                if ks == "NotEnabled":
```
Arsene12358: **Blocker: a disabled operation is being converted into a calibration miss.** `ks == "NotEnabled"` is a no-op decision, not missing empirical coverage. Because SOL/SOL_FULL/EMPIRICAL dispatch occurs before the later zero guard, EMPIRICAL reaches this branch and ultimately raises, while SOL reports a positive hypothetical communication cost.

Please move kernel selection and the no-op return ahead of every mode branch, after `node_num` and `op_name` validation:

```python
kernel_source = cls._select_alltoall_kernel(
    database, quant_mode, moe_ep_size, topk, moe_backend=moe_backend
)
if kernel_source == "NotEnabled":
    if database_mode == common.DatabaseMode.SOL_FULL:
        return (0.0, 0.0, 0.0)
    source = "sol" if database_mode == common.DatabaseMode.SOL else "empirical"
    return PerformanceResult(0.0, energy=0.0, source=source)
```

Then remove this `_slice()` special case and the late duplicate guard, and pass/reuse `kernel_source` in `get_empirical_from_sol()`. Please test both disabled backends across all five modes and assert that `util_empirical.estimate()` is never called for `NotEnabled`.

Arsene12358 (changes requested):
Requesting changes for inconsistent handling of disabled TRT-LLM AllToAll kernels across database modes. For both `moe_backend="DEEPGEMM"` and `"CUTE_DSL"`, the selector returns `NotEnabled`, but the SOL, SOL_FULL, and EMPIRICAL branches return before the zero guard at lines 2308–2316. A mode-matrix reproduction produced zero for SILICON/HYBRID, a positive latency for SOL/SOL_FULL, and `EmpiricalNotImplementedError` for EMPIRICAL. Using actual DeepSeek-V3 dimensions on GB200, the direct API returned `0.000892 ms` in SOL, zero in SILICON/HYBRID, and raised in EMPIRICAL.

The normal `MoEDispatch` graph currently has an outer gate and falls back to NCCL for DeepGEMM/CuteDSL, so the immediate scope is the public operation API, diagnostics, and alternate/future callers. The query should nevertheless enforce its own kernel-eligibility invariant.

Please normalize the mode, compute `node_num`, validate `op_name`, select the AllToAll kernel, and handle `NotEnabled` before dispatching to any mode. Preserve `(0.0, 0.0, 0.0)` for SOL_FULL, return a zero `PerformanceResult` with `source="sol"` for SOL and `source="empirical"` for EMPIRICAL/SILICON/HYBRID, remove the now-duplicate late guard, and reuse the selected kernel in the empirical path. Add a backend × mode regression matrix for DEEPGEMM/CUTE_DSL and an enabled-kernel control.

Arsene12358 (approved):

Issue #901, opened 2026-04-23
[BUG] Qwen/Qwen3.5-397B-A17B 256K stress profile on sglang/h200_sxm hard-fails instead of returning no results


mvillmow:
### Describe the bug

`aiconfigurator cli default` hard-fails for the public model `Qwen/Qwen3.5-397B-A17B` on `h200_sxm + sglang 0.5.9` when I use the `256K` long-context RL stress profile:

- `ISL=229376`
- `OSL=32768`
- `total-gpus=8`
- `database-mode=HYBRID`
- relaxed latency bounds: `TTFT=60000`, `TPOT=1000`

On the same setup, the `128K` stress profile succeeds:

- `ISL=114688`
- `OSL=16384`

So this looks like a long-context failure mode rather than a generic model/backend incompatibility.

The failure mode is not a clean "unsupported / no results" outcome. Instead, `cli default` throws:

- `KeyError: 229376` in `sglang_backend.py` when looking up agg cache coverage
- then falls back into a path that eventually hits
  - `AssertionError: We don't enable the path for non-wideep SGLang to support TP>1 and DP>1 for attn simultaneously`
- both agg and disagg then fail with
  - `RuntimeError: No results found for any parallel configuration. Showing last exception: We don't enable the path for non-wideep SGLang to support TP>1 and DP>1 for attn simultaneously`

The net effect is that a profile-aware sweep aborts instead of producing a clean "no result for this profile" outcome.

### Steps to reproduce

Environment:

- AIConfigurator commit: `a8b0934c4717b993c41dd3938fee15ae6fc0d8c6`
- backend: `sglang`
- backend version: `0.5.9`
- system: `h200_sxm`
- model: `Qwen/Qwen3.5-397B-A17B`

Repro command:

```bash
$AIC_VENV/bin/python $AIC_ROOT/src/aiconfigurator/main.py cli default \
  --model Qwen/Qwen3.5-397B-A17B \
  --system h200_sxm \
  --total-gpus 8 \
  --backend sglang \
  --backend-version 0.5.9 \
  --generated-config-version 0.5.9 \
  --database-mode HYBRID \
  --isl 229376 \
  --osl 32768 \
  --ttft 60000 \
  --tpot 1000 \
  --save-dir /tmp/aic_qwen_issue_256k
```

Contrast case that succeeds:

```bash
$AIC_VENV/bin/python $AIC_ROOT/src/aiconfigurator/main.py cli default \
[...]

PR #1239, opened 2026-06-23
fix: report expected "no results" sweep outcomes cleanly

tianhaox:
## Summary

Intentional "no results" sweep outcomes — out-of-memory, KV-cache capacity, and missing perf data — were raised as a bare `RuntimeError` and printed with a full Python traceback. An expected failure therefore looked like a crash and was indistinguishable from a real bug in the logs.

This adds a `NoResultsError(RuntimeError)` base and groups the expected no-result exceptions under it (`NoFeasibleConfigError`, `PerfDataNotAvailableError`, plus new `InsufficientMemoryError` / `KVCacheCapacityError` for the OOM / KV-cache cases). The CLI experiment handler now classifies failures with a chain-walking `is_expected_no_result_cause`: expected outcomes are logged cleanly, while genuine bugs keep their traceback.

Error messages are unchanged and `NoResultsError` subclasses `RuntimeError`, so existing `except RuntimeError` / `except Exception` callers are unaffected.

## Testing

- New unit tests for the exception hierarchy and the chain-walking recognizer (direct member, generic wrapper chained from a family member, wrapper around a real bug, cycle-safety).
- Manual repro: an OOM sweep now prints 0 tracebacks (was a full traceback), still exits non-zero with an actionable message; a missing-perf-data sweep stays clean; a genuine bug still surfaces its traceback.

Relates to #901

## Summary by CodeRabbit

* **New Features**
  * Better classification of “no results” sweep outcomes, with clearer error handling for memory limits and infeasible configurations.
  * More cases now report a clean failure message instead of a full traceback when the outcome is expected.

* **Bug Fixes**
  * Improved detection of expected failure chains, including wrapped and nested exceptions.
  * Prevented incorrect handling of cyclic exception references during failure detection.

* **Tests**
  * Added coverage for the new no-results error categories and sweep outcome handling.


src/aiconfigurator/sdk/sweep.py
```diff
@@ -510,12 +514,12 @@ def sweep_agg(
...
         ) from exceptions[-1]
     if all_configs_oom:
-        raise RuntimeError(
+        raise InsufficientMemoryError(
             "sweep_agg: no results — model does not fit in GPU memory for any parallel config. "
             "Try increasing --total-gpus, using a quantized model, or a system with more VRAM per GPU."
         )
     if all_kv_cache_oom:
-        raise RuntimeError(
+        raise KVCacheCapacityError(
```
jasonqinzhou, line 528: **[P1] Preserve separate model-OOM, KV-cache, and SLA outcomes before assigning these exception types.**

`_sweep_one_parallel_agg` currently treats `summary.check_oom() or summary.check_kv_cache_oom()` as one `all_oom` bit, while this caller clears `all_kv_cache_oom` only after an SLA-feasible row exists. That reverses the important terminal cases: pure KV-cache exhaustion reaches the earlier `InsufficientMemoryError`, while evaluated non-OOM points that merely miss TTFT/TPOT reach this `KVCacheCapacityError`; the final `NoFeasibleConfigError` is unreachable on the normal no-exception/no-results path. I reproduced the SLA case by returning `(empty_df, False)` from the helper.

Please have the helper return whether it saw any model-fit point and any fully memory-fit point, then OR those signals across parallel configurations here. The key helper logic should be equivalent to:

```python
model_oom = summary.check_oom()
kv_cache_oom = summary.check_kv_cache_oom()
saw_model_fit |= not model_oom
saw_memory_fit |= not model_oom and not kv_cache_oom
if model_oom or kv_cache_oom:
    break
```

The terminal mapping then becomes: no model-fit point -> `InsufficientMemoryError`; model fit but no fully memory-fit point -> `KVCacheCapacityError`; at least one fully memory-fit point but no SLA row -> `NoFeasibleConfigError`. Please lock down those three cases with focused tests. This spans the helper, caller, and tests, so a one-hunk GitHub suggestion would not be safe.

src/aiconfigurator/sdk/errors.py
```diff
@@ -4,12 +4,64 @@
...
+        current = stack.pop()
+        if id(current) in seen:
+            continue
+        if isinstance(current, NoResultsError):
+            return True
+        seen.add(id(current))
+        if current.__cause__ is not None:
+            stack.append(current.__cause__)
+        if current.__context__ is not None:
+            stack.append(current.__context__)
```
jasonqinzhou, line 91: **[P2] Respect the effective exception chain before suppressing a traceback.**

This currently searches `__context__` even when an explicit cause replaces it or `raise ... from None` suppresses it. A real `KeyError` raised while handling `InsufficientMemoryError` is therefore classified as expected and logged without its traceback. Mirroring Python's chaining rules preserves intentional explicit wrappers while ignoring hidden context.

```suggestion
        if current.__cause__ is not None:
            stack.append(current.__cause__)
        elif not current.__suppress_context__ and current.__context__ is not None:
            stack.append(current.__context__)
```

tests/unit/sdk/test_errors.py
```diff
@@ -0,0 +1,83 @@
...
+    except RuntimeError as exc:
+        assert is_expected_no_result_cause(exc)
+
+
+def test_recognizer_rejects_wrapper_around_real_bug() -> None:
+    # A genuine bug surfacing as the last exception must keep its traceback.
+    try:
+        raise RuntimeError("no results for any config") from KeyError("unexpected")
+    except RuntimeError as exc:
+        assert not is_expected_no_result_cause(exc)
```
jasonqinzhou, line 99: **Add a regression test proving that a genuine bug with suppressed expected context keeps its traceback classification.**

This complements the implementation suggestion in `errors.py` and directly protects the PR's promise that genuine bugs remain visible.

```suggestion
        assert not is_expected_no_result_cause(exc)


def test_recognizer_ignores_suppressed_expected_context() -> None:
    try:
        try:
            raise InsufficientMemoryError("expected no-result")
        except InsufficientMemoryError:
            raise KeyError("real bug while handling expected result") from None
    except KeyError as exc:
        assert exc.__suppress_context__
        assert not is_expected_no_result_cause(exc)
```

src/aiconfigurator/sdk/sweep.py
tianhaox: Thanks for catching this. Addressed in 5e197f50: `_sweep_one_parallel_agg` now returns separate `saw_model_fit` and `saw_memory_fit` signals, and `sweep_agg` ORs them across runtime and parallel configurations before mapping the terminal error. Added a focused parameterized test covering model OOM, KV-cache OOM, and memory-fit/SLA-only misses, including preservation of an earlier fit signal across a later OOM configuration.

src/aiconfigurator/sdk/errors.py
tianhaox: Thanks, fixed in 5e197f50. `is_expected_no_result_cause` now follows an explicit `__cause__` first, otherwise only an unsuppressed `__context__`, matching the effective exception chain used by Python. Regression coverage now verifies explicit-cause precedence and suppressed-context handling while preserving implicit-context traversal.

tests/unit/sdk/test_errors.py
tianhaox: Added in 5e197f50. The regression test reproduces `raise KeyError(...) from None` while handling `InsufficientMemoryError`, asserts that context is suppressed, and verifies the exception remains classified as a genuine error. I also added coverage for explicit-cause precedence and the positive implicit-context case.

src/aiconfigurator/cli/main.py
```diff
@@ -1564,7 +1568,9 @@ def _execute_tasks(
             logger.warning(msg)
             failure_messages.append(msg)
         except Exception as exc:
-            if perf_database.has_perf_data_not_available_cause(exc):
+            if is_expected_no_result_cause(exc) or perf_database.has_perf_data_not_available_cause(exc):
```
jasonqinzhou, line 1571: **[P1] Apply effective-chain semantics to the perf-data recognizer before suppressing tracebacks.** This new OR still classifies a genuine error as expected whenever it was raised while handling `PerfDataNotAvailableError`, because `has_perf_data_not_available_cause` follows `__context__` even when an explicit cause replaces it or `raise ... from None` suppresses it. At this head I reproduced a suppressed `KeyError` with `is_expected_no_result_cause=False` and `has_perf_data_not_available_cause=True`, so this condition is true and `logger.exception` is skipped. Please update that helper to follow `__cause__` first, otherwise only unsuppressed `__context__`, and add explicit-cause/suppressed-context regression tests. Until both classifiers use the same effective chain, the PR promise that genuine bugs retain tracebacks is not met.

jasonqinzhou (commented):
The three earlier findings are fixed. One related traceback-classification path still needs correction before approval.

src/aiconfigurator/cli/main.py
jasonqinzhou: Agreed to accept this as a scoped follow-up. I am resolving this thread and approving/merging #1239 because its intended fixes are otherwise verified. I will immediately open a dedicated PR that updates `has_perf_data_not_available_cause` to use effective-chain semantics and adds explicit-cause plus `raise ... from None` regression coverage, then link that PR back here.

jasonqinzhou (approved):
Approved. The reported sweep-classification and exception-chain fixes are verified, targeted tests and all required checks pass, and the remaining pre-existing perf-data chain-walker issue is explicitly deferred to an immediate follow-up PR.

jasonqinzhou:
Follow-up opened: #1270 fixes the deferred perf-data exception-chain issue and adds the explicit-cause plus suppressed-context regression coverage discussed above.

PR #1250, opened 2026-06-25
fix(dsa): extrapolate per-backend sub-dicts instead of backend-keyed outer dict

Magitoneu:
#### Overview:

`ContextDSAModule._extrapolate` and `GenerationDSAModule._extrapolate` were passing the `{dsa_backend: {...}}` arch_dict directly to `interpolation.extrapolate_data_grid`, treating backend names (`"flashmla_kv"`, `"trtllm"`) as grid axis keys instead of descending into each backend's `{num_heads: ...}` sub-dict. This caused silently incorrect extrapolation at wrong nesting levels.

#### Details:

- **Root cause:** The `dsa_backend` level was added to the data structure in both `load_context_dsa_module_data` (dsa.py:1243) and `load_generation_dsa_module_data` (dsa.py:1293), but `_extrapolate` was never updated to descend through it. The query path already handled this correctly via `_select_dsa_backend` (dsa.py:99).
- **Why no KeyError:** `extrapolate_data_grid` is designed to tolerate missing data — it uses `nearest_1d_point_helper` to find existing keys and logs warnings instead of raising. With the wrong nesting level, it silently produced incorrectly extrapolated data.
- **Fix:** Added an inner loop `for dsa_backend_key in list(arch_dict.keys())` in both `_extrapolate` methods so extrapolation operates on each backend's sub-dict.
- **Docstrings:** Updated stale docstrings in both `_extrapolate` methods and both `load_*_dsa_module_data` functions to include the `dsa_backend` level.

- `src/aiconfigurator/sdk/operations/dsa.py` — `_extrapolate` methods (lines ~255 and ~906)
- Verify the data structure at `load_context_dsa_module_data` (dsa.py:1243) and `load_generation_dsa_module_data` (dsa.py:1293)

- Relates to GitHub issue: #1223

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed DSA extrapolation for datasets containing multiple backends, ensuring results stay correctly grouped under each backend.
  * Updated extrapolation to traverse the full nested structure (including backend and grid levels), with correct handling of prefix-specific extrapolation when applicable.
* **Documentation**
  * Refreshed CSV/data-loading docs to match the updated key layout, including the backend dimension.
* **Tests**
  * Added regression tests covering multi-backend extrapolation for both context and generation scenarios, verifying backend-specific outputs at the expected coordinates.


YijiaZhao:
This fixes a **silent**-correctness bug (extrapolation ran at the wrong nesting level — backend names treated as grid keys — and `extrapolate_data_grid` only warns, never raises), but there's no regression test guarding it.

Could we add a unit test that:
1. builds a `data_wrapper` with **two** `dsa_backend` keys (e.g. `flashmla_kv` and `trtllm`) under one `arch`, each with its own `{num_heads: ...}` grid;
2. runs `_extrapolate`;
3. asserts each backend's grid is extrapolated **independently** over `num_heads` (and that the backend names are not consumed as the grid x-axis / num_heads)?

Same for both `ContextDSAModule._extrapolate` and `GenerationDSAModule._extrapolate`. Since the bug was silent, a test is really what locks the fix in and prevents regressing back to the wrong nesting level.

jasonqinzhou (approved):

PR #1294, opened 2026-07-02
fix(generator): preserve disagg system name in Task bridge

simone-chen:
#### Overview:

The Task→generator bridge set `NodeConfig.system_name` from the shared top-level `task_config.system_name`. For disaggregated Task v2 requests that field is empty (the selected system lives in `prefill_system_name`/`decode_system_name`), so `NodeConfig.system_name` came out blank, `request_resolution` found no system, and hardware facts (node selectors, NCCL/UCX env) were dropped — workers could schedule on the wrong architecture (e.g. a `--system gb200` request ran on H200).

#### Details:

- Read `task_config.primary_system_name` instead, which returns the top-level system for agg and the prefill-side system for disagg.
- Same-system disagg only; heterogeneous prefill/decode placement remains a separate future concern (as noted in the bug report).

- `src/aiconfigurator/generator/module_bridge.py` — one-line change near the end of `task_config_to_generator_config`.
- `tests/unit/generator/test_module_bridge_system_name.py`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected system name propagation into generated template node configuration, preventing missing system details in disaggregated setups.
  * Improved disaggregated validation to ensure prefill and decode system identifiers are consistent; mismatches now return a clear error.
* **Tests**
  * Added/updated unit tests covering aggregated vs. disaggregated behavior to verify node system naming and the new mismatch error handling.

### DIFF
--- a/src/aiconfigurator/generator/module_bridge.py
+++ b/src/aiconfigurator/generator/module_bridge.py
@@ -275,8 +275,25 @@ def task_config_to_generator_config(
     )
 
     params = _deep_merge(params, overrides.get("Params"))
-    # Expose SDK's system identifier to templates via NodeConfig.system_name
-    params.setdefault("NodeConfig", {})["system_name"] = task_config.system_name
+    # Expose SDK's system identifier to templates via NodeConfig.system_name.
+    # Use primary_system_name: for disagg, the shared top-level system_name is
+    # empty (phase-specific prefill/decode systems carry the value), so reading
+    # system_name here would blank NodeConfig and drop all hardware facts.
+    #
+    # NodeConfig.system_name (and the hardware facts derived from it) is global,
+    # while disagg prefill/decode can request different systems. Heterogeneous
+    # placement is out of scope here, so fail fast rather than silently applying
+    # the prefill system's node selectors/env to both workers.
+    system_name = task_config.primary_system_name
+    if task_config.serving_mode == "disagg":
+        prefill_system = getattr(task_config, "prefill_system_name", system_name)
+        decode_system = getattr(task_config, "decode_system_name", system_name)
+        if prefill_system != decode_system:
+            raise ValueError(
+                "Generator artifacts currently require matching prefill/decode systems; "
+                f"got prefill={prefill_system!r}, decode={decode_system!r}"
+            )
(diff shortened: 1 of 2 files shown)


src/aiconfigurator/generator/module_bridge.py
```diff
@@ -275,8 +275,11 @@ def _build_worker_params(prefix: str, extra_overrides: dict | None) -> tuple[dic
     )
 
     params = _deep_merge(params, overrides.get("Params"))
-    # Expose SDK's system identifier to templates via NodeConfig.system_name
-    params.setdefault("NodeConfig", {})["system_name"] = task_config.system_name
+    # Expose SDK's system identifier to templates via NodeConfig.system_name.
+    # Use primary_system_name: for disagg, the shared top-level system_name is
+    # empty (phase-specific prefill/decode systems carry the value), so reading
+    # system_name here would blank NodeConfig and drop all hardware facts.
+    params.setdefault("NodeConfig", {})["system_name"] = task_config.primary_system_name
```
jasonqinzhou: `primary_system_name` is always the prefill system for disagg, while these hardware facts are global. With GB200 prefill and H200 decode, both generated workers receive the GB200/arm64 selectors. Since heterogeneous placement is explicitly out of scope here, can we fail fast instead of silently generating the wrong placement?

```suggestion
    system_name = task_config.primary_system_name
    if task_config.serving_mode == "disagg":
        prefill_system = getattr(task_config, "prefill_system_name", system_name)
        decode_system = getattr(task_config, "decode_system_name", system_name)
        if prefill_system != decode_system:
            raise ValueError(
                "Generator artifacts currently require matching prefill/decode systems; "
                f"got prefill={prefill_system!r}, decode={decode_system!r}"
            )
    params.setdefault("NodeConfig", {})["system_name"] = system_name
```

Please add a regression test for differing prefill/decode systems.

simone-chen: Done — added the fail-fast: for disagg, if `prefill_system_name != decode_system_name` we now raise `ValueError` (heterogeneous placement stays out of scope) instead of silently applying the prefill system to both workers. Added `test_disagg_heterogeneous_systems_raise`. Also merged latest main.

jasonqinzhou (approved):

PR #1295, opened 2026-07-02
fix(generator): carry multimodal image workload into BenchConfig

simone-chen:
#### Overview:

For Qwen3-VL image workloads (`--image-height/--image-width/--num-images`), the image config was preserved in `exp_config.yaml` but lost entering the generator bridge. The bridge built `BenchConfig` only from explicit overrides, so `image_batch_size` fell back to the schema default `0` and `bench_run.sh` / `k8s_bench.yaml` ran a text-only benchmark.

#### Details:

- Seed `BenchConfig` from the Task image fields (`image_height`, `image_width`, `num_images_per_request`) when both dimensions are set.
- Deep-merge explicit `BenchConfig` overrides afterward so user values still win.
- Fields read via `getattr` so text-only and duck-typed callers are unaffected (`image_batch_size` stays 0).

- `src/aiconfigurator/generator/module_bridge.py` — `bench_cfg` seeding block.
- `tests/unit/generator/test_module_bridge_image_workload.py` — image, override-wins, and text-only cases.

## Summary by CodeRabbit

* **New Features**
  * Image-based workloads now auto-populate generated benchmark configuration with image batch size and dimension mean values (when image height and width are positive).

* **Bug Fixes**
  * Benchmark configuration overrides are now deep-merged on top of image-derived defaults, preserving derived fields not explicitly overridden.
  * Text-only and zero-image cases keep image batch settings at zero and avoid adding image dimension mean values.

* **Tests**
  * Added unit tests covering default propagation, override precedence, and zero-image/text-only edge cases.

### DIFF
--- a/src/aiconfigurator/generator/module_bridge.py
+++ b/src/aiconfigurator/generator/module_bridge.py
@@ -252,7 +252,27 @@ def task_config_to_generator_config(
         "tpot": _safe_float(_series_val(result_df, "tpot", task_config.tpot), task_config.tpot),
     }
     sla_cfg = _deep_merge(sla_cfg, overrides.get("SlaConfig"))
-    bench_cfg = overrides.get("BenchConfig")
+
+    # Seed BenchConfig from the Task's multimodal image workload so image args
+    # reach bench_run.sh / k8s_bench.yaml. Without this, image_batch_size falls
+    # back to the schema default 0 and the templates emit a text-only benchmark
+    # even for image workloads. Explicit BenchConfig overrides win via merge.
+    image_height = getattr(task_config, "image_height", 0) or 0
+    image_width = getattr(task_config, "image_width", 0) or 0
+    # Default only a missing/None image count to 1. A deliberate
+    # num_images_per_request=0 ("disable the image encoder", used with 448x448
+    # dimensions by the web UI) must survive so the templates omit image args.
+    image_batch_size = getattr(task_config, "num_images_per_request", None)
+    if image_batch_size is None:
+        image_batch_size = 1
+    bench_cfg: dict[str, Any] = {}
+    if image_height > 0 and image_width > 0:
+        bench_cfg = {
+            "image_batch_size": image_batch_size,
+            "image_width_mean": image_width,
+            "image_height_mean": image_height,
+        }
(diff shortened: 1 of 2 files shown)


src/aiconfigurator/generator/module_bridge.py
```diff
@@ -252,7 +252,21 @@ def _build_worker_params(prefix: str, extra_overrides: dict | None) -> tuple[dic
...
+    # Seed BenchConfig from the Task's multimodal image workload so image args
+    # reach bench_run.sh / k8s_bench.yaml. Without this, image_batch_size falls
+    # back to the schema default 0 and the templates emit a text-only benchmark
+    # even for image workloads. Explicit BenchConfig overrides win via merge.
+    image_height = getattr(task_config, "image_height", 0) or 0
+    image_width = getattr(task_config, "image_width", 0) or 0
+    bench_cfg: dict[str, Any] = {}
+    if image_height > 0 and image_width > 0:
+        bench_cfg = {
+            "image_batch_size": getattr(task_config, "num_images_per_request", 1) or 1,
```
jasonqinzhou: An explicit `num_images_per_request=0` is supported as “disable the image encoder” (and is used with 448×448 dimensions by the web UI). `or 1` changes that zero to one and emits image arguments in both benchmark artifacts. Default only missing/`None` values so zero survives.

```suggestion
    image_batch_size = getattr(task_config, "num_images_per_request", None)
    if image_batch_size is None:
        image_batch_size = 1
    bench_cfg: dict[str, Any] = {}
    if image_height > 0 and image_width > 0:
        bench_cfg = {
            "image_batch_size": image_batch_size,
```

Please add a regression test with dimensions set and image count zero, asserting the image blocks remain absent.

simone-chen: Done — now default only a missing/None count to 1, so an explicit `num_images_per_request=0` (with dimensions set) survives and `image_batch_size` stays 0, leaving the image args omitted. Added `test_explicit_zero_image_count_disables_encoder_even_with_dimensions` (448x448 + count 0). Also merged latest main.

jasonqinzhou (approved):

PR #1303, opened 2026-07-04
refactor(sdk): perf-table interpolation engine v2 (perf_interp)

tianhaox:
## What

Clean-room redesign of the SILICON perf-table interpolation/extrapolation layer, in BOTH engines: one shared N-axis resolver (`sdk/perf_interp/` in Python, `perf_database/perf_interp.rs` in the compiled core) driven by declarative per-op configs. **All op families resolve on it** — GEMM, context/generation/encoder attention, the full MLA family (incl. MLAModule + WideEP), DSA (4-axis past-KV), DeepSeek-V4 CSA/HCA + mHC + MegaMoE, MoE main/wideep/DeepEP/alltoall, Mamba2/GDN, NCCL/custom-allreduce, MLA-BMM, GEMM scales. Load-time grid pre-expansion (`extrapolate_data_grid`), the dsv4 robust-lookup layer, and the old scalar interp family (both languages) are deleted entirely.

**Scope evolution during review-readiness** (each item detailed in its section below):
1. Python redesign + validation campaign (the original scope).
2. Rust port of the same engine — required because the compiled engine-step backend is the wheel's default; shipping the Python side alone would make the wheel disagree with the SDK.
3. Two real cross-engine bugs found and fixed by per-op accounting along the way: the MoE `attention_dp` token multiplier (independently also fixed upstream in #1313 — deduped after the merge) and the silently-wrong cp>1 evaluation (now guarded for sparse attention; dense CP is modeled via #1313).
4. Reconciled with upstream #1313 (merged in): their `PerfDbSources` shared-layer loaders + dense-CP Phase 1 as the base, this PR's v2 query semantics on top. Their shared-layer loaders closed the last known python/rust residual.

Follow-ups tracked in #1333 (sparse-CP port, CP calib data, parity-suite hardening).

## Goals (priority order)

- **P0-1 — robust ragged/asymmetric handling, especially GEMM.** GEMM's (n,k) are scattered real matmul shapes; dense m-sweeps form no Cartesian product. The engine ingests them natively (site curves + nearest-site util transfer) and degrades gracefully on missing points instead of crashing or rectangularizing.
- **P0-2 — utilization-preserving extrapolation.** Beyond the collected range: hold the boundary util (`util = SOL/latency`, k_tail-median anchor), let the analytic SOL carry the growth. Unbounded by design. This replaced several hand-rolled copies of the same idea (DSA `boundary_util_value`, MoE `_estimate_overflow_with_last_token_util`, scale_matrix's SOL-ratio clamp) and several raw-linear extrapolations that could undershoot launch floors or go negative.
- **P1** — per-axis value transforms, energy-as-power blending, bounded caches.

## Design (full rationale in `perf_interp/config.py`)

Resolution: **exact-hit (measured points verbatim — preserves GEMM-m sawtooth) -> in-data resolve -> boundary util-hold -> structured miss**.

| Resolver | Shape | Notes |
|---|---|---|
| `ScatteredSites(site_axes, curve_axis)` | GEMM (n,k)+m | exact site -> own curve; unknown shape -> log-space IDW util transfer (coverage-filtered, 2.0-octave gate) |
| `Grid()` | everything else, 1..N axes | nested bracket+blend, ragged-branch drop; truncated corners = ordinary out-of-range util-hold |

Key empirical findings baked in:
- **Curvature is per-axis** (`transform_axis`): sqrt on seq only — global sqrt 9.4% vs per-axis 2.0% interior on real data; sqrt wins on the seq axis itself (3.2% vs raw 6.0% in seq-row folds). Recipe: sqrt-on-seq for context-type ops (~seq²); RAW everywhere else.
- **DSA/CSA top-k regime piecewise removed** after archaeology (PR #903's motivation was cubic ringing + pre-expansion pollution — both root causes are structurally gone) and LOO on every real config (plain linear crossing ties/beats it; CSA gb200: 1.72% vs 1.92%).
- Same op can be 3-axis (legacy) or 4-axis (past-KV) per slice — detected at query time per (quant, arch, backend) slice.
- Distinct beyond-range contracts preserved where deliberate: compute_scale (a quantization DELTA) stays FLAT-clamped; mamba falls back to plain SOL; MoE singleton-underflow still raises.

## Leave-one-out evidence (real tables; truth = measured latency)

| Fold | median / p90 |
|---|---|
| GEMM interior / frontier util-hold / **site-holdout** (drop a whole shape) | 4.87% / **0.96%** / **3.63%** (0 miss) |
| attention interior / seq-row (sqrt vs raw) / frontier | 2.00% / **3.21% vs 5.99%** / 6.6% |
| context MLA interior / frontier | 2.19% / 5.47% |
| context DSA (4-axis leaf holdout) interior / frontier | 5.4% / 10.8% (0 miss; sparse s-grid around the top-k knee — the signed knee fold shows no systematic bias: +1.0% vs +1.2% control) |
| CSA context (gb200): plain vs regime-aware | **1.72% vs 1.92%** (knee-just-above +0.57% vs −2.94%) |

Validation contract: LOO vs measured only; deviation from the previous system is a sanity signal (single-digit % expected), not a guardrail.

## Net effect on the codebase

- **Deleted**: `extrapolate_data_grid` + 12 dead `_extrapolate` methods + 18 pre-expansion grid constants + the top-k piecewise pair + the MoE overflow estimator + `_dsv4_robust_3d_lookup`/`_dsv4_lookup_prefix_resolved` + the raw-vs-expanded table juggling (deepcopies -> aliases) + 15 white-box tests of removed machinery (their contracts are engine-native, covered by `test_perf_interp_engine.py`).
- **`interpolation.py` is down to the error class + `get_value` (548 → 31 lines) and scipy is out of the wheel dependencies.** The last consumer, `_lookup_sparse_kernel` (dsv4 CP kernel-DELTA correction), turned out to HAVE a natural SOL — the kernel scores (past + isl/2) KV pairs per query token, so `SOL = b·(past·isl + isl²/2)` follows from the code's own premises. The migration fixes a real bug: the sparse sweep stops at isl=8k while the CP path queries `mqa_full` at full sequence length (128k+). Hold-out A/B across all 288 collected slices (gb200/gb300/b200/b300, every heads/tp/past_kv/batch), predicting the frontier from a table truncated to max_isl/4 (4× extrapolation — still short of the 16× CP asks for): legacy linear-trend median **−49%** (p10 −70%), flat-hold **−86%**, util-hold under the pair-count SOL **−1%** (measured latency/isl² flattens from isl≥2048; the linear-trend error grows with extrapolation distance). The legacy largest-lower-batch × bs/bp scaling becomes measured batch brackets. Also retired with it: `_extracted_metrics_cache` plumbing and the dead `_is_dsa_interpolation_miss` recognizer.
- Site-index cache is a bounded LRU (cap 32) with a documented table-immutability contract.
- **Task-level wall time: ~5.5x faster.** `aic default` (Qwen3-32B, h100_sxm/sglang, 8 GPUs, full agg+disagg sweep): 26-29s on main -> **4.5-4.9s** on this branch, byte-comparable results (pareto rows match at median 0.3% / max 1.8% delta -- inside the interpolation-change envelope). Repeated in-process runs scale the same on both (no hidden warm-cache dependency). The win is query-path cost: scipy cubic/griddata over pre-expanded grids -> dict-walk bracket+blend.
- `interpolation.py` is deleted outright: `InterpolationDataNotAvailableError` -> `sdk/errors.py`, `get_value` -> `perf_interp`.

## Final validation campaign (what flipped this to ready)

All folds vs measured latency on real tables (full table in `docs/design/perf_interp_design.md` section 4.1):

- **Corner truncation**: artificially delete the upper-right (large-seq x large-batch) triangle of every [s][b] plane and query the deleted points — attention 5.7% / MLA 4.8% median, **0 misses in 620 queries**.
- **Knee capture** (DSA top-k @2048, signed, 118 folds across h100+gb200 and both archs): knee-adjacent **+1.0%** vs control +1.2% — linear crossing does not over/under-shoot the knee.
- **Prefix (past-KV) axis** (dsv4 sparse table, 26 past_kv values, whole-row holdout): **2.5%** median, 0 misses.
- **One-sided coverage**: where only one bracket side covers the query, the old survivor-clamp scored **-41%/-44% signed**; the new single-survivor SOL-ratio semantic lands at **10.7%/8.3%** (this campaign found and fixed it).
- **Task-level equivalence**: full `aic default` sweep matches the old engine at median 0.3% / max 2.1% while running **~5.5x faster**.

Plus an adversarial code review pass over the engine; its confirmed findings (neighbour-poisoning, distance-gate leak, cache-key collision, NaN guards, config validation gaps) are fixed with locking tests.

## Rust port (the compiled engine-step backend speaks v2 too) — post-#1313-merge state

The wheel bundles the pyo3 extension and the rust engine-step backend is the DEFAULT, so the Rust perf-database layer is ported in the same PR — merging them separately would ship a wheel whose compiled path answers with the old semantics while the Python SDK answers with v2.

- New `rust/aiconfigurator-core/src/perf_database/perf_interp.rs`: mirror of `sdk/perf_interp/engine.py` (same resolution chain incl. the single-survivor SOL-ratio semantic; deliberate deltas documented in the module doc: scalar-latency leaves, u32 keys + f64 coords, site index built at load).
- All 12 Rust op families rewired; the mirrored legacy machinery (topk piecewise, dsv4 robust-lookup + b/bp batch scaling, generation-attention target-lattice densification, op-level MoE overflow estimator, `interpolation.rs` interp family) is deleted. Per-op SOL rooflines ported from the Python op sources.
- **Parity**: 58+ cross-language anchors in unit tests (Python v2 SILICON values on the same shipped parquet, 1e-9 rel) across exact/interior/util-hold/site-transfer paths; full `aic default` sweep with `--engine-step-backend python` vs `rust` matches at max 0.0007% with identical pareto rows.
- Rust diff: 20 files, +3732/−1889. Crate suite: 131 passed; the 6 failures are pre-existing on main (stale b200 spec expectation + pyo3-env tests), verified by stash A/B.
- **Fixed in this PR (pre-existing, root-caused by per-op accounting):** the rust `MoeOp` dropped the `attention_dp_size` token multiplier that Python's `MoE.query` applies (all dp ranks funnel into the shared expert pool). On DSV3 (tp1/dp8/moe_tp8, h200 sglang) 12 of 13 context ops matched python EXACTLY and only `context_moe` diverged 4.67x — the entire 858-vs-560ms TTFT gap. `MoeOp` now carries `attention_dp_size` (serde default 1; dense/dp=1 sweeps bit-identical) and the `_moe` opspec emits it; regression test locks `query(dp=4, t) == query(dp=1, 4t)`. After the fix DSV3 TPOT matches python exactly (19.444ms both).
- **CP guard (this PR):** the compiled engine has no context-parallel prefill model (Python's `_query_cp` full/cp sparse-kernel swap + CP all-gathers is not ported, nor are the sparse tables). Before the guard, cp_size>1 configs on the rust backend silently evaluated the NON-CP composition (measured: 3392ms returned where Python fail-louds on missing sparse tables). Spec conversion now raises `OpConversionError` for cp_size>1, matching Python's fail-loud contract; CP sweeps use `--engine-step-backend python` until the CP model is ported (follow-up).
- Remaining known gap (pre-existing, separate feature): the Rust loaders do not merge **shared-layer** rows, so a curve whose boundary point comes from the shared layer anchors its beyond-range util-hold one collected point short (DSV3 TTFT: rust 983 vs python 858 — fully accounted: with the shared layer disabled Python reproduces rust's 8.2618 ms/layer exactly). Provenance of the borrowed point: same-framework cross-VERSION sharing — sglang **0.5.6.post2**'s 20480-token row (same kernel `sglang_fused_moe_triton`, latency 3.9748) anchors the 0.5.10 curve, and it is *lower* than 0.5.10's own 16384 point (4.2300), so which side is closer to ground truth here is genuinely unclear. Shared-layer support in `aiconfigurator-core` is its own follow-up.

## Reconciliation with upstream #1313 (merge commit 6fd9d556)

#1313 (rust dense-CP Phase 1 + `PerfDbSources` shared-layer loaders) landed on main mid-flight and rewrote the same rust files. Resolution policy: **their loader/source plumbing and CP machinery as the base; this PR's perf_interp query resolution, SOL rooflines, and parity tests on top.** Notable outcomes:

- Their shared-layer loaders are the missing piece this PR had documented as a known gap: with them, **DSV3 tp1/dp8/moe_tp8 now matches EXACTLY on both backends** (TTFT 858.138 / TPOT 19.444) — the rust side finally sees shared-layer boundary points.
- The MoE `attention_dp` multiplier was fixed independently on both sides (same bug, different symptoms); deduped to one spec-driven field with exactly one multiply.
- The sparse-CP guard is narrowed: dense-GQA/MoE CP flows through their Phase 1 (verified: dense CP at 64k matches bit-for-bit at cp=1/2/4); DSA/DSV4 sparse CP still fails loud (`OpConversionError`) because neither the model nor its sparse tables are ported/collected.
- Their loader merge orders (dsa first-wins, dsv4/mla-module/mhc last-wins) are preserved verbatim per their Python-parity annotations.
- Their 227-case `rust-engine-step-parity` CI suite passes green on the merged tree.

## Substitute review of #1313's surviving code (it was self-merged unreviewed)

Verified correct: the CP zigzag math (chunking, fused extras, correction ordering — line-equivalent to Python), the shared-layer source ordering/filters, and the token-major CP op transforms. Three cross-engine bugs found and fixed here (commit 608c5f3b):

1. **cp>1 decode MoE combine zeroed** — Python zeroes only pre-dispatch; combine queries allreduce. Restored (was slipping the 1% parity gate as a small decode term).
2. **WideEPContextMLA cp>1 silently un-sharded** — opspec dropped `cp_size`; ported the zigzag with serde-default compat.
3. **DSA loader duplicate policy inverted** (first-row-wins vs Python's last-row-wins-within-file) — live divergence on b300/vllm's 16.6k duplicated coordinates; regression test pins a real coordinate to the Python oracle.

Latent/minor review findings are tracked in #1333 (section 4.5), and a parity-suite hardening todo list (anti-vacuous guard, sparse-CP guard assertion, dense long-context CP, beyond-range shared-layer case) is in #1333 (section 5).

## Logical-parity completion (owner's principle: same data + mirrored consumers => identical engines; symmetry locked by synthetic tests, not gated on data)

The remaining "guarded asymmetries" were eliminated by porting the consumers instead of guarding them (commits 8a530ae3 + 6e22160e):

- **Sparse-CP (`_query_cp`) ported for both GLM/DSA and DSV4** (chunked in-grid mqa walk, topk full/per-card swap, CSA/HCA branches, AG terms). The `_reject_cp` guard is retired; the specs carry `cp_size`/`window_size`. Verified end-to-end: a cp=2 DeepSeek-V3.2 config now fails on BOTH backends with the byte-identical sparse-tables message (rust originally returned a silently-wrong number, then a guard error). Synthetic composition anchors match python exactly (4575.0 / 1060.0) and a real-data in-grid anchor matches at 1e-9 on gb200.
- **CSA topk DELTA correction ported** (same `AIC_DSV4_TOPK_CORRECTION` env gate, shared calib load, python-oracle tests at 1e-9) — parity holds automatically the day calib data ships, superseding the interim fail-loud guard.
- **Non-SILICON databases route to the python step** (parity by delegation): the compiled engine is SILICON-only, so HYBRID/EMPIRICAL configs get the python answer on both backend settings instead of capability divergence.

With these, code-level parity debt is zero: every remaining #1333 item is data/calibration work, and when calibration corrects the model both implementations change together under the parity tests.

## Final cross-engine parity summary

| Check | Result |
|---|---|
| 58+ per-family cross-language anchors (unit tests) | 1e-9 |
| Task-level `aic default` sweep, Qwen3-32B dense | max 0.0007%, rows identical |
| DSV3 MLA+MoE, tp1/dp8/moe_tp8 (h200 sglang) | **exact** (858.138 / 19.444 both) |
| Dense CP cp=1/2/4 @ isl=64k (h100 sglang) | **exact** |
| Upstream 227-case engine-step parity suite | all pass |
| cargo suite | 139 passed + the 6 pre-existing-on-main failures |
| python suite | 1287 passed |


docs/design/perf_interp_design.md
```diff
@@ -0,0 +1,275 @@
...
+
+`extrapolate_data_grid`, its 12 per-op `_extrapolate` methods and 18
+`_*_TARGET_*` grid constants (~30k synthesized nodes per quant-mode,
+order-dependent interpolation-of-interpolation, and the source of the
+"Skipping interpolation" warning spam on ragged inputs) are deleted. Also
+gone: raw-linear extrapolations that could undershoot launch-overhead floors
+or go negative below the smallest collected size (NCCL/allreduce message
+sizes, MoE token underflow, DSV4 decode below min `s_total`).
+
+### 5.6 The DSV4 sparse-kernel helper (the last scipy consumer)
```
tianhaox, line 214: @YijiaZhao please help to review this part

src/aiconfigurator/sdk/perf_interp/config.py
```diff
@@ -0,0 +1,274 @@
...
+                raise ValueError(f"curve_axis {self.resolver.curve_axis!r} cannot also be a site axis")
+            if len(set(self.resolver.site_axes)) != len(self.resolver.site_axes):
+                raise ValueError(f"duplicate site axes in {self.resolver.site_axes}")
+            if len(self.resolver.site_axes) + 1 != len(self.axes):
+                raise ValueError("every axis must be either a site axis or the curve axis")
+
+
+# ---------------------------------------------------------------------------
+# Example records (illustrative — the real sol_fn is injected per (op, database)
+# from the op's existing get_sol). Adding an op = adding a record here.
```
simone-chen, line 206: We might want to update add_a_new_model.md to update the record here when adding a new op.

src/aiconfigurator/sdk/perf_interp/engine.py
```diff
@@ -0,0 +1,403 @@
...
+    keys = sorted(node)
+    c = coords[-1]
+    k_tail = cfg.resolver.k_tail
+    if c > keys[-1]:
+        tail = keys[-k_tail:]
+    elif c < keys[0]:
+        tail = keys[:k_tail]
+    else:  # innermost is in range; an OUTER axis was snapped
+        tail = [min(keys, key=lambda k: abs(k - c))]
+
```
simone-chen, line 401: The code won't report error even if `len(keys) < k_tails` as `keys[-k_tail:]` or `keys[:k_tail]` will just be trimmed to `keys[:]`.
Should we actually raise an error because the data is so sparse that we don't even have `k_tails` number of data points?

src/aiconfigurator/sdk/perf_interp/config.py
```diff
@@ -0,0 +1,274 @@
...
+    #: decode-only site with m<=64 must not answer an m=8192 query). If no site
+    #: covers it, fall back to nearest sites with per-site curve-end util-hold.
+    require_curve_coverage: bool = True
+    #: Boundary-util anchor = median of the last k_tail curve points (sawtooth-
+    #: robust). 1 = plain boundary point.
+    k_tail: int = 3
+
+
+@dataclass(frozen=True)
+class Grid:
```
simone-chen, line 142: Grid resolver doesn't seem to have defined a standard for "too sparse". It will always return a value except for a coord is totally absent. Even if there is only one data point for a dimension, it will still return some value from grid hold. I saw that `Extrapolation is UNBOUNDED by design` - how much accuracy we can achieve from it?

jasonqinzhou (approved):

src/aiconfigurator/sdk/perf_interp/engine.py
tianhaox: Intentional — and now documented in the k_tail docstrings (ec0561ae). k_tail is a robustness knob (median over the boundary tail to resist the GEMM-m wave-quantization sawtooth), not a data-sufficiency gate: with fewer collected points, the median of what exists is the best available anchor, degrading naturally to the plain boundary point (k_tail=1 semantics, which is Grid's default anyway).

Raising here would also be inconsistent with the engine's existing contract: a single-point curve already routes to util-hold (`_eval_curve`, `len(curve) < 2`), and step 4 of the resolution order only raises when there is *nothing* to anchor on — turning 'we have 2 points instead of 3' into a hard miss would regress currently-answerable queries across the SDK. Sparse-data policing belongs to the collector / LOO harness at collection time, not query-time refusal.

src/aiconfigurator/sdk/perf_interp/config.py
tianhaox: Grid-hold isn't an arbitrary value: it snaps to the nearest collected path, holds the boundary util (SOL/latency), and lets the analytic SOL carry the growth — outside the data the roofline is the only signal we have, and a distance cap would just turn a physics-anchored estimate into a hard miss, which aborts the whole task-level sweep (the SDK's job is precisely to answer in regions the collector didn't reach).

On measured accuracy (validation section of `docs/design/perf_interp_design.md`): the corner-truncation fold (delete the entire large-seq x large-batch corner of every [s][b] plane, query the deleted points) lands at 5.7% median near the cut edge / 9.9% deep into the corner, 0 misses in 620 queries; one-sided coverage scores 10.7%/8.3% after the SOL-ratio correction; generation frontier 9.7%/2.4%. The honest tail: the sparsest DSA fold shows p90 ~52%, concentrated in n<=2-head overhead-dominated slices — that's a collection-density problem (re-collection campaign in flight), not something a query-time gate can fix.

For the one-point-per-dimension case: bracketing is impossible, so holding that point's util is the only option that uses all the information we have. If we want a sparsity quality signal, the right home is LOO-harness / collector-side reporting at collection time, not query-time refusal.

src/aiconfigurator/sdk/perf_interp/config.py
tianhaox: Done in ec0561ae — add_a_new_model.md Situation 3 now has the perf_interp step: declare an OpInterpConfig factory in `sdk/perf_interp/config.py` (resolver + axes + sol_fn) and register it in `OP_CONFIG_FACTORIES`.

simone-chen (approved):

simone-chen:
FWIW, the rust <> python parity drift is expected as this PR refactors the perf DB in python but didn't touch the rust part. We can fix it in the same PR or I can open another PR to fix the parity issue.

tianhaox:
> FWIW, the rust <> python parity drift is expected as this PR refactors the perf DB in python but didn't touch the rust part. We can fix it in the same PR or I can open another PR to fix the parity issue.

I'll implement it in this PR.

YijiaZhao (commented):
First off: the perf_interp v2 engine itself looks like a clear improvement, and the legacy DSV4 sparse-kernel extrapolation bug is real (we reproduce the ~-49% raw-linear-trend error exactly on our own B300 table). The comments below are all about the DSV4/CP modeling story built on top of it — none of them should block the interp refactor itself, but several should land before the CP path is treated as usable.

---

### 1. [design/data-strategy] Root cause first: the sparse-sweep isl cap is a hardcoded planning constant, not an sglang-derived limit — and the CP extrapolation story is built on top of it

The PR justifies the scored-pair SOL on `_lookup_sparse_kernel` with "CP queries live far beyond the collected isl sweep (8192), so the extrapolation direction is the common case". But the 8192 cap is not a physical or runtime constraint:

- The collector's **runtime filter** already reads `server_args.chunked_prefill_size` from sglang (`runtime_limits.py::runtime_chunk_size`, sglang-derived; 16384 on B200/B300-class GPUs). The **case grid**, however, is pinned by a static `sparse_chunk_prefill_size: 8192` in the cases yaml — so actual coverage is set by a hardcoded number that is *more conservative than the production runtime value on the very GPUs the tables target*. This inverts the repo's own "runtime limits are read from sglang, never hardcoded" principle. First fix: derive the case-grid token budget from the target system's sglang-derived chunk size (collector change, so presumably via #1333, but the justification in this PR's comments should not present the 8192 boundary as a given).

Even with that fixed, CP asks for full-isl (128k+) anchors, and the right answer there is per-term, not blanket extrapolation:

- **mqa never needs to leave the grid.** The pair count is additive over prefill chunks: `mqa_full(isl, past) = Σ_k mqa(chunk, past + k·chunk)` holds *exactly* under the PR's own SOL premise (`b·(past·isl + isl²/2)`), every summand stays at collected shapes (`isl ≤ chunk`, `past ≤ 1M`) and sits in the region where measured util has actually flattened (see #2) — and it is literally the kernel sequence production chunked prefill executes.
- **topk is the term that genuinely needs full-isl data** — no ×cp identity (sub-quadratic, data-dependent, branch-switching on full context c4_len), which is exactly how the GLM-5 CP design handled it (docs/CONTEXT_PARALLEL_DSA_MODELING.md §3.2, §9.1: the full-isl topk reference is collected, not scaled). The topk/mqa calib collectors are *standalone kernel benches* (no model forward), so they are not physically bound by `chunked_prefill_size` at all — the calib sweep can be decoupled from the chunk cap and extended to CP ranges cheaply. That extension must be **branch-aware**: the kernel switches implementation by full-context c4_len (16k / 32k / cluster thresholds), and the top-k saturation knee `index_topk·compress_ratio − prefix` is a *diagonal* in (isl, step) coordinates, so the grid needs densifying along `past+isl` near the boundaries (or re-keying by c4_len).

With those two in place, beyond-range util-hold becomes a last-resort fallback rather than the designed path. Since #1333 already tracks "CP calib data", it would be good to record this asymmetry there: grid budget from sglang; mqa via in-grid chunk decomposition; topk via extended, branch-aware standalone collection.

Related consequence: within one `_query_cp` sum, `mqa_full` now silently extrapolates 16x while `tl_full` still raises `PerfDataNotAvailableError` beyond the grid — so the motivating scenario (CP at 128k) still cannot complete end-to-end, and the same quantity-role (full-isl sparse anchor) has two different beyond-range contracts. GLM-5's `ContextDSAModule._query_cp` keeps fail-loud for both, so the contract is now also inconsistent across models.

### 2. [validation] The "-1% median" util-hold claim does not reproduce; small-prefix anchors are unsaturated

We ran the PR's own holdout protocol (truncate to max_isl/4 = 2048, predict isl=8192) on the shipped `b300_sxm/sglang/0.5.10/dsv4_paged_mqa_logits_module_perf.parquet`:

| method | median | p10 | p90 |
|---|---|---|---|
| legacy raw-linear trend | -48.6% | -69.9% | -7.8% |
| SOL util-hold | **+4.1%** | -7.3% | **+23.5%** |

The legacy -49% claim reproduces exactly; the -1% does not (at least not on this system). The bad tail is structural, not noise: "util under isl² flattens from isl~2048" only holds for **step ≥ 4096** (where the `past·isl` term keeps the kernel fed). At `step=0` util is still climbing at the table boundary (7.0e6 at isl=256 → 2.18e8 at isl=6144, never plateauing), and the isl=8192/step=0 point shows a util *drop* (latency 0.087→0.204 ms for a 1.78x pair-count increase). Holding an unsaturated or anomalous boundary util systematically over-predicts (+21% in the worked example), and nothing validates what util does between 8k and 128k — the campaign only covers 4x, the CP path asks for 16x.

Asks: (a) report the holdout **signed and stratified by step** (pooled medians hide exactly this tail); (b) exclude / special-case unsaturated anchors (e.g. require the boundary util slope over the k_tail to be below a threshold before holding it, else borrow the plateau util from larger-step slices); (c) since the legacy path is deleted, consider a debug log on beyond-range util-hold resolutions (anchor util, extrapolation distance) — there is currently no escape hatch or observability if this misbehaves in a sweep.

### 3. [model-vs-implementation] The CP all-gather terms don't match the SGLang implementation, and the new test locks them in

`_query_cp` models the CP comms as `AG(isl·index_head_dim) + AG((isl/4)·head_dim)` for CSA and `AG(min(isl,window)·head_dim) + AG((isl/128)·head_dim)` for HCA, and `test_cp_dsv4_modeling.py` now asserts these exact volumes. Reading the actual SGLang DSV4 CP implementation (main @ 3d3a7ec03, `models/deepseek_v4.py` + `layers/attention/dsv4/compressor.py`), the per-layer gathers are structurally different:

- The KV latent all-gather moves the **full-isl** bf16 latent (`kv_lora_rank + rope_head_dim` per token) — the gather happens *before* the cache write / compression, so a `(isl/ratio)`-length "compressed entries" gather does not exist in the implementation. The model sized the AG by what is *stored*, not what is *communicated*.
- The second gather is the compressor's `kv_score` (`2·(1+overlap)·head_dim` per token), also full-isl; after it, **every rank runs the compressor over the full token set** — this replicated compute is neither in the per-card base (which only saw isl/cp tokens) nor modeled as a separate term.
- HCA needs no separate windowed-KV gather — the same latent gather feeds the SWA cache (`store_cache(swa_k=kv)` post-gather).

Element-count sanity check: the model totals ~256 elements/token for a CSA layer (128 indexer + 512/4 compressed) vs ≥1600 elements/token actually gathered (576 latent + ≥1024 kv_score) — a ~6x under-count, in the direction of under-predicting CP comm. Notably GLM-5's instrument-verified CP model gathers 704 elements/token, and DSV4 gathers *more* than GLM-5 in reality, yet the model claims less — these numbers cannot have come from a timeline.

Asks: calibrate the AG terms against a real DSV4 CP timeline before the CP path is considered usable (the GLM-5 numbers were pinned by instrumenting sglang's `cp_utils`), and annotate the test/model with the sglang commit the comm model corresponds to — if it targeted an earlier sglang CP implementation, that provenance is currently unrecorded.

### 4. [cross-engine] Rust `Dsv4Table` lacks the topk DELTA correction — silent divergence the moment calib data ships

The rust parity oracle is generated with `AIC_DSV4_TOPK_CORRECTION=0` (per the test doc-comment), i.e. parity is proven against Python *with the correction disabled*. That currently holds only because no system ships `dsv4_csa_topk_calib` — the Python DELTA is a no-op today. The moment calib data lands, Python subtracts `flat − top_last` on CSA module queries and rust doesn't: the **default** engine-step backend silently over-predicts CSA context attention (our B300 measurements put the dummy-flat topk bias at roughly half the per-layer CSA module time), and the two engines diverge — exactly the class of bug this PR's own per-op accounting hunted down elsewhere (the MoE `attention_dp` multiplier).

Asks: (a) either port the DELTA subtraction to `Dsv4Table` (it is a small lookup: two rows per shape, one subtraction) or make rust fail loud / warn when a calib table exists on disk — parity must not depend on data happening to be absent; (b) minor, same area: the `cp_size>1 → OpConversionError` guard is the right call, but the error should point users at `--engine-step-backend python`; and in #1333, please order the sparse-CP rust port *after* the CP data/comm items (#1 and #3 above) — porting first would translate an unvalidated model into a second language and the parity suite would pass with both engines wrong together.

### 5. [minor] The scored-pair SOL is not scoped to the kernel it is valid for

`_lookup_sparse_kernel`'s `sol_fn = b·(past·isl + isl²/2)` is installed generically for whatever `kernel` is requested, and the same sidecar dict also carries `hca_attn` — a *windowed* kernel whose work is window-capped (linear beyond the window), for which a quadratic SOL would badly over-predict any beyond-range resolve. Today only `paged_mqa_logits` flows through this function, but nothing enforces that. Suggest `assert kernel == "paged_mqa_logits"` (or a per-kernel SOL map) so the next caller doesn't inherit the wrong physics silently.

### 6. [minor] `_query_cp`'s correctness silently depends on `AIC_DSV4_TOPK_CORRECTION` being enabled

The CP composition subtracts `top_last(per_card)` against a base obtained from the standard module query — which is only algebraically equivalent to the GLM-5 form (`raw − flat(per_card) + top_last(full)/cp`) because that base has already had the `flat − top_last` DELTA subtracted. With `AIC_DSV4_TOPK_CORRECTION=0`, the CP result silently keeps a `flat(per_card) − top_last(per_card)` over-estimate. Two smaller notes in the same area: the branch docstring ("does NOT reuse the non-CP topk-DELTA correction") is misleading — the branch *relies* on that correction through the base, and a literal-minded refactor would silently break the algebra; and the `+top_last(per_card)` inside the base's DELTA vs the `−tl_perc` in the CP branch resolve through two different interpolators (`_dsv4_topk_delta_ms`'s prefix→isl→bs chain vs `_bs_slice`+`_lookup_2d`), so the cancellation is only exact on grid hits. Worth an assert/warning when the correction is disabled, and ideally making the flat→top_last swap explicit inside the CP branch (query the flat row directly against a raw base) so the formula is self-contained in one function.

tianhaox:
@yijia thanks — this is exactly the review this PR needed, and two of your points corrected the record. Addressed in fa0dfd62 (data-collection items deferred to #1333 per the data owner):

**pt.1 (adopted, and you were right about the inversion):** `_query_cp` now resolves `mqa_full`/`mqa_perc` via the in-grid chunk decomposition — `mqa(isl, past) = Σₖ mqa(chunk, past + k·chunk)` with chunk=8192 — every lookup in-grid (past_kv is collected to ~1M), no extrapolation, and mqa/topk return to a uniform fail-loud beyond-range contract (GLM-5 form). The 16x util-hold is no longer load-bearing anywhere on the CP path. The mqa/topk asymmetry (in-grid decomposition vs branch-aware full-isl collection, knee-diagonal densification or c4_len re-keying) is recorded in #1333's CP-data item.

**pt.2 (confirmed — my number was wrong):** the pooled −1% was a cancellation artifact. Stratified signed rerun (288 slices, max_isl/4 truncation): step=0 **+25.6%** (unsaturated anchors, exactly your diagnosis), 0<step<4096 +10.2%, step≥4096 −8.1%. Design doc §5.6 rewritten with the stratified table and the pooling lesson. The unsaturated-anchor guard (util-slope threshold / borrow plateau util from larger-step slices) is tracked in #1333 — with the decomposition in place this path has no production caller, so we'd rather land the guard with its own LOO validation than rush it. Beyond-range holds now emit a DEBUG line (coords, anchor util, distance) for sweep observability.

**pt.3 (adopted as ordered follow-up):** agreed the AG terms are formula-derived, not implementation-calibrated — #1333 now orders: sglang cp_utils-style instrumentation on a real DSV4 CP timeline (pin latent-gather + kv_score volumes + the replicated compressor term, version-annotated) BEFORE the comm model or its test is treated as a contract, and before any sparse-CP rust port.

**pt.4:** (a) tracked in #1333 with your suggested ordering (rust topk DELTA or fail-loud-on-calib-present, after the data/comm items); (b) the guard message already points at `--engine-step-backend python`.

**pt.5 (adopted):** `_lookup_sparse_kernel` now rejects any kernel other than `paged_mqa_logits` (guard + test); the head-selection test moved onto the supported kernel.

**pt.6 (adopted):** `_query_cp` warns when `AIC_DSV4_TOPK_CORRECTION=0` and documents the DELTA-corrected-base coupling.

tianhaox:
Update on pt.4's ordering: after discussion with the owner we went further than the earlier reply — the sparse-CP compositions (`_query_cp` for GLM/DSA + DSV4) and the CSA topk DELTA are now PORTED (8a530ae3/6e22160e) rather than guard-deferred, under the principle that parity and validation are separate tracks: both engines read the same data, so the consumers are mirrored and symmetry is locked by synthetic tests (DSA 4575.0 / DSV4 CSA 4575.0 / HCA 1060.0 anchors + a real-data in-grid `_mqa_chunked` anchor at 1e-9), with byte-identical fail-loud messages while the tables are absent. Your "both engines wrong together" concern is handled by keeping validation (the AG instrumentation + topk collection you outlined) as its own #1333 track — when calibration corrects the model, both implementations change together under the parity tests, and the engines never disagree in the meantime. Non-SILICON databases now also delegate to the python step, so HYBRID configs can't capability-diverge either.

YijiaZhao:
Thanks for the fast and thorough follow-through — all six points are addressed to our satisfaction (fa0dfd62 / 8a530ae3 / 6e22160e), and the parity-vs-validation two-track framing on pt.4 is a fair resolution of the ordering concern: engines that are wrong *together* and covered by 1e-9 parity anchors are a strictly better place to wait for calibration from than engines that diverge.

Independent backing: we ran an A/B of this branch (at fa0dfd62) against main on our B200 measured-alignment datasets — GLM-5 CP, GLM-5.2 CP and GLM-5.2 TP single-step traces (~3.3–3.5k prefill cases each). Prefill MAPE is unchanged on all three (3.44/3.44, 7.21/7.21, 8.49/8.45 mean %), per-case prediction drift ≤0.14% on the CP paths and ≤1.5% max on TP; decode moved slightly in the PR's favor. No regression on any GLM sparse path.

What remains is the #1333 validation track — the DSV4 CP timeline instrumentation (AG volumes + replicated compressor term) and the topk/collector data items — which is on our side as data owner. We'll pick those up with the B300 collection work.

PR #1314, opened 2026-07-06
feat(cross-node-ep): DeepEP WideEP normal + low-latency collectors, data, and SDK enablement

Harrilee:
Add SGLang WideEP DeepEP dispatch/combine perf collection (both DeepEP modes) and the SDK plumbing needed to actually consume it, so `cli default --enable-wideep` produces configs for DeepSeek-family MoE models on SGLang.

## Collectors

- `deepep_normal` (NEW): prefill normal-mode dispatch/combine, wrapping `deepep/test_intranode.py` (single node, `node_num=1`).
- `deepep_ll`: decode low-latency dispatch/combine, wrapping `deepep/test_low_latency.py`.
- Both registered in `collector/wideep/sglang/registry.py` and run as full-node collective jobs through `collect.py`'s fullnode path. `test_intranode` / `test_low_latency` gain `do_check` / `metrics_out` so the collectors can harvest timings instead of scraping stdout.
- `framework_manifest.yaml` pins the `comm` family of `wideep_sglang` to sglang 0.5.12 on `lmsysorg/sglang:v0.5.12-cu130`, which ships deep_ep 1.2.1+d28bd67 (`kNumMaxTopK` 11). The DeepSeek-V4 images the rest of `wideep_sglang` uses ship deep_ep 1.1.0 (`kNumMaxTopK` 9) and cannot run topk>9 shapes such as `(4096, 512, 10)`. `wideep_moe` deliberately stays on the DeepSeek-V4 default because its DSv4 module support on the 0.5.12 image is unverified.

## Data: both DeepEP tables land at 0.5.12, and the 0.5.10 copies are deleted

Collected on H100 / H200 / B200 / B300 / GB200 / GB300. Two commits carry the landing:

- `chore(perf-database): regenerate the op-kernel-source manifest` — regeneration only, no data change. The committed manifest had drifted from the tree it is generated from, in ways this branch did not cause: 123 groups were missing and 57 of the 82 recorded groups had a wrong `total_raw_rows`, some off by more than 4x. Doing this separately keeps the data commit's manifest hunk to the two lines a reviewer can check.
- `feat(data): recollect the SGLang DeepEP tables on 0.5.12` — twelve new tables, six new sidecars, twelve deletions, two manifest lines.

The 0.5.10 tables are **deleted, not kept as a fallback**, so nothing can silently resolve back to them. `custom_allreduce_perf` is untouched, so each `comm/sglang/0.5.10/` directory and its sidecar survive with that one entry.

### Why recollect

**Low-latency coverage on GB200 and GB300 was truncated, and the sidecar said `complete` anyway.** Both carried 124 rows where the 8-rank systems carried 154 over the same 14 shapes. The four largest shapes stop early — `(6144, 4, 128)` and `(6144, 8, 256)` at 8 tokens, `(7168, 6, 384)` and `(7168, 8, 384)` at 4 — while every other shape runs the token ladder out to 1024. That is the shape of a run that died partway, not a kernel limit. The 0.5.12 collection reaches 1024 on all four, so `status: complete` is now true by measurement rather than by assertion.

**Two Mixtral shapes were being filtered out.** `(4096, 2, 8)` and `(6144, 2, 8)` are now collected: +22 low-latency rows and +276 normal rows per system.

**68 keys in the committed B200 normal table are 2–13x slower than any healthy node reproduces.** They are concentrated in two shapes, `(6144, 4, 128)` and `(4096, 8, 256)`. At hidden 6144, topk 4, 12 SMs, 16384 tokens the committed file reads **25733 µs** of `dispatch_transmit_us` where new B200, new B300, committed B300 and new GB200 all cluster at **1913–1976 µs**. Counting `combine_transmit_us` too the affected set is 95 keys, 99 distinct rows out of 2208. **B300 has zero such keys on either column**, so this is not a version effect — that slice looks to have been collected on a degraded node. Deleting the file is the only way to stop it being read.

**Reviewers should expect large ratios on B200 and small ones everywhere else.** Median new/old latency ratio over the shared keys is 0.994–1.003 on all six systems. B200's p5 is 0.575 and its minimum 0.078 (a 12.8x speed-up), entirely from the degraded slice above; every other system's p5 sits between 0.87 and 0.99. GB200's median is 1.001 — that is the point. Outside the B200 slice this recollection buys provenance coherence and the new shapes, **not** corrected numbers.

### The new tables are strict supersets

Every one of the twelve is a strict superset of the 0.5.10 table it replaces, checked key by key on `(hidden_size, num_topk, num_experts, num_token, dispatch_sms, combine_sms)`: **zero missing keys on all twelve**. No query that resolved before this change stops resolving.

| system | GPUs/node | LL rows | LL shapes | normal rows | normal shapes |
|---|---|---|---|---|---|
| gb200 | 4 | 176 | 16 | 2622 | 19 |
| gb300 | 4 | 176 | 16 | 2622 | 19 |
| h100_sxm | 8 | 176 | 16 | 2484 | 18 |
| h200_sxm | 8 | 176 | 16 | 2484 | 18 |
| b200_sxm | 8 | 176 | 16 | 2484 | 18 |
| b300_sxm | 8 | 176 | 16 | 2484 | 18 |

Manifest `total_raw_rows`, re-derived with `check_kernel_source.py --op-file` rather than by hand:

- `wideep_deepep_ll_perf`: 1054 → **1246** = 6 × 176 new, plus the 2 × 95 that H100 and H200 still carry at 0.5.6.post2.
- `wideep_deepep_normal_perf`: 14398 → **16054** = 15180 new, plus 2 × 437 at 0.5.6.post2.

### Coverage gaps, excluded rather than synthesized

- **hidden 8192** (Nemotron-Ultra): unsupported at every rank count — DeepEP's TMA per-warp budget — so absent from every table.
- **`(2688, 6, 128)`** (Nemotron-Nano): fails DeepEP's own BF16 combine assertion at 8 ranks, `payload_diff` around 2e-3 against a 5e-6 tolerance, deterministic across eleven attempts on nine nodes and two clusters. It passes at 4 ranks because combine sums across ranks and BF16 rounding grows with rank count. Present on the two 4-GPU-per-node systems only, which is the entire 19-vs-18 shape difference in the table above, and the committed 0.5.10 baselines already showed the same split.

## SDK enablement (what makes `--enable-wideep` actually run)

- `task_v2.py`: for WideEP DeepSeek (sglang + `deepep_moe`), resolve `fmha_quant_mode` → `fp8_block` and `kvcache_quant_mode` → `fp8`, matching the labels the collector writes for the `wideep_*_mla` tables. Fixes `Unsupported wideep_context_mla quant mode 'bfloat16'`. Non-WideEP keeps the narrow-EP bf16 downgrade.
- `operations/mla.py`: resolve the WideEP MLA `kernel_source` by what was actually collected — Blackwell (SM100) uses `trtllm_mla`, Hopper uses fa3/flashinfer — instead of hard-requiring flashinfer. Fixes the GB200/GB300 MLA lookups.
- `_attention_op_keys` now keys the MLA op tables on the same model-dispatch predicate the model layer and the quant resolution already used (`moe_backend == deepep_moe`), not on `enable_wideep`. Three sites had to agree and one did not, so the intra-node DeepEP sweeps validated a WideEP `fp8_block` label against the narrow-EP `context_mla` table and both DeepEP arms of the support matrix died on every DeepSeek-V3/R1 run that reached them.
- `op_kernel_source_manifest.yaml`: register `wideep_deepep_ll` across the six systems.

### Multi-node DeepEP: substituted, marked, never silent

A multi-node request (`node_num > 1`) is answered from the exact node scale when it was measured. When it was not, the `node_num=1` rows for the same `(hidden, topk, num_experts)` are substituted **and the result is marked an estimate** — it can no longer pass as a measurement.

Two commits on this branch touch this and a reviewer reading the history will see them disagree. `fix(sdk): fail closed on a missing WideEP DeepEP multi-node scale` made the miss raise; `fix(sdk): mark WideEP DeepEP multi-node estimates instead of refusing them` replaced that with the marked substitution. **The marked substitution is the end state.** Raising was correct about the dishonesty and wrong about the price: every DeepEP table in the tree carries `node_num=1` only, except H100/H200 at 0.5.6.post2, and GB200/GB300 hold 4 GPUs per node, so GB300 needs `node_num=2` at 8 GPUs while DeepSeek-V3 does not fit in 4. Measured on this HEAD, the documented headline command below lost every candidate in both experiments and exited non-zero; raising removed WideEP from four of the six systems that carry DeepEP data at all.

The mark rides `PerformanceResult.source = "estimated"`, the existing member for a value modeled from measured components, so it reaches the user through plumbing that already exists. **What a user actually sees** on `gb300 --total-gpus 64 --enable-wideep`:

- every affected row's rank reads `1*` … `5*` in both the agg and the disagg ranked table, with a footnote under each table explaining that the row's latency includes an op modeled from measurements at a different scale and is optimistic;
- the `Overall Best Configuration` block carries `* ESTIMATED: these numbers include extrapolated WideEP DeepEP multi-node cost, not direct measurements, and are optimistic.` next to the throughput and latency figures it applies to;
- `best_config_topn.csv` gains an `_estimated` boolean column, for both agg and disagg, so a reader outside the CLI is not left without the asterisk;
- `per_ops_source.json` in each top-N directory names the affected ops (`context_moe_pre_dispatch` / `generation_moe_pre_dispatch` = `estimated`);
- one `logger.warning` per `(table, node scale)` rather than one per table, so a sweep touching node 2, 4, 8 and 16 gets four lines per table quantifying four different errors instead of one line hiding all of them.

**What the marking does not do.** It does not correct the numbers. The latency returned is still the single-node measurement, which has no cross-node RDMA leg and is optimistic; the mark makes that visible, nothing more. The measured node-N/node-1 ratios the warning quotes — LL 2.2x / 3.1x / 3.5x and normal 1.8x / 3.1x / 6.2x at node 2 / 4 / 8 — come from H100/H200 sglang 0.5.6.post2, **the only campaign in the tree that ever measured more than one node scale**, and no latency is scaled by them: a correction factor would be a node-scaling model owing its own validation, and it would have to explain why normal degrades to 6.2x where LL only reaches 3.5x. Past node 8 nothing has been measured, so `node_num=16` quotes the node-8 figure explicitly as a lower bound rather than inventing a number. Sub-node scales (`node_num < 1`, a config using fewer GPUs than one node holds) also reuse the `node_num=1` rows but are **not** marked: that substitution stays intra-node in both directions and over-estimates, since a partial node's all-to-all is smaller than a full one. Where real multi-node measurements exist the sweep uses them and nothing is marked. Real multi-node coverage — collected or a validated scaling model — is tracked as a `TODO(perf enhancement)` on the resolver.

## Verification on this HEAD

- `tools/perf_database/check_collector_data.py`: all six rules pass (R1 sidecar coverage, R2 reuse validity, R3 comm exclusion, R4 family placement, R5 manifest v2 identity, R6 no legacy markers).
- A full `check_kernel_source.py` regeneration over the landed tree reproduces the committed manifest byte for byte, so the two-line hand edit in the data commit is provably the whole of the change.
- A reverse sidecar sweep over the whole data root — 581 sidecars, 983 table claims — finds no claim without a file. `check_collector_data.py` has no such rule, and stripping the two entries from the 0.5.10 sidecars is exactly the kind of edit that would leave one behind.
- No `wideep_deepep_{ll,normal}_perf.parquet` remains at 0.5.10 anywhere in the tree, and no source list built at 0.5.10, 0.5.12 or 0.5.14 on any of the six systems reaches one.
- The 0.5.6.post2 H100/H200 tables, the only real multi-node DeepEP data in the tree, are untouched and still resolve `node_num` 1/2/4/8.
- `pytest -m unit`: **2836 passed, 10 skipped, 0 failed**. `ruff check .` and `ruff format --check .` clean.
- **The documented headline command produces ranked configs again, all marked.** `cli default --system gb300 --model deepseek-ai/DeepSeek-V3 --total-gpus 64 --enable-wideep --backend sglang` exits 0 with **715 agg** and **104 disagg** results; disagg wins at **1145.67 tokens/s/gpu**, 1.46x better than agg. All five rows of both ranked tables carry the `*`, both footnotes render, the summary block carries the `ESTIMATED` caveat, and `_estimated` is `True` for all ten rows in the two CSVs. Eight warnings fire — `normal` and `ll` at node 2, 4, 8 and 16 — each quoting the ratio for its own scale. On the fail-closed commit the same command exits 1 with `sweep_agg: no results for any parallel configuration`.
- **Where real multi-node data exists, nothing changes and nothing is marked.** `cli default --system h200_sxm --model deepseek-ai/DeepSeek-V3 --total-gpus 64 --enable-wideep --backend sglang` resolves both DeepEP tables to `node_num` 1/2/4/8 (the new 0.5.12 single-node rows plus the real 0.5.6.post2 multi-node rows), returns 688 agg and 127 disagg results, emits **zero** extrapolation warnings, ranks `1`–`5` with no asterisk and no footnote, tags the dispatch ops `silicon` in `per_ops_source.json`, and reports `_estimated = False` on every row. Its top-5 output is identical to the fail-closed commit's across all 43 agg and 62 disagg shared CSV columns; the only difference is the new `_estimated` column.
- **The estimate marking changed no result counts.** `cli default --model-path deepseek-ai/DeepSeek-R1 --system gb300 --backend sglang --perf-db-version 0.5.12 --total-gpus 32` ranks agg 2226, `agg_deepep` **278**, disagg 140, `disagg_deepep` **64** — identical on the fail-closed commit and on this HEAD.

## Tests

- `tests/unit/sdk/database/test_moe_dispatch.py`: exact-scale preference, the marked multi-node substitution, the unmarked sub-node fallback, one warning per node scale, and that the normal table shares the one resolver. (This module needs `torch` to collect, which the dev extras do not install; its 13 WideEP tests pass against a `torch.xpu` stub. Note that CI currently skips the whole module.)
- `tests/unit/cli/test_estimate_marking.py` (NEW, 15 tests): the `*` and footnote on agg and disagg ranked tables, the summary-block caveat, the `_estimated` CSV column including the out-of-band disagg path, and that an unmarked run stays unmarked.
- `tests/unit/sdk/task_v2/test_task_config.py`: WideEP DeepSeek quant resolution.

## Manual test

```
aiconfigurator cli default --system gb300 --model deepseek-ai/DeepSeek-V3 --total-gpus 64 --enable-wideep --backend sglang
```

## Summary by CodeRabbit

* **New Features**
  * Added WideEP DeepEP full-node performance collectors for both low-latency (LL) and normal-mode runs.
  * Added new WideEP DeepEP LL/normal perf output artifacts and marked new completed performance tables.
* **Improvements**
  * Improved WideEP MLA data selection to use the resolved kernel source for more accurate table lookups.
  * Enhanced DeepEP performance fallback to reliably prefer exact multi-node data, with safe single-node fallback and clearer errors.
* **Bug Fixes**
  * Preserved explicitly provided quantization settings during task resolution.
* **Tests**
  * Added unit coverage for DeepEP LL node-number fallback and WideEP DeepSeek fp8 label resolution.


src/aiconfigurator/sdk/task_v2.py
```diff
@@ -753,13 +756,29 @@ def _resolve_quant_modes(self) -> None:
...
             # The decode role uses generation attention, which keeps fp8.
-            if (
+            elif (
                 role != "decode"
                 and not fmha_explicit.get(role, False)
-                and self._architecture in ("DeepseekV3ForCausalLM", "KimiK25ForConditionalGeneration")
+                and is_deepseek_mla
                 and fmha == common.FMHAQuantMode.fp8
             ):
                 self._set_role_attr(role, "fmha_quant_mode", common.FMHAQuantMode.bfloat16)
```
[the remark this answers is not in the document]
venkywonka: this seems to be a legitimate bug. this PR itself introduces an inconsistency by keying the new quant resolution on `deepep_moe` alone while the validation keys on `enable_wideep`

aic-core/src/aiconfigurator_core/sdk/operations/moe.py
```diff
@@ -1014,7 +1019,40 @@ def get_empirical(num_tokens: int, topk: int, num_experts: int) -> float:
...
+                            "wideep_deepep_ll: multi-node (node_num>1) configs reuse node_num=1 "
+                            "single-node data (APPROXIMATION: ignores cross-node all-to-all, "
+                            "under-estimates latency at large batch -- see TODO perf enhancement). "
+                            "Logged once per run."
+                        )
+                    data = fallback
+            assert data, (
+                f"wideep_deepep_ll data missing for node_num={node_num} "
+                f"(and node_num=1 fallback) hidden={hidden_size} topk={topk} experts={num_experts}"
+            )
```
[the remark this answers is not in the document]
Harrilee: Check again now

collector/collect.py
```diff
@@ -1122,21 +1122,72 @@ def collect_sglang(
...
+    # every GPU at once, so they cannot go through the per-GPU worker pool in
+    # collect_ops. Intercept them here and run each directly; collect.py's
+    # finalize step then converts the produced wideep_deepep_*_perf.txt files to
+    # parquet. These ops only run when explicitly requested (never as part of a
+    # collect-all with ops=None).
+    fullnode_ops = ("deepep_ll", "deepep_normal")
+    run_deepep_ll = bool(ops) and "deepep_ll" in ops
+    run_deepep_normal = bool(ops) and "deepep_normal" in ops
+    run_any_fullnode = run_deepep_ll or run_deepep_normal
+    ops_for_pool = [op for op in ops if op not in fullnode_ops] if run_any_fullnode else ops
```
jasonqinzhou: [P1] Keep the full-node operations inside the Collector V3 provenance contract

The static R1 and hash-closure gates pass, but the execution path still removes both operations before `build_collections()`. A current-head invocation of `collect_sglang(..., ops=["deepep_ll"])` with the GPU runner mocked returned `errors=[]` and `provenance_ctx["collections"]=[]`; finalization therefore still has neither a registry-to-table mapping nor `ResumeCheckpoint` case IDs from which to derive a V3 sidecar. The manually added `provenance: legacy` entries only cover the checked-in files and do not make future collections attestable. Please keep the full-node execution mechanism, but integrate both operations with collection mapping, checkpoint case IDs, and finalize-time `complete`/`partial` derivation.

aic-core/src/aiconfigurator_core/sdk/operations/moe.py
```diff
@@ -977,12 +981,71 @@ def load_data(cls, database: PerfDatabase) -> None:
...
+                    cls._wideep_fallback_logged.add(table)
+                    logger.warning(
+                        "wideep_deepep_%s: multi-node (node_num>1) configs reuse node_num=1 "
+                        "single-node data (APPROXIMATION: ignores cross-node all-to-all, "
+                        "under-estimates latency at large batch -- see TODO perf enhancement). "
+                        "Logged once per run.",
+                        table,
+                    )
+                lookup_node = 1
+                node_data = fallback
```
jasonqinzhou: [P1] Do not rank multi-node configurations with single-node communication timings

The refactor correctly centralizes the lookup and replaces the bare assertions with `PerfDataNotAvailableError`, but this assignment still substitutes node-1 NVLink measurements for every multi-node miss. The PR's advertised 64-GPU path therefore receives systematically optimistic dispatch/combine latency, and the helper's own docstring says the data is not directly usable for accurate multi-node sizing. A warning does not prevent those values from affecting configuration ranking. Please fail closed when the exact node scale is unavailable, or put the approximation behind an explicit non-production mode that cannot enter normal recommendations, until real internode data or a validated node-scaling model exists.

collector/wideep/sglang/collect_deepep_ll.py
```diff
@@ -0,0 +1,388 @@
...
+                    num_tokens,
+                    hidden,
+                    num_experts,
+                    num_topk,
+                    rank,
+                    num_ranks,
+                    group,
+                    buffer,
+                    seed=1,
+                    do_check=False,
```
jasonqinzhou: [P1] Keep route masking out of performance collection

`do_check=False` skips correctness assertions but does not disable the ten random assignments that replace routes with `-1` in `test_low_latency.py`. With the fixed seed, the smallest top-k-8 case retains only 2/8 routes, while the 1/2/4-token top-k-1 cases retain zero routes. The committed H200 `(hidden=2048, token=1, topk=8, experts=128)` row still proves the mismatch: `8.467 us * 0.502657 GB/s * 1000 = 4256 bytes`, exactly two 2128-byte FP8 selections. Please gate masking on correctness mode (or disable it for collection) and assert that every performance sample contains exactly `num_tokens * num_topk` valid selections.

aic-core/src/aiconfigurator_core/systems/data/gb200/comm/sglang/0.5.10/collection_meta.yaml
```diff
@@ -9,3 +9,7 @@ runtime:
 tables:
   custom_allreduce_perf:
     status: complete
+  wideep_deepep_ll_perf:
+    status: complete
```
jasonqinzhou: [P1] Do not mark incomplete collections complete

This status says `complete`, but the GB200 LL parquet contains only 1 of the 14 structural shapes present in the other 4-GPU dataset, while GB300 has four shapes truncated to only 3-4 of the 11 token points. The underlying LL and normal runners still append failed shapes and then return normally whenever any output file exists; buffer-allocation and divisibility skips are also not surfaced. Please preserve the partial rows, mark GB200 and GB300 truthfully, and return an aggregate partial-collection error so future sidecars derive `partial` from observed failures instead of allowing incomplete output to finalize as complete.

collector/wideep/sglang/collect_deepep_normal.py
```diff
@@ -0,0 +1,447 @@
...
+
+    sms_list = _sms_list()
+    if not sms_list:
+        raise RuntimeError("No SM counts resolved for DeepEP normal collection")
+
+    # Resolve the device name inside the worker (rank 0) so the parent never
+    # creates a CUDA context that would persist across the per-shape spawns.
+    device_name = None
+    # DeepEP normal kernels come from deep_ep (independent of the sglang runtime),
+    # so the recorded sglang version is a dataset bucket label. Allow an explicit
```
jasonqinzhou: [P1] Keep the runtime version as truthful provenance

All 12 added parquets are stored under `comm/sglang/0.5.10`, every embedded `version` value is `0.5.12`, and the sidecars also declare the runtime as 0.5.10. The loader admits the active directory's rows without filtering their embedded version, so a 0.5.10 request silently consumes 0.5.12 measurements. Please keep the files under the measured runtime and declare any evidence-backed reuse explicitly, or recollect/verify them on 0.5.10 and make the row, directory, and sidecar provenance agree. A version directory cannot be an arbitrary bucket label.

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -785,6 +788,28 @@ def _resolve_quant_modes(self) -> None:
...
                 self._set_role_attr(role, key, resolved)
 
+        # WideEP DeepSeek/Kimi remaps FMHA/KV labels to match collector tagging on
+        # the wideep_*_mla tables (see collect_mla_module.py). Must run before the
+        # data-driven fp8->bf16 fallback so those roles are not downgraded.
+        for role in roles:
+            backend_name = self._role_attr(role, "backend_name")
+            is_deepseek_mla = self._architecture in (
+                "DeepseekV3ForCausalLM",
+                "KimiK25ForConditionalGeneration",
```
jasonqinzhou: [P1] Do not apply WideEP MLA labels to Kimi K2.5

`DeepSeekModel.create()` still returns ordinary `DeepSeekModel` for `KIMIK25` before the WideEP dispatch. On this head, a Kimi K2.5 SGLang/DeepEP task resolves `fmha=fp8_block` and `kvcache=fp8`, but `get_model()` constructs `DeepSeekModel` with regular `ContextMLA`; querying H200 then fails because that regular table has no `fp8_block` slice. Please derive label rewriting and validation from the same predicate as model construction: exclude Kimi here unless it gains a WideEP model implementation, and add a test that builds and queries the actual attention operation after Task validation.

collector/wideep/sglang/collect_deepep_ll.py
```diff
@@ -0,0 +1,388 @@
...
+    * ``hidden`` not in the DeepEP ``SWITCH_HIDDEN`` allowlist -- the LL kernels
+      are template-specialized per hidden size (see DEEPEP_LL_SUPPORTED_HIDDEN).
+    * ``topk > kNumMaxTopK`` (see DEEPEP_LL_MAX_TOPK) -- the LL dispatch kernel
+      asserts on larger top-k.
+    * missing ``topk`` -- dense / non-MoE configs.
+    """
+    allow = _hidden_allowlist()
+    combos = set()
+    for hidden, num_experts, topk in _iter_moe_configs():
+        if hidden % HIDDEN_DIVISOR != 0:
```
jasonqinzhou: [P1] Keep model-shape exclusions visible to the Collector plan

The Collector policy requires a queued case to execute or raise; framework kernel limits may not be silently encoded as collector-side filters. With current model configs, the raw inventory has 22 unique MoE shapes, but this getter silently returns 16 LL shapes and the normal getter returns 20, with no case IDs, failure records, or counted memory drops. The workers also `continue` on divisibility and buffer-allocation failures. Please populate these operations through the public case plan and use execute-or-raise failure accounting; reserve the sanctioned generation-time filter for a measured memory-footprint check with counted drops, and put true node-killing hangs in the dated denylist.

aic-core/src/aiconfigurator_core/systems/op_kernel_source_manifest.yaml
```diff
@@ -571,9 +571,16 @@ groups:
   - op_file: wideep_deepep_ll_perf.parquet
     kernel_source: 'deepep'
     tier: shared
-    systems: [h100_sxm, h200_sxm]
+    systems: [b200_sxm, b300_sxm, gb200, gb300, h100_sxm, h200_sxm]
+    frameworks: [sglang]
+    total_raw_rows: 1054
```
jasonqinzhou: [P2] Make this generated manifest reproducible for the DeepEP schemas

The six committed LL files contain 751 rows, not the recorded 1,054. More importantly, `audit_kernel_source.py` recognizes the LL `combine_avg_t_us` column but none of the normal schema's four component-latency columns; a targeted current-head audit scans 154/154 H200 LL rows but skips all 2,208 normal rows with `No latency column found`. A fresh instructed regeneration therefore drops the normal group entirely. Please teach the audit how to compose/select the normal latency metric (while excluding component metrics from the shape key), then regenerate this manifest instead of maintaining these counts by hand.

jasonqinzhou (dismissed):
**REQUEST_CHANGES · 3.8/10 · high confidence**

## What this PR does

This PR adds SGLang DeepEP low-latency and normal-mode full-node collectors, packages six systems of dispatch/combine data, and wires WideEP MLA/MoE queries into the SDK so DeepSeek-family multi-node configurations can be generated.

This follow-up cleanly consolidates the duplicated DeepEP node lookup and replaces optimization-sensitive assertions with the SDK's typed data-availability error while preserving the legacy/core exception identity.

## Why this score

3.8/10, REQUEST_CHANGES. The incremental refactor fixes the duplicated lookup and bare-assert concern, but its direct helper contract still returns node-1 NVLink data for a node-4 miss, so inaccurate single-node communication timings continue to influence multi-node ranking. The other selected-finding surfaces are byte-identical to the prior head, leaving the full-node provenance gap, masked-route measurements, falsely complete partial data, runtime-version mismatch, Kimi MLA mismatch, silent Collector shape exclusions, and unreproducible kernel manifest unresolved.

aic-core/src/aiconfigurator_core/sdk/operations/moe.py
Harrilee: Could you clarify it? The goal was to use single node data to represent multi node data as a fallback. But you are asking me to fail it when data is not available?

Yiming992 (dismissed):
(Withdrawn — superseded by the consolidated review below.)

Arsene12358 (changes requested):
## Verdict: REQUEST_CHANGES — with a concrete path to merge

First, what holds up well: I ran the headline flow end-to-end on this branch (`deepseek-ai/DeepSeek-V3`, `h200_sxm`, sglang, 64 GPUs, `enable_wideep=True`) and it produces results; the normal-mode dataset is genuinely strong (16–17 shapes × 23 token points × 6 SM counts — beyond the sm=20 legacy minimum); the per-shape `mp.spawn` isolation in the collectors is a thoughtful containment design; and `_resolve_wideep_deepep_node_data`'s exact-node-first precedence is the right shape for the future — once real multi-node parquets land they win automatically, with zero SDK changes.

I independently verified the earlier findings on this PR (including @jasonqinzhou's set) with local reproductions on `329bcfe4`; everything below is confirmed with an evidence chain, not relayed. One design decision needs to be made explicitly, and four items are hard blockers. @Harrilee — your question on the fallback thread ("you are asking me to fail when data is not available?") is answered concretely in the proposal below: not fail-everything — fail-closed by default, with an explicit opt-in.

### The design decision: gate the node-1 substitution behind an explicit opt-in

**Now:** any multi-node miss silently reuses node-1 rows, with a once-per-process log. **Proposed:** the same fallback exists, but the user has to ask for it, results are annotated, and the miss path picks the closest measured scale instead of always node-1.

The substitution error is directly measurable from this repo's own data: H200 `0.5.6.post2` carries real node 2/4/8 LL measurements next to node-1, same campaign, same DeepSeek shape (7168/8/256). Dispatch+combine, µs:

| tokens | node1 | node2 | node4 | node8 | node1-for-node8 error |
|---|---|---|---|---|---|
| 8 | 30.4 | 59.8 | 73.8 | 78.0 | 2.6× |
| 64 | 68.8 | 160.2 | 234.6 | 254.3 | 3.7× |
| 256 | 203.1 | 520.5 | 764.7 | 880.7 | 4.3× |

Median true node8/node1 scaling is 3.5×, and the error grows with tokens. On GB200/GB300 the substitution is structurally further off: 4-rank NVLink measurements (collected with `allow_mnnvl=False`, `num_qps_per_rank=experts/4`) standing in for a `node_num=16` EP64 collective across an NVL72 fabric. Because the substituted cost is frozen in the EP-scale dimension, it does not just shift absolute TPOT — it systematically flatters larger-EP configs in the ranking. And on Blackwell this is not an edge case: no multi-node data exists there at all, so the fallback is the primary path for 100% of multi-node queries behind the PR's headline command.

Proposed change (small diff):

1. Add an explicit opt-in, default OFF — e.g. `Task(..., wideep_allow_node1_fallback: bool = False)` plus a CLI flag `--wideep-allow-node1-fallback` (naming flexible).
2. Default (opted out) + exact-node miss → raise `PerfDataNotAvailableError` whose message names the missing slice and the flag, so when a Blackwell sweep comes back empty the operator sees exactly why and what the two options are (collect multi-node data, or accept the approximation explicitly).
3. Opted in → substitute from the **largest measured `node_num` ≤ requested** rather than jumping straight to node-1 (today an H200 `node_num=16` query would skip the real node-8 data and use node-1 — strictly worse for no reason), keep the once-per-table log, and annotate the affected configs/rows as estimated so downstream ranking stays honest.
4. Track the real fix where it already belongs — the Cross Node EP PoC estimation work (single→multi-node scaling validated against the H100/H200 node-scaling truth above). Note for GB systems: the NVL72 domain makes EP16–64 an intra-fabric all-to-all, so a 16-node × 4-GPU run of the same benchmark with `allow_mnnvl=True` measures it directly — no `test_internode` 8-local-rank constraint.

Yes, this means `gb300 --total-gpus 64` needs the flag until GB multi-node data lands. That is the honest state of the data, and one flag is a low bar for an explicit "I accept estimates" contract.

### Hard blockers (all independently reproduced on this branch)

**1. The committed LL data was measured with masked routes — it needs re-collection.** `test_low_latency.py` unconditionally masks 10 random `topk_idx` positions with `-1` (not gated on `do_check`), and the bandwidth accounting counts only surviving selections. Committed H200 row `(hidden=2048, tokens=1, topk=8, experts=128)`: implied dispatch payload = 8.467 µs × 0.502657 GB/s = 4256 B = exactly **2.00 of 8** routes. Across the six committed LL files, **35 rows have 0.0 GB/s bandwidth** — the benchmark dispatched nothing and recorded the latency of an empty collective — all at small-token/low-topk points, i.e. exactly the low-batch decode regime LL mode exists to model, and the error direction is optimistic. **Now:** masking always on, payload unchecked. **Should:** skip masking for perf collection (or gate it on `do_check`), assert the recorded payload equals `tokens × topk` selections, re-collect the LL tables.

**2. `status: complete` is false and the manifest is stale.** GB200's LL parquet holds **1 shape of 14 (11 rows, DeepSeek-V3 only)**; GB300 has 4 shapes truncated to ≤8-token sweeps (including the Kimi-K2 geometry 7168/8/384); both metas say `complete`. The manifest's `total_raw_rows: 1054` matches neither the six committed files (751) nor all versions combined (941) — but 941 − 11 + 124 = 1054, i.e. GB200's file was cut from 124 rows to 11 **after** the manifest was generated, and nothing was regenerated. `status:` is load-bearing in the V3 loader (`partial` excludes a version dir from loading/reuse), so this label does real work and it is wrong. **Now:** partial data labeled complete, stale manifest, PR-body coverage claims that describe only the normal table (LL is 154/124/11 rows per system, not "2208/2346"). **Should:** finish or re-run GB200+GB300 LL (same campaign as the masking fix), or mark truthfully (splitting tables if needed so a partial LL doesn't exclude the good normal data), regenerate the manifest, correct the PR body.

**3. Version provenance is misfiled.** Every committed row embeds `version=0.5.12`; the dirs and sidecars say `0.5.10`; the branch carries `chore: move data from 0512 to 0510`. The loaders never read the embedded version column, so the directory is the only version truth — and it now lies. **Now:** row says 0.5.12, dir/meta say 0.5.10. **Should:** file under the measured runtime (0.5.12) and, if 0.5.10 compatibility is the intent, express that through the declared-reuse mechanism instead of relocating files — row, dir, and sidecar must agree.

**4. The Kimi K2.5 remap targets a model that cannot consume it, and there are now three different WideEP predicates.** The new remap keys on `moe_backend == "deepep_moe"` and includes `KimiK25ForConditionalGeneration`, but `DeepSeekModel.create()` routes KIMIK25 to the plain model ("WideEP variants are DEEPSEEK-V3-specific") and validation keys on `enable_wideep`. Reproduced: on this branch, Kimi + `enable_wideep` passes validation and then dies mid-sweep with `sweep_agg: no results for any parallel configuration` (no fp8_block slice in the regular MLA table); on main the same task fails fast at validation with a clear `UnsupportedWideepConfigError`. `moe_backend='deepep_moe'` without `enable_wideep` likewise fails on a different error than main. **Now:** remap, validation, and model construction each use a different predicate. **Should:** derive all three from one shared predicate; exclude Kimi from the remap until a WideEP Kimi model exists; add a test that builds the constructed model and queries it (attribute assertions on `Task` don't catch this class of bug).

### Smaller items (non-blocking, please track)

- `audit_kernel_source.py` recognizes the LL latency columns but none of the normal schema's four component columns (`*_transmit_us`, `*_notify_us`) — a manifest regeneration would silently drop the whole normal group. Teach it to compose the normal metric.
- The full-node interception in `collect_sglang` keys on `ops` only, but `_registry_with_requested_wideep` also admits `case_plan.ops` — a case-plan-requested `deepep_ll` would flow into the per-GPU worker pool and every worker would spawn a full-node job (N² processes). Cheap guard: exclude `fullnode_ops` on the case-plan path too.
- The full-node runs bypass `build_collections`, so no `ResumeCheckpoint` case IDs and no V3 sidecar exist; the committed metas are hand-written `provenance: legacy`. Either wire these ops into the provenance path or document the exemption where Collector V3 expects it.
- 6 of 22 LL shapes and 2 of 22 normal shapes are filtered client-side. The kernel-limit justifications in the code comments are legitimate and well documented — but surface them as accounted exclusions (case plan / PR body); the PR body currently discloses only two of them.
- Trivia: the fallback log is `logger.warning` while the comments/PR body say INFO; `run_deepep_normal_fullnode` defaults `perf_filename` to a raw string where the LL twin uses `PerfFile`; the empirical MLA paths accept any `attention_backend` and can silently resolve an invalid value to `trtllm_mla` (the silicon paths validate against `{flashinfer, fa3}`).

### Merge path

(1) gate + annotate the fallback with closest-measured-scale selection (this PR, small diff); (2) fix the masking and re-collect LL in one campaign, with truthful metas, regenerated manifest, and agreeing version provenance; (3) unify the WideEP predicate and drop Kimi from the remap; (4) file the PoC follow-ups (validated node-scaling estimation; GB in-fabric multi-node collection). With those, this lands well — the collector architecture and the normal-mode dataset are already in good shape.

---
🤖 Review compiled with [Claude Code](https://claude.com/claude-code); all reproductions run locally against `harrli/cross-node-moe` @ `329bcfe4`.

collector/wideep/sglang/collect_deepep_normal.py
Harrilee: All labels updated to 0.5.10.

src/aiconfigurator/sdk/task_v2.py
Harrilee: You're right, and thanks for catching this — there was a genuine inconsistency. `task_v2.py` was labelling Kimi as a WideEP model while deepseek.py has always routed `KIMIK25` away from `WideEPDeepSeekModel`, so the resolved quant labels described a configuration the model builder never constructs.

I've dropped `KimiK25ForConditionalGeneration` from the WideEP label-remap predicate, so Kimi again resolves exactly as it does on main. I left `deepseek.py` untouched, since it was never part of this branch, and added a regression test that pins Kimi out of the WideEP path so it can't be re-enabled silently.

Kimi WideEP support is deferred to its own PR, as it's more than a label fix: the WideEP MLA ops hardcode DeepSeek-V3's 128 attention heads (`num_head = 128 // tp_size`, about a dozen live sites in `mla.py`, plus `engine.py:550,562` and two Rust mirrors). Until that's parameterized, Kimi's 64 heads can't be modelled correctly.

aic-core/src/aiconfigurator_core/systems/op_kernel_source_manifest.yaml
Harrilee: Agreed on the substance — fixed in `a0dc216f`.

You're right that the audit can't handle the normal schema. `_pick_latency_column` resolves a table's metric from the first known latency column in the header, and the normal table has none: its latency is split across `dispatch_transmit_us`, `dispatch_notify_us`, `combine_transmit_us` and `combine_notify_us`. So a regeneration didn't correct that entry, it **deleted** it.

The fix resolves the metric from a group of columns to sum, using the same sums the SDK applies at load time (`moe.py`, `load_wideep_deepep_ll_data` / `load_wideep_deepep_normal_data`): `dispatch + combine` for the low-latency table, the four transmit/notify components for the normal one. All component columns leave the shape key, and a row with any unparseable component is skipped rather than composed from a partial sum. Note this also changes the LL metric, which was previously keyed on `combine_avg_t_us` alone.

One correction on the count: `1054` was already right, so it doesn't move. The six `0.5.10` files hold 864 rows (154×4 + 124×2 — GB200 has since been backfilled from 11 to 124); the other 190 are the two legacy `0.5.6.post2` h100/h200 tables, which carry the same `deepep` kernel_source and so aggregate into the same group. The entry that was actually wrong is `wideep_deepep_normal_perf`: `13524` hand-written against an actual `14398`.

While verifying this I also hit a reproducibility bug worth fixing, since the whole point here is that these numbers be re-derivable. Parsed files were merged in thread-completion order while `record_row` is last-write-wins per `(shape, framework)`, so `median_pct_divergence` varied across identical runs — `gdn_perf` / `causal_conv1d_update` gave 25.84, 32.49 and 34.29 on three runs against unchanged data. Merging in submission order fixes it; four consecutive runs are now byte-identical.

A scoped regeneration now emits blocks byte-identical to what's committed, and both counts agree with a direct pyarrow count over the eight contributing tables. I added `--op-file` so one table can be re-derived in seconds instead of the ~9 minutes a full walk takes, plus unit tests pinning the composition and the merge order.

I scoped the commit to the two DeepEP groups deliberately. A full-tree regeneration rewrites ~800 lines; that drift predates this branch, which touches no perf table other than these two, and the `op-kernel-source-manifest` daily-run workflow owns folding it back in. That workflow looks like it's behind.

collector/wideep/sglang/__init__.py
```diff
@@ -2,3 +2,33 @@
...
+
+        return get_collector_runtime("sglang", workload="wideep").version
+    except Exception:
+        pass
+    try:
+        from importlib.metadata import version as get_version
+
+        return get_version("sglang")
+    except Exception:
+        return "unknown"
```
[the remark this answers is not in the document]
Harrilee: Stale — the claim does not hold at HEAD. dataset_version_label is defined exactly once, at collector/wideep/sglang/__init__.py:9, and both collectors already import it (collect_deepep_ll.py:330, collect_deepep_normal.py:376). There are no local copies to remove.

collector/collect.py
Harrilee: Agreed. I updated the SGLang full-node DeepEP path so `deepep_ll` and `deepep_normal` still use the full-node runners, but no longer disappear from the Collector V3 bookkeeping.

Concretely, `collect_sglang()` now builds the normal collection mapping first, keeps the full-node collections in `provenance_ctx["collections"]`, and then splits execution into:

- regular ops → existing per-GPU worker pool
- `deepep_ll` / `deepep_normal` → full-node runner, executed once

The full-node path now creates normal `ResumeCheckpoint` case IDs for the selected cases. If the runner reports failed cases, those cases are marked failed in the checkpoint so finalize-time sidecar generation can derive `partial` instead of relying on hand-written `provenance: legacy` entries.

I also fixed the adjacent case-plan routing issue: a case-plan-requested full-node op is no longer allowed to fall through into the per-GPU worker pool, so it cannot spawn one full-node job per worker.

Validation:
- `ruff check` / `ruff format --check` / `git diff --check` passed locally.
- Ran a smoke on `ocics001` with a temporary checkout and mocked full-node runner: direct `ops=["deepep_ll"]` produced a non-empty collection mapping and checkpoint for one selected case, and the `case_plan.ops=["deepep_ll"]` path bypassed the per-GPU pool as expected.

collector/collect.py
```diff
@@ -100,6 +100,43 @@ def setup_warning_filters():
 # Failures of one (model, dtype) group within an op before the summary flags
 # it as systemic (a fix-me warning; nothing is skipped).
 SYSTEMIC_GROUP_THRESHOLD = 5
+SGLANG_FULLNODE_OPS = {
```
jasonqinzhou: let's move full node collector logic to a separate file.

aic-core/src/aiconfigurator_core/systems/data/b300_sxm/comm/sglang/0.5.10/collection_meta.yaml
```diff
@@ -9,3 +9,7 @@ runtime:
 tables:
   custom_allreduce_perf:
     status: complete
+  wideep_deepep_ll_perf:
```
jasonqinzhou: can we confirm that current AIC can reuse these wide ep data points for newer versions if newer version doesn't have these data points.

Arsene12358:
Checked the latest pushes (through `2c3ad738`) against my review — status from my side.

**Resolved on the code layer — nice work:**

- The `num_max_dispatch_tokens_per_rank` root-cause (`465be748`) is excellent. It explains both why `do_check` had to be off and where the zero-bandwidth rows in the committed LL tables came from, and it reaches a stronger conclusion than my masked-routes finding did: the samples weren't just partially masked, they were collected in a configuration no deployment runs. The follow-up (`0b348f66`) that corrects the mechanism claim after reading the DeepEP source while keeping the measured finding is exactly the right standard for perf-data provenance.
- Route masking is now disabled for perf samples with correctness asserted unconditionally — that closes the collection half of hard blocker 1.
- The comm-family 0.5.12 image pin (`bbd90683`) settles the version-provenance direction for future collections, digest-pinned and honestly labeled. Thanks — that's what blocker 3 asked for.
- The `collect.py` full-node provenance work + the case-plan routing guard close the ResumeCheckpoint/sidecar gap and the case-plan N² spawn hole. I'll defer to @jasonqinzhou's structural comment on splitting it into its own file.
- The Kimi drop (`4fb59d32`) verifies on my side (fail-fast restored, matching main), and the sms cache-corruption find/fix (`1a8f352d`) was a genuine bug and a good catch.

**Still open before I can flip my review:**

1. **The committed parquets are unchanged** — and by your own capacity analysis they were measured in an invalid configuration. Is a re-collection on the pinned 0.5.12 image planned for all six systems within this PR? One campaign would close blocker 1 (zero-bandwidth rows), blocker 2 (GB200/GB300 token-truncated shapes + truthful `status:`), and blocker 3 (rows and directories land under 0.5.12) together. If re-collection is deferred instead, the current tables shouldn't ship labeled as `complete` silicon data.
2. **The node-1 fallback gating proposal hasn't had a response yet** — it's the central ask of my review. Concretely: fail-closed by default with an actionable error naming the flag, an explicit `--wideep-allow-node1-fallback` opt-in, substitution from the largest measured `node_num ≤ requested` rather than always node-1, and estimated-annotation on affected configs. If you disagree with any part, happy to discuss here — measured node8/node1 on H200 is 2.1–4.3×, so silent substitution isn't a state I can approve, but I'm flexible on the mechanism.
3. Two smaller items: (a) the remap/validation predicate still diverges for `moe_backend='deepep_moe'` without `enable_wideep` (the remap applies fp8_block while validation checks the regular table → hard `ValueError`; one shared predicate closes it); (b) the sms cache-corruption regression test from `1a8f352d` was dropped in `89dcfbd5` — agreed a standalone module was heavy, but that bug class deserves a pin; a small case inside the existing `test_moe_dispatch.py` would do.

---
🤖 Follow-up compiled with [Claude Code](https://claude.com/claude-code); all verifications run locally against `harrli/cross-node-moe` @ `2c3ad738`.

aic-core/src/aiconfigurator_core/sdk/operations/moe.py
jasonqinzhou: ignore this comment.

Harrilee: Ignored after discussion.

collector/wideep/sglang/collect_deepep_normal.py
Harrilee: Recollected 0.5.12

aic-core/src/aiconfigurator_core/sdk/operations/mla.py
```diff
@@ -48,6 +48,29 @@
...
+    requested backend when the table actually carries it; otherwise fall back to
+    the kernel source that was collected (e.g. ``trtllm_mla`` on Blackwell) so the
+    lookup does not miss purely on a name mismatch. If neither is present, return
+    the requested backend and let ``require_data_slice`` raise the standard
+    coverage error.
+    """
+    try:
+        if attn_backend in data_wrapper:
+            return attn_backend
+        if "trtllm_mla" in data_wrapper:
```
jasonqinzhou, line 67: **[P1] Resolve the effective backend before selecting perf data**

Could we avoid resolving this based on whichever table key happens to exist? In SGLang, `trtllm_mla` is the DeepSeek/SM100 auto-selection only when no attention backend is specified. Here `Task.attention_backend=None` is converted to `flashinfer` before model construction, so this helper cannot distinguish `auto` from an explicit `flashinfer`/`fa3` request and can rank an explicitly requested backend using `trtllm_mla` measurements.

Please preserve `None`/`auto`, resolve the effective backend once from framework/version/model/GPU policy, then require exact perf coverage for that resolved backend (fail closed for explicit requests). The same effective backend must reach the generated launch configuration. Regression tests should cover auto Blackwell -> `trtllm_mla`, explicit `flashinfer`/`fa3` not falling back, Hopper passthrough, and missing coverage.

aic-core/src/aiconfigurator_core/systems/data/gb200/comm/sglang/0.5.12/collection_meta.yaml
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+schema_version: 1
+provenance: legacy
```
jasonqinzhou: **[P1] Record trustworthy provenance for the recollected tables**

These are newly recollected SGLang 0.5.12 tables, but all six new sidecars declare `provenance: legacy`. They contain no image digest, collector hash, case-plan hash, collection date, or row counts, so the artifacts cannot be tied back to the campaign that produced them. Please publish real collection provenance for each new table rather than using the legacy escape hatch.

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -497,8 +497,16 @@ def get_worker_candidates(
...
                     if not summary.check_oom() and not summary.check_kv_cache_oom():
                         all_configs_oom = False
+                        candidate_df = summary.get_summary_df().copy()
+                        # Out-of-band column (leading underscore, like
+                        # _per_ops_source): carries op provenance forward without
+                        # touching the declared ColumnsStatic schema. Disagg rows
+                        # are composed arithmetically from these candidates and
+                        # have no summary of their own, so this is the only place
+                        # the flag can be captured for them.
+                        candidate_df[common.ESTIMATED_COLUMN] = summary.has_estimated_source()
```
jasonqinzhou: **[P1] Preserve the estimated flag through every disaggregated result builder**

This captures the flag on worker candidates, but two supported disaggregated paths later rebuild results with `columns=common.ColumnsDisagg`, which drops this out-of-band column: autoscale in `src/aiconfigurator/sdk/picking.py:504` and the public `DisaggInferenceSession` path in this file at line 814. Consequently, multi-node DeepEP estimates can still appear unmarked. Please OR `_estimated` from the prefill/decode inputs into the rebuilt rows in both paths and cover both with regression tests.

collector/wideep/sglang/collect_deepep_normal.py
```diff
@@ -0,0 +1,675 @@
...
+
+    Checked per shape, so shapes legitimately absent from the CSV -- skipped for
+    ``num_experts % num_ranks`` or dropped on a buffer-allocation failure, both of
+    which the worker logs -- do not raise here. Whole-shape loss is accounted for by
+    the caller against the orchestrator's own succeeded/failed tally; this is about
+    an axis being short *within* the shapes that did run.
+    """
+    import csv
+
+    want_sms, want_tokens = set(sms_list), set(tokens)
```
jasonqinzhou: **[P1] Validate the complete `(sms, token)` grid before certifying the table**

Tracking `got_sms` and `got_tokens` independently does not prove the full Cartesian grid: a missing pair passes whenever that SMS and token each appear in other rows. In addition, whole-shape loss at line 659 only warns, after which the run still reports `collection complete` and downstream provenance can attest `status: complete`. Please validate the exact requested pair set for every planned shape, and fail or mark the table partial when any pair or shape is absent.

jasonqinzhou (commented):
Three remaining correctness and artifact-provenance issues:

aic-core/src/aiconfigurator_core/systems/data/gb200/comm/sglang/0.5.12/collection_meta.yaml
Harrilee: Updated

aic-core/src/aiconfigurator_core/sdk/common.py
```diff
@@ -683,6 +683,14 @@ def get_default_models() -> set[str]:
     "power_w",  # NEW: E2E weighted average power in watts
 ]
 
+# Out-of-band result column: True when the row's latency includes an op tagged
+# PerformanceResult.source == "estimated". Deliberately NOT a member of
+# ColumnsStatic / ColumnsAgg / ColumnsDisagg -- those are the declared result
+# schemas the sweep constructs against, and a DataFrame built with
+# columns=<schema> would drop this, so it is attached after construction the
+# way "_per_ops_source" is.
+ESTIMATED_COLUMN = "_estimated"
```
jasonqinzhou: **[P1] Scope this marker to the WideEP communication fallback**

The _estimated column and source="estimated" are too broad for this behavior. This path specifically means that WideEP DeepEP communication reused node_num=1 data for an unmeasured multi-node scale; generic estimated sources already exist for DSA, DSV4, and static-FP8 GEMM. Please name the source, column, and helper wideep_comm_node1_fallback and scope the CLI warning to that exact provenance so unrelated modeled operations are not mislabeled.

jasonqinzhou: Addressed in fdc3bb91. The source, column, and helpers now use wideep_comm_node1_fallback; generic estimated sources no longer trigger the WideEP communication warning. Aggregate, disaggregated, autoscale, CSV, CLI, API documentation, and regression coverage were updated. Ruff passed and 151 focused tests passed.

aic-core/src/aiconfigurator_core/sdk/operations/moe.py
```diff
@@ -983,6 +983,69 @@ def clear_cache(cls) -> None:
...
+    def _wideep_comm_node1_fallback(
+        data_by_node,
+        *,
+        table: str,
+        requested_node_num: float,
+        hidden_size: int,
+        topk: int,
+        num_experts: int,
+    ) -> dict:
+        """Return matching node-1 WideEP communication data or fail clearly."""
```
jasonqinzhou, line 1039: Removing `_MEASURED_NODE_SCALING_SOURCE` and `_MEASURED_NODE_SCALING` here is intentional. Those hard-coded ratios came from the H100/H200 SGLang `0.5.6.post2` campaign and were used only to format the old warning; they never scaled the returned latency. With node-1 reuse now defined as the supported WideEP estimate (`source="estimated"`) and missing node-1 rows failing explicitly, retaining the old runtime constants would be dead and potentially misleading beside the `0.5.12` data.

Arsene12358 (approved):
## Verdict: APPROVE — all blockers from my review are resolved, re-verified locally at `4f35efed`

Closing the loop on my request-changes review and the follow-up checklist; every item was re-verified against the current head, not just read from commit messages:

- **Blocker 1 (LL data measured with masked routes / invalid buffer config): closed.** The recollected 0.5.12 tables carry 16 shapes × 11 full token sweeps on all six systems with **zero** zero-bandwidth rows (the old tables had 42), collected with capacity=128 and correctness asserted unconditionally. Coverage even grew — deep_ep 1.2.1 unlocks the topk=10 shapes.
- **Blocker 2 (false `complete`, stale manifest): closed.** GB200/GB300 are fully covered with no token-truncated shapes, `status:` labels are now truthful per table (and I confirmed partial-marked dirs still load), and the manifest matches the committed files.
- **Blocker 3 (version provenance): closed, beyond what I asked.** The relabeled 0.5.10 tables are gone; rows, directories, and sidecars all agree on 0.5.12; and the sidecars now record image digest, collector ref/hash, case-plan hash, collection date, and row counts. This is what trustworthy perf-data provenance looks like.
- **Blocker 4 (Kimi remap / predicate divergence): closed.** Kimi fails fast at validation again, and after `4762e129` the remap, validation, and model construction key on one predicate — the `moe_backend='deepep_moe'`-without-`enable_wideep` combo that previously died on a labeling mismatch now runs a full sweep green on my machine.
- **Gating proposal: resolved by a mechanism I'm happy to accept.** I asked for an opt-in flag; what landed is annotate-by-default — node-1 reuse is typed `source="wideep_comm_node1_fallback"` end to end (agg/disagg/autoscale/CSV/CLI/API) with a scoped warning, and a true miss fails closed. That achieves the substance of the ask — no silent substitution, honest ranking provenance — and the precisely-scoped source name makes downstream filtering trivial, so I'm conceding the flag. The follow-up that makes the estimate itself trustworthy (validated node-scaling, and ideally in-fabric EP16–64 collection on NVL72) remains important future work.

Two non-blocking notes for the record: the sms cache-corruption guard from `1a8f352d` still has no regression test (a small case in `test_moe_dispatch.py` would pin a subtle bug class), and I'm deliberately not speaking to the open MLA auto-vs-explicit backend thread — that one is @jasonqinzhou's to close.

Thanks @Harrilee — the capacity root-cause, the honest re-collection, and the provenance sidecars turned this from a PR I had to block into one that raises the bar for perf-data changes in this repo.

---
🤖 Review compiled with [Claude Code](https://claude.com/claude-code); verifications run locally against `harrli/cross-node-moe` @ `4f35efed`.

PR #1323, opened 2026-07-07
feat(afd): add AFD default CLI mode built on the v2 Task architecture  

gehujun:
Summary

    This PR adds an **AFD (Attention–FFN Disaggregated)** default CLI mode
    built directly on the v2 `Task` dataclass + `sweep_afd` pipeline, without
    routing through the legacy `TaskConfig` / `TaskRunner` path.

    It brings AFD to feature parity with the existing `default` mode for
    Dense / PD deployments: users can now run

   bash
    aiconfigurator cli default --serving-mode afd 
   and get a Pareto sweep, SLA-aware picking, and a full report/summary,
   end-to-end through the v2 Task stack.

   Motivation

   - The legacy AFD path relied on _LegacyAFDTaskAdapter and duplicated
   configuration handling on top of the old TaskConfig.
   - v2 Task is now the canonical entry point for CLI/SDK; AFD was the
   last serving mode still using the legacy adapter.
   - Removing the adapter unblocks further refactors (unified validation,
   unified sweep interface) and keeps AFD in sync with future Task/sweep
   improvements automatically.

   Changes

   v2 Task (sdk/task_v2.py)

   - Extend Task with serving_mode='afd' and 21 AFD-specific fields.
   - Add _resolve_afd_search, sweep_afd_kwargs, _validate_afd.
   - Migrate build_afd_parallel_lists and normalization helpers off the
   legacy module.

   Sweep pipeline (sdk/sweep.py, sdk/pareto_analysis.py)

   - Add sweep_afd() as a thin wrapper around a new afd_pareto().
   - afd_pareto() provides:
     - KV-cache-derived batch sizing
     - Combined-with-PD static prefill
     - SLA constraint enforcement on the AFD frontier

   CLI (cli/main.py, cli/report_and_save.py, cli/utils.py)

   - Remove _LegacyAFDTaskAdapter; drive AFD through v2 Task in both
   default and experiment modes.
   - Add AFD-specific report rendering and summary comparison.

   Tests

   - tests/unit/cli/test_afd_default_cli.py — CLI integration
   - tests/unit/sdk/test_afd_parallel_lists.py — parallel-list builder
   - tests/unit/sdk/test_afd_pareto.py — Pareto core
   - tests/unit/sdk/test_afd_pareto_review_actions.py — SLA / review

   Backward Compatibility

   - CLI surface is unchanged; --serving-mode afd continues to accept the
   same flags as before.
   - Legacy TaskConfig-based AFD entry points are removed only from the
   CLI path; SDK users on the legacy API are unaffected.
   - No changes to on-disk perf data or system YAML schema.

   Testing

   - pytest -m unit — all AFD unit tests pass locally
   - aiconfigurator cli default --serving-mode afd --model-path Qwen/Qwen3-32B-FP8 --total-gpus 8 --system h200_sxm
   produces a valid AFD Pareto report end-to-end.

   Checklist

   - Code compiles and lints (ruff check / ruff format --check)
   - New/changed code is covered by unit tests
   - No changes to public CLI flags or config schema
   - Docs / help strings updated where user-visible behavior changed
   - Follow-up: extend experiment mode presets with AFD examples

## Summary by CodeRabbit

* **New Features**
  * Added `--serving-mode afd` to the default workflow, including AFD Pareto sweep (`sweep_afd`/`afd_pareto`) and AFD experiment-task support.
  * Added new `h20` (H20-3e) hardware profile.
* **Improvements**
  * Enhanced AFD reporting (AFD vs disagg classification, richer GPU breakdown, optional “AFD vs best non-AFD” ratio) and extended AFD metrics/schemas (e.g., `request_rate`, `parallel`), including AFD-aware result merging.
* **Bug Fixes**
  * Added stronger AFD feasibility validation and refined AFD decode latency modeling; skip AFD deployment artifact generation during result saving.
* **Tests**
  * Added unit coverage for AFD CLI wiring, task construction, Pareto/sweep behavior, and result handling.


src/aiconfigurator/cli/main.py
```diff
@@ -220,9 +220,18 @@ def _add_default_mode_arguments(parser):
         default=common.BackendName.trtllm.value,
         help="Backend name. Use a specific backend (trtllm, vllm, sglang) or 'auto' to sweep "
         "across all supported backends for the given system and compare results side by side. "
-        "When 'auto' is used, both agg and disagg results are merged across backends and the "
+        "When 'auto' is used, selected serving-mode results are merged across backends and the "
         "globally optimal configuration is selected. Default: trtllm.",
     )
+    parser.add_argument(
```
jasonqinzhou, line 329: --serving-mode appears to be introduced by this PR. Previously, default always compared agg and disagg, while AFD was available only through single-point estimate mode.
Instead of adding a new auto/all selector, could we:
Keep serving_mode: afd available in exp for explicit or customized runs.
Have default automatically include AFD when it is supported and the GPU budget is at least two nodes; otherwise, continue comparing only agg and disagg.
This seems more consistent with the purpose of default: automatically compare all feasible serving strategies. Unsupported AFD configurations should be skipped with a clear warning rather than failing the entire default run.
We should also update the documentation/API accordingly and correct the PR description’s statement that the CLI surface is unchanged, since --serving-mode is new in this PR.

src/aiconfigurator/cli/main.py
```diff
@@ -220,9 +220,18 @@ def _add_default_mode_arguments(parser):
...
+        "--serving-mode",
+        choices=["auto", "all", "agg", "disagg", "afd"],
+        type=str,
+        default="auto",
+        help="Serving modes to sweep and compare. 'auto' (default) compares agg and disagg; "
+        "'all' compares agg, disagg, and afd; pick a single mode to restrict the search. "
+        "afd requires at least 2 nodes worth of GPUs and is skipped otherwise.",
```
jasonqinzhou: Is the two-node minimum an intentional deployment limitation?

tianhaox (changes requested):
Reviewed this PR in depth (all findings below were independently re-verified against the PR branch before posting). The overall direction is right — making AFD a first-class v2 `Task` serving mode and deleting the legacy adapter is the correct architecture. However, verification found **8 confirmed correctness defects**, two of which break features this PR itself advertises (pinned single-point estimation, large-cluster sweeps), so I'm requesting changes.

## Confirmed defects (ranked by severity)

**1. `sdk/task_v2.py` (~L2066) — pinned single-point AFD crashes 100% of the time.**
`_run_afd_single_point` calls `AFDConfig(..., n_a_workers=n_a_workers, ...)`, but `AFDConfig.n_a_workers` is declared `field(init=False)` (derived in `__post_init__`, `config.py:260`). Reproduced on this branch: `TypeError: AFDConfig.__init__() got an unexpected keyword argument 'n_a_workers'`. Any run with `afd_n_a_nodes`/`afd_n_f_nodes`/`afd_tp_a` set crashes — this path has no test coverage.

**2. `cli/main.py` (~L1562) — the AFD experiment-YAML branch silently drops every `afd_*` field and `nextn`.**
Instead of `Task.from_yaml` (used by the agg/disagg branch one line below), the AFD branch builds `Task(**afd_kwargs)` from a hard-coded 22-key whitelist. Pinned topology, `afd_combined_with_pd`, `afd_total_batch_size`, calibration factors, and `nextn`/`nextn_accept_rates` are all dropped with no warning — and this bypasses `from_yaml`'s unknown-key `ValueError` guard, so users get a default full sweep that silently ignores their YAML. Since v2 `Task` already supports `serving_mode='afd'`, the whole branch can collapse to `Task.from_yaml(task_config, **overrides)`.

**3. `generator/module_bridge.py` (~L191) — AFD results produce fabricated deployment artifacts.**
`task_config_to_generator_config` only branches agg vs everything-else(=disagg). AFD rows have no `(p)*`/`(d)*` columns (`ColumnsAFD` uses `(a)*`/`(f)*`), and `_series_val` silently falls back to defaults, yielding a fictitious tp=1/pp=1/bs=1 disagg topology. `save_results` generates artifacts for **every** experiment unconditionally (`report_and_save.py:845` → `:1015`), so with `--save-dir` set a bogus `generator_config.yaml` is written whenever AFD is feasible — AFD doesn't even need to win the comparison. At minimum, skip artifact generation for AFD with a warning.

**4. `sdk/task_v2.py` (~L399) — default `afd_max_candidates=2000` + `candidate_overflow="error"` makes AFD fail on ≥128-GPU clusters.**
Counted with the actual enumeration logic: 128 GPUs dense = 2040 candidates > 2000 (MoE overflows at 96 GPUs with 3000+). The CLI exposes no override for `afd_max_candidates`/`afd_candidate_overflow`. Result: `--serving-mode afd` dies with `SystemExit(1)` ("No task configs could be built"); `--serving-mode all` silently drops AFD from the comparison. Suggest defaulting to truncate (or exposing the knobs on the CLI).

**5. `sdk/pareto_analysis.py` (~L1284) — the quick balance-ratio prefilter probes the F pool at the wrong batch size, pruning genuinely balanced topologies.**
The prefilter queries both pools at the same per-A-worker `probe_bs`, but the real evaluation feeds the F pool the aggregated tokens `b_micro_total = n_a_workers * a_micro_batch_size` (`inference_session.py:991`). `t_f` is underestimated by up to `n_a_workers`×, so a truly balanced candidate gets `quick_ratio < 0.3` and is rejected (`rejection_counts["balance"]`) before `run_afd` ever runs — the optimal AFD topology can be silently excluded from the sweep. The probe's F-side batch should be scaled by the A-worker count.

**6. `sdk/task_v2.py` (~L1714) — combined-with-PD prefill enumeration applies the wrong backend's parallel filters.**
`iter_parallel("prefill")` reads `prefill_backend_name` (dataclass default `"trtllm"`), and `_resolve_afd_search` never copies the task's `backend_name` into it. For vLLM/SGLang AFD MoE tasks this applies trtllm's `dp>1 and tp>1` pruning (dropping valid configs) while skipping vLLM's `moe_tp>1 and moe_ep>1` unsupported filter (admitting configs the backend can't run). Fix: set `prefill_backend_name = backend_name` (and `prefill_enable_wideep`) during AFD resolution.

**7. `cli/main.py` (~L1279) — regression (non-AFD): HYBRID mode lost the decode-system availability skip in `--backend auto`.**
The pre-PR code used `_database_mode_requires_declared_perf_database` (covers SILICON **and** HYBRID) and warn-skipped backends whose decode system lacked data. The new `_disagg_backend_available` early-returns True for anything != SILICON. Under `--backend auto --database-mode HYBRID` with a decode system missing that backend's data, a disagg Task is now built with `decode_backend_version=None` and fails deep inside `sweep_disagg(decode_database=None)` with a raw AttributeError instead of the old clean skip. The condition should be restored to `_database_mode_requires_declared_perf_database(database_mode)`.

**8. `sdk/task_v2.py` (~L1813) — "empty candidate list" is overloaded as the pinned-topology sentinel.**
`build_afd_parallel_lists` silently filters invalid user candidates (e.g. `afd_tp_a_candidates=[3]` on an 8-GPU/node system → `[]`, log line only). An unpinned task with a legitimately empty search space is then misrouted into `_run_afd_single_point` (`if not self._afd_parallel_config_list`), which crashes with `TypeError` on `n_f_nodes * gpus_per_node` (None × int) instead of raising a clear no-feasible-config error. Use an explicit pinned flag and raise `NoFeasibleConfigError` for an empty candidate list.

## Additional issues worth addressing

- **`generator/facts/hardware.yaml` was rewritten via a YAML dump**: 87 of the 117 removed lines are comments — the SPDX header and all provenance/guard comments (WIDEEP-crashes-on-Hopper rationale, NCCL env sources, node_selector provenance) are stripped. The h20/h20_3e addition is also unrelated to AFD; please hand-append just the new block (preserving comments) and consider splitting the hardware support into its own PR.
- **Docs**: the new `--serving-mode` flag is not documented in `docs/cli_user_guide.md`.
- **Third copy of the rate-matching calibration**: `_AFD_PREFILL_DEGRADATION` / `_AFD_TTFT_CORRECTION_FACTOR` and the combine formula duplicate `picking.py`/`sweep.py`, which are locked together by a parity test — the AFD copies sit outside that lock and will silently drift when the calibration is re-tuned.
- **Performance**: `afd_pareto` nests the full candidate sweep inside the SLA-pair loop, rebuilding SLA-invariant per-candidate state (`get_model`×2, partitions, memory probes) per runtime config; each batch-size probe also constructs a fresh `AFDInferenceSession` (re-running `_build_models`). Caching per candidate / `topo_key` would cut wall-clock substantially on large sweeps.

Findings 1, 2, and 4 indicate the pinned-topology path, YAML experiment path, and large-cluster path have no regression coverage — please add tests for those alongside the fixes.

src/aiconfigurator/sdk/pareto_analysis.py
```diff
@@ -780,10 +794,11 @@ def _derive_a_batch_size(
         include_kvcache=True,
         kvcache_multiplier=kvcache_multiplier,
         free_gpu_memory_fraction=free_gpu_memory_fraction,
-        align_to=8,
+        align_to=1,
     )
+    max_total_bs = _afd_total_batch_capacity(max_micro_bs, kvcache_multiplier, align_to=8)
 
-    return max(min(max_bs, 256), 32), a_model, a_partition
+    return max(min(max_total_bs, 256), 32), a_model, a_partition
```
yunyang1999: The 256 cap clips exactly the region where AFD wins. At long context, with the (now-correct)
capacity accounting, the optimal *total* A-pool batch commonly lands in the 300–1000 range —
e.g. Qwen3-235B-A22B isl=32K on 32×H200: the KV-capacity-feasible optimum is ~384 total, and
capping at 256 costs ~20% throughput on the winning configs, enough to flip several
AFD-vs-baseline verdicts near the crossover.

Suggest raising the cap to ≥1024, or making it a searchable/CLI knob. (The floor-to-32
behavior is fine as documented — the caller's per-candidate OOM check handles it.)

src/aiconfigurator/sdk/pareto_analysis.py
yunyang1999: With this floor the effective batch grid is {1, 2, 4} ∪ {≥32}: nothing between 5 and 31 is
ever evaluated. At long context that hole covers the feasible optimum — e.g. Qwen3-235B
isl=64K, TPOT≤50ms: per-user KV is multiple GiB, so the best feasible total batch is 16–24.
The search then either returns an infeasible ≥32 point (rejected later) or falls back to the
far-inferior ≤4 point, and the scenario is misreported as "AFD loses" when it's actually near
parity.

Suggest bs_min = 8 (our sweeps converge with that), or extending the low-side grid to
{1, 2, 4, 8, 16, 24}.

src/aiconfigurator/sdk/pareto_analysis.py
```diff
@@ -780,10 +794,11 @@ def _derive_a_batch_size(
...
 
 
 _AFD_BALANCE_RATIO_THRESHOLD = 0.3
```
yunyang1999: Thanks for lowering this from 0.5 — that removes most of the mis-pruning we had measured.
One residual concern: `_quick_balance_ratio` probes A and F at the same bs≤32 and doesn't
account for the n_a_workers×M token aggregation on the F side (nor comm), so its ratio can
deviate ~0.2 absolute from the full session — we measured quick=0.92 vs actual=0.72 on
DeepSeek-V3 with a 1:1 node split. At threshold 0.3 the surviving risk is small but nonzero
(deep-MoE models with heavy aggregation).

Two cheap hardening options: (a) log pruned candidates at INFO with their quick-ratio so a
suppressed region is visible in the sweep summary, or (b) make the prune advisory-only and
let the SLA filter do the rejecting.

src/aiconfigurator/sdk/pareto_analysis.py
yunyang1999, line 1571: Candidate failures are only visible at DEBUG. If a whole tier fails systematically (e.g. an
op-name change breaking one tp_a), the search space silently narrows and the sweep still
"succeeds" with worse results. Suggest logging at WARNING (or keeping DEBUG per-candidate but
emitting a per-sweep summary counter: "N/M AFD candidates failed — top reasons: ...").

yunyang1999: `ceil(decode_rate / per_worker)` only ever generates prefill pools that keep up with decode, and the min-GPU selection below then picks the smallest such pool. Prefill-BOUND deployments — fewer prefill workers than rate-match, decode partially idle, system throughput set by prefill — are never generated, yet they win whenever prefill is the scarce resource (e.g. models whose prefill is only feasible at tp8, or short-osl workloads at tight budgets).

Suggested change: enumerate num_workers ∈ [1..ceil(...)], compute the combined row for each, keep the best by end-to-end tokens/s/gpu (min-GPU selection is only optimal under exact rate-matching).

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: Thanks. The hard-coded 256 cap limits the auto-batch search before TPOT/OOM evaluation, so valid AFD candidates above 256 can be missed.
We’ll replace it with a configurable per A-worker search limit, afd_max_a_batch_size, defaulting to 1024, expose it through Task/CLI, and add a regression test covering a derived capacity of 384.

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: Thanks for pointing this out. I confirmed that the current auto-batch path evaluates {1, 2, 4} separately and starts the 8-aligned binary search at 32, so 8/16/24 are never evaluated. If 32 violates the TPOT constraint while 16 or 24 is feasible, the search can incorrectly fall back to <=4 and bias the AFD comparison.
We will lower both the binary-search entry threshold and its lower bound to 8, while retaining the 1/2/4 low-latency probes. We will also add regression coverage for a case where batch 32 fails TPOT but batch 24 passes, plus a case where the analytical capacity itself is below 32.
We have not independently reproduced the exact Qwen3-235B 64K performance delta yet, but the search-space hole itself is confirmed from the code.

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: Good point. We’ll take option (b): make the quick balance ratio advisory-only and let the full SLA evaluation reject candidates. This avoids mis-pruning caused by the simplified A/F batch and communication model.

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: confirmed. afd_pareto() currently catches per-candidate exceptions and only logs the details at DEBUG, so a partially successful sweep can return normally even when a topology tier has failed systematically. We’ll keep the per-candidate traceback at DEBUG to avoid log flooding, but add a single aggregate WARNING at the end of the sweep with the failed/attempted count and top exception reasons. We’ll also add coverage for the partial-success case so exception-based narrowing is no longer silent.

jieliu1997: Good catch — we’ll fix this by enumerating prefill worker counts from 1 to the rate-matched count and selecting the configuration with the best end-to-end tokens/s/GPU.

jieliu1997:
> Reviewed this PR in depth (all findings below were independently re-verified against the PR branch before posting). The overall direction is right — making AFD a first-class v2 `Task` serving mode and deleting the legacy adapter is the correct architecture. However, verification found **8 confirmed correctness defects**, two of which break features this PR itself advertises (pinned single-point estimation, large-cluster sweeps), so I'm requesting changes.
> 
> ## Confirmed defects (ranked by severity)
> **1. `sdk/task_v2.py` (~L2066) — pinned single-point AFD crashes 100% of the time.** `_run_afd_single_point` calls `AFDConfig(..., n_a_workers=n_a_workers, ...)`, but `AFDConfig.n_a_workers` is declared `field(init=False)` (derived in `__post_init__`, `config.py:260`). Reproduced on this branch: `TypeError: AFDConfig.__init__() got an unexpected keyword argument 'n_a_workers'`. Any run with `afd_n_a_nodes`/`afd_n_f_nodes`/`afd_tp_a` set crashes — this path has no test coverage.
> 
> **2. `cli/main.py` (~L1562) — the AFD experiment-YAML branch silently drops every `afd_*` field and `nextn`.** Instead of `Task.from_yaml` (used by the agg/disagg branch one line below), the AFD branch builds `Task(**afd_kwargs)` from a hard-coded 22-key whitelist. Pinned topology, `afd_combined_with_pd`, `afd_total_batch_size`, calibration factors, and `nextn`/`nextn_accept_rates` are all dropped with no warning — and this bypasses `from_yaml`'s unknown-key `ValueError` guard, so users get a default full sweep that silently ignores their YAML. Since v2 `Task` already supports `serving_mode='afd'`, the whole branch can collapse to `Task.from_yaml(task_config, **overrides)`.
> 
> **3. `generator/module_bridge.py` (~L191) — AFD results produce fabricated deployment artifacts.** `task_config_to_generator_config` only branches agg vs everything-else(=disagg). AFD rows have no `(p)*`/`(d)*` columns (`ColumnsAFD` uses `(a)*`/`(f)*`), and `_series_val` silently falls back to defaults, yielding a fictitious tp=1/pp=1/bs=1 disagg topology. `save_results` generates artifacts for **every** experiment unconditionally (`report_and_save.py:845` → `:1015`), so with `--save-dir` set a bogus `generator_config.yaml` is written whenever AFD is feasible — AFD doesn't even need to win the comparison. At minimum, skip artifact generation for AFD with a warning.
> 
> **4. `sdk/task_v2.py` (~L399) — default `afd_max_candidates=2000` + `candidate_overflow="error"` makes AFD fail on ≥128-GPU clusters.** Counted with the actual enumeration logic: 128 GPUs dense = 2040 candidates > 2000 (MoE overflows at 96 GPUs with 3000+). The CLI exposes no override for `afd_max_candidates`/`afd_candidate_overflow`. Result: `--serving-mode afd` dies with `SystemExit(1)` ("No task configs could be built"); `--serving-mode all` silently drops AFD from the comparison. Suggest defaulting to truncate (or exposing the knobs on the CLI).
> 
> **5. `sdk/pareto_analysis.py` (~L1284) — the quick balance-ratio prefilter probes the F pool at the wrong batch size, pruning genuinely balanced topologies.** The prefilter queries both pools at the same per-A-worker `probe_bs`, but the real evaluation feeds the F pool the aggregated tokens `b_micro_total = n_a_workers * a_micro_batch_size` (`inference_session.py:991`). `t_f` is underestimated by up to `n_a_workers`×, so a truly balanced candidate gets `quick_ratio < 0.3` and is rejected (`rejection_counts["balance"]`) before `run_afd` ever runs — the optimal AFD topology can be silently excluded from the sweep. The probe's F-side batch should be scaled by the A-worker count.
> 
> **6. `sdk/task_v2.py` (~L1714) — combined-with-PD prefill enumeration applies the wrong backend's parallel filters.** `iter_parallel("prefill")` reads `prefill_backend_name` (dataclass default `"trtllm"`), and `_resolve_afd_search` never copies the task's `backend_name` into it. For vLLM/SGLang AFD MoE tasks this applies trtllm's `dp>1 and tp>1` pruning (dropping valid configs) while skipping vLLM's `moe_tp>1 and moe_ep>1` unsupported filter (admitting configs the backend can't run). Fix: set `prefill_backend_name = backend_name` (and `prefill_enable_wideep`) during AFD resolution.
> 
> **7. `cli/main.py` (~L1279) — regression (non-AFD): HYBRID mode lost the decode-system availability skip in `--backend auto`.** The pre-PR code used `_database_mode_requires_declared_perf_database` (covers SILICON **and** HYBRID) and warn-skipped backends whose decode system lacked data. The new `_disagg_backend_available` early-returns True for anything != SILICON. Under `--backend auto --database-mode HYBRID` with a decode system missing that backend's data, a disagg Task is now built with `decode_backend_version=None` and fails deep inside `sweep_disagg(decode_database=None)` with a raw AttributeError instead of the old clean skip. The condition should be restored to `_database_mode_requires_declared_perf_database(database_mode)`.
> 
> **8. `sdk/task_v2.py` (~L1813) — "empty candidate list" is overloaded as the pinned-topology sentinel.** `build_afd_parallel_lists` silently filters invalid user candidates (e.g. `afd_tp_a_candidates=[3]` on an 8-GPU/node system → `[]`, log line only). An unpinned task with a legitimately empty search space is then misrouted into `_run_afd_single_point` (`if not self._afd_parallel_config_list`), which crashes with `TypeError` on `n_f_nodes * gpus_per_node` (None × int) instead of raising a clear no-feasible-config error. Use an explicit pinned flag and raise `NoFeasibleConfigError` for an empty candidate list.
> 
> ## Additional issues worth addressing
> * **`generator/facts/hardware.yaml` was rewritten via a YAML dump**: 87 of the 117 removed lines are comments — the SPDX header and all provenance/guard comments (WIDEEP-crashes-on-Hopper rationale, NCCL env sources, node_selector provenance) are stripped. The h20/h20_3e addition is also unrelated to AFD; please hand-append just the new block (preserving comments) and consider splitting the hardware support into its own PR.
> * **Docs**: the new `--serving-mode` flag is not documented in `docs/cli_user_guide.md`.
> * **Third copy of the rate-matching calibration**: `_AFD_PREFILL_DEGRADATION` / `_AFD_TTFT_CORRECTION_FACTOR` and the combine formula duplicate `picking.py`/`sweep.py`, which are locked together by a parity test — the AFD copies sit outside that lock and will silently drift when the calibration is re-tuned.
> * **Performance**: `afd_pareto` nests the full candidate sweep inside the SLA-pair loop, rebuilding SLA-invariant per-candidate state (`get_model`×2, partitions, memory probes) per runtime config; each batch-size probe also constructs a fresh `AFDInferenceSession` (re-running `_build_models`). Caching per candidate / `topo_key` would cut wall-clock substantially on large sweeps.
> 
> Findings 1, 2, and 4 indicate the pinned-topology path, YAML experiment path, and large-cluster path have no regression coverage — please add tests for those alongside the fixes.

Thanks for pointing these out. I confirmed that all seven are valid issues and have addressed them:
- Fixed pinned single-point AFD construction and added topology validation.
- Switched AFD experiment loading to Task.from_yaml() so AFD, nextn, and quantization fields are no longer silently dropped.
- Disabled AFD deployment artifact generation with an explicit warning until the generator schema can represent A/F workers correctly.
- Raised and exposed the AFD candidate limit and overflow controls; the reported 128-GPU dense and 96-GPU MoE cases now complete without default truncation.
- Propagated the AFD backend/version/WideEP settings to the combined static prefill path while preserving explicit candidate lists.
- Restored decode-system availability checks for HYBRID mode.
- Added an explicit pinned-topology state and now report an empty unpinned search as NoFeasibleConfigError.

src/aiconfigurator/cli/main.py
jieliu1997: Thanks for the suggestion. We discussed this trade-off earlier and, for this version, intentionally treat AFD as an additional, opt-in search path orthogonal to the existing agg/disagg comparison. This preserves the existing default behavior and avoids changing its runtime, result set, or recommendation for current users. Therefore, default will continue to compare agg and disagg unless AFD is explicitly requested.
You are right that the statement “CLI surface is unchanged” is inaccurate because --serving-mode is newly introduced. We will correct the PR description accordingly.

src/aiconfigurator/cli/main.py
jieliu1997: Yes, this is intentional for the current AFD topology model. The search allocates A and F pools at node granularity and requires at least one full node for each, so the minimum is two nodes. This is a limitation of the current deployment model rather than a fundamental AFD requirement; splitting A and F workers within a single node is not supported yet. We’ll clarify this in the CLI help text.

src/aiconfigurator/cli/main.py
```diff
@@ -1300,9 +1319,28 @@ def build_default_tasks(
                     database_mode,
                 )
 
-    # v2 Task uses a flat schema. Global (both-mode) fields stay top-level;
-    # worker-spec fields are top-level for agg but must be fanned out to
-    # prefill_* / decode_* for disagg (v2 forbids shared top-level worker fields).
+    def _disagg_backend_available(backend_name: str) -> bool:
```
tianhaox, line 1623: Per-mode backend skipping breaks `save_results` under `--backend auto`. Previously a backend missing decode-system data was dropped from **both** agg and disagg; now `_disagg_backend_available` skips only disagg, so `tasks` can contain `agg_<b>` without `disagg_<b>` (same asymmetry for `afd_<b>` via the `except ValueError` skip at L1458). `save_results` (report_and_save.py:878) then does `tasks[f"{exp_name}_{b}"]` for every backend present in `tasks` → `KeyError`, swallowed by the function-wide `except Exception` — after a full sweep, **no CSV/exp_config/deployment artifacts are written**, with only a one-line "Failed to save results".

Repro: `--backend auto --system h200_sxm --decode-system h100_sxm` with one backend lacking the decode-system DB.

Please make `save_results` tolerate per-mode absence (derive backends per mode bucket), or restore the whole-backend drop.

tests/unit/sdk/test_afd_pareto_review_actions.py
```diff
@@ -0,0 +1,652 @@
...
+            runtime_config=RuntimeConfig(isl=128, osl=32, tpot=25.0),
+            database=_FakeDatabase(),
+            backend_name="trtllm",
+            afd_parallel_config_list=[(1, 1, 2, 2, 3, "optimistic")],
+            gpus_per_node=8,
+            combined_with_pd=True,
+        )
+
+
+def test_static_prefill_options_inherit_base_model_config(monkeypatch):
```
tianhaox, line 598: This new test fails deterministically (locally: 1 failed / 72 passed across the AFD test files). The new `resolve_context_fmha_compat` call in `_enumerate_afd_prefill_options` (pareto_analysis.py:488) isn't mocked and performs a live HF fetch for the fictional repo `Qwen/Qwen3-MoE`; the resulting `HuggingFaceDownloadError` is swallowed by the blanket `except Exception: continue` at :526, every candidate is dropped, and `assert options` fails.

The same path is a production fragility: any transient/offline HF failure silently empties the static prefill pool (DEBUG-only logs) and surfaces as a misleading `NoFeasibleConfigError` about prefill options. Please mock the call in the test, and let download errors propagate (or at least WARN) rather than falling into the blanket except.

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -1700,5 +2096,90 @@ def run_single_disagg(
...
+            derived = self.afd_total_batch_size // n_a_workers
+            if self.afd_a_batch_size is not None and self.afd_a_batch_size != derived:
+                raise ValueError(
+                    f"afd_a_batch_size={self.afd_a_batch_size} conflicts with "
+                    f"afd_total_batch_size={self.afd_total_batch_size} / n_a_workers={n_a_workers} = {derived}."
+                )
+            a_batch_size = derived
+
+        # Determine f_moe_ep_size (default: tp_f for MoE, 1 for dense)
+        f_moe_ep_size = tp_f if self._is_moe else 1
```
tianhaox, line 2358: Pinned path defaults `f_moe_ep_size = tp_f` without the `num_experts % ep == 0` check that the swept path enforces (`build_afd_parallel_lists`, task_v2.py:375). E.g. 256 experts with `afd_n_f_nodes=3` on 8-GPU nodes → ep=24: `models/moe.py` only asserts magnitude and the expert math floor-divides, so we silently model 240/256 experts (wrong perf estimates); `tp_f > num_experts` instead dies in a deep `AssertionError`. Please validate divisibility in the pinned path too and raise a clear error.

tianhaox (commented):
Verified that all items from the previous review round landed (configurable `afd_max_a_batch_size` + 384 regression test, batch-search lower bound 8, advisory-only balance ratio, aggregate failure WARNING, prefill-worker enumeration, two-node help text, and the earlier CodeRabbit majors) — thanks, that all checks out.

This pass found 8 further issues, left inline. The first two (silent loss of all saved results under `--backend auto` with asymmetric task sets, and a deterministically failing new regression test due to an unmocked HF fetch) should be fixed before merge; the rest are ordered by severity, and the last one is a perf follow-up.

src/aiconfigurator/sdk/pareto_analysis.py
```diff
@@ -333,6 +374,1235 @@ def get_working_list(working_list, max_constraint):
...
+
+    ttft = best_option["ttft"] * ttft_correction_factor
+    # End-to-end throughput is capped by the slower stage.
+    effective_prefill_seq_s = best_workers * best_option["seq_s"] * prefill_degradation
+    seq_s = min(effective_decode_seq_s, effective_prefill_seq_s)
+    tokens_s = seq_s * osl
+    request_latency = ttft + decode_time
+    num_total_gpus = decode_gpus + best_workers * best_option["num_gpus"]
+    decode_power = float(row.get("power_w", 0.0) or 0.0)
+    power_w = (
```
tianhaox: `power_w` for AFD rows is misleading as-is: the decode-side power is hardcoded `0.0` (`AFDInferenceSession._build_summary`, inference_session.py:1922), so combined rows collapse to ~`prefill_power*ttft/request_latency` (a few % of the true value) and pure-AFD rows are exactly 0, yet they land in the same `show_power` table/CSV as disagg rows.

We're fine **not** supporting power for AFD in this PR — but then make it explicit: blank/NaN the `power_w` column for AFD rows and exclude them from `show_power`, instead of reporting near-zero numbers next to disagg.

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -1441,6 +1777,58 @@ def sweep_disagg_kwargs(self, *, prefill_database, decode_database) -> dict[str,
...
+
+        return {
+            "model_path": self.model_path,
+            "runtime_config": runtime_config,
+            "database": database,
+            "backend_name": self.backend_name,
+            "model_config": self.build_model_config(role="agg"),
+            "afd_parallel_config_list": [tuple(c) for c in self._afd_parallel_config_list],
+            "gpus_per_node": self._afd_gpus_per_node,
+            "total_gpus": self.total_gpus,
```
tianhaox: `afd_total_gpus` and `total_gpus` diverge: enumeration and pinned validation use `afd_total_gpus or total_gpus` (:1146), but `sweep_afd_kwargs` passes `total_gpus=self.total_gpus`, which `afd_pareto` uses for the `gpu_budget` rejection and prefill combining. With `afd_total_gpus=32, total_gpus=16`: the sweep rejects every candidate → `NoFeasibleConfigError` despite the explicit 32-GPU budget; a pinned 32-GPU topology passes validation but picking computes `replicas = 16 // 32 = 0`. Also `_validate_afd` (:1463) still requires `total_gpus` even when `afd_total_gpus` is set, contradicting the field comment "defaults to total_gpus". Suggest passing the effective budget into the sweep and aligning validation.

src/aiconfigurator/generator/facts/hardware.yaml
```diff
@@ -1,168 +1,90 @@
...
-#     and collector/sglang/dsv4_megamoe/README.md (verified GPU product label pattern).
-#   - nccl_env: derived from platform hardware configs (§9 hardware-level silicon facts).
-#   - shared_memory: standard values from k8s deployment baseline tests and guard rails.
-#   - moe_backend: per design doc §9.3 (CUTLASS on Hopper sm90; WIDEEP on Blackwell sm100/GB200;
-#     sglang deepep_moe is a Blackwell/wide-EP selection only — see webapp/events/event_fn.py
-#     moe_backend="deepep_moe" when enable_wideep+sglang, cli/exps/deepseek_wideep_trtllm.yaml).
-#
-# SCOPE: hardware-only facts. No NIC names, storage classes, or cluster secrets.
-
 profiles:
```
tianhaox, line 18: This file was re-serialized wholesale: the `SPDX-FileCopyrightText` / `Apache-2.0` header and all provenance comments (node_selector/toleration/moe_backend sources) were deleted, and quoting churned — while the actual values are unchanged except the new `h20` profile. Please restore the original file content and hand-append only the `h20` block, keeping the license header and provenance.

src/aiconfigurator/cli/report_and_save.py
```diff
@@ -138,14 +138,77 @@ def _plot_worker_setup_table(
...
+            field_names.append("power_w")
+        table.field_names = field_names
+        for i, row in enumerate(top_configs.to_dict("records")):
+            a_gpus = int(row["(a)workers"]) * int(row["(a)tp"])
+            f_gpus = int(row["(f)workers"])
+            prefill_workers = row.get("(p)workers")
+            has_prefill_pool = prefill_workers is not None and not pd.isna(prefill_workers)
+            if has_prefill_pool:
+                p_workers = int(prefill_workers)
+                p_gpus = p_workers * int(row.get("(p)tp", 1))
```
tianhaox: `p_gpus = p_workers * (p)tp` ignores pp/dp/cp. MoE prefill candidates include dp>1 (e.g. tp=1, dp=4 → 4 GPUs/worker), so the `(=A..+F..+P..)` breakdown no longer sums to `num_total_gpus` and understates the prefill pool — an operator deploying from the table would under-provision. The correct per-worker count is already on the row as `(p)num_gpus`; use that instead.

src/aiconfigurator/sdk/pareto_analysis.py
```diff
@@ -333,6 +374,1235 @@ def get_working_list(working_list, max_constraint):
...
+        "gpu_budget": 0,
+        "low_batch_oom": 0,
+    }
+    exceptions: list[Exception] = []
+    exception_counts: Counter[str] = Counter()
+    candidate_attempts = 0
+    # Track topologies that OOMed at a given microbatch count; higher counts
+    # use a larger KV-cache multiplier and can be pruned for the same topology.
+    _oom_at_mb: dict[tuple[int, int, int, int], int] = {}
+    for eval_runtime_config in runtime_configs_to_evaluate:
```
tianhaox, line 1114: Perf note (fine as a tracked follow-up): all per-candidate setup — 2× `deepcopy`, 2× `get_model`, partition builds, memory sampling via `_derive_a_batch_size` (whose signature takes no ttft/tpot) — sits inside this outer `for eval_runtime_config` loop, yet the runtime configs differ only in ttft/tpot. With `--request-latency` set, `enumerate_ttft_tpot_constraints` yields ~14 pairs → ~14 identical setups per candidate; on top of that, `_evaluate_auto_batch` rebuilds the session/models for every batch probe. Hoisting candidate setup out of the SLA loop (or caching model/partition by `topo_key`) should cut AFD sweep wall time by roughly an order of magnitude.

src/aiconfigurator/sdk/pareto_analysis.py
```diff
@@ -333,6 +374,1235 @@ def get_working_list(working_list, max_constraint):
...
+            candidate_attempts += 1
+            try:
+                # --- Topology-level checks (independent of a_batch_size) ---
+                tp_f = int(n_f_nodes) * int(gpus_per_node)
+                if int(f_moe_ep_size) <= 0 or tp_f % int(f_moe_ep_size) != 0:
+                    continue
+                f_moe_tp = tp_f // int(f_moe_ep_size)
+                n_a_workers = (int(n_a_nodes) * int(gpus_per_node)) // int(tp_a)
+
+                a_model_config = copy.deepcopy(base_model_config)
```
tianhaox, line 1133: BTW (pre-existing in `AFDInferenceSession`, not introduced by this PR — fine as a follow-up): **MTP modeling in the AFD decode path is incomplete**, and this PR makes it the default comparison path.

MTP in this codebase is modeled in two halves: (1) the model layer scales generation ops by `_mtp_scale_factor = 1/(1+E)·(L+nextn)/L` (the speedup), and (2) the session layer widens the op-query batch to `bs*(nextn+1)` for the verify step (`base_backend._run_generation_phase`, base_backend.py:359 — the main cost). The AFD session only gets half (1): `a_model_config`/`f_model_config` here are deepcopies that keep `nextn`, so the ops carry the ÷(1+E) speedup, but `_integrate_decode_phase` queries A/F ops at the unwidened batch and decode comm uses `tokens_per_req = 1` — `inference_session.py` has zero `nextn` references. AFD thus takes MTP's full benefit while skipping most of its cost.

Net effect at nextn=1 / E≈0.85: A-side attention per-token time understated ~2×, F-side GEMM 1.2–2×, comm ~(nextn+1)/(1+E). Since `task_v2` defaults `nextn=1` for the MTP families (DeepSeek/Kimi/Qwen3.5) — exactly AFD's target models — the default AFD-vs-disagg comparison is biased toward AFD, and the auto-batch TPOT search runs on the optimistic tpot.

Suggested follow-up: apply the `(nextn+1)` width in the AFD decode queries (A/F ops + comm tokens) to match base_backend.py:359, or in the meantime WARN / force `nextn=0` for AFD sweeps.

tianhaox (commented):
One more note (BTW-level), left inline: AFD's MTP modeling.

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -1139,6 +1133,30 @@ def _pipeline_tcycle(self, t_a: float, t_f: float, t_a2f: float, t_f2a: float) -
         # conservative K=2
         return max(t_a + t_a2f, t_f + t_f2a), False
```
yunyang1999: **`pipeline_model="serial"` is accepted by validation but silently computed as K=2.**

`AFDConfig.__post_init__` validates against `("optimistic", "conservative", "serial")` (config.py:265), but `_pipeline_tcycle` only branches on `"optimistic"` and otherwise falls through to this conservative return. There is no `serial` branch anywhere in `inference_session.py`. A `serial` entry in the search config, or `--pipeline-model serial`, therefore yields K=2 numbers with no error and no warning.

Either implement the tier —

```python
if cfg.pipeline_model == "serial":
    return t_a + t_a2f + t_f + t_f2a, False
```
— or drop "serial" from valid_pipeline_models until it is.

yunyang1999:
This is not high priority for this PR and can be addressed later, once our code deployment framework is more mature. Just leave a comment here as a reminder: please remember to update AIC exsiting docs with AFD implementation and guidance.

`--serving-mode`, roughly fifteen `--afd-*` flags and 21 new `Task` fields land here with nothing under `docs/`. Both `docs/cli_user_guide.md` and `docs/advanced_tuning.md` currently document `serving_mode` for agg/disagg only, so a user following the official guide has no path to discover this mode exists, let alone its YAML schema. 

At minimum:

- **`docs/cli_user_guide.md`** — the new mode, a table of the `--afd-*` flags, and two things that will otherwise turn straight into issues: the **two-node minimum**, and exactly what `tokens/s/gpu` counts for an AFD row (A + F pools, and whether the paired static prefill pool is in the denominator).
- **`docs/advanced_tuning.md`** — the YAML schema for the AFD fields, mirroring the existing agg/disagg examples.
- **The `exp` example YAML** — one `serving_mode: afd` block.

src/aiconfigurator/sdk/pareto_analysis.py
yunyang1999: Thanks for catching this — it’s a subtle one. One small thing I ran into while looking at the fix, in case it flips the result: `GenerationAttention.query` keys only on `batch_size` and `s` and ignores `x`, so querying at `bs*(nextn+1)` prices that many *independent* sequences each with a full KV, while MTP's k verify positions share one KV per sequence. Measured at s=32K / tp=1 the op scales essentially linearly with batch (1.00 / 1.99 / 2.98 / 3.97 at batch 32/64/96/128), so a uniform widening over-charges the attention core by ~k. It may hit AFD harder since the A pool is nearly all attention.

Probably still worth doing it your way for now — matching base_backend removes the asymmetry, which is the bigger problem, and a proper split would need `GenerationAttention` to take a query-multiplicity argument. Just thought it was worth recording so the residual bias is a known quantity rather than a surprise later.

+1 on the interim WARN. Might also be worth surfacing `nextn` in the AFD result row — with `_MTP_DEFAULT_FAMILIES` covering DeepSeek / Kimi-K2.5 / Qwen3.5 / MiniMax, a default-on `nextn=1` is invisible in the output today.

src/aiconfigurator/cli/main.py
jieliu1997: I’ll make `save_results` derive backends per mode and add a regression test for asymmetric tasks under `--backend auto`.

tests/unit/sdk/test_afd_pareto_review_actions.py
jieliu1997: Thanks. I’ll mock `resolve_context_fmha_compat` in this test and update the candidate error handling so Hugging Face download failures propagate instead of being silently treated as infeasible candidates.

src/aiconfigurator/sdk/task_v2.py
jieliu1997: Good catch, thanks. I’ll add the same expert-divisibility validation to the pinned path, raise a clear `ValueError` for invalid `f_moe_ep_size`, and add regression coverage for both non-divisible and oversized EP configurations.

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: AFD power is not modeled yet, so reporting `0.0` or a prefill-only weighted value is misleading. I’ll set `power_w` to `NaN` for both pure and combined AFD rows, exclude AFD rows from power-availability checks, and add regression coverage.

src/aiconfigurator/sdk/task_v2.py
jieliu1997: OK，I’ll consistently use the effective AFD budget (`afd_total_gpus` when set, otherwise `total_gpus`) across enumeration, pinned validation, sweep execution, picking, and reporting. I’ll also align validation so either field can provide the AFD budget and add regression coverage for the divergent-budget cases.

src/aiconfigurator/generator/facts/hardware.yaml
jieliu1997: Thanks, I’ll restore the original `hardware.yaml` content verbatim, including the SPDX header, provenance comments, quoting, and formatting, and then append only the new H20 profile to keep the diff minimal.

src/aiconfigurator/cli/report_and_save.py
jieliu1997: The breakdown should use the existing `(p)num_gpus` value rather than deriving the per-worker GPU count from `(p)tp`. A regression test with `dp > 1` will verify that the A/F/P components sum to `num_total_gpus`.

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: Agreed that the candidate setup is invariant across the SLA pairs and is unnecessarily repeated inside the outer loop. Hoisting or caching the model, partition, and memory setup by `topo_key` will be tracked as a performance follow-up.

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: Thanks both. The current AFD path is optimistic because it applies the MTP acceptance-rate benefit without the corresponding verify-width cost. Uniform batch widening would restore parity with `base_backend`, but it is only an approximation for attention because the verify positions share KV state. Proper query-multiplicity modeling will be tracked separately; in the meantime, the AFD path should warn when `nextn > 0` and surface `nextn` in its results.

src/aiconfigurator/sdk/inference_session.py
jieliu1997: ok, `serial` should not silently fall through to the conservative model. An explicit serial branch using `t_a + t_a2f + t_f + t_f2a` will be added, together with regression coverage for all three pipeline models. Unexpected values should also fail explicitly instead of inheriting conservative semantics.

src/aiconfigurator/cli/main.py
jieliu1997: Fixed in `a0cac510`. Auto-backend result rows now retain their originating `_task_key`, and `save_results` derives the applicable tasks from each serving-mode bucket instead of assuming every backend exists for every mode. Row-to-task resolution prefers the exact task key and fails explicitly on ambiguous/unknown mappings. Regression coverage: `test_auto_result_tasks_are_scoped_to_mode_and_preserve_variant_identity` and `test_save_results_auto_handles_asymmetric_backends_by_mode`. The affected CLI/AFD/SDK/generator suite passes (`471 passed, 4 skipped`).

tests/unit/sdk/test_afd_pareto_review_actions.py
jieliu1997: Fixed in `a0cac510`. Context-FMHA compatibility is now resolved once, before entering the per-candidate exception boundary. Model metadata and Hugging Face download failures therefore propagate to the caller instead of silently dropping every candidate; failures that are genuinely candidate-specific remain isolated. The fixture mocks the current `resolve_context_fmha_by_data` call. Regression coverage: `test_static_prefill_model_compatibility_failure_is_not_swallowed` and `test_static_prefill_candidate_failure_does_not_drop_other_candidates`. The affected suite passes (`471 passed, 4 skipped`).

src/aiconfigurator/sdk/task_v2.py
jieliu1997: Fixed in `a0cac510`. Pinned and swept AFD topologies now share `_is_valid_afd_moe_ep_size`, which requires a positive EP size, EP within available F ranks and expert count, and divisibility of both F ranks and `num_experts`. Invalid pinned configurations now fail with a clear `ValueError` before model construction. Tests cover EP=24 with 256 experts, EP larger than the expert count, and a valid divisor. The affected suite passes (`471 passed, 4 skipped`).

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: Fixed in `a0cac510`. Pure AFD summaries now emit `power_w=NaN`; combined AFD+prefill rows preserve unknown power whenever either component is unknown. Power availability ignores NaN rows, and table rendering leaves unknown power blank rather than formatting it as zero. Regression coverage includes pure AFD summaries, combined rows, availability checks, CSV persistence, and table output. The affected suite passes (`471 passed, 4 skipped`).

src/aiconfigurator/sdk/task_v2.py
jieliu1997: Fixed in `a0cac510`. `Task.effective_total_gpus` is now the single AFD budget source (`afd_total_gpus` when set, otherwise `total_gpus`) across validation, topology enumeration, pinned checks, `sweep_afd_kwargs`, picking, summary throughput/replica calculations, and reporting. AFD validation now accepts either field as the budget. Regression tests cover the 32-vs-16 divergent-budget case, a pinned 32-GPU topology, `afd_total_gpus` without `total_gpus`, fallback behavior, and over-budget rejection. The affected suite passes (`471 passed, 4 skipped`).

src/aiconfigurator/generator/facts/hardware.yaml
jieliu1997: Fixed in `a0cac510`. The file now comes from current `main` with the SPDX header, provenance/guard comments, quoting, and existing profile formatting intact. Relative to `main`, the entire diff is the hand-appended H20 profile only: 19 inserted lines and no unrelated rewrites. Generator-related tests are included in the affected suite (`471 passed, 4 skipped`), and Ruff checks also pass.

src/aiconfigurator/cli/report_and_save.py
jieliu1997: Fixed in `a0cac510`. The combined-prefill table now computes P-pool GPUs as `p_workers * row["(p)num_gpus"]`, so TP/PP/DP/CP are all represented by the already-resolved per-worker count. `test_afd_worker_table_uses_prefill_num_gpus_per_worker` covers TP=1, DP=4 and verifies the rendered `A8+F8+P8` breakdown sums to 24 GPUs. The affected suite passes (`471 passed, 4 skipped`).

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: Agreed and now tracked as [#1426](https://github.com/ai-dynamo/aiconfigurator/issues/1426). The issue records the repeated model/partition/memory setup, requires an explicit stable topology cache key, isolates mutable SLA-specific state, and includes Pareto parity tests plus before/after setup-count and wall-time benchmarks. This remains a performance follow-up rather than a correctness blocker for this PR.

src/aiconfigurator/sdk/pareto_analysis.py
jieliu1997: Fixed in `a0cac510`, with the remaining modeling limitation tracked in [#1425](https://github.com/ai-dynamo/aiconfigurator/issues/1425). AFD now applies `nextn + 1` verify width to A-side/F-side decode queries and decode communication, matching the existing base-backend convention; `nextn=0` remains a no-op. The effective `nextn` is included in AFD result rows, and `nextn>0` emits an explicit warning that batch widening is only a shared-KV approximation. Regression tests cover widths 1 and 2, A/F compute, communication, result output, and the warning. The follow-up issue separates independent sequence count from query multiplicity for a proper shared-KV model. The affected suite passes (`471 passed, 4 skipped`).

src/aiconfigurator/sdk/inference_session.py
jieliu1997: Fixed in `a0cac510`. `serial` now computes the strict stage sum `t_a + t_a2f + t_f + t_f2a` with no hidden communication, `conservative` retains K=2 behavior, and unexpected values raise `ValueError` instead of falling through. Regression coverage includes per-layer serial timing, global-step serial timing, and invalid model rejection. The affected suite passes (`471 passed, 4 skipped`).

jieliu1997:
@yunyang1999 Thanks for marking the documentation work as non-blocking. It is now tracked in [#1427](https://github.com/ai-dynamo/aiconfigurator/issues/1427), with explicit scope for `docs/cli_user_guide.md`, `docs/advanced_tuning.md`, and `src/aiconfigurator/cli/example.yaml`. The acceptance criteria require every command/YAML example to parse successfully and call out the two-node topology minimum, effective GPU-budget semantics, the A+F(+P) throughput denominator, pipeline/MTP guidance, and current artifact/power limitations. The correctness fixes for this PR are in `a0cac510`; the docs will remain a separately reviewable follow-up.

tianhaox (approved):

tianhaox:
approved, please resolve the merge conflict. Thanks!

tianhaox:
> approved, please resolve the merge conflict. Thanks!

Seems like recent merges introduced additional conflicts. please help to resolve. Since we have mulitple PRs ongoing, auto-merge is enabled for this PR.

jieliu1997:
> > approved, please resolve the merge conflict. Thanks!
> 
> Seems like recent merges introduced additional conflicts. please help to resolve. Since we have mulitple PRs ongoing, auto-merge is enabled for this PR.

@tianhaox The conflicts have been resolved and all checks are passing. However, auto-merge did not take effect. Could you please merge the PR manually? Thanks!

PR #1330, opened 2026-07-08
fix: build the aiconfigurator wheel as manylinux_2_28

nv-anants:
The wheel carries a native Rust/PyO3 (abi3) extension but was built on ubuntu:jammy and tagged linux_x86_64 -- a bare platform tag that PyPI rejects and that effectively pins the wheel to the build host's glibc. Build it as a portable manylinux_2_28_x86_64 wheel instead (glibc >= 2.28; RHEL 8+ / Ubuntu 20.04+).

Move the build stage onto quay.io/pypa/manylinux_2_28_x86_64 (glibc 2.28) with an explicit rustup toolchain and uv, pinned to the image's CPython 3.12. maturin's built-in auditwheel checks the manylinux_2_28 policy and stamps the tag -- no separate auditwheel step. uv build makes maturin's PEP517 backend force --compatibility off, so the target is re-asserted via MATURIN_PEP517_ARGS on the build command. README notes the new tag.

## Summary by CodeRabbit

* **Documentation**
  * Updated the “Install from PyPI” guidance with the wheel’s `manylinux_2_28_x86_64` compatibility and `glibc >= 2.28` requirement.
  * Clarified that wheels are not published for macOS, Windows, or other architectures (including Linux aarch64); those platforms should install from source.

* **Chores**
  * Improved the Linux release-wheel build to use a `manylinux_2_28` multi-arch setup and deterministic compatibility tagging.
  * Updated the wheel verification step to use the pinned bundled CPython, and adjusted the `maturin` build constraint for consistency.


docker/Dockerfile
```diff
@@ -15,15 +15,34 @@ RUN mkdir /opt/aiconfigurator && \
...
-# crate sources, a C linker (gcc), and ca-certificates so cargo can fetch crates
-# from crates.io over TLS. Rust itself is bootstrapped by maturin's build backend.
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends ca-certificates gcc libc6-dev && \
-    rm -rf /var/lib/apt/lists/*
+# The `aiconfigurator` wheel (Rust extension + SDK/data payload + CLI, generator,
+# webapp, Spica) is built in the PyPA manylinux_2_28 image so the native PyO3
+# `.so` links glibc 2.28 and maturin's auditwheel stamps the portable
+# `manylinux_2_28_<arch>` tag; `aiconfigurator-core` is a pure-Python metapackage.
+# We add `uv` and a rustup toolchain so cargo can fetch crates over TLS.
```
simone-chen, line 22: Do we still define `aiconfigurator-core` as a pure-Python metapackage after https://github.com/ai-dynamo/aiconfigurator/pull/1329?
It might be more accurate to describe it as a Python shim over the `aiconfigurator-core` Rust crate? 🙏

simone-chen (approved):
Thanks for the PR!

docker/Dockerfile
simone-chen: Saw it fixed in a later commit.

PR #1336, opened 2026-07-09
ci: enable prediction regression gate as PR check, demote accuracy job to nightly

tianhaox:
## What

PR 2 of 2 for the prediction-regression-gate rollout (#1289, design doc §5). Trigger-only — no harness changes:

- **`prediction-regression-gate.yml`**: add the `pull_request` trigger. The `AIC Prediction Regression Gate` report job is the check to mark **required** in branch protection after merge (fail-closed logic already shipped in #1289).
- **`accuracy-regression-testing.yml`**: demote to nightly `schedule:` (2 PM UTC, same slot as the support matrix daily run). The scheduled run's value is the new-side prediction-vs-silicon MAPE in the uploaded artifacts; old-vs-new comparison stays meaningful on manual dispatches with an explicit `old-ref`. Also: resolve `old-ref` with a shell fallback (schedule events carry no inputs — the previous expression would resolve to an empty string and fail the worktree add), and drop the two PR-comment feeder steps that only ran on `pull_request`.
- **Remove `accuracy-regression-comment.yml`**: its only trigger path was the feeder artifact from the PR-time accuracy job; dead once that trigger is gone.

## Validation

- **Post-merge dispatch validation of #1289 (rollout step 1)**: [run 29018012385](https://github.com/ai-dynamo/aiconfigurator/actions/runs/29018012385) — main vs main, both collect sides + report green.
- **This PR gates itself end-to-end**: the `pull_request` trigger runs the gate on this PR's own merge ref (old = base `68fc4519`, which has the harness).

## Revert story

Reverting this PR restores the dispatch-only status quo without touching the harness.

## Post-merge admin step

Mark **`AIC Prediction Regression Gate`** as a required status check on `main` (and un-require `AIC Accuracy Regression Testing` if it is currently required).

## Summary by CodeRabbit

* **New Features**
  * Added a required pull request check for prediction regression gating.
  * Shifted accuracy regression testing to run nightly via scheduled automation.

* **Bug Fixes**
  * Updated regression runs to correctly determine the “old revision” for scheduled executions and manual comparisons.

* **Automation Changes**
  * Removed the automated posting/updating of PR comparison comments for completed accuracy regression runs.


Arsene12358 (approved):
## Review verdict: approve — no merge-blocking issues found

This PR does exactly what the design doc §5 prescribes for PR 2 (trigger-only: enable the gate's `pull_request` trigger, demote the accuracy job to nightly, delete the dead commenter), and it validated itself end-to-end: the newly enabled trigger ran the full gate on this PR's own merge ref and passed in ~15 min ([run 29020719871](https://github.com/ai-dynamo/aiconfigurator/actions/runs/29020719871)), with the two earlier pushes showing `cancelled` runs that confirm the stale-run concurrency cancellation works as intended.

### Blocker hunt — candidates examined, none survived

- **Fail-closed invariant (design §7's explicit PR-2 reviewer mandate): survived.** The `report` job runs under `if: always()` with an explicit `exit 1` when `needs.collect.result != 'success'`, which also covers the refs-fails → collect-skipped path (`result == 'skipped'`), so the required check can never silently skip-pass.
- **No stranded PRs from removing the accuracy job's PR trigger.** The active branch rules on `main` (rulesets API) contain no required-status-check rule today, so nothing currently requires `AIC Accuracy Regression Testing`.
- **The deleted commenter workflow is genuinely dead.** Its only job is gated on `workflow_run.event == 'pull_request'`, which can no longer occur, and code search finds no other producer/consumer of the feeder artifact. Removing it is also a net security win — it was the repo's privileged `workflow_run` workflow (contents/issues/PR write) parsing artifacts produced by untrusted PR code.
- **The nightly `old-ref` fallback is a real fix, not cosmetics.** On `schedule` events the `inputs` context is empty, so the previous expression would have fed `""` into `git worktree add` and failed every nightly run; `${OLD_REF_INPUT:-main}` resolves correctly (the scheduled checkout lands on local `main`, and `--detach` avoids the already-checked-out error).
- **No path filter on the gate trigger is correct** for a future required check (a path-filtered required check strands PRs that don't touch the paths); the 2× compute cost is explicitly accepted in design §7.
- **Untrusted fork code runs safely**: all gate jobs pin `permissions: contents: read`, checkouts use `persist-credentials: false`, and no step references a secret; new interpolations are routed through `env:` rather than inlined into `run:` bodies.

### Non-blocking observations

1. **Nightly is a main-vs-main self-comparison** that still builds the full old-side venv (improvement metrics ≡ 0). The file-header comment already documents this as interim until the `run_silicon.py` convergence, so no change requested; if the convergence stalls, guarding the four old-side steps with `if: github.event_name != 'schedule'` is a cheap intermediate.
2. **`pull_request.base.sha` can lag the merge ref's actual base** (harness logic from #1289, merely activated here): the old side is the base tip at event time while `refs/pull/N/merge` is recomputed against the current base tip, so a fast-moving `main` can attribute unrelated transitions to a PR. Transient noise only (a push re-fires `synchronize` with a fresh `base.sha`); an exact follow-up fix is resolving the old ref as `git rev-parse HEAD^1` on the merge-ref checkout.
3. Cosmetic: the two workflows' `old-ref` dispatch inputs default to `""` vs `"main"` — behavior is identical through the shared `:-main` fallback; only worth aligning if either file is touched again.

### Post-merge admin steps (echoing the PR body, both check out)

1. Mark **`AIC Prediction Regression Gate`** (the report job's name) as a required status check on `main`; until then the gate runs but is advisory — a documented rollout gap, not a defect.
2. Rulesets show no required-check rule, but classic branch protection isn't readable with a non-admin token — an admin should confirm `AIC Accuracy Regression Testing` isn't required there before merging.

PR #1338, opened 2026-07-10
fix(sdk): skip MoE workspace for dense Gemma 4

zhuangqh:
#### Details:

Reproduce with:

  aiconfigurator cli default --model-path google/gemma-4-31B-it --total-gpus 2 --system h100_sxm --database-mode SOL --top-n 1

Before this fix, both agg and disagg reject every parallel configuration after memory accounting raises:

  TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'

Dense Gemma 4 variants have topk=0 and num_experts=None, but shared GEMMA4MIX memory accounting multiplied by num_experts. Only charge MoE dispatch workspace when routed experts are present.

- closes GitHub issue: #xxx

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory usage estimates for mixture-of-experts models by excluding unnecessary workspace overhead when no experts are configured.
  * Corrected activation and total memory calculations for dense model variants, resulting in more accurate backend resource estimates.

### DIFF
--- a/tests/unit/sdk/test_utils.py
+++ b/tests/unit/sdk/test_utils.py
@@ -7,11 +7,13 @@ Unit tests for SDK utility functions.
 Tests HuggingFace config parsing and model config retrieval.
 """
 
+from types import SimpleNamespace
 from unittest.mock import patch
 
 import pytest
 
 from aiconfigurator.sdk import common, config
+from aiconfigurator.sdk.backends.base_backend import BaseBackend
 from aiconfigurator.sdk.models import Gemma4MixModel, HybridMoEModel
 from aiconfigurator.sdk.utils import (
     _parse_hf_config_json,
@@ -829,6 +831,26 @@ class TestGemma4MixModelBuilder:
         )
         gen_names = {op._name for op in model.generation_ops}
         assert not any("moe" in n.lower() or "router" in n.lower() for n in gen_names)
+
+        database = SimpleNamespace(
+            system_spec={
+                "misc": {
+                    "nccl_mem": {1: 0},
+                    "other_mem": 0,
+                }
+            }
+        )
+        memory = BaseBackend()._get_memory_usage(
+            model,
+            database,
+            batch_size=1,
+            beam_width=1,
+            isl=1,
+            osl=1,
+        )
+
+        assert memory["activations"] == pytest.approx(BaseBackend.MIN_ACTIVATION_BYTES / (1 << 30))
+        assert memory["total"] > 0
 
     def test_dense_variant_rejects_moe_ep_gt_1(self):
         """Dense Gemma 4 has no experts, so any moe_ep_size > 1 must be rejected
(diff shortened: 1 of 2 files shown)


src/aiconfigurator/sdk/backends/base_backend.py
```diff
@@ -1572,7 +1572,7 @@ def _get_memory_usage(
 
         # MoE block-scale dispatch workspace (only for families that pay this cost).
         # 128 = block scale; 4 = float bytes.
-        if family in self.MOE_WORKSPACE_FAMILIES:
+        if family in self.MOE_WORKSPACE_FAMILIES and model._topk and model._num_experts:
```
tianhaox: Non-blocking: the `model._topk` conjunct is redundant for correctness — `topk` is normalized to an int upstream (`utils.py` maps `None` → 0) and the workspace term multiplies by `_topk` anyway, so `model._num_experts` alone is what prevents the `float * None` crash:

```python
if family in self.MOE_WORKSPACE_FAMILIES and model._num_experts:
```

Fine to keep the two-operand form if you prefer it as a readable "has routed experts" predicate. Longer term this duplicates `Gemma4MixModel._is_dense` (gemma4.py) — a shared `has_routed_experts` property on the model would keep the op-construction gate and the memory gate in sync, but that's out of scope for this fix.

tests/unit/sdk/test_utils.py
```diff
@@ -830,6 +832,26 @@ def test_dense_variant_builds_without_moe_ops(self):
...
 
+        database = SimpleNamespace(
+            system_spec={
+                "misc": {
+                    "nccl_mem": {1: 0},
+                    "other_mem": 0,
+                }
+            }
+        )
+        memory = BaseBackend()._get_memory_usage(
```
tianhaox, line 855: Nit (non-blocking): this makes `test_dense_variant_builds_without_moe_ops` cover two unrelated behaviors (op construction + memory accounting), so a memory-path regression would fail under a misleading test name. Consider extracting these assertions into a dedicated `test_dense_variant_memory_usage_no_crash` (a small helper for the dense-model construction would avoid duplicating the setup).

tianhaox (commented):
Thanks for the fix — I traced it end-to-end and it looks correct and complete:

- The workspace term is proportional to `_num_experts * _topk`, so whenever either is 0 the old code already added exactly 0 bytes; the new guard only changes behavior for the `num_experts=None` crash case. None of the real MoE models in `MOE_WORKSPACE_FAMILIES` can reach this line with falsy `_topk`/`_num_experts`, so there's no under-counting risk.
- Checked the remaining `_num_experts`/`_topk` consumers on the dense-Gemma4 path (`inference_session.py`, engine MoE-op paths, pareto enumeration) — all already guarded, so the CLI repro is fully fixed by this one change.
- The positive side of the guard (real MoE still pays the dispatch workspace) is already pinned by `test_sglang_deepseek_v4_pro_moe_workspace_uses_residual_hidden_size`, so no coverage gap there.

Two non-blocking nits inline.

Arsene12358:
@zhuangqh Could you help address the existing comments and fix the conflict. After that I think we are ready to merge. Thanks

zhuangqh:
@tianhaox  @Arsene12358 rebased and comments addressed. PTAL

tianhaox (approved):

PR #1339, opened 2026-07-10
chore: replace mislabeled vLLM v0.19 FP16 AR data with real bf16

wenpengw-nv:
#### Overview:
The existing vLLM v0.19 AllReduce tables were labeled as bfloat16, but the original measurements were collected with FP16 inputs and later relabeled without recollection.
#### Details:

vLLM v0.19 AllReduce dispatch is dtype-sensitive: Torch Symmetric Memory accepts BF16 but not FP16. For the same GPU, TP size, and payload, the old FP16 run may therefore use CustomAR or PyNCCL while a real BF16 run uses SymmMem two-shot or multimem.

Recollect `h100_sxm/h200_sxm/b200_sxm/b300_sxm/gb200/gb300` on real bf16 dtype.

Comparison with the old data
TP=2
image
TP=4
image
TP=8
image

- closes GitHub issue: #xxx


tianhaox:
Thanks for the recollection — verified the new parquets against the old ones: coverage grids are identical across all 6 systems and the deltas are consistent with the SymmMem dispatch story.

One measurement glitch though, in `h100_sxm/vllm/0.19.0/custom_allreduce_perf.parquet`:

**TP=4, graph mode, message_size=128 → 29.8µs**, while all neighboring sizes are ~4.5µs, and it's even slower than the eager-mode value at the same point (25.4µs):

```
tp4 vLLM_custom_graph: (128, 0.0298), (256, 0.0046), (512, 0.0045), (1024, 0.0045), (2048, 0.0047), ...
```

This looks like measurement jitter rather than a real dispatch effect, and it would skew interpolation for small payloads at TP=4 on h100. I scanned all 6 systems for >3x spikes vs neighbors — this is the only one. Could you re-measure just this point (h100_sxm, TP=4, graph, 128B) and update the parquet? Everything else LGTM.

wenpengw-nv:
@tianhaox Thanks for catching this. I re-collected the H100 data and updated the parquet. The measurement glitch is now fixed.

tianhaox (approved):

PR #1340, opened 2026-07-10
feat(sdk): model EPD (encoder disaggregation) for VL disagg sweep

### LINKED ISSUE #1546: [Design] EPD (encoder disaggregation) modeling for vision-language models
### Motivation

  VL serving stacks increasingly run the vision encoder on dedicated instances (SGLang `--encoder-only`, Dynamo EPD recipes, vLLM EC disaggregation), because encode compute
  contends with LM prefill/decode and scales on a different axis (images/s vs tokens/s). aiconfigurator can only model colocated encoders, so it cannot answer the sizing questions
  EPD deployments pose: should the encoder be split out, on how many workers, with which TP/batch, and at what TTFT/throughput cost or gain.

  ### Design (agreed shape)

  - One switch, `enable_epd` (CLI `--enable-epd`): the disagg experiment becomes E+P+D, the agg experiment becomes E+agg. No behavior change when off.
  - Encode workers are encoder-only instances (`EncoderOnlyModel`, no LLM weights; satisfies the compiled-engine spec contract with empty LM op lists). ViT weight-sharded TP with encoder DP pinned off — the encoder-instance default of both engines; DP's benefit is expressed by sweeping multiple small-TP encode workers.
  - P/agg workers become language-only (`ModelConfig.language_only=True`): vision tokens still extend context; no ViT compute/memory on LM workers.
  - Candidates: encoder TP × batch (batch capped at 8 = SGLang's `SGLANG_ENCODER_MAX_BATCH_SIZE` default, fail-loud above), VRAM-gated with the same framework overhead as the P/D checks.
  - 3-pool rate matching: cell throughput = min of degraded E/P/D pool capacities; encode GPUs count toward the per-GPU objective and the replica budget. E+agg keeps, per encoder choice, the best rate per admissible cell GPU count as a nondominated (cell gpus, cell rate) frontier — the cluster-aware floor-packing picker makes the final call.
  - TTFT composition: encode is sequential per request; the queueing correction mirrors each baseline (disagg corrects the encode stage like its prefill TTFT; agg adds raw encode latency), keeping EPD-vs-baseline apples-to-apples.
  - Fail-closed generator: EPD rows skip deployment artifacts with a warning until the bridge maps the encode pool.
  - Single-point APIs: `run_single_agg`/`run_single_disagg` accept `encoder_tp`/`encoder_batch_size`/`encoder_num_workers`.

  ### Cross-module contract changes (requesting explicit approval per repo-guide)

  - aic-core: `EncoderOnlyModel` + `build_encoder_ops` (models/blocks/vit), `run_encoder_static` (base_backend).
  - SDK: `ColumnsAggEpd` row schema; `(e)bs` column on disagg rows.
  - CLI: `--enable-epd`, `--encoder-tp`, `--encoder-system`, `--encoder-latency-correction`; estimate-mode EPD args.
  - Generator bridge: fail-closed guard for EPD rows.

  ### Out of scope / follow-ups

  - E→P embedding transfer time; encode-side queueing / host CPU overhead.
  - Generator mapping of the encode pool into deployment topologies.

  Implementation: PR #1340

wenpengw-nv:
#### Overview:

Adds EPD (encoder/prefill/decode disaggregation) modeling for vision-language models, mirroring how real EPD deployments work (SGLang `--encoder-only` / language-only instances). One switch, both serving modes: `--enable-epd` turns the disagg experiment into **E+P+D** and the agg experiment into **E+agg**.

#### Details:

**Topology**

* `--enable-epd` (Task field `enable_epd`): the vision encoder runs on dedicated encode workers instead of colocated with prefill/agg.
* **Encode (E) workers** are encoder-only instances built directly from the encoder config (`EncoderOnlyModel`, no LLM weights; it satisfies the compiled-engine spec contract with empty LM op lists, so the default Rust engine path works). ViT weight-sharded over the worker's TP with encoder DP off — the encoder-instance default of both engines (vLLM  `mm_encoder_tp_mode="weights"`, SGLang `mm_enable_dp_encoder=False`). `enable_epd` pins the Task-level encoder-DP knob off accordingly; the DP layout's benefit is expressed by sweeping multiple small-TP encode workers instead.
* **P/agg workers become language-only** (`ModelConfig.language_only=True`): vision tokens still extend the context, but no ViT compute/memory lands on the LM worker.

**Search space & rate matching**

* Encode-worker candidates: encoder TP (default `1/2/4/8`) × batch (default `1/2/4/8`); per-worker throughput = batch / encode latency. Batch is capped at 8 (SGLang's `SGLANG_ENCODER_MAX_BATCH_SIZE` default) and enforced fail-loud at Task validation, the candidate helper, and the single-point args. Candidates are OOM-gated against the encoder system's GPU memory with the same framework overhead the P/D checks charge.
* **3-pool rate matching**: cell throughput = min of the degraded E/P/D pool capacities — the encode pool can itself be the bottleneck. Pool size is swept up to `max_encoder_workers` (default 32, mirroring the P/D worker lists); over-provisioned pools are enumerated too, since they can be the only counts that land on an exact per-replica GPU grid. Encode GPUs count toward the per-GPU objective and the replica GPU budget.
* **E+agg**: each row is a rate-matched cell of `(a)workers` language-only agg workers plus `(e)workers` encode workers (`common.ColumnsAggEpd` schema). Per encoder choice, the best rate for every admissible cell GPU count is kept as a nondominated (cell gpus, cell rate) frontier, so the cluster-aware floor-packing picker — not a per-cell efficiency score — makes the final call. Disagg rows fill the existing `(e)*` columns (+ new `(e)bs`).

**TTFT composition**

* Encode → prefill is sequential per request: each encode choice spends its batch latency out of the TTFT budget before the LM-side SLA filter.
* The queueing correction mirrors each baseline: disagg corrects the encode stage like its prefill TTFT (×1.8, since the inline baseline corrects its combined E+P TTFT); agg adds the raw encode latency (run_agg adds the inline encoder outside its queueing factor) — so EPD-vs-baseline comparisons stay apples-to-apples in both modes.

**Knobs & APIs**

* Task: `encoder_tp_candidates`, `encoder_batch_candidates`, `max_encoder_workers`, `encoder_latency_correction`, `encoder_system_name` (hetero encoder: encode pool on its own GPU type; missing perf DB fails loud instead of silently falling back). All encoder knobs require `enable_epd`.
* CLI: `--enable-epd`, `--encoder-tp`, `--encoder-system`, `--encoder-latency-correction`.
* **Single-point evaluation**: `run_single_agg` / `run_single_disagg` accept `encoder_tp` / `encoder_batch_size` / `encoder_num_workers` to spot-check one fixed EPD cell (raw-TTFT convention, args validated fail-loud).
* Docs: `docs/cli_user_guide.md` (CLI controls, default + estimate modes), `docs/advanced_tuning.md` (Task tuning fields), `docs/generator_overview.md` (EPD rows are fail-closed in the generator).

**Behavior with EPD off**

* Text-only paths are byte-identical to main (locked by the rate-match parity test).
* Two **in-scope VL fidelity fixes** apply regardless of EPD (before/after on packaged h200_sxm/sglang, Qwen3-VL-8B):

| Fix | Case | main | this PR | Why |
| -- | -- | -- | -- | -- |
| Smart-resize rounding (floor → round) | 1010×1010 estimate | 961 tok/img; TTFT 431.58 ms | 1024 tok/img; TTFT 441.14 ms | HF `smart_resize` rounds each side to the *nearest* multiple of 32; flooring undercounts the tokens the deployed preprocessor produces. 32-aligned resolutions and text-only: unchanged. |
| Plain-agg ctx grid on effective ISL | default agg sweep, 768²×2 imgs | ctx grid 2048; top1 2177.01 tok/s/gpu | ctx grid 3200; top1 2101.66 tok/s/gpu | Vision embeddings occupy LLM context — the engine prefills 3200 tokens/request; the text-only grid understated per-iteration prefill work (legacy `find_best_agg` behavior restored). |

Disagg pareto rows are unchanged (only the new empty `(e)bs` schema column differs).

**Not modeled (out of scope for this PR)**

* E→P embedding transfer time and encode-side queueing/host CPU overhead.

## Usage

```bash
aiconfigurator cli default --model-path Qwen/Qwen3-VL-8B-Instruct \
  --total-gpus 32 --system h200_sxm --backend sglang \
  --isl 2000 --osl 200 --ttft 250 --tpot 20 \
  --image-height 1344 --image-width 1344 --num-images 2 \
  --enable-epd
```

* `src/aiconfigurator/sdk/sweep.py` — the EPD section: `_get_encoder_worker_candidates` (candidate enumeration + VRAM gate), `_overlay_encoder_stage` (single composition point for both topologies), `_rate_match_agg_epd` (E+agg frontier), and the 3-pool `_match_workers` closure in `sweep_disagg`.
* `src/aiconfigurator/sdk/task_v2.py` — `enable_epd` + encoder knobs, `_normalize_epd_encoder_dp`, and the `run_single_*` EPD path.
* `aic-core`: `models/blocks/vit.py` (`EncoderOnlyModel`, `build_encoder_ops`; `vit_ops.py` is now a compatibility shim) and `backends/base_backend.py` (`run_encoder_static`);`common.py` (`ColumnsAggEpd`, `(e)bs`).
* Real-database regressions: `tests/unit/sdk/task_v2/test_epd_real_database.py` (default-Rust path, agg + disagg).

#### Related Issues:

* Closes ai-dynamo/aiconfigurator#1546

## Summary by CodeRabbit

* **New Features**
  * Added optional dedicated vision encoder workers for vision-language inference.
  * Added controls for encoder tensor parallelism, batch sizing, latency adjustment, and GPU system selection.
  * Added encoder-aware aggregated and disaggregated performance sweeps, including worker matching and resource reporting.
  * Extended CLI and Python APIs with encoder configuration options.
* **Bug Fixes**
  * Improved visual token estimation by rounding image dimensions correctly.
  * Workers without vision encoders now handle vision context consistently.
  * Reports now include encoder worker details and batch sizes when applicable.

### DIFF
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -439,6 +439,12 @@ def _add_default_mode_arguments(parser):
         "Defaults to the prefill/agg side's system.",
     )
     parser.add_argument(
+        "--encoder-latency-correction",
+        type=float,
+        default=1.0,
+        help="Latency correction scale for EPD encode workers (requires --enable-epd). Default: 1.0.",
+    )
+    parser.add_argument(
         "--ttft",
         type=float,
         default=2000.0,
@@ -1469,6 +1475,7 @@ def build_default_tasks(
     enable_epd: bool = False,
     encoder_tp: list[int] | None = None,
     encoder_system: str | None = None,
+    encoder_latency_correction: float = 1.0,
     ttft: float = 2000.0,
     tpot: float = 30.0,
     request_latency: float | None = None,
@@ -1696,6 +1703,7 @@ def build_default_tasks(
             enable_epd=enable_epd,
             encoder_tp_candidates=encoder_tp,
             encoder_system_name=encoder_system,
+            encoder_latency_correction=encoder_latency_correction,
             **global_kwargs,
         )
 
@@ -1715,6 +1723,7 @@ def build_default_tasks(
             enable_epd=enable_epd,
             encoder_tp_candidates=encoder_tp,
             encoder_system_name=encoder_system,
+            encoder_latency_correction=encoder_latency_correction,
             **global_kwargs,
         )
 
@@ -2894,6 +2903,7 @@ def main(args):
             enable_epd=args.enable_epd,
             encoder_tp=args.encoder_tp,
             encoder_system=args.encoder_system,
+            encoder_latency_correction=args.encoder_latency_correction,
             ttft=args.ttft,
             tpot=args.tpot,
             request_latency=args.request_latency,
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -849,6 +849,30 @@ def _add_estimate_mode_arguments(parser):
         "(vLLM mm_encoder_tp_mode='data' / SGLang --mm-enable-dp-encoder semantics).",
     )
     parser.add_argument(
+        "--enable-epd",
+        action="store_true",
+        help="EPD single point: overlay a fixed encode-worker pool on the agg/disagg point; "
+        "the LM side becomes language-only. Supports --estimate-mode agg/disagg.",
+    )
+    parser.add_argument(
+        "--encoder-tp",
+        type=int,
+        default=None,
+        help="EPD encode-worker TP (required with --enable-epd).",
+    )
+    parser.add_argument(
+        "--encoder-batch-size",
+        type=int,
+        default=1,
+        help="EPD encode-worker batch size. Default: 1.",
+    )
+    parser.add_argument(
+        "--encoder-num-workers",
+        type=int,
+        default=1,
+        help="EPD encode workers in the pool. Default: 1.",
+    )
+    parser.add_argument(
         "--batch-size",
         "--bs",
         dest="batch_size",
@@ -2403,6 +2427,104 @@ def _print_per_ops_latency(per_ops_data: dict) -> None:
             _print_per_ops_section("AFD Transfer (per layer, a2f + f2a)", directional)
 
 
+def _run_estimate_epd(args, estimate_mode: str) -> None:
+    """EPD single-point estimate via Task.run_single_* (dedicated encode pool)."""
+    from aiconfigurator.sdk.task_v2 import Task
+
+    if estimate_mode not in ("agg", "disagg"):
+        raise SystemExit("--enable-epd supports --estimate-mode agg or disagg only.")
+    workload = dict(
+        enable_epd=True,
+        backend_version=args.backend_version,
+        database_mode=args.database_mode,
+        isl=args.isl,
+        osl=args.osl,
+        image_height=args.image_height,
+        image_width=args.image_width,
+        num_images_per_request=args.num_images,
+        gemm_quant_mode=args.gemm_quant_mode,
+        moe_quant_mode=args.moe_quant_mode,
+        kvcache_quant_mode=args.kvcache_quant_mode,
+        fmha_quant_mode=args.fmha_quant_mode,
+        comm_quant_mode=args.comm_quant_mode,
+        nextn=args.nextn,
+        nextn_accepted=args.nextn_accepted,
+    )
+    encoder_kwargs = dict(
+        encoder_tp=args.encoder_tp,
+        encoder_batch_size=args.encoder_batch_size,
+        encoder_num_workers=args.encoder_num_workers,
+    )
+    if estimate_mode == "agg":
+        task = Task.from_cli(
+            serving_mode="agg",
+            model_path=args.model_path,
+            system_name=args.system,
+            backend_name=args.backend,
+            **workload,
+        )
+        row = task.run_single_agg(
+            tp=args.tp_size,
+            pp=args.pp_size,
+            dp=args.attention_dp_size,
+            moe_tp=args.moe_tp_size,
+            moe_ep=args.moe_ep_size,
+            batch_size=args.batch_size,
+            ctx_tokens=args.ctx_tokens,
+            **encoder_kwargs,
+        )
+    else:
+        task = Task.from_cli(
+            serving_mode="disagg",
+            prefill_model_path=args.model_path,
+            decode_model_path=args.model_path,
+            prefill_system_name=args.system,
+            decode_system_name=args.decode_system or args.system,
+            prefill_backend_name=args.backend,
+            decode_backend_name=args.backend,
+            **workload,
+        )
+        row = task.run_single_disagg(
+            prefill_tp=args.prefill_tp_size,
+            prefill_pp=args.prefill_pp_size or 1,
+            prefill_dp=args.prefill_attention_dp_size or 1,
+            prefill_moe_tp=args.prefill_moe_tp_size or 1,
+            prefill_moe_ep=args.prefill_moe_ep_size or 1,
+            prefill_batch_size=args.prefill_batch_size,
+            prefill_num_workers=args.prefill_num_workers,
+            decode_tp=args.decode_tp_size,
+            decode_pp=args.decode_pp_size or 1,
+            decode_dp=args.decode_attention_dp_size or 1,
+            decode_moe_tp=args.decode_moe_tp_size or 1,
+            decode_moe_ep=args.decode_moe_ep_size or 1,
+            decode_batch_size=args.decode_batch_size,
+            decode_num_workers=args.decode_num_workers,
+            **encoder_kwargs,
+        )
+    logger.info("EPD %s single-point estimate:", estimate_mode)
+    keys = (
+        "ttft",
+        "tpot",
+        "encoder_latency",
+        "request_latency",
+        "seq/s",
+        "tokens/s/gpu",
+        "num_total_gpus",
+        "(a)workers",
+        "(p)workers",
+        "(d)workers",
+        "(e)workers",
+        "(e)tp",
+        "(e)bs",
+        "(e)memory",
+        "power_w",
+    )
+    for key in keys:
+        if key in row:
+            value = row[key]
+            logger.info("  %-16s %s", key, f"{value:.3f}" if isinstance(value, float) else value)
+
+
 def _run_estimate_mode(args):
     """Run the estimate mode to predict TTFT, TPOT, and power for a single config."""
     from aiconfigurator.cli.api import cli_estimate
@@ -2420,6 +2542,10 @@ def _run_estimate_mode(args):
     )
 
     _resolve_and_validate_nextn(args)
+
+    if args.enable_epd:
+        _run_estimate_epd(args, estimate_mode)
+        return
 
     # Resolve --detail before running the estimate so time detail can compare
     # against a second SOL-mode result.
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -2427,19 +2427,10 @@ def _print_per_ops_latency(per_ops_data: dict) -> None:
             _print_per_ops_section("AFD Transfer (per layer, a2f + f2a)", directional)
 
 
-_ESTIMATE_QUANT_ENUMS = {
-    "gemm_quant_mode": common.GEMMQuantMode,
-    "moe_quant_mode": common.MoEQuantMode,
-    "kvcache_quant_mode": common.KVCacheQuantMode,
-    "fmha_quant_mode": common.FMHAQuantMode,
-    "comm_quant_mode": common.CommQuantMode,
-}
-
-
 def _run_estimate_epd(args, estimate_mode: str) -> None:
     """EPD single-point estimate via Task.run_single_* (dedicated encode pool)."""
-    from aiconfigurator.cli.api import POWER_DATA_COVERAGE_THRESHOLD
-    from aiconfigurator.sdk.task_v2 import Task
+    from aiconfigurator.cli.api import apply_row_power_coverage_gate
+    from aiconfigurator.sdk.task_v2 import _QUANT_ENUM_TABLES, Task
 
     if estimate_mode not in ("agg", "disagg"):
         raise SystemExit("--enable-epd supports --estimate-mode agg or disagg only.")
@@ -2461,9 +2452,7 @@ def _run_estimate_epd(args, estimate_mode: str) -> None:
         nextn=args.nextn,
         nextn_accepted=args.nextn_accepted,
     )
-    workload.update(
-        {name: enum[getattr(args, name)] for name, enum in _ESTIMATE_QUANT_ENUMS.items() if getattr(args, name)}
-    )
+    workload.update({name: getattr(args, name) for name in _QUANT_ENUM_TABLES if getattr(args, name, None)})
     encoder_kwargs = dict(
         encoder_tp=args.encoder_tp,
         encoder_batch_size=args.encoder_batch_size,
@@ -2496,6 +2485,11 @@ def _run_estimate_epd(args, estimate_mode: str) -> None:
         def _role(value, shared):
             return value if value is not None else shared
 
+        # Shared quant/version args map to the per-role fields: disagg Tasks
+        # reject top-level worker fields and silently ignore backend_version.
+        role_shared = {"backend_version": workload.pop("backend_version", None)}
+        for name in [k for k in workload if k.endswith("quant_mode")]:
+            role_shared[name] = workload.pop(name)
         task = Task.from_cli(
             serving_mode="disagg",
             prefill_model_path=args.model_path,
@@ -2504,6 +2498,8 @@ def _run_estimate_epd(args, estimate_mode: str) -> None:
             decode_system_name=args.decode_system or args.system,
             prefill_backend_name=args.backend,
             decode_backend_name=args.backend,
+            **{f"prefill_{k}": v for k, v in role_shared.items()},
+            **{f"decode_{k}": v for k, v in role_shared.items()},
             **workload,
         )
         row = task.run_single_disagg(
@@ -2523,6 +2519,7 @@ def _run_estimate_epd(args, estimate_mode: str) -> None:
             decode_num_workers=args.decode_num_workers,
             **encoder_kwargs,
         )
+    row = apply_row_power_coverage_gate(row)
     logger.info("EPD %s single-point estimate:", estimate_mode)
     keys = (
         "ttft",
@@ -2545,8 +2542,8 @@ def _run_estimate_epd(args, estimate_mode: str) -> None:
         if key not in row:
             continue
         value = row[key]
-        if key == "power_w" and row.get("power_coverage", 1.0) < POWER_DATA_COVERAGE_THRESHOLD:
-            logger.info("  %-16s unavailable (%.0f%% power-data coverage)", key, row["power_coverage"] * 100)
+        if key == "power_w" and value is None:
+            logger.info("  %-16s unavailable (%.0f%% power-data coverage)", key, row.get("power_coverage", 0.0) * 100)
             continue
         logger.info("  %-16s %s", key, f"{value:.3f}" if isinstance(value, float) else value)
 
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -418,6 +418,20 @@ def _add_default_mode_arguments(parser):
         "(vLLM mm_encoder_tp_mode='data' / SGLang --mm-enable-dp-encoder semantics).",
     )
     parser.add_argument(
+        "--enable-epd",
+        action="store_true",
+        help="EPD (vision-language models): run the vision encoder on dedicated encode workers "
+        "instead of colocated with prefill. Turns the disagg experiment into EPD; requires an "
+        "image workload (--image-height/--image-width).",
+    )
+    parser.add_argument(
+        "--encoder-tp",
+        type=int,
+        nargs="+",
+        default=None,
+        help="EPD encode-worker TP sizes to sweep (requires --enable-epd). Default: 1 2 4 8.",
+    )
+    parser.add_argument(
         "--ttft",
         type=float,
         default=2000.0,
@@ -1445,6 +1459,8 @@ def build_default_tasks(
     image_width: int = 0,
     num_images: int = 1,
     enable_encoder_dp: bool = True,
+    enable_epd: bool = False,
+    encoder_tp: list[int] | None = None,
     ttft: float = 2000.0,
     tpot: float = 30.0,
     request_latency: float | None = None,
@@ -1685,6 +1701,8 @@ def build_default_tasks(
             decode_backend_version=backend_version,
             prefill_enable_chunked_prefill=enable_chunked_prefill,
             moe_backend=moe_backend_value,
+            enable_epd=enable_epd,
+            encoder_tp_candidates=encoder_tp,
             **global_kwargs,
         )
 
@@ -2861,6 +2879,8 @@ def main(args):
             image_width=args.image_width,
             num_images=args.num_images,
             enable_encoder_dp=not args.disable_encoder_dp,
+            enable_epd=args.enable_epd,
+            encoder_tp=args.encoder_tp,
             ttft=args.ttft,
             tpot=args.tpot,
             request_latency=args.request_latency,
--- a/src/aiconfigurator/cli/main.py
+++ b/src/aiconfigurator/cli/main.py
@@ -2427,8 +2427,18 @@ def _print_per_ops_latency(per_ops_data: dict) -> None:
             _print_per_ops_section("AFD Transfer (per layer, a2f + f2a)", directional)
 
 
+_ESTIMATE_QUANT_ENUMS = {
+    "gemm_quant_mode": common.GEMMQuantMode,
+    "moe_quant_mode": common.MoEQuantMode,
+    "kvcache_quant_mode": common.KVCacheQuantMode,
+    "fmha_quant_mode": common.FMHAQuantMode,
+    "comm_quant_mode": common.CommQuantMode,
+}
+
+
 def _run_estimate_epd(args, estimate_mode: str) -> None:
     """EPD single-point estimate via Task.run_single_* (dedicated encode pool)."""
+    from aiconfigurator.cli.api import POWER_DATA_COVERAGE_THRESHOLD
     from aiconfigurator.sdk.task_v2 import Task
 
     if estimate_mode not in ("agg", "disagg"):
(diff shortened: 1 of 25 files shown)


tianhaox (commented):
Reviewed at 88cc2dfde. The modeling core holds up well — I specifically chased the "does this change non-EPD behavior" angles (visual tokens entering the agg isl guards and the prefill-batch divisor, the smart_resize rounding) and they all check out as intentional consistency fixes aligned with the pre-existing VL accounting (`isl_eff` in the legacy agg sweep, `inference_session`'s effective-ISL), pinned by tests. Text-only paths are bit-identical.

The main gap is outside the diff: **the generator is not EPD-aware, so `--enable-epd` + `--save-dir` silently emits deployment configs that contradict the recommendation** (`src/aiconfigurator/generator/module_bridge.py` is untouched by this PR):

- **E+agg**: `task_config_to_generator_config`'s agg branch computes `agg_workers = effective_total_gpus // (tp*pp*dp)` (module_bridge.py:216) and never reads `(a)workers`/`(e)workers`, while `_overlay_encoder_stage` folds encoder GPUs into `num_total_gpus` (sweep.py:882-886). Example: a top cell of 3× agg tp2 + 2× encode tp1 (8 GPUs) renders as **4 colocated-encoder agg workers and 0 encode workers**.
- **E+P+D**: the disagg branch's `gpus_per_replica = prefill_workers*p_gpus + decode_workers*d_gpus` (module_bridge.py:238-242) ignores the encoder pool, so replica scaling over-provisions p/d onto the GPUs the row reserved for encoders — and again no encode workers are rendered. Encode workers still only come from manual `--generator-set Workers.encode.*` overrides; the comment there ("the encode worker is not produced by the SDK sweep") is now stale.

`save_results` feeds every best row into the generator unconditionally (the only EPD rejection in main.py is on the auto-recommend path without `--total-gpus`). Suggest either teaching module_bridge the `(a)`/`(e)` columns in this PR, or failing loud / skipping generator artifacts for EPD rows until it does.

Inline comments cover the rest: an `--encoder-tp` KeyError crash, a power_coverage gap, a div-by-zero guard, and three consolidation nits.

src/aiconfigurator/sdk/sweep.py
```diff
@@ -653,6 +718,261 @@ def _get_disagg_worker_candidates(
...
+    saw_oom = False
+    # Same OOM gate as the P/D candidates (set_memory_and_check_oom), charging
+    # the same framework overhead as _get_memory_usage (nccl_mem + other_mem).
+    mem_capacity_gib = database.system_spec["gpu"]["mem_capacity"] / (1 << 30)
+    misc_spec = database.system_spec["misc"]
+    other_mem = misc_spec["other_mem"] * (1.0 + backend.OTHERS_OVERHEAD_FRAC)
+    # The dominance filter below relies on ascending batch order.
+    b_schedule = sorted(set(b_list or _DEFAULT_ENCODER_BATCH_SCHEDULE))
+    for etp in sorted(set(tp_list or _DEFAULT_ENCODER_TP_LIST)):
+        overhead_gib = (misc_spec["nccl_mem"][min(etp, 8)] + other_mem) / (1 << 30)
```
tianhaox, line 859: `--enable-epd --encoder-tp 3` (or 5/6/7) crashes with a raw `KeyError: 3` here (reproduced in a worktree): shipped system YAMLs only define `nccl_mem` keys {1,2,4,8}, `min(etp, 8)` doesn't help for 3/5/6/7, and this lookup runs **before** the `try/except ValueError` around `build_encoder_ops` below that would otherwise gracefully skip non-divisible tp values. Task.validate only requires positive ints and the CLI arg is an unrestricted `type=int, nargs="+"`. Either validate encoder_tp against the supported set up front, or move this lookup after the divisibility guard so the value is skipped with the existing mechanism.

src/aiconfigurator/sdk/sweep.py
```diff
@@ -653,6 +718,261 @@ def _get_disagg_worker_candidates(
...
+        best_seq_s = 0.0
+        for b in b_schedule:
+            latency, power_w, memory = backend.run_encoder_static(
+                model, database, runtime_config, b, latency_correction_scale=latency_correction
+            )
+            if memory.get("total", 0.0) + overhead_gib >= mem_capacity_gib:
+                # Memory grows with batch: larger batches also OOM.
+                saw_oom = True
+                break
+            seq_s = b * 1000.0 / latency
```
tianhaox, line 886: Minor: `latency` can be `0.0` per `run_encoder_static`'s own contract (it guards its power division with `if raw_latency > 0 else 0.0`). The realistic zero routes (no images / zero tokens-per-image) are already pre-empted by the visual-token `ValueError` above, but a degenerate or corrupt perf DB would take down the whole sweep with `ZeroDivisionError` here. Cheap to guard: skip the candidate (or raise the same fail-loud error as missing data) when `latency <= 0`.

src/aiconfigurator/sdk/sweep.py
```diff
@@ -59,6 +61,8 @@
 # (locked in via parity test; do not change without updating picking.py too).
 _RATE_MATCH_PREFILL_DEGRADATION = 0.9
 _RATE_MATCH_DECODE_DEGRADATION = 0.92
+# EPD: the encode pool takes the same class of loss as prefill.
+_RATE_MATCH_ENCODER_DEGRADATION = 0.9
```
tianhaox, line 69: This is hardcoded while its two siblings are configurable Task fields (`rate_match_prefill_degradation` / `rate_match_decode_degradation`, task_v2.py:450-451, exposed in example.yaml and threaded through the sweep kwargs). All three encoder call sites (lines ~877/921/1340) read the module constant directly, so anyone calibrating rate-match degradations against a real deployment — the reason the sibling knobs exist — cannot tune the encoder pool without a code edit. Suggest a `rate_match_encoder_degradation` Task field wired the same way.

src/aiconfigurator/sdk/sweep.py
```diff
@@ -841,12 +1211,38 @@ def sweep_disagg(
...
-    max_prefill_batch_size = prefill_max_num_tokens // runtime_config.isl
+    # The token budget divides by the effective ISL (text + vision context
+    # tokens); Task._prefill_effective_isl builds the budget the same way,
+    # so the caller's batch intent round-trips.
+    try:
+        enc_cfg = get_model_config_from_model_path(model_path).get("extra_params")
+    except Exception:
+        logger.debug("Could not resolve model config for the effective ISL; using text ISL", exc_info=True)
+        enc_cfg = None
+    prefill_effective_isl = runtime_config.isl + BaseBackend._visual_context_tokens_from_encoder_config(
```
tianhaox: This effective-ISL block is now the third copy of the same logic: `inference_session.py:827-838` (pre-existing) and `Task._prefill_effective_isl` (added in this PR) are semantically identical, and the three are coupled only by "must match" comments/docstrings. All three sites have `(model_path, runtime_config)` in hand, so a single shared helper (e.g. next to `_visual_context_tokens_from_encoder_config` on BaseBackend) would drop in at all three call sites — task_v2 already lazy-imports BaseBackend, so no import-cycle issue. Worth consolidating before the next VL token-accounting change silently diverges the prefill batch budget between the three.

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -1808,7 +2020,22 @@ def run_single_disagg(
         # --- Rate-match the pair ---
         p_dict = p_summary.get_summary_df().iloc[0].to_dict()
         d_dict = d_summary.get_summary_df().iloc[0].to_dict()
-        return _rate_match_dict(p_dict, prefill_num_workers, d_dict, decode_num_workers)
+        row = _rate_match_dict(p_dict, prefill_num_workers, d_dict, decode_num_workers)
+        row["power_coverage"] = min(p_summary.get_power_data_coverage(), d_summary.get_power_data_coverage())
```
tianhaox: `power_coverage` ignores the encoder pool entirely: both LM summaries here are built `language_only=True`, so with full LM energy data but no ViT energy data this reports 1.0 — while `run_encoder_static` silently returns `power_w = 0.0` (its energy sum defaults each op to 0.0) and `_overlay_encoder_stage` blends that 0 W into `power_w` weighted by the encode-time share. The understated power then passes the coverage gate as trustworthy. Same on the `run_single_agg` path (coverage stamped before the overlay), and sweep rows get no `power_coverage` at all (the CLI gate defaults to 1.0). Suggest folding an encoder coverage term into the min() — or zeroing/NaN-ing coverage when encoder energy data is absent, matching the fail-loud philosophy the encoder DB lookup already follows.

src/aiconfigurator/cli/main.py
```diff
@@ -2285,6 +2348,129 @@ def _print_per_ops_latency(per_ops_data: dict) -> None:
             _print_per_ops_section("AFD Transfer (per layer, a2f + f2a)", directional)
 
 
+_ESTIMATE_QUANT_ENUMS = {
```
tianhaox: Three pieces of `_run_estimate_epd` duplicate infrastructure that already has an owner: (1) this table is byte-identical to `task_v2._QUANT_ENUM_TABLES`, which also owns `_resolve_quant_str` (used by `Task.from_yaml`) — a new quant knob added there is silently dropped here; (2) the coverage check below re-inlines `api._apply_power_coverage_gate`'s rule (it even imports the threshold from api.py, duplicating only the logic); (3) the disagg required-param checks and `_role()` fallbacks mirror `cli_estimate`, as the comment itself says. Short-term: import the table/resolver and share a coverage-gate helper. Longer-term: `Task.run_single_*` docstrings already claim to subsume cli_estimate's per-point use case, so routing the EPD estimate through one arg-resolution layer would avoid every future estimate knob being added twice — and would give EPD estimates an API surface beyond print-only output.

src/aiconfigurator/sdk/sweep.py
```diff
@@ -636,6 +694,186 @@ def _get_disagg_worker_candidates(
...
+            if enc["encoder_latency"] + r["ttft"] >= ttft_target:
+                continue
+            encoder_capacity = float(enc["seq/s"]) * _RATE_MATCH_ENCODER_DEGRADATION
+            best: tuple[tuple[float, int], int, int] | None = None
+            for a_num in range(1, _MAX_AGG_WORKERS_EPD + 1):
+                e_num = max(1, math.ceil(rate_one * a_num / encoder_capacity))
+                num_gpu = gpus_one * a_num + enc["num_total_gpus"] * e_num
+                key = (rate_one * a_num / num_gpu, -num_gpu)
+                if best is None or key > best[0]:
+                    best = (key, a_num, e_num)
```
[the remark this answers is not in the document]
tianhaox: Addressed at head: E+agg cells are now gated by the per-replica GPU budget — `Task._replica_num_gpu_list()` (num_gpu_per_replica filtered by max_gpu_per_replica) is passed into the sweep as `num_gpu_list` (task_v2.py:1584), and `_rate_match_agg_epd` skips any cell whose total GPU count is not in that set (sweep.py:942; `_epd_e_num_candidates` also derives e-counts from it). An E+agg cell is budgeted exactly like a disagg replica. Resolving.

tianhaox:
Heads-up, related to this PR's sweep results: while studying EPD benefit scenarios on this branch I found a pre-existing GPU-accounting bug that EPD unmasks — filed as #1476.

`_rate_match_dict` (and its parity twin `picking._build_disagg_summary_dict`) compute prefill GPUs as `pp*tp*dp`, missing **cp**, and the report omits cp from the `(p)parallel` string. On main this stays mostly hidden for VL because the colocated encoder pushes cp-prefill rows over tight TTFT filters; with `--enable-epd` the language-only prefill removes that penalty, cp rows become selectable, and their inflated `tokens/s/gpu` (a cp8 worker counted as 1 GPU) dominates the disagg ranking. Concrete example on this branch: Qwen3-VL-8B, 2×1344², ttft=150 → the winning E+P+D row reports 1179 tokens/s/gpu / "3 replicas in 32 GPUs", while the actual shape (2× cp8 prefill + 4 decode + 4 encode = 24 GPUs/replica) needs 72 GPUs and honestly lands at ~264 tokens/s/gpu.

Not caused by this PR, but worth being aware of when validating EPD sweep output — disagg EPD wins that involve cp prefill candidates are currently overstated. Fix incoming for the accounting itself.

src/aiconfigurator/sdk/sweep.py
wenpengw-nv: Fixed: unsupported tp is skipped with a warning before the `nccl_mem` lookup; if all tps are skipped it falls through to the existing `NoFeasibleConfigError`. The single-point/estimate paths share the same helper, so they're covered too.

src/aiconfigurator/sdk/sweep.py
wenpengw-nv: Now raises `ValueError`  fail-loud over skip.

src/aiconfigurator/sdk/sweep.py
wenpengw-nv: Added `rate_match_encoder_degradation` as a Task field, wired like its two siblings (sweep kwargs, overlay, single-point, example.yaml). The `((seq/s x deg) x workers)` association order now lives in a shared helper, so matcher and overlay stay bit-identical.

src/aiconfigurator/sdk/sweep.py
wenpengw-nv: Consolidated into `BaseBackend.effective_prefill_isl`; all three sites delegate. Exception fallback is equivalent (`visual_tokens(None) == 0`), outputs byte-identical pre/post.

src/aiconfigurator/sdk/task_v2.py
wenpengw-nv: `run_encoder_static` now returns an encoder `power_coverage` (mirrors `get_power_data_coverage`); single-point paths blend it time-weighted into the row coverage and gate fail-closed (missing/NaN hides power).

src/aiconfigurator/cli/main.py
wenpengw-nv: (1)(2)fixed: quant table/resolver now shared via `Task.from_cli` + `_QUANT_ENUM_TABLES`, coverage gate shared via `apply_row_power_coverage_gate`. This also fixed a real bug — top-level quant modes crashed disagg estimates; they now route per-role. (3) agreed, left for the follow-up.

wenpengw-nv:
Generator fix takes the fail-loud/skip option: `save_results` skips EPD rows with a warning, and the bridge raises for direct callers; proper `(a)/(e)` mapping stays a follow-up. All other comments are addressed inline.

aic-core/src/aiconfigurator_core/sdk/models/vit_ops.py
```diff
@@ -221,3 +223,18 @@ def build_encoder_ops(enc_cfg: common.VisionEncoderConfig, tp_size: int, enable_
...
+
+    Mirrors an encoder-only instance (e.g. SGLang ``--encoder-only``); only
+    ViT-side rules govern its tensor parallelism.  Duck-types the slice of
+    ``BaseModel`` the encoder phase reads: ``encoder_ops``,
+    ``encoder_config`` and ``config``.
+    """
+
+    encoder_ops: list
+    encoder_config: common.VisionEncoderConfig
+    config: object
```
Arsene12358: **Blocking — the default EPD path crashes with this duck type.** A real packaged PerfDatabase selects the Rust evaluator by default. That evaluator creates a cached full-engine handle, and `build_engine_spec_json` unconditionally reads `model.context_ops` and `model.generation_ops`. The documented CLI command therefore fails for both agg and disagg with `AttributeError: EncoderOnlyModel has no attribute context_ops`; changing only `engine_step_backend` to `python` succeeds.

The minimum fix I verified is to satisfy the full-engine contract with empty language-op lists:

```suggestion
    encoder_ops: list
    encoder_config: common.VisionEncoderConfig
    config: object
    context_ops: list = dataclasses.field(default_factory=list)
    generation_ops: list = dataclasses.field(default_factory=list)
```

Decoupling ad-hoc op-list evaluation from full-engine compilation would also solve the contract mismatch. Either way, please add a regression using a real packaged PerfDatabase with the default backend, covering both agg and disagg plus explicit Rust. Synthetic databases route to Python and cannot catch this failure.

src/aiconfigurator/sdk/sweep.py
```diff
@@ -686,6 +760,298 @@ def _get_disagg_worker_candidates(
...
+                    num_gpu_set or set(),
+                    e_workers_bound,
+                ):
+                    num_gpu = gpus_one * a_num + enc_worker["num_total_gpus"] * e_num
+                    if num_gpu_set and num_gpu not in num_gpu_set:
+                        continue
+                    cell_rate = min(agg_rate, encoder_capacity * e_num)
+                    key = (cell_rate / num_gpu, -num_gpu)
+                    if best is None or key > best[0]:
+                        best = (key, a_num, e_num)
```
Arsene12358: **Blocking — this local argmax can discard the fixed-cluster optimum.** The score here maximizes one-cell efficiency, but downstream picking maximizes packed throughput using `floor(total_gpus / cell_gpus)`. Those objectives diverge when GPUs are stranded.

Concrete case: an 8-GPU agg worker at 10 req/s and a 4-GPU encoder at 6 req/s, with allowed cells `{16, 24}`. This code retains `1a + 2e = 16 GPUs @ 10/s` because 10/16 is larger than 12/24, and discards `2a + 2e = 24 GPUs @ 12/s`. On a 24-GPU cluster, the discarded cell is the actual winner: 12/s versus 10/s.

Please retain the best candidate for every allowed cell GPU count, or retain the nondominated `(cell_gpus, cell_rate)` frontier, until the cluster-aware picker runs. Add the 16-vs-24 GPU counterexample as a regression test and assert that the 24-GPU cell wins for `total_gpus=24`.

src/aiconfigurator/sdk/sweep.py
```diff
@@ -686,6 +760,298 @@ def _get_disagg_worker_candidates(
...
+        )
+    rows: list[dict] = []
+    saw_oom = False
+    # Same OOM gate as the P/D candidates (set_memory_and_check_oom), charging
+    # the same framework overhead as _get_memory_usage (nccl_mem + other_mem).
+    mem_capacity_gib = database.system_spec["gpu"]["mem_capacity"] / (1 << 30)
+    misc_spec = database.system_spec["misc"]
+    other_mem = misc_spec["other_mem"] * (1.0 + backend.OTHERS_OVERHEAD_FRAC)
+    # The dominance filter below relies on ascending batch order.
+    b_schedule = sorted(set(b_list or _DEFAULT_ENCODER_BATCH_SCHEDULE))
```
Arsene12358, line 844: **Blocking — an explicit batch list bypasses the advertised maximum of 8.** The default is `[1, 2, 4, 8]`, but any non-empty `b_list` replaces it without a cap; Task validation only checks positive integers. With the real `h200_sxm / sglang / 0.5.14` database and the Python escape hatch, `[9, 16]` returns valid candidates at both batch sizes. The single-point path accepts the same unsupported values.

Please centralize the supported encoder batch maximum and fail loudly for explicit values above it in Task validation and this lower-level helper, including single-point evaluation. Add boundary tests proving 8 is accepted and 9 is rejected. If the limit is intended to vary by backend or deployment configuration, expose that capability explicitly rather than silently widening the stated SGLang search space.

Arsene12358 (changes requested):
Requesting changes because the advertised EPD workflow fails on the default execution path, and two additional defects can produce suboptimal or unsupported recommendations.

## Spec

- The PR's exact <code>--enable-epd</code> usage command fails for both agg and disagg against the packaged <code>h200_sxm / sglang / 0.5.16</code> database. A real <code>PerfDatabase</code> selects Rust by default, but <code>EncoderOnlyModel</code> does not satisfy the full-engine contract used by the ad-hoc Rust evaluator.
- E+agg collapses feasible cell sizes by per-cell efficiency before the fixed-cluster picker runs. Because whole-cell packing uses floor division, this can remove the actual cluster-throughput winner.
- Explicit encoder batch candidates are not capped at 8 as the PR promises. With the Python escape hatch and a real SGLang database, batches 9 and 16 are both returned as candidates.

I left inline comments with concrete reproductions, suggested fixes, and the regression tests needed to close each item.

The statement that plain PD/agg paths are unchanged when EPD is off also needs reconciliation. This diff unconditionally changes VL image-token rounding and the plain-agg context grid. Those may be valid fixes, but they should be explicitly in scope with before/after prediction evidence, or split from this feature.

## Standards

- <code>CONTRIBUTING.md</code> requires a design issue and agreed design for significant new functionality, and <code>.claude/rules/repo-guide.md</code> requires explicit approval for cross-module contract changes. This PR spans aic-core, SDK, CLI, schemas, and the generator bridge, but the body still contains <code>#xxx</code>, has no linked issue, and has no approving review. Please link the actual design issue and record the cross-module approval.
- <code>CONTRIBUTING.md</code> and Generator Rule 3 require documentation for user-visible features. Please document the new CLI controls in <code>docs/cli_user_guide.md</code>, Task tuning fields in <code>docs/advanced_tuning.md</code>, and the fail-closed EPD generator limitation in <code>docs/generator_overview.md</code>.

## Verification

I reproduced the documented CLI failure locally at this head. The focused changed-area suite passed (<code>222 passed</code>) and the broad CPU unit suite passed (<code>3,324 passed, 10 skipped</code>), which confirms the current tests do not exercise the real-database default-Rust path. All GitHub checks are green, and the head merges cleanly with current <code>main</code>; these are behavioral and policy blockers rather than CI or merge-conflict failures.

src/aiconfigurator/sdk/sweep.py
wenpengw-nv: Fixed in 3111077a: `_MAX_ENCODER_BATCH = 8` is now a single source (SGLang's `SGLANG_ENCODER_MAX_BATCH_SIZE` default), enforced fail-loud at Task validation, the candidates helper, and the single-point args. Boundary tests prove 8 accepted / 9 rejected at each door

aic-core/src/aiconfigurator_core/sdk/models/vit_ops.py
wenpengw-nv: Fixed in 7b96d482 with the suggested empty `context_ops`/`generation_ops`. Regression added in `tests/unit/sdk/task_v2/test_epd_real_database.py`: real packaged h200_sxm/sglang DB, agg + disagg, default and explicit-rust backends. Verified locally against a freshly built extension — the documented command now completes on the default path.

src/aiconfigurator/sdk/sweep.py
wenpengw-nv: Fixed in 3111077a: the matcher now keeps the best rate per admissible cell GPU count and emits the nondominated (cell_gpus, cell_rate) frontier — the cluster-aware picker decides. Your 16-vs-24 counterexample is the regression test and asserts the 24-GPU cell wins at total_gpus=24.

wenpengw-nv:
@Arsene12358 All three blocking items fixed (7b96d482, 3111077a — details inline); docs added in 33c32f0c (`cli_user_guide.md`, `advanced_tuning.md`, `generator_overview.md`); design issuefiled as #1546  and linked from the body.

A self-review pass added three small fixes in the same fail-loud pattern (3f0330b7): the E+agg gpus/replica decomposition now derives worker GPUs from `WORKER_GPU_DIMS` (it dropped cp, printing arithmetic that didn't sum to the row total), estimate mode rejects `--encoder-*` flags without `--enable-epd` instead of silently ignoring them, and `serving_mode='afd'` rejects `enable_epd` instead of silently dropping it.

Arsene12358 (approved):
Re-reviewed at 6058dc4. All previously requested functional fixes are verified, and the linked #1546 design is accepted with the agreed one-time process exception. Commit 6058dc4 adds the missing unit markers so CI now executes the 12 real-database and generator regression tests; the exact selector passes locally, the broader local unit/build gate passed with 3,919 tests, and the remote Build and Test (unit) job is green. I found no remaining merge blockers.

Issue #1333, opened 2026-07-09
Follow-ups from the perf_interp v2 redesign (PR #1303): rust CP model, rust shared-layer, DSV4 CP calib data, minor gaps


tianhaox:
Tracking issue for the known gaps documented in #1303. All are pre-existing or deliberately deferred; none block that PR (each has a guard or is data-blocked).

## 1. ~~Compiled engine: no SPARSE-attention context-parallel (CP) prefill model~~ RESOLVED  
**[RESOLVED in #1303, commits 8a530ae3/6e22160e]** The `_query_cp` compositions (GLM/DSA + DSV4) and the CSA topk DELTA are ported; the `_reject_cp` guard is retired. Both engines compute identically when sparse tables exist and fail loud with the byte-identical message when they don't (symmetry locked by synthetic tests + one real-data in-grid anchor). Remaining work here is purely the DATA/CALIBRATION track (items below); when calibration corrects the model, both implementations change together under the parity tests.  
**[UPDATED 2026-07-09]** Dense-GQA/MoE CP is now modeled upstream (#1313 Phase 1) and flows through the compiled engine; the #1303 guard is narrowed to sparse-attention modules (DSA/DSV4) only.
The Rust engine has no port of Python's `_query_cp` (per-card base + full/cp swap of the super-linear sparse sub-kernels + CP all-gathers) and does not load the sparse-kernel tables it needs. Before #1303 it **silently evaluated the non-CP composition** for `cp_size > 1` (measured: 3392 ms returned where Python fail-louds); #1303 added a spec-conversion guard (`OpConversionError`), so CP sweeps must use `--engine-step-backend python`. Porting = sparse-kernel loaders (`paged_mqa_logits`, `csa_topk_calib`) + `_query_cp` for DSA + DSV4 (CSA/HCA branches) + the CSA topk-DELTA calibration.

## 2. ~~Compiled engine: no shared-layer merge~~ RESOLVED  
**[RESOLVED 2026-07-09]** #1313's `PerfDbSources` multi-source loaders implement the shared-layer merge; reconciled with the v2 query layer in #1303. Verified: DSV3 tp1/dp8/moe_tp8 now matches exactly on both backends (TTFT 858.138 / TPOT 19.444). The sharing-predicate fidelity note below still stands as a data-quality question.
[...]

PR #1351, opened 2026-07-13
fix(rust-core): restore engine-step speedup (SiteIndex::resolve) + add perf regression gate

simone-chen:
#### Overview:

Fix a Rust engine-step performance regression that had made the compiled core slower than pure-Python, and add a CI gate so it cannot silently recur.

**Root cause:** the regression was introduced by **#1303** (`refactor(sdk): perf-table interpolation engine v2 (perf_interp)`), which rewrote the Rust perf-database interpolation. 

#### Details:

- **Fix (`perf_interp.rs`)**: the GEMM scattered-sites resolver (`SiteIndex::resolve`) sorted **every** collected site by distance on each query — recomputing the distance (a sqrt over the site axes) **inside the sort comparator** — an `O(n log n)` per-query cost just to keep the `nn_sites` nearest. It dominated the engine step for query-heavy models. Now: cache each site's distance once per query, cache curve-coverage bounds at build, and gate-then-`select_nth` the nearest sites over a single `(distance, index)` buffer. **Exact parity preserved** (index tie-break reproduces the prior stable-sort order); the 231-case parity gate stays green. `nemotron-nas` context **1180→100µs (0.18× → 2.12× vs Python)**; every engine-step surface is faster in Rust again.
- **Perf gate (`test_engine_step_perf.py`)**: the perf analog of the parity suite — asserts Rust-vs-Python p50 speedup ≥ a per-case floor (all **≥1.0**; nemotron-nas 1.5×). Machine-independent (both timed on the same runner, so the ratio cancels absolute speed), wired into the `rust-engine-step-parity` CI job.
- **CI**: also gate the existing `test_compile_engine_parity.py` suite (47 tests, passing, but previously orphaned from CI).
- **Docs**: replace the stale `benchmarks.md` with a dated/commit-stamped `perf-speedup-report.md` (fresh full-family run, all 9 families 1.08–2.54×); remove completed one-time scan diagnostics and Rust-migration planning docs; de-prefix and refresh the two Phase-2 docs to current repo status; consolidate the two living reports into `docs/`.

`rust/aiconfigurator-core/src/perf_database/perf_interp.rs` (the fix), then `rust/aiconfigurator-core/parity_tests/test_engine_step_perf.py` (the gate).

#### Related Issues:

- Regression introduced by #1303 (`perf_interp` v2).
- Relates to the SILICON-only empirical-layer port (#1333), noted in `docs/python-dedup-plan.md` as the gate on a global Rust default flip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved performance interpolation by using precomputed coverage and distance data, supporting more consistent nearest-site selection.

* **Tests**
  * Added additional CI parity checks for engine compilation round-trips and Rust vs Python runtime behavior.
  * Added an engine-step performance regression gate to ensure Rust is not slower than Python.

* **Documentation**
  * Updated parity scan and runbook materials, deduplication plan, and added a generated engine-step speedup report.
  * Removed obsolete migration plans, benchmark records, and outdated diagnostic documentation.


simone-chen:
Addressed all 5 CodeRabbit findings in 7ab80e1de:

1. **BLAS thread caps** — added `OMP_NUM_THREADS=1` / `MKL_NUM_THREADS=1` env to the perf-gate CI step (matches the benchmark docs; stabilises the Python-side p50).
2. **"two gates" → three** — `perf-speedup-report.md` now lists engine-step parity, compile-engine parity, and perf.
3. **Stale DRIFT count** — replaced `16` with the completed-scan count `4` in `python-dedup-plan.md` (risk table + acceptance criteria).
4. **compile-engine command** — README now says the Rust extension must be built (`maturin develop --release`) first, not pytest alone from a clean checkout.
5. **"machine-independent" wording** — softened to "more comparable across machines" (ratio divides out most machine-speed variance but can shift across architectures) in the perf test docstring, README, and perf-speedup-report.

jasonqinzhou (approved):

PR #1352, opened 2026-07-14
feat: support OSS-backed model deployment

yangeer:
## What changed

- add an AIConfigurator service entry point and Docker image
- discover versioned models from S3-compatible OSS storage
- generate vLLM DynamoGraphDeployment manifests that load models through ModelExpress
- expose OSS model selection and deployment controls in the service UI
- add unit coverage and deployment documentation

## Why

The service previously assumed local or Hugging Face model paths. This adds an OSS-backed workflow for environments where model artifacts are stored in an S3-compatible object store and streamed into vLLM through ModelExpress.

## Impact

Users can browse models from OSS, estimate configurations, and generate/deploy vLLM manifests without embedding credentials or mounting a local model directory. Existing non-OSS generator behavior remains the fallback.

## Validation

- `TMPDIR=/tmp .venv/bin/pytest -s -q tests/unit/generator/test_aggregators.py tests/unit/generator/test_vllm_oss_deployment.py tests/unit/sdk/test_s3_models.py tests/unit/service` (24 passed)
- targeted `ruff check` for changed Python modules and tests
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added a web-based AIConfigurator service for estimates, Kubernetes artifact generation, deployment management, run history, and health checks.
  * Added containerized service deployment with non-root execution and readiness monitoring.
  * Added support for discovering and loading model configurations from S3-compatible storage.
  * Added Kubernetes deployment support for S3-hosted vLLM models and ModelExpress workflows.
  * Added configurable deployment names, node selectors, storage endpoints, regions, and secrets.

* **Bug Fixes**
  * Improved model name resolution for local and S3 model paths.
  * Filtered unsupported vLLM deployment options to produce safer Kubernetes configurations.

* **Documentation**
  * Added Kubernetes and S3-compatible storage deployment guidance.

### DIFF
--- a/docker/Dockerfile.service
+++ b/docker/Dockerfile.service
@@ -1,0 +1,59 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+FROM ghcr.io/astral-sh/uv:0.11.3 AS uv
+FROM registry.k8s.io/kubectl:v1.31.0 AS kubectl
+
+FROM python:3.12-slim-bookworm AS build
+
+COPY --from=uv /uv /usr/local/bin/uv
+WORKDIR /build
+COPY pyproject.toml uv.lock README.md LICENSE ./
+COPY src/ ./src/
+
+# Do not publish an image with Git LFS pointer stubs instead of performance data.
+RUN set -eu; \
+    bad="$(grep -rlE '^version https://git-lfs' /build/src/aiconfigurator/systems 2>/dev/null || true)"; \
+    if [ -n "$bad" ]; then \
+      echo "ERROR: Git LFS objects are not materialized:" >&2; \
+      echo "$bad" >&2; \
+      exit 1; \
+    fi; \
+    uv build --wheel --out-dir /dist
+
+FROM python:3.12-slim-bookworm AS runtime
+
+ARG AIC_MODEL_S3_BUCKET=aiplat
+ARG AWS_ENDPOINT_URL=https://oss-s3.haiercash.com
+ARG AWS_DEFAULT_REGION=cn-east-1
+
+COPY --from=uv /uv /usr/local/bin/uv
+COPY --from=kubectl /bin/kubectl /usr/local/bin/kubectl
+COPY --from=build /dist /wheelhouse
+
+RUN set -eu; \
+    wheel="$(find /wheelhouse -name '*.whl' -print -quit)"; \
+    uv pip install --system "${wheel}[service]"; \
+    rm -rf /wheelhouse; \
+    useradd --create-home --uid 10001 aiconfigurator; \
+    mkdir -p /data/model-configs /tmp/aiconfigurator-service; \
+    chown -R aiconfigurator:aiconfigurator /data/model-configs /tmp/aiconfigurator-service
+
+ENV AIC_MODEL_SOURCE=s3 \
+    AIC_MODEL_S3_BUCKET=${AIC_MODEL_S3_BUCKET} \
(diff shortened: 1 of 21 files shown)


tianhaox:
Thanks for the contribution — the OSS → ModelExpress workflow fills a real gap for air-gapped / enterprise environments where models live in an S3-compatible store. Before line-level comments, I'd like to offer an architecture-level suggestion: this PR bundles three pieces with very different reusability, and splitting them would make the valuable parts much easier to review and land.

## Suggested split

**1. S3-compatible model source (`sdk/s3_models.py`) — worth upstreaming as its own small PR**
- Make `AIC_MODEL_SOURCE=s3` an explicit opt-in. Currently an *unset* env var defaults to `"s3"`, which flips every existing install into S3 mode — `get_default_models()` has no fallback, so the untouched Gradio webapp and `/api/v1/options` break on any unconfigured machine.
- Drop the vendor-specific fallbacks (`oss-s3.haiercash.com`, bucket `aiplat`) — fail fast when unconfigured instead.
- Keep the `@cache` on `get_default_models()` (this PR removes it), and consider `Delimiter`-based listing — the current code pages through every object key (incl. all weight shards) to find `config.json`.

**2. ModelExpress / S3 deployment path (generator + vllm template) — the most valuable piece for the Dynamo ecosystem**
- Make `K8sConfig.oss_enabled` an explicit typed flag instead of inferring from the `s3://` prefix. The new default expressions reference `ServiceConfig`, which is not in the eval context on the CLI path (`module_bridge.apply_defaults`), so the Jinja fallback returns the raw expression *string* — truthy — and **every plain CLI vllm k8s render gets the OSS branch enabled with a garbage image** (reproducible with a stock HF model).
- Restore the neutral `nvcr.io` image default — the private `parties-hub.haiercash.com` registry currently appears in 6 source locations and is unpullable for anyone else.
- Factor the duplicated worker/frontend bootstrap script into a Jinja macro, and tolerate `tokenizer.model`-only models — a missing `tokenizer.json` currently crash-loops both pods under `set -e`.
- `_filter_vllm_k8s_cli_args` applies to **all** vllm k8s renders (not just OSS) and silently strips tuned flags (`--max-num-seqs`, `--block-size`, `--enforce-eager`, `--cudagraph-capture-sizes`) while `run.sh` keeps them — the two artifacts now disagree. Suggest dropping it or gating it to the OSS path with a warning.

**3. `service/` + web UI — suggest keeping downstream or moving to `tools/`**
This is the most site-specific part (localized UI, kubectl-apply assumptions, run-store design). `tools/` already hosts this kind of integration (`simple_sdk_demo`, `automation`, …) without shipping in the wheel. The REST facade may well have upstream value later — probably best revisited once there's a second consumer, rather than fixing the API shape via a demo.

## Service-layer bugs worth fixing wherever it lands
- Apply endpoint: after a successful `kubectl apply`, a health-check timeout or RBAC failure raises 502/504, so `_mark_deployed_dgd` never runs — the deployment succeeds but is reported failed and never recorded. Also `healthy` only checks `exists`, which both code paths hardcode to `True`.
- `_run_estimate`: requests **without** `systems_paths` skip `_SYSTEMS_PATHS_LOCK` entirely, so they can run against another concurrent request's temporarily-swapped global systems paths and return results from the wrong perf data.
- The deployment `name` field is never validated against RFC 1123 (service, UI, or aggregator), so e.g. `"qwen3_0.6B"` renders a manifest `kubectl apply` rejects.

## Process
- Commits are missing the DCO `Signed-off-by` trailer required by CONTRIBUTING.md.

Happy to discuss the split — items 1+2 together would be roughly a third of this PR's size and much faster to get merged.

yangeer:
Closing this draft PR; the changes will be split into focused PRs based on review feedback.

PR #1353, opened 2026-07-14
perf: bound throughput by corrected ttft

Spycsh:
#### Overview:

bound throughput by corrected ttft

Observed from local disagg corr study but may not be the case for Nvidia. Check with CI.

- closes GitHub issue: [#1424](https://github.com/ai-dynamo/aiconfigurator/issues/1424)

## Summary by CodeRabbit

- **Performance Improvements**
  - Improved capacity/throughput and disaggregated latency estimates by incorporating prefill time-to-first-token constraints alongside prefill and decode capacity when calculating effective request rate.

- **Testing**
  - Enhanced accuracy regression monitoring with nightly metrics plus opt-in pull-request runs for performance regressions.
  - Added/updated an automated accuracy regression comment flow that generates summaries and comparison plots from trusted workflow artifacts.


Spycsh:
Hi @tianhaox @jasonqinzhou, we observed a larger gap between the measured and predicted `tokens/s/gpu` and `req/s` in the 1P3D case compared to the others (2P2D and 3P1D), which seems unusual. Specifically, the measured vs. predicted `tokens/s/gpu` differs by about 40%, with the prediction being higher than the measurement, whereas the gap is below 20% for 2P2D and 3P1D.

I think the reason is that AIC is currently performing rate matching only between prefill and decode:

```py
seq_s = min(
    prefill_summary_dict["seq/s"] * prefill_num_worker * prefill_degradation_factor,
    decode_summary_dict["seq/s"] * decode_num_worker * decode_degradation_factor,
)
```

However, if the TTFT correction factor (`_AUTOSCALE_TTFT_CORRECTION_FACTOR = 1.8`) is taken into account (https://github.com/ai-dynamo/aiconfigurator/blob/main/src/aiconfigurator/sdk/picking.py#L38), the prefill capacity should also be limited by:

```python
prefill_summary_dict["bs"] * prefill_num_worker * 1000.0 / prefill_summary_dict["ttft"]
```

In particular, for the 1P3D case where the prefill request rate is lower than the decode request rate, we need to use the **corrected** `prefill_summary_dict["ttft"]` to limit the prefill request rate. Since the corrected `prefill_summary_dict["ttft"]` is higher, the effective prefill request rate should be lower, resulting in a lower AIC-predicted system `seq/s`.

With this PR, the correlation MAPE for `tokens/s/gpu` and `req/s` on XPU drops from **11.77%/11.88%** to **5.91%/6.04%**.

I'm not sure whether this will also help on the CUDA side.

PR #1369, opened 2026-07-15
feat(cli): GPU recommender mode with auto-recommend and sweep optimizations

natoscott:
## Summary

Adds a recommend capability for procurement sizing: given a performance<br>target, find the minimum GPUs and optimal deployment configuration.<br>Inverts the default mode flow -- instead of "given N GPUs, estimate<br>performance" it answers "given a performance target, find minimum GPUs."

Two ways to access it:

```bash
# Explicit recommend subcommand:
aiconfigurator cli recommend --model-path Qwen/Qwen3-32B \
  --system h200_sxm --target-request-rate 10

# Or omit --total-gpus in default mode (auto-recommend):
aiconfigurator cli default --model-path Qwen/Qwen3-32B \
  --system h200_sxm --target-request-rate 10
```

The auto-recommend path was inspired by @jasonqinzhou's suggestion<br>to consolidate modes.

## Details

**SDK API** (`cli/api.py`):

* New `recommend()` function with the same inputs as `cli_default`<br>minus `total_gpus`, plus `target_request_rate` / `target_concurrency`<br>(mutually exclusive).
* Sweeps all valid per-worker TP/DP/EP configurations using<br>`build_default_tasks`, then applies `pick_load_match` to compute<br>`replicas_needed = ceil(target_rate / per_replica_rate)` and<br>`total_gpus_needed = replicas * gpus_per_replica`.
* Auto-escalates GPU budget (8->16->32->64) when the model does not fit<br>on a single node. Escalation uses structured `ExperimentOutcome` to<br>retry only on OOM/KV-cache errors, not SLA or data-gap failures.

**CLI** (`cli/main.py`):

* `recommend` subparser with `--target-request-rate` /<br>`--target-concurrency` mutually exclusive group.
* `default` mode gains the same load target args; omitting<br>`--total-gpus` with a load target auto-dispatches to recommend.

**Generator bridge** (`module_bridge.py`):

* `task_config_to_generator_config` prefers `total_gpus_needed` from<br>the result row over the escalation budget, so saved deployment<br>artifacts have correct replica/worker counts.

**Webapp**:

* Sizing Recommender tab (quant/eplb params removed -- auto-inferred).
* `/recommend` endpoint added to sla_service.py demo FastAPI service.

## Review feedback addressed

* \[tianhaox HIGH\] Generator artifacts sized from recommendation, not budget
* \[tianhaox HIGH\] PP search uses default PP=1 (no PP>1 data; see ai-dynamo/aiconfigurator#1385)
* \[tianhaox MED\] Structured ExperimentOutcome replaces SystemExit(1)
* \[tianhaox MED\] Dead quant/eplb webapp params removed
* \[tianhaox LOW\] Unreachable ValueError removed
* \[CodeRabbit\] README mode count, input validation, load target wording

## Sweep performance optimizations (commits 3-7)

\~31% faster for MoE, \~18% faster for dense (measured on same PP=1 base):

* `copy.deepcopy` -> `dataclasses.replace` in sweep hot paths
* Lazy DataFrame construction in `run_agg` and `run_static`
* Batch disagg results instead of incremental `pd.concat`
* Opt-in `parallel_experiments` via ThreadPoolExecutor

## Test plan

- [X] 254 unit tests pass
- [X] Profiled: MoE 5.63s -> 3.89s, dense 1.78s -> 1.46s
- [X] Integration tests for recommend monotonicity
- [X] New tests: ExperimentOutcome, is_gpu_retriable, module_bridge<br>total_gpus_needed, non-retriable escalation stop, partial failure

Generated with [Claude Code](<https://claude.com/claude-code>)

## Summary by CodeRabbit

* **New Features**
  * Added a `recommend` mode to size the minimum GPUs needed to meet request-rate or concurrency targets and SLA constraints.
  * Added a Sizing Recommender tab with ranked recommendations, debugging info, and CSV download.
  * Added a `/recommend` endpoint for programmatic retrieval of sizing results.
* **Documentation**
  * Updated README, CLI user guide, and Python API examples with `recommend` usage and walkthroughs.
* **Bug Fixes**
  * Improved recommendation reporting and “top config” selection when replica-based sizing is available.
* **Tests**
  * Added integration and unit coverage for `recommend` behavior and escalating sizing.

## Summary by CodeRabbit

* **New Features**
  * Added a `recommend` CLI mode to size GPUs from throughput targets with SLA inputs, returning ranked GPU and replica requirements.
  * Exposed Python support via `cli_recommend` and added a `/recommend` service endpoint for retrieving the top recommendation.
* **Documentation**
  * Updated README and CLI user guide with the new mode, parameters, and examples.
* **Bug Fixes**
  * Improved recommendation handling when no valid configs are found, including clearer retry/escalation behavior for retriable GPU-capacity failures.
  * Refined reporting/ranking when GPU sizing is driven by `*_needed` fields.
* **Tests**
  * Added integration and unit coverage for parsing, escalation logic, and recommendation correctness.

### DIFF
--- a/src/aiconfigurator/cli/__init__.py
+++ b/src/aiconfigurator/cli/__init__.py
@@ -7,13 +7,22 @@ CLI module for aiconfigurator.
 Provides both command-line interface and programmatic Python API.
 
 Python API usage:
-    from aiconfigurator.cli import cli_default, cli_exp, cli_generate, cli_support
+    from aiconfigurator.cli import cli_default, cli_exp, cli_generate, cli_recommend, cli_support
 
     # cli_default: Run agg vs disagg comparison
     result = cli_default(
         model_path="Qwen/Qwen3-32B",
         total_gpus=8,
         system="h200_sxm",
+    )
+
+    # cli_recommend: Find minimum GPUs for a performance target
+    result = cli_recommend(
+        model_path="Qwen/Qwen3-32B",
+        system="h200_sxm",
+        target_request_rate=50.0,
+        ttft=2000,
+        tpot=30,
     )
 
     # cli_exp: Run experiments from YAML or dict config
@@ -42,6 +51,7 @@ from aiconfigurator.cli.api import (
     cli_default,
     cli_exp,
     cli_generate,
+    cli_recommend,
     cli_support,
 )
 
@@ -50,5 +60,6 @@ __all__ = [
     "cli_default",
     "cli_exp",
     "cli_generate",
+    "cli_recommend",
     "cli_support",
 ]
--- a/src/aiconfigurator/cli/__init__.py
+++ b/src/aiconfigurator/cli/__init__.py
@@ -16,7 +16,7 @@ Python API usage:
         system="h200_sxm",
     )
 
-    # cli_recommend: Find minimum GPUs for a performance target
+    # recommend: Find minimum GPUs for a performance target
     result = cli_recommend(
         model_path="Qwen/Qwen3-32B",
         system="h200_sxm",
--- a/src/aiconfigurator/cli/api.py
+++ b/src/aiconfigurator/cli/api.py
@@ -126,6 +126,7 @@ def _execute_and_wrap_result(
     strict_sla: bool = False,
     target_request_rate: float | None = None,
     target_concurrency: float | None = None,
+    parallel_experiments: bool = False,
 ) -> CLIResult:
     """Execute task configs using main.py's function and wrap result in CLIResult."""
     chosen_exp, best_configs, pareto_fronts, best_throughputs, best_latencies, outcomes = _execute_tasks_internal(
@@ -135,6 +136,7 @@ def _execute_and_wrap_result(
         strict_sla=strict_sla,
         target_request_rate=target_request_rate,
         target_concurrency=target_concurrency,
+        parallel_experiments=parallel_experiments,
     )
 
     return CLIResult(
@@ -519,6 +521,7 @@ def cli_recommend(
             strict_sla=strict_sla,
             target_request_rate=target_request_rate,
             target_concurrency=target_concurrency,
+            parallel_experiments=True,
         )
         retriable = [
             name
--- a/src/aiconfigurator/cli/api.py
+++ b/src/aiconfigurator/cli/api.py
@@ -27,7 +27,7 @@ from aiconfigurator.sdk.config import ModelConfig
 from aiconfigurator.sdk.config_builders import apply_nextn as _apply_nextn
 from aiconfigurator.sdk.config_builders import build_model_config as _build_model_config
 from aiconfigurator.sdk.config_builders import resolve_nextn_auto as _resolve_nextn_auto
-from aiconfigurator.sdk.errors import ExperimentOutcome, is_gpu_retriable
+from aiconfigurator.sdk.errors import ExperimentOutcome, NoFeasibleConfigError, is_gpu_retriable
 from aiconfigurator.sdk.models import check_is_moe, resolve_context_fmha_by_data, resolve_dsv4_moe_arch
 from aiconfigurator.sdk.speculative import (
     SpeculativeDecodingProfile,
@@ -304,7 +304,7 @@ def cli_default(
 
     result = _execute_and_wrap_result(tasks, mode="default", top_n=top_n, strict_sla=strict_sla)
     if not result.best_configs:
-        raise SystemExit(1)
+        raise NoFeasibleConfigError("No feasible configurations found for the given parameters.")
 
     if save_dir:
         # Create a mock args object for save_results compatibility
@@ -537,7 +537,7 @@ def cli_recommend(
         )
 
     if not result.best_configs:
-        raise SystemExit(1)
+        raise NoFeasibleConfigError("No feasible GPU configuration found for the given load target.")
 
     if save_dir:
 
@@ -664,7 +664,7 @@ def cli_exp(
 
     result = _execute_and_wrap_result(tasks, mode="exp", top_n=top_n)
     if not result.best_configs:
-        raise SystemExit(1)
+        raise NoFeasibleConfigError("No feasible configurations found for the given experiments.")
 
     if save_dir:
         # Create a mock args object for save_results compatibility
--- a/src/aiconfigurator/cli/api.py
+++ b/src/aiconfigurator/cli/api.py
@@ -120,10 +120,17 @@ def _execute_and_wrap_result(
     mode: str,
     top_n: int = 5,
     strict_sla: bool = False,
+    target_request_rate: float | None = None,
+    target_concurrency: float | None = None,
 ) -> CLIResult:
     """Execute task configs using main.py's function and wrap result in CLIResult."""
     chosen_exp, best_configs, pareto_fronts, best_throughputs, best_latencies = _execute_tasks_internal(
-        tasks, mode, top_n=top_n, strict_sla=strict_sla
+        tasks,
+        mode,
+        top_n=top_n,
+        strict_sla=strict_sla,
+        target_request_rate=target_request_rate,
+        target_concurrency=target_concurrency,
     )
 
     return CLIResult(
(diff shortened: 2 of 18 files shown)


natoscott:
@coderabbitai Thanks for the thorough review. Here's what we addressed:

**Fixed:**
- **README mode count** — corrected to six modes (added `estimate`)
- **Unreachable `ValueError`** — moved the `gpus_per_node > _MAX_GPUS_PER_WORKER` guard before the escalation loop
- **Quant/eplb params silently dropped** — removed the unused quant and eplb inputs from `run_recommender` and event handler wiring. These are auto-inferred by Task from the HF model config and were never forwarded to `cli_recommend`; showing them in the UI was misleading
- **`min_gpus` test helper** — now computes the minimum across all modes, not just the first non-empty one
- **Display table GPU count** — `display_total_gpus` now uses `row["total_gpus_used"]` when `replicas_needed` is present, instead of the search-ceiling `total_gpus` parameter. Applied to both disagg (line 188) and agg (line 282) rendering sites
- **Ranking label** — updated to show "Ranked by total_gpus_needed" when load-match sorting is active, instead of the incorrect "Ranked by tokens/s/gpu"

**Kept as-is (intentional):**
- **`target_request_rate` default of 10** — a form that works on first click is better UX than two empty fields that require the user to figure out which to fill. If a user fills in concurrency without clearing rate, the "exactly one of..." error is self-explanatory
- **Argument parser duplication** — this is the established pattern across all modes (`default`, `estimate`, `exp`, `generate`). Extracting shared args is a worthwhile refactor but out of scope for this PR

tianhaox (commented):
Thanks — the layering here is sound (thin CLI/API/webapp wrappers over the existing `pick_load_match` machinery, which this PR is the first to actually expose end-to-end), and the display fixes in `report_and_save.py` check out against the SDK. Test coverage for parsing/wiring/escalation is good.

I traced the full call chain (`_execute_tasks` → `pick_load_match` → `save_results` → `task_config_to_generator_config`) and found **two issues I'd consider merge-blocking**, plus a few smaller ones — all inline below:

1. **[HIGH]** With `--save-dir`, the generated deployment artifacts are sized from the internal escalation budget (`task.total_gpus` = 8/16/32/64), not `total_gpus_needed` — so the saved `generator_config.yaml`/artifacts silently contradict the recommendation. The table/summary path was fixed but the artifact path wasn't.
2. **[HIGH]** `_build_recommend_tasks` unconditionally overrides PP candidates to `[1, 2, 4]`, which re-opens configurations that `task_v2` deliberately pins to `pp=[1]` for wideep/megamoe MoE paths.

Things I explicitly verified as fine: `_execute_tasks` already accepts the load-match params; omitting `max_total_gpus` matches the unconstrained-output intent; `pick_load_match` produces `replicas_needed`/`total_gpus_needed` as assumed; `dataclasses.replace` re-runs `__post_init__` so the budget clamping works; all webapp component factories/handlers referenced actually exist.

src/aiconfigurator/cli/api.py
```diff
@@ -300,6 +307,258 @@ class _MockArgs:
...
+        mock_args.ttft = ttft
+        mock_args.tpot = tpot
+        mock_args.request_latency = request_latency
+        mock_args.top_n = top_n
+        mock_args.generated_config_version = None
+        mock_args.generator_set = None
+        mock_args.generator_config = None
+        mock_args.generator_dynamo_version = None
+
+        save_results(
```
tianhaox, line 568: **[HIGH] Saved deployment artifacts are sized from the search budget, not the recommendation.**

`save_results` generates `generator_config.yaml` + backend artifacts per top config via `task_config_to_generator_config`, which computes worker/replica counts from `task_config.total_gpus` (`generator/module_bridge.py:193-224`: `agg_workers = total_gpus // gpus_per_replica`).

In recommend mode, `task.total_gpus` is the internal escalation budget (8/16/32/64) — not `total_gpus_needed`. So if the recommendation says "96 GPUs, 12 replicas" but the successful budget was 8 and the best config uses tp2, the saved artifacts describe a 4-worker deployment: neither one replica nor the recommended size, but an internal number the user never chose.

The display path (`report_and_save.py`) was fixed to use `replicas_needed`, but this artifact path wasn't. Suggest either patching the tasks passed to `save_results` per-row (`dataclasses.replace(task, total_gpus=<row total_gpus_needed>)`), or disabling artifact generation for recommend mode (CSV/plots only) until it's wired up. There's currently no test covering the `save_dir` path for recommend mode.

src/aiconfigurator/cli/api.py
```diff
@@ -300,6 +307,258 @@ class _MockArgs:
...
+    """Widen the search space for recommend mode: enable PP and scale GPU candidates."""
+    import dataclasses
+
+    num_gpu_candidates = [1 << i for i in range(total_gpus.bit_length()) if (1 << i) <= total_gpus]
+
+    rebuilt = {}
+    for name, task in base_tasks.items():
+        overrides = {
+            "total_gpus": total_gpus,
+            "agg_pp_candidates": [1, 2, 4],
```
tianhaox: **[HIGH] Unconditional PP override breaks the wideep/megamoe search-space constraints.**

`task_v2.py:_resolve_agg_search` deliberately pins `agg_pp_candidates` to `[1]` for trtllm/sglang wideep and sglang megamoe (its `_set` helper only applies when the field is `None`). Forcing `[1, 2, 4]` here re-opens PP>1 combinations for those paths — configurations with no collector data and questionable backend support. In HYBRID mode these get extrapolated estimates and can surface as "cheapest" recommendations, which is the worst failure mode for a procurement tool.

Suggest only widening PP when the resolved default would have allowed it, e.g. skip the override when the task is wideep/megamoe (or preserve the task's resolved candidates and only widen `num_gpu_candidates`/`total_gpus`). Same applies to the prefill/decode overrides below.

Related minor point: with `tp <= 8` and `pp <= 4`, the max reachable per-worker size is 32 GPUs for dense models, so the final 64-GPU escalation step buys nothing there.

src/aiconfigurator/cli/api.py
```diff
@@ -300,6 +307,258 @@ class _MockArgs:
...
+            result = _execute_and_wrap_result(
+                tasks,
+                mode="default",
+                top_n=top_n,
+                strict_sla=strict_sla,
+                target_request_rate=target_request_rate,
+                target_concurrency=target_concurrency,
+            )
+            break
+        except SystemExit as exc:
```
tianhaox: **[MED] Escalation only triggers on total failure, and `SystemExit(1)` conflates several causes.**

`_execute_tasks` raises `SystemExit(1)` only when *all* experiments fail. Two consequences:

1. Partial failure doesn't escalate: if agg succeeds within the current budget but disagg needs a larger one, disagg silently disappears from the results instead of being retried at the next budget.
2. `SystemExit(1)` also covers "SLA too tight" and "no perf data in the database" — the latter re-runs up to 4 full sweeps pointlessly before failing.

At minimum, worth a log line acknowledging the other causes; ideally escalate per-experiment rather than on the all-failed signal.

src/aiconfigurator/cli/api.py
```diff
@@ -300,6 +307,258 @@ class _MockArgs:
...
+                budgets[attempt + 1],
+            )
+
+    if result is None:
+        raise ValueError(
+            f"System {system} has {gpus_per_node} GPUs/node which exceeds "
+            f"the maximum per-worker search limit ({_MAX_GPUS_PER_WORKER})."
+        )
+
+    # Re-select chosen_exp by fewest total GPUs (not highest throughput).
```
tianhaox: **[LOW] This re-selection is redundant.** Under load-match, `process_experiment_result` (`cli/utils.py`) already returns `best_throughput = 1/total_gpus_needed`, so `_execute_tasks`'s `chosen_exp = max(best_throughputs)` is already the fewest-GPU experiment. Harmless as a defensive measure, but the PR description's claim that the default path picks by highest throughput isn't accurate for load-match mode — either drop this block or leave a comment noting it's defensive.

src/aiconfigurator/cli/api.py
```diff
@@ -300,6 +307,258 @@ class _MockArgs:
...
+        except SystemExit as exc:
+            if exc.code != 1 or attempt == len(budgets) - 1:
+                raise
+            logger.info(
+                "Recommend: no valid config at %d GPUs/worker; escalating to %d.",
+                gpu_budget,
+                budgets[attempt + 1],
+            )
+
+    if result is None:
```
tianhaox: **[LOW] Dead code with a misleading message.** `budgets` always has at least one element, so the loop either sets `result` or re-raises on the last attempt — this branch is unreachable, and the "exceeds the maximum per-worker search limit" message doesn't describe anything that can happen.

src/aiconfigurator/webapp/events/event_fn.py
```diff
@@ -1389,3 +1400,86 @@ def generate_combined_csv(record_df_prefill, record_df_decode):
         csv_filename = "data.csv"
         combined_df.to_csv(csv_filename, index=False)
         return csv_filename
+
+    @staticmethod
+    def run_recommender(
```
tianhaox: **[MED] The quant-mode and eplb inputs are dead parameters.** `gemm_quant_mode`, `kvcache_quant_mode`, `fmha_quant_mode`, `moe_quant_mode`, `comm_quant_mode`, and `enable_eplb` are accepted here but never passed to `cli_recommend` — which couldn't take them anyway (`build_default_tasks` infers quantization from the model config). So the tab renders a quant panel whose controls have zero effect on the results, with no indication to the user.

Suggest removing the quant panel from this tab (and these params/handler inputs), or marking it read-only as "inferred from model config".

src/aiconfigurator/cli/main.py
```diff
@@ -385,6 +385,168 @@ def _add_default_mode_arguments(parser):
     )
 
 
+def _add_recommend_mode_arguments(parser):
```
tianhaox, line 460: **[LOW] ~160 lines duplicated from `_add_default_mode_arguments`, with subtle divergence already** (e.g. help text wording). Given this repo's history of "fixed in one mode, forgot the other" bugs, suggest extracting the shared workload/SLA/model args into a helper both modes call, keeping only `--total-gpus` vs the target group as mode-specific.

src/aiconfigurator/cli/api.py
natoscott: This has been removed.

src/aiconfigurator/cli/api.py
natoscott: Dropped this block.

src/aiconfigurator/cli/api.py
natoscott: Removed that, thanks.

src/aiconfigurator/cli/api.py
natoscott: Yep, fixed in module_bridge.py where task_config_to_generator_config now reads total_gpus_needed from the result row.  Will add to the testing heretoo  - thanks.

src/aiconfigurator/cli/api.py
natoscott: Completely reworked this aspect now, thanks - should be resolved.

src/aiconfigurator/webapp/events/event_fn.py
natoscott: Removed, thanks.

src/aiconfigurator/cli/api.py
natoscott: New test verifies that recommend() passes the DataFrame with total_gpus_needed through to save_results, which is what task_config_to_generator_config reads in the bridge. Combined with test_module_bridge_recommend_gpus.py (which verifies the bridge uses that column), the full save path should be covered now.

src/aiconfigurator/cli/main.py
natoscott: Opened issue #1386 tracking this separately.  With the auto-recommend feature (auto-dispatch to _run_recommend when --total-gpus is omitted), this is less pressing I guess - users can use default mode for both paths, so the recommend subparser is becoming a secondary entry point. The duplication will resolve naturally if/when we consolidate to just the 2 modes.

src/aiconfigurator/webapp/components/readme_tab.py
```diff
@@ -52,6 +52,12 @@ def create_readme_tab(app_config):
                     This tab is used to do comparison between different pareto estimation results.
                 #### Use cases
                     1. comparing agg vs. disagg, to understand whether disagg might help in your case and find the best deployment strategy of your target SLA. Disagg is not always better. You need to specify your SLA.
+
+
+                ### Sizing Recommender tab
+                    Given a model, system, backend, workload, and a throughput target (request rate or concurrency),
+                    find the minimum number of GPUs and optimal deployment configuration (agg vs disagg, TP/PP/DP).
+                    Designed for procurement sizing — the output is unconstrained.
```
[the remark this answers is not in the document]
natoscott: Fixed.

src/aiconfigurator/cli/main.py
```diff
@@ -385,6 +385,167 @@ def _add_default_mode_arguments(parser):
...
+        type=float,
+        default=None,
+        help="Target system request rate in req/s. Find minimum GPUs to serve this rate.",
+    )
+    load_group.add_argument(
+        "--target-concurrency",
+        type=float,
+        default=None,
+        help="Target number of concurrent users. Find minimum GPUs to serve this concurrency.",
+    )
```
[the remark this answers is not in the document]
natoscott: Added a positive-float validator to handle this aspect now, thanks.

natoscott:
@coderabbitai resume

src/aiconfigurator/cli/__init__.py
```diff
@@ -43,6 +52,7 @@
     cli_exp,
     cli_generate,
     cli_support,
+    recommend,
 )
```
Arsene12358: README.md:149/170 and docs/cli_user_guide.md:384 import `cli_recommend`, which doesn't exist — only `recommend` is exported, so the documented examples raise `ImportError`. Adding an alias here is the minimal fix and also matches the sibling `cli_*` naming convention:

```suggestion
    recommend,
)

# Alias matching the sibling cli_* naming used throughout the docs.
cli_recommend = recommend
```

src/aiconfigurator/cli/__init__.py
```diff
@@ -51,4 +61,5 @@
     "cli_exp",
     "cli_generate",
     "cli_support",
+    "recommend",
```
Arsene12358: Export the alias as well so `from aiconfigurator.cli import cli_recommend` (as shown in README.md and docs/cli_user_guide.md) works:

```suggestion
    "recommend",
    "cli_recommend",
```

src/aiconfigurator/cli/__init__.py
```diff
@@ -7,7 +7,7 @@
 Provides both command-line interface and programmatic Python API.
 
 Python API usage:
-    from aiconfigurator.cli import cli_default, cli_exp, cli_generate, cli_support
+    from aiconfigurator.cli import cli_default, cli_exp, cli_generate, recommend, cli_support
```
Arsene12358: With the `cli_recommend` alias exported, this usage docstring can match the README/user-guide spelling too (the example on lines 19–26 below likewise):

```suggestion
    from aiconfigurator.cli import cli_default, cli_exp, cli_generate, cli_recommend, cli_support
```

Arsene12358 (changes requested):
**The documented Python API name `cli_recommend` doesn't exist — the README and user-guide examples raise `ImportError`.**

README.md (lines 149 and 170) and docs/cli_user_guide.md (line 384) both do `from aiconfigurator.cli import cli_recommend`, but the package only exports `recommend`:

```
README.md:149:            from aiconfigurator.cli import cli_default, cli_exp, cli_generate, cli_recommend, cli_support
README.md:170:            result = cli_recommend(model_path="Qwen/Qwen3-32B-FP8", ...)
docs/cli_user_guide.md:384:  from aiconfigurator.cli import cli_recommend
src/aiconfigurator/cli/__init__.py exports: CLIResult, cli_default, cli_exp, cli_generate, cli_support, recommend
```

So copy-pasting the documented quick-start fails on the import line. Interestingly the docs' spelling is the one that matches the package convention — every sibling API is `cli_*` (`cli_default`, `cli_exp`, `cli_generate`, `cli_support`).

The inline suggestions add a `cli_recommend` alias in `cli/__init__.py`, which is the smallest change that makes the documented examples true while keeping the existing `recommend` callers (`cli/main.py`, `sla_service.py`, webapp) working. Alternatives if you prefer: rename `recommend` -> `cli_recommend` outright in `api.py` and update the internal imports, or keep `recommend` only and fix the two docs — any of the three resolves it, as long as docs and exports agree. A tiny doc-import smoke test (import each name shown in the README) would keep this from regressing.

src/aiconfigurator/cli/__init__.py
natoscott: @Arsene12358 yep, these are all the same issue - originally the code was using cli_recommend throughout.  Then there was a discussion that we want to end up with just 2 modes in the end, estimate and recommend, so it seemed the right thing to do was to push ahead with "recommend" naming ... but now I'm not sure, I may have misinterpreted the earlier discussion.  I'm happy with either name (but ideally just one right?) - should I change it over to cli_recommend throughout?

Arsene12358: @natoscott that history makes sense, thanks. My vote: **`cli_recommend` throughout, single name, no alias.** Reasoning:

- Consistency should track the code as it is today, not the planned end-state. Right now every exported sibling is `cli_*` (`cli_default`, `cli_exp`, `cli_generate`, `cli_support`), and there's no `estimate` Python API yet — so a lone `recommend` is the odd one out, and the docs in this PR already say `cli_recommend` anyway.
- Since `recommend` is introduced by this PR (never shipped in a release), renaming it now has zero compat cost — which is exactly why doing it now beats doing it later. Agreed on ideally just one name: skip the alias I suggested and just rename.
- When the 2-mode consolidation actually lands, that's the right moment to rename the whole API family in one deliberate PR with deprecation aliases for the old `cli_*` names — a single new function jumping ahead of that migration just splits the convention in the meantime.

So: rename to `cli_recommend` in `api.py` + `__init__.py` + the internal imports (`cli/main.py`, `sla_service.py`, webapp `event_fn.py`, tests), and the README/user-guide examples become correct as written.

natoscott: @Arsene12358 done.  (plus a rebase, so webapp bits here are now dropped)

src/aiconfigurator/cli/__init__.py
natoscott: Done.

src/aiconfigurator/cli/__init__.py
natoscott: Resolved, everything using cli_recommend now.

src/aiconfigurator/cli/report_and_save.py
```diff
@@ -391,11 +403,18 @@ def log_final_summary(
 
     if not best_config_df.empty:
         best_conf_details = best_config_df.iloc[0]
-        summary_box.append(f"    - Best Throughput: {best_throughput * chosen_task.total_gpus:,.2f} tokens/s")
+        if load_match and "replicas_needed" in best_conf_details.index:
+            display_total_gpus = int(best_conf_details["total_gpus_needed"])
+        else:
+            display_total_gpus = chosen_task.total_gpus
+        summary_box.append(f"    - Best Throughput: {best_throughput * display_total_gpus:,.2f} tokens/s")
         summary_box.append(f"    - Per-GPU Throughput: {best_throughput:.2f} tokens/s/gpu")
```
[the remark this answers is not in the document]
natoscott: Good catch - fixed.

Yiming992 (approved):
Approving — all three blocking items from my earlier review are resolved on this head: the API is now `cli_recommend` throughout so the README/user-guide examples work as written, the no-feasible-sizing path raises a catchable `NoFeasibleConfigError` instead of `SystemExit`, and the `get_static_info()` call-order regression lost its live consumer with the webapp removal (I re-verified all three against `bc57aa5`).

Non-blocking follow-ups we'll pick up after merge:

- The five open CodeRabbit minors are all still valid on this head: `min_gpus` cross-mode minimum in the integration test, the "exactly one (mutually exclusive)" doc wording, the silent drop of a load target when `--total-gpus` is also given, non-positive target parse tests, and a disagg case in the bridge test.
- `get_static_info()` still asserts on `_summary_df` while `run_static` defers it — the remaining caller (`tools/simple_sdk_demo/cli/main.py`) happens to call `get_summary_df()` first, but it's a trap for the next caller; a one-line materialize-before-assert would harden it.
- The `mode == "recommend"` branch in `_execute_tasks`'s no-results hint is unreachable (`cli_recommend` passes `mode="default"`), so recommend users still get the `--total-gpus` suggestion; keying the hint off the load target would fix it.

Arsene12358 (approved):
Requested changes landed — cli_recommend throughout, verified on bc57aa5.

PR #1370, opened 2026-07-15
feat(collector): Collector V3 design + manifest v2 per-op runtime pins (AIC-1217, AIC-1500)

Arsene12358:
PR 1 of 4 of the **Collector V3: op-centric collector management** project (Linear: [AIC-1217](https://linear.app/nvidia/issue/AIC-1217/finalize-collector-v3-design) design + [AIC-1500](https://linear.app/nvidia/issue/AIC-1500/collector-v3-a-manifest-v2-per-op-runtime-pins-resolver-validation) manifest v2). It carries the reviewed design contract and the first implementation slice: per-op runtime pins. The remaining PRs (family-tree data reorg → provenance/`reuse.yaml`/`changed_ops` + CI audit → loader reuse rules + evidence policy) build on this one; the full delivery plan is §10 of the design doc.

## What's in here

- **`docs/perf_database/collector-v3-op-centric-design.md`** — the Collector V3 design contract (op identity chain, family-first physical layout, manifest v2, provenance, three-channel reuse, changed-op manifest, evidence policy, quarterly operation). Mirrored in Linear on the project; this copy is canonical.
- **`collector/framework_manifest.yaml` → `schema_version: 2`** — per-framework `default` runtime + explicit per-family override slots; wideep flattened to peer frameworks (`wideep_sglang`, `wideep_trtllm`) with `base_framework`/`data_backend`; all public images pinned by **multi-arch index digest** (`name:tag@sha256:...`; bare internal image names exempt). Pins are unchanged: sglang 0.5.14, trtllm 1.3.0rc10, vllm 0.24.0, wideep-sglang 0.5.10.
- **`collector/framework_manifest.py`** — v2 validation (incl. the digest rule), `resolve_op_runtime(framework, op)` implementing `op → PerfFile table → catalog family → families[family] or default`, per-op `require_collector_runtime` (public signature preserved — `collector/collect.py` is deliberately untouched; mixed-pin requests fail with the exact op→version split), and `validate_resolution()` returning one error per unresolvable registry op.
- **`collector/op_catalog.py`** — loader for the table→family map from `collector/op_backend_catalog.yaml`. The catalog does not exist on main yet (it arrives with #1345), so the loader returns `None` when absent: all-default resolution works today, while declaring a `families:` override without the catalog is a hard error.
- **CI gate**: `test_real_manifest_resolves_every_registry_op` asserts `validate_resolution() == []` in the standard unit suite — every op in all five registries must resolve to exactly one pinned runtime, fail-closed.
- README version-management section rewritten for v2, including: pin the **index** digest, never a platform-child digest (a child digest pins one architecture and breaks the rest of the fleet).

## Coordination notes

- **#1345 (op-backend facts):** the moment its `op_backend_catalog.yaml` lands, the CI gate here starts requiring every registry op's table to have a catalog family (including `wideep_moe_perf`, used by both wideep registries). Three currently-missing tables are already flagged on that PR ([comment](https://github.com/ai-dynamo/aiconfigurator/pull/1345#issuecomment-4979252717)); whichever PR rebases second will see the gate fire if the catalog misses a registry table — that is intended fail-closed behavior.
- **GitLab auto-collect pipeline:** if it parses `frameworks.<fw>.version` (v1 schema), it must learn the v2 path `frameworks.<fw>.default.version` before its next run against a checkout containing this change.

## Known limitation (deliberate)

`require_collector_runtime` detects mixed pins by version string only; two family overrides sharing a version but pinning different images would not be split. Unreachable with today's manifest (no overrides yet) — noted here so PR 4's reviewer sees it with fresh eyes.

## Test plan

`uv run pytest tests/unit/collector -m unit` → 428 passed, 5 skipped (includes the real-manifest CI gate, all five v1 `require_collector_runtime` regression cases, fail-closed gates for override-without-catalog / table-without-family / unknown-op, malformed-catalog hardening, and wideep validator branches). All five image digests were verified against the live registries as manifest-list/index digests, twice (implementation + independent review pass).

## Summary by CodeRabbit

* **New Features**
  * Added Collector V3 per-operation runtime resolution using `(framework, op)` → table → catalog family mappings.
  * Introduced workload-aware runtime selection (including WideEP compatibility) and variant-based runtime image access.
* **Documentation**
  * Updated Version Management for manifest schema v2 and added the op-centric Collector V3 design spec.
* **Bug Fixes**
  * Enforced fail-closed resolution for unknown identities and stricter manifest validation; tightened public container images to require digest pinning.
* **Tests**
  * Added/expanded unit tests for op catalog loading, schema v2 validation, WideEP/override rules, and `validate_resolution` error reporting.

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -139,8 +139,9 @@ ops cannot generate the needed data points.
 Each backend (trtllm, vllm, sglang) has a **registry** (`registry.py`) that maps ops to collector modules, and a **version resolver** (`version_resolver.py`) that picks the right module at runtime. Individual collector files declare their compatibility via `__compat__`. The current collector framework versions and runtime images are declared in `framework_manifest.yaml`.
 
 ```text
-framework_manifest.yaml — current collector framework versions and images
-framework_manifest.py   — manifest loader/validator
+framework_manifest.yaml — current collector framework versions and images (schema v2)
+framework_manifest.py   — manifest loader/validator/resolver
+op_catalog.py         — table→family identity map (op_backend_catalog.yaml)
 model_cases.py       — collector v2 model/SM case-plan resolver
 registry.py          — declares which module handles which version range
 version_resolver.py  — routes runtime version → module (packaging.version)
@@ -152,8 +153,36 @@ wideep/*/registry.py — WideEP-only ops appended when the v2 plan requests them
 network/             — collective communication collectors and Slurm network jobs
 ```
 
-WideEP entries in `framework_manifest.yaml` describe independent runtimes.
-`collect.py` requires the exact public version and rejects mixed-pin runs.
+`framework_manifest.yaml` is `schema_version: 2`: each framework entry has a
+`default` runtime (`version` + `images`) plus an optional `families:` map of
+per-family overrides. Resolution for a given `(framework, op)` walks
+`table → catalog family → families[family] or default` — exactly one runtime
+per op, always. `families:` overrides require the op catalog
+(`op_backend_catalog.yaml`); without it, family-scoped pins fail closed rather
+than silently falling back to `default`.
+
+WideEP runtimes (`wideep_sglang`, `wideep_trtllm`) are flattened into peer
+framework entries rather than nested under a `wideep:` key, so the same
+per-op resolution logic applies uniformly. Each WideEP entry adds
+`base_framework` (the stock framework it forks from), `data_backend` (which
+table namespace its output is written under), and `collector_dir`.
+`collect.py` requires the exact public version for a run and rejects
+mixed-pin selections across ops.
+
+Image digests are policy, not decoration: any image reference that names a
+public registry path (contains `/`) must be pinned with `@sha256:<64 hex>`;
+bare internal image names (no `/`, e.g. `deepseek-v4-blackwell`) are exempt.
+When bumping a version pin, re-fetch and update the digest for every image
+variant touched — a stale digest silently keeps pulling the old image.
+
+Use `collector.framework_manifest.resolve_op_runtime(framework, op)` to
+resolve a single op's runtime, and
+`collector.framework_manifest.validate_resolution()` to check the whole
+manifest against every backend registry at once — it returns a list of error
+strings (empty = valid) and is the fail-closed CI gate asserting every
+registered op resolves to exactly one pinned runtime. See
+`docs/perf_database/collector-v3-op-centric-design.md` §4 for the full
+design.
 
 ## File Naming Convention
 
--- a/collector/README.md
+++ b/collector/README.md
@@ -169,9 +169,10 @@ table namespace its output is written under), and `collector_dir`.
 `collect.py` requires the exact public version for a run and rejects
 mixed-pin selections across ops.
 
-Image digests are policy, not decoration: any image reference that names a
-public registry path (contains `/`) must be pinned with `@sha256:<64 hex>`;
-bare internal image names (no `/`, e.g. `deepseek-v4-blackwell`) are exempt.
+Image digests are policy, not decoration: the rule is syntactic — any image
+reference containing `/` must be pinned with `@sha256:<64 hex>`; only bare
+internal image names (no `/`, e.g. `deepseek-v4-blackwell`) are exempt. In
+practice that means every public-registry reference is digest-pinned.
 When bumping a version pin, re-fetch and update the digest for every image
 variant touched — a stale digest silently keeps pulling the old image. Always
 pin the manifest-list (index) digest, never a platform-child digest —
--- a/collector/README.md
+++ b/collector/README.md
@@ -59,13 +59,17 @@ Today we only collect intra-node comm. This script will collect custom allreduce
 It will also collect nccl allreduce, all_gather, all2all, reduce_scatter using nccl.
 The generated files are nccl_perf.txt, oneccl_perf.txt, and custom_allreduce_perf.txt.
 
-# Collector v2: model-centric cases
+# Model-centric cases and healing runs
 
-Collector v2 is model-centric. A healing run collects cases for a specific
-model/GPU pair instead of running every op bucket and hoping the support
-matrix improves. Use `--model-cases-full` when you want collector v2 YAML to
-define a full model-centric run. Omitting all model-case flags runs the
-backend registry directly without a collector v2 case plan.
+The model-centric case layer (introduced as "collector v2") is how collection
+coverage is declared, and it is unchanged under Collector V3 — V3 adds
+op-centric *management* on top (per-family runtime pins, provenance sidecars,
+declared reuse; see `docs/perf_database/collector-v3-op-centric-design.md`),
+it does not replace how cases are declared or healed. A healing run collects
+cases for a specific model/GPU pair instead of running every op bucket and
+hoping the support matrix improves. Use `--model-cases-full` when you want
+the case YAML to define a full model-centric run. Omitting all model-case
+flags runs the backend registry directly without a model-centric case plan.
 
 ```bash
 # Heal one model on one GPU type.
@@ -85,10 +89,10 @@ python3 collect.py --backend trtllm \
   --gpu b200_sxm \
   --plan-only
 
-# Collector v2 full run: aggregate base ops plus every model case YAML file.
+# Full model-centric run: aggregate base ops plus every model case YAML file.
 python3 collect.py --backend trtllm --model-cases-full
 
-# Raw backend registry run: no collector v2 case plan.
+# Raw backend registry run: no model-centric case plan.
 python3 collect.py --backend trtllm
 
 # Ephemeral subset filter (healing): substring match, repeatable.
@@ -142,7 +146,7 @@ Each backend (trtllm, vllm, sglang) has a **registry** (`registry.py`) that maps
 framework_manifest.yaml — current collector framework versions and images (schema v2)
 framework_manifest.py   — manifest loader/validator/resolver
 op_catalog.py         — table→family identity map (op_backend_catalog.yaml)
-model_cases.py       — collector v2 model/SM case-plan resolver
+model_cases.py       — model-centric model/SM case-plan resolver
 registry.py          — declares which module handles which version range
 version_resolver.py  — routes runtime version → module (packaging.version)
 collect.py/collect_ops — validates __compat__ and fails incompatible ops
--- a/collector/README.md
+++ b/collector/README.md
@@ -173,7 +173,10 @@ Image digests are policy, not decoration: any image reference that names a
 public registry path (contains `/`) must be pinned with `@sha256:<64 hex>`;
 bare internal image names (no `/`, e.g. `deepseek-v4-blackwell`) are exempt.
 When bumping a version pin, re-fetch and update the digest for every image
-variant touched — a stale digest silently keeps pulling the old image.
+variant touched — a stale digest silently keeps pulling the old image. Always
+pin the manifest-list (index) digest, never a platform-child digest —
+`docker buildx imagetools inspect <image>` prints the index digest; a child
+digest pins one architecture and breaks the other platforms.
 
 Use `collector.framework_manifest.resolve_op_runtime(framework, op)` to
 resolve a single op's runtime, and
--- a/collector/framework_manifest.py
+++ b/collector/framework_manifest.py
@@ -194,6 +194,8 @@ def require_collector_runtime(
                 f"{framework} ops resolve to multiple runtime versions ({split}); "
                 "run each version group in its own container"
             )
+        if not by_version:
+            raise KeyError(f"{key} registry has none of the requested ops: {sorted(ops)}")
         resolved[key] = next(iter(by_version.values()))
 
     wideep_key = f"wideep_{normalized}"
@@ -228,11 +230,11 @@ def validate_resolution(
     """Return one error per registry op that does not resolve to a pinned runtime."""
     try:
         manifest = load_manifest(manifest_path)
-    except (TypeError, ValueError) as error:
+    except (TypeError, ValueError, yaml.YAMLError) as error:
         return [f"manifest: {error}"]
     try:
         family_map = load_family_map(catalog_path)
-    except ValueError as error:
+    except (ValueError, yaml.YAMLError) as error:
         return [f"op catalog: {error}"]
 
     errors: list[str] = []
--- a/collector/framework_manifest.py
+++ b/collector/framework_manifest.py
@@ -220,6 +220,34 @@ def require_collector_runtime(
     return runtime
 
 
+def validate_resolution(
+    *,
+    manifest_path: str | Path = MANIFEST_PATH,
+    catalog_path: str | Path = CATALOG_PATH,
+) -> list[str]:
+    """Return one error per registry op that does not resolve to a pinned runtime."""
+    try:
+        manifest = load_manifest(manifest_path)
+    except (TypeError, ValueError) as error:
+        return [f"manifest: {error}"]
+    try:
+        family_map = load_family_map(catalog_path)
+    except ValueError as error:
+        return [f"op catalog: {error}"]
+
+    errors: list[str] = []
+    for framework_key in _REGISTRY_MODULES:
+        if framework_key not in manifest["frameworks"]:
+            errors.append(f"{framework_key}: registry exists but the manifest has no entry")
+            continue
+        for entry in _registry_entries(framework_key):
+            try:
+                _resolve_from(manifest, family_map, framework_key, entry)
+            except (KeyError, LookupError, ValueError) as error:
+                errors.append(f"{framework_key}:{entry.op}: {error}")
+    return errors
+
+
 def validate_manifest(manifest: dict[str, Any]) -> None:
     if manifest.get("schema_version") != 2:
         raise ValueError("collector framework manifest schema_version must be 2")
--- a/collector/framework_manifest.py
+++ b/collector/framework_manifest.py
@@ -152,6 +152,19 @@ def resolve_op_runtime(
     raise KeyError(f"{framework_key} registry has no op {op!r}")
 
 
+def _runtime_identity(runtime: CollectorRuntime) -> tuple[str, tuple[tuple[str, str], ...]]:
+    """Executor identity: version plus the pinned images. Two runtimes with the
+    same package version but different images are different containers. `family`
+    stays out — it is routing metadata, not an executor property."""
+    return runtime.version, tuple(sorted(runtime.images.items()))
+
+
+def _describe_runtime(runtime: CollectorRuntime) -> str:
+    """Version plus images, for errors where the version alone cannot distinguish."""
+    images = ", ".join(f"{variant}={image}" for variant, image in sorted(runtime.images.items()))
+    return f"{runtime.version} [{images}]"
+
+
 def require_collector_runtime(
     framework: str,
     installed_version: str,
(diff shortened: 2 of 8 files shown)


collector/framework_manifest.py
```diff
@@ -45,49 +54,158 @@ def get_collector_runtime(
...
             raise RuntimeError(
-                f"Stock {framework} and WideEP ops require different runtime versions "
-                f"({runtime.version} != {wideep_runtime.version}); run them in separate containers"
+                f"{framework} ops resolve to multiple runtime versions ({split}); "
+                "run each version group in its own container"
             )
-        runtime = wideep_runtime
+        if not by_version:
+            raise KeyError(f"{key} registry has none of the requested ops: {sorted(ops)}")
+        resolved[key] = next(iter(by_version.values()))
```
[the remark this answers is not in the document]
Arsene12358: Fixed in ae8127e3. A missing-ops check now runs before resolution: any requested op absent from the framework's registry raises `KeyError("<key> registry has no op(s): [...]")`, including the mixed valid+typo case (regression test `test_typo_mixed_with_real_op_fails_closed` added). The registry-import half is fixed on the validate_resolution side per the sibling comment.

collector/framework_manifest.py
```diff
@@ -104,25 +222,64 @@ def require_collector_runtime(
...
+    errors: list[str] = []
+    for framework_key in _REGISTRY_MODULES:
+        if framework_key not in manifest["frameworks"]:
+            errors.append(f"{framework_key}: registry exists but the manifest has no entry")
+            continue
+        for entry in _registry_entries(framework_key):
+            try:
+                _resolve_from(manifest, family_map, framework_key, entry)
+            except (KeyError, LookupError, ValueError) as error:
+                errors.append(f"{framework_key}:{entry.op}: {error}")
```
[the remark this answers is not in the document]
Arsene12358: Fixed in ae8127e3. `_registry_entries` is now called inside a try/except ImportError that appends `"<key>: registry import failed: ..."` and continues, so the validator keeps its returns-error-list contract instead of crashing (test `test_validator_reports_registry_import_failure` added, monkeypatching a bogus module path).

collector/README.md
```diff
@@ -152,8 +153,39 @@ wideep/*/registry.py — WideEP-only ops appended when the v2 plan requests them
...
+framework entries rather than nested under a `wideep:` key, so the same
+per-op resolution logic applies uniformly. Each WideEP entry adds
+`base_framework` (the stock framework it forks from), `data_backend` (which
+table namespace its output is written under), and `collector_dir`.
+`collect.py` requires the exact public version for a run and rejects
+mixed-pin selections across ops.
+
+Image digests are policy, not decoration: any image reference that names a
+public registry path (contains `/`) must be pinned with `@sha256:<64 hex>`;
+bare internal image names (no `/`, e.g. `deepseek-v4-blackwell`) are exempt.
```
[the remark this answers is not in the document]
Arsene12358: Fixed in ae8127e3. README, design doc §4, and design doc §12.2 now all state the rule as implemented: syntactic — any image reference containing `/` must carry `@sha256:<64 hex>`, only bare internal names exempt — with "public registry" kept as the in-practice gloss rather than the rule.

docs/perf_database/collector-v3-op-centric-design.md
```diff
@@ -0,0 +1,481 @@
...
+- Revalidating historical framework versions after each upgrade.
+- Forcing all ops to one version during out-of-cycle model onboarding.
+- Unifying op naming across frameworks into one global op (identity is always
+  the pair *(framework, family)*).
+
+## 2. Op identity
+
+One identity chain, single-sourced, three levels:
+
+```
```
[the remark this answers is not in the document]
Arsene12358: Fixed in ae8127e3 — the three bare fences (identity chain, filesystem layout, source-order example) are now ```text.

Arsene12358:
History rewritten in place (no content change — `git diff` old↔new head is empty, trees byte-identical): commit author/committer/sign-off corrected from an unset-git-config machine-local identity to `Yiming Liu <yimingl@nvidia.com>`. The review-fix commit previously referenced as `ae8127e3` is now `d75271e7`.

Arsene12358:
Rebased onto main now that #1345 is merged. This is the integration moment the design set up: with `op_backend_catalog.yaml` present, the fail-closed identity gate (`test_real_manifest_resolves_every_registry_op`) is **armed** — and green. `validate_resolution()` reports 0 errors across all five registries against the merged catalog, per-op resolution now carries family identity (e.g. `attention_context → family='attention'`), and the full collector unit suite passes (430). CI on this push exercises the armed gate for the first time.

docs/perf_database/collector-v3-op-centric-design.md
```diff
@@ -0,0 +1,484 @@
...
+  image: "lmsysorg/sglang:v0.5.14"
+  image_digest: "sha256:..."
+tables:
+  moe_perf:
+    collector_ref: 0b077da5        # repo SHA the collector ran from
+    collector_hash: "sha256:..."   # content hash of the op's module closure
+    case_plan_hash: "sha256:..."   # hash of the resolved case set
+    collected_at: 2026-07-20
+    rows: 12345
+    status: complete               # complete | partial
```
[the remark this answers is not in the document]
Arsene12358: Verified against the implementation on the stacked branches (#1372's `collector/provenance.py` + `collect.py` finalize): `status` is already derived, not self-asserted, and not inferred from provenance-field presence. Finalize marks a table `complete` iff its producing ops' run checkpoints hold zero unresolved failed cases and no `ModuleCollectionFailure` was recorded (`derive_table_status`); anything else is `partial`. `case_plan_hash` attests the case set the run actually attempted, so a filtered subset run remains distinguishable from a full-plan run.

The documentation gap was real, though — §5 showed the schema without stating that contract. Fixed in ed5eb089: two new §5 bullets document the derivation rule and the scope of `status`.

The heavier part of the suggestion — CI re-validating collected coverage against the manifest's required systems/cases before granting `complete` — is deliberately not adopted, for two reasons now stated in the doc: (1) CI cannot re-expand a case plan (case generation consults live device memory; §8 locks `case_plan_hash` as a collection-time attestation, never recomputed in CI); (2) per-sidecar `status` is per-system-dir by construction, so "required systems" is a different layer — the §8 changed-op manifest names the required systems and the §9 evidence gate (AIC-1214) rejects a pin/collector change whose evidence doesn't cover them. That layering is where an incomplete collection run fails to look supported.

tianhaox (commented):
Reviewed this PR (correctness pass over the v2 validation, `resolve_op_runtime` chain, `require_collector_runtime` v1-regression behavior, and cross-file call sites; plus a cleanup pass).

**No findings survived verification — LGTM from my side.**

- The fail-closed paths (family override without catalog → hard error, mixed-pin rejection with the op→version split, `validate_resolution` one-error-per-op) match the PR description as implemented.
- The disclosed limitation (mixed-pin detection by version string only) is real but unreachable with today's manifest — agree with deferring it to PR 4's reviewer.
- Two nits not worth acting on: `op_versions` in the mixed-pin check is built eagerly but only used in the error message, and `_runtime_from_spec` re-derives the workload from `framework_key` — both trivial, fine as is.

collector/op_catalog.py
```diff
@@ -0,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Table→family identity map from the op-backend catalog (Collector V3 §2).
+
+The catalog file is introduced by the op-backend-facts registry (PR #1345).
+Until it lands on main, ``load_family_map`` returns ``None`` and callers must
```
simone-chen: > Until it lands on main

#1345 has been merged to main so this may be no longer valid

collector/README.md
simone-chen: Do we still keep collector v2 notes here in readme? Or maybe add a note:
 
```suggestion
# Collector v2: model-centric cases (Warning: to be deprecated soon)
```

docs/perf_database/collector-v3-op-centric-design.md
```diff
@@ -0,0 +1,499 @@
...
+Every op resolves to exactly one pinned, reproducible runtime (framework default
+plus explicit per-op overrides). Collected data carries enough provenance to know
+which collector, runtime, and case plan produced it. Any change to a pin or
+collector produces an exact changed-operation manifest saying what must be
+recollected; everything else is reused deterministically under authored,
+backward-safe rules. CI and the support-matrix healer decide from manifest +
+provenance alone whether required silicon evidence exists — anything unknown
+fails closed.
+
+### Success criteria (from the Linear project)
```
simone-chen:
```suggestion
### Success criteria
```

docs/perf_database/collector-v3-op-centric-design.md
```diff
@@ -0,0 +1,499 @@
...
+
+1. A collector change identifies exactly which operations and data slices require
+   recollection.
+2. Unchanged compatible operation data remains reusable.
+3. CI rejects missing provenance, missing required silicon data, and unsupported
+   silent fallback.
```
simone-chen, line 26: Do we also want to gate on collector error rate or LOO MAPE? High error rate -> high data point sparsity -> potentially high LOO MAPE.

docs/perf_database/collector-v3-op-centric-design.md
```diff
@@ -0,0 +1,499 @@
...
+have been invalidated by it, so backward fill preserves historical
+reproducibility; forward fill silently answers "how fast is 0.5.14" with
+0.5.15's kernels.
+
+### 6.3 Channel 2 — declared reuse (`reuse.yaml`, same backend, any direction)
+
+When we *know* data is valid for a version we never collected — typically a
+framework upgrade where a family's kernels did not change — we say so
+explicitly. Replaces the presence-only `SHARED_LAYER_REUSE.txt`. A
+declared-reuse version dir contains no parquet of its own, only:
```
simone-chen: >  A declared-reuse version dir contains no parquet of its own

What does this line mean? In the attached example (checked out from the 4-th PR), both 0.22.0 and 0.24.0 have parquet files, but 0.22.0 declares reuse `from_version: 0.24.0`.

Image

docs/perf_database/collector-v3-op-centric-design.md
```diff
@@ -0,0 +1,499 @@
...
+
+- a parquet table without a matching provenance entry;
+- a `reuse.yaml` pointing at nonexistent data or violating §6 rules;
+- any registry op that does not resolve through manifest v2;
+- a table filed under the wrong family;
+- pin/collector/case-plan changes without the evidence the policy demands (§9).
+
+The CI audit is the primary gate; loader strict mode is the backstop.
+
+## 9. Evidence policy (AIC-1219)
```
simone-chen: We should remove all mentioning of linear tickets in the doc if it persists in the repo, but can keep them if we will drop this doc in the future.

docs/perf_database/collector-v3-op-centric-design.md
```diff
@@ -0,0 +1,499 @@
...
+   pin-bump commit emits the quarter's work list mechanically: every
+   *(framework, family)* whose pin moved, with tables and systems. No human
+   enumerates recollection scope.
+2. **The work list splits into "collect" and "declare".** Per changed family,
+   the evidence policy (§9) allows exactly two moves: recollect the family's
+   tables inside the new pinned image (new
+   `data/<system>/<family>/<backend>/<new_version>/` dirs with provenance), or
+   declare reuse (`reuse.yaml` ← previous quarter's version) backed by
+   kernel-identity evidence from the #1345 facts plus a spot benchmark per SM
+   generation. The quarter's GPU bill is proportional to what the framework
```
simone-chen: This leaves some ambiguity in execution - which one takes precedence? Do we always try reusing, and if it doesn't work, switch to collection?
Ideally, `reuse.yaml` is never used if we do a solid quarterly collection campaign. (As I can't think of any case when reuse can be better than a fresh collection ?)

docs/perf_database/collector-v3-op-centric-design.md
simone-chen: Do we allow partial reuse? i.e. borrow data points from an older version in case of incomplete parquet data point coverage. (I assume no)

collector/op_catalog.py
Arsene12358: Right — stale since #1345 merged. Fixed in bb123e8d: the docstring now says the catalog is on main, and keeps documenting the `None`-when-absent contract for stripped/partial checkouts (the fail-closed caller behavior is unchanged).

collector/README.md
Arsene12358: The section stays, but the suggested deprecation warning would be inaccurate — the model-centric case layer is not going away. V3 changes *management* (per-family runtime pins, provenance sidecars, declared reuse), not how collection coverage is declared or healed: the case equation (base grids ∪ model shapes), the healing flags, and the case YAML surface are all unchanged and remain the case layer *under* V3.

What was misleading is the "v2" version label implying supersession. Fixed in bb123e8d: the section is retitled "Model-centric cases and healing runs" and opens by stating the relationship to V3 explicitly (with a pointer to the design doc).

docs/perf_database/collector-v3-op-centric-design.md
Arsene12358: Taken as suggested in bb123e8d (part of the same sweep that removed all Linear references from the doc — see the thread at §9).

docs/perf_database/collector-v3-op-centric-design.md
Arsene12358: Not as a merge gate in v0.11, deliberately — but the observability for it already lands, and it's a good candidate for the health-schema layer:

- **Error rate**: every run's failure records (`errors_<module>.json`, `collection_summary_<backend>.json`, grouped by (model, dtype)) are retained as CI artifacts on the data PR (§9 retention). The collector doctrine intentionally keeps failure as *recorded data with triage thresholds* (~10% investigate, whole-group = stop and fix) rather than a hard gate — a gate on raw error rate would pressure exactly the wrong behaviors (skips, retries, reduced coverage) that the failure-handling rules forbid.
- **LOO MAPE**: agreed on the causal chain (error rate → sparsity → interpolation error). That's an *interpolation-quality* metric over the assembled database, which is the support-matrix health classifier's plane (it consumes `data_provenance` and the sidecars), not the per-run evidence gate's. The v0.11 evidence gate checks collection deltas (`parquet_diff` median > 5% escalates) and completeness (provenance/status) — quality-of-coverage gating on top of that is a natural follow-up there.

If you want it tracked, I'd file it as a follow-up on the health-schema work rather than widen this design — happy to do that.

docs/perf_database/collector-v3-op-centric-design.md
Arsene12358: Good catch — the sentence described only the typical case and the l40s dirs you found are the documented exception. Two things going on in your screenshot:

1. **Typical declared-reuse dir**: no parquet, only `reuse.yaml` (a version we never collected, borrowing wholesale from one we did).
2. **Self-overlap (your l40s example)**: a dir holding BOTH its own parquet and a declaration for the same table. Per the §6.1 merge substrate, the dir's own rows always win the shapes they cover — the declaration only forward-fills shapes its own data is missing. These l40s entries are mechanical migrations of the legacy `SHARED_LAYER_REUSE.txt` semantics (the old marker did exactly this implicitly); each carries a `reason` field disclosing that they're missing-shape forward-fill only, not kernel-identity-reviewed.

The doc now says this explicitly — §6.3 gained a self-overlap paragraph in bb123e8d ("In the typical case… A dir may also hold BOTH…").

Arsene12358: Yes — partial (per-shape) reuse is not just allowed, it's the substrate everything else builds on (§6.1): the loader merges by shape key across the admitted sources, first source wins, later sources only fill gaps. "Borrow data points from an older version to cover incomplete coverage" is exactly channel 1 (implicit backward fill, §6.2) — free, always on, and safe because a donor can never shadow the primary's own rows.

One distinction worth keeping separate: dirs *marked* partial (`collection_meta.yaml` with a `status: partial` table, or the legacy `INCOMPLETE.txt`) are excluded from acting as sources at all — they can't be primaries or donors. So: missing shapes get filled per-shape from healthy older versions; explicitly-partial datasets don't participate.

docs/perf_database/collector-v3-op-centric-design.md
Arsene12358: Agreed the doc left this ambiguous — now stated explicitly in §11.2 (bb123e8d). The rule:

- **Precedence is decided at plan time, not runtime.** There is no "try reuse, fall back to collection" — the quarter's data PR ships one move or the other per family, and the evidence gate checks whichever was chosen.
- **Default = recollect.** Declaring reuse is the exception, permitted only when the kernel-identity facts (#1345) show the family's kernels did not change across the pin bump AND the per-generation spot benchmark confirms it.
- On your last point — you're right that reuse is never *better* than fresh collection on pure data quality. It exists for exactly one reason: when kernels are unchanged, recollection produces statistically identical rows at fleet-wide GPU cost. The spot benchmark is the check that the "identical" claim holds. So in a quarter where everything changed, `reuse.yaml` is indeed never written — the ideal you describe is the design's default path, and reuse is the priced escape hatch, not a shortcut.

docs/perf_database/collector-v3-op-centric-design.md
Arsene12358: Agreed — the doc persists in the repo (it's the durable design contract), and since the repo is public, internal Linear links don't resolve for most readers anyway. Done in bb123e8d (+ two small follow-ups on the stacked branches for references their amendments had introduced): all Linear ticket IDs and links are gone from the doc — replaced with self-contained wording — across all four PRs. GitHub PR references (#1345, #1322) stay, since they're public and load-bearing. The internal Linear mirror of this doc keeps its ticket links, so tracking context isn't lost — the repo copy is now the self-contained one.

docs/perf_database/collector-v3-op-centric-design.md
Arsene12358: Follow-up: the precedence paragraph was reworded as an explicit three-step decision procedure in 65ad9427 (it previously read as justification rather than instruction):

1. **Default: recollect** — always valid, needs no extra justification.
2. **Reuse only if BOTH hold**: kernel-identity facts show the family's kernels unchanged across the pin bump, AND the per-generation spot benchmark confirms it within threshold.
3. **Either check fails → recollect.** Plan-time decision; the data PR ships one move or the other, the evidence gate verifies whichever shipped, and the loader never switches between them at query time.

jasonqinzhou (approved):

collector/framework_manifest.py
```diff
@@ -45,49 +54,162 @@ def get_collector_runtime(
...
+        entries = [e for e in _registry_entries(key) if not ops or e.op in ops]
+        if ops:
+            missing = ops - {e.op for e in entries}
+            if missing:
+                raise KeyError(f"{key} registry has no op(s): {sorted(missing)}")
+        by_version: dict[str, CollectorRuntime] = {}
+        op_versions: dict[str, str] = {}
+        for entry in entries:
+            runtime = _resolve_from(manifest, family_map, key, entry)
+            by_version.setdefault(runtime.version, runtime)
```
jasonqinzhou: **Please key this by the full executable runtime identity, not only `version`.**

A family override can intentionally keep the same package version while selecting a different image/digest. In that case `setdefault(runtime.version, runtime)` collapses both runtimes and one image wins by registry order, so one container can be accepted for operations that require different images.

Minimal fix:

- define the identity as `(runtime.version, tuple(sorted(runtime.images.items())))`;
- use that identity for this per-framework grouping and for the stock/WideEP comparison below;
- keep `family` out of the identity, because it is routing metadata rather than an executor property;
- add a regression where the default and a family override share a version but have different `images.default` values; `require_collector_runtime()` should reject the mixed request and report the operation-to-runtime split.

No committed manifest entry triggers this today, but leaving it version-only makes the new per-family override surface unsafe as soon as such an override is added.

jasonqinzhou (commented):
One actionable runtime-identity issue remains before the per-family override surface is safe to use.

collector/framework_manifest.py
Arsene12358: Confirmed and fixed in 8822629c, following your recipe exactly: `_runtime_identity() = (version, tuple(sorted(images.items())))` is now the key at both sites — the per-framework grouping and the stock/WideEP comparison — with `family` excluded as routing metadata. The versions-differ error messages are preserved byte-for-byte; a version-tie with differing images now rejects with a distinguishable message reporting the operation-to-runtime split including the image values. Both regression shapes from your comment (default vs family override sharing a version with different `images.default`; the stock/WideEP analogue) were red pre-fix — `DID NOT RAISE`, one image silently winning by registry order — and now reject.

Arsene12358:
Doing a final support matrix check. If no further comments, will start merging today

PR #1372, opened 2026-07-15
feat(collector): committed provenance, authored reuse, changed-op manifest, fail-closed audit (AIC-1502)

Arsene12358:
PR 3 of 4 of the **Collector V3: op-centric collector management** project (Linear: [AIC-1502](https://linear.app/nvidia/issue/AIC-1502/collector-v3-c-provenance-sidecar-collection-metayaml-authored)), implementing design §5, §6.3, and §8. **Stacked on #1371** (base `yimingl/aic-1501-family-layout-reorg`); retargets when the stack merges. This PR **declares an approved cross-module contract change** per `.claude/rules/collector/layer_permissions.md`: collector finalize + aic-core loader + tools + one new workflow, coordinated.

## What this does

Every table in the perf tree now carries committed provenance, reuse is authored instead of implied, and change tracking + enforcement are wired into CI:

- **`collection_meta.yaml`** (design §5): collector finalize writes it flat beside staged parquet — producing runtime (framework/version/image/digest from the manifest), per-table `collector_ref` (repo SHA), `collector_hash` (content hash over an authored closure: the op's module + a fixed shared core + its case files, `collector/hash_closures.yaml`, 34/34 registry modules covered by a two-direction CI-riding gate), `case_plan_hash` (attested from the run's checkpoint case ids), rows, `status: complete|partial` (derived from checkpoint failures). The external promotion step carries the sidecar with the parquet; the CI audit makes forgetting it impossible.
- **`reuse.yaml`** (design §6.3): replaces `SHARED_LAYER_REUSE.txt`. The tree conversion (via the reviewed `tools/perf_database/migrate_markers.py`, plan/execute/verify, idempotent): 255 markers → per-table declarations with **derived effective donors** (newest sibling under the legacy ordering — behavior-preserving; every `reason` field discloses the mechanical derivation), 27 comm-family markers dropped **without** declarations (design §6.5 rule 5 — comm never participates in sibling reuse), 7 `INCOMPLETE.txt` → `status: partial` sidecars, and **527 pre-V3 dirs backfilled with `provenance: legacy` sidecars** (design §5 legacy tier, amended this PR) so the sidecar-coverage gate is total over the tree. Zero legacy markers remain. Discovery parity held through every data commit: 10 systems / 89 (backend, version) pairs, byte-identical.
- **`tools/perf_database/changed_ops.py`** (design §8): diffs two revisions per (framework, family) on three GPU-free signals — manifest pin, closure hash (revision-pure: even `SHARED_CORE` is ast-read from each revision), case-input files — emitting the locked manifest schema consumed by the evidence resolver (PR 4), the CI gate (AIC-1214), and the support-matrix healer. Fail-closed: unknown ops, `VersionRoute` shapes (unsupported → raise), and pre-V3 baselines (dedicated exit code 3, mapped to a neutral skip in CI) never pass silently.
- **`tools/perf_database/audit_collector_data.py` + `.github/workflows/collector-audit.yml`**: six fail-closed rules on every PR touching data/collector/tools — sidecar coverage, reuse validity (donors must exist with real data, same backend), comm exclusion, family placement, manifest resolution, no legacy markers. All six pass on the real tree as committed. Evidence *gating* against the changed-ops manifest remains AIC-1214 scope (the workflow uploads it as an artifact).
- **Loader + 7 consumers** read the structured files yaml-first with legacy `.txt` fallback for one transition window (once-per-tree deprecation warnings), including a 4-way predicate-parity test pinning the one deliberate divergence (hot-path leniency in `operations/base.py`).

## ⚠️ Maintainer sign-off requested: runtime gate extension

Populating provenance required capturing the resolved `CollectorRuntime` in `collect_vllm`/`collect_trtllm` — which means **`require_collector_runtime`'s exact-version fail-closed gate now applies to vLLM and TRT-LLM runs (previously sglang-only), including the vLLM XPU path which was fully ungated before**. A container whose installed framework version differs from the manifest pin now hard-fails with instructions to use the pinned image. This is design §4's uniform enforcement and is zero-behavior-change against the committed manifest, but it is a real new failure path for out-of-pin containers: please confirm explicitly in review.

## Migration notes for reviewers

- **12 `reuse.yaml` files (17 entries), all under `l40s`**, declare tables that also exist as parquet in the declaring dir (identification predicate: self-overlap between declared table and local data). Their original marker text asserted channel-3 (cross-framework) intent; the conversion is by marker type, per the migration algorithm. Inert in this PR (declarations are not yet consumed for ordering) and safe under §6.1 (primary always wins), but **PR 4 must re-derive or drop these before its ordering flip treats declarations as reviewed evidence** — they deviate from §6.3's "a declared-reuse dir contains no parquet of its own".
- The 527-dir `status: complete` backfill makes explicit what the absence of `INCOMPLETE.txt` already meant implicitly; every backfilled sidecar is honestly tagged `provenance: legacy` (no fabricated hashes). Strict mode (PR 4) treats legacy as warn-not-fail for one release.
- Known gaps, documented in-code: XPU registries are invisible to provenance/change tracking ([AIC-1509](https://linear.app/nvidia/issue/AIC-1509/collector-v3-follow-up-xpu-registry-tracking-registry-xpu-invisible-to) filed); `VersionRoute` registry entries unsupported by the manifest tool (raises).

## Test plan

`uv run pytest tests/unit -m unit -n 2` green (collector suite 469, tools 221, sdk incl. 22 discovery + 8 parity tests; the finalize-glue writer has dedicated coverage incl. the legacy-tier merge guard) · real-tree audit: all six rules OK, exit 0 · `changed_ops --base HEAD --head HEAD`: 30 pairs unchanged; `--base <pre-V3>`: exit 3 with the documented message · discovery parity 10/89 unchanged across all four data commits · every fix wave re-reviewed.


Arsene12358:
**Merge protocol (stacked PR):** please do not merge this PR until its base PR has merged to main and GitHub has auto-retargeted this one to `main` (the repo's delete-branch-on-merge triggers that automatically). Merging it as-is would fold this PR into its base PR's branch instead of main. After each upstream merge I'll rebase this branch onto main (squash-merge cascade) — the diff you review is unaffected. Merge order: #1370 → #1371 → #1372.

Arsene12358:
CI fix pushed (branch rebased onto the updated #1371 base, new head `2b73e491`): in addition to the base-branch fixes, this branch adds `src/aiconfigurator_core/systems/data/**/*.yaml` to the core wheel's maturin includes — the `reuse.yaml`/`collection_meta.yaml` sidecars this PR introduces were not being packaged, which would have made installed-wheel behavior silently diverge from repo behavior. The wheel-layer check now probes a sidecar and a reuse declaration explicitly.

Arsene12358:
CI fixes pushed (head `cf0dc28e`, rebased onto the updated #1371 branch — cascade per the stack protocol):

- **Validate PR title**: retitled to single-scope `feat(collector)` (the title-check action cannot parse comma-separated scopes; see #1371 comment for details).
- **copyright-checks**: the check covers `.yaml` and flagged exactly the 762 sidecars this PR introduces (534 `collection_meta.yaml` + 228 `reuse.yaml`). All 762 now carry the repo-standard SPDX header (YAML comments are invisible to `yaml.safe_load`, so loader/audit parsing is byte-for-byte unchanged — real-tree audit still 6/6 OK), and both writers emit the header going forward so quarterly collections and backfill re-runs stay compliant: `collector/provenance.py::write_collection_meta` and the `tools/perf_database/migrate_markers.py` renderers. Renderer tests pin the header with `startswith` assertions (the yaml round-trips are comment-blind).
- **Build and Test (unit)**: inherits the git-less-container test fix from #1371.

Arsene12358:
Round 2 (head `b0eb31f7`, cascade rebase onto #1371): the remaining unit failure was `test_changed_ops.py` erroring on the git-less test container (`FileNotFoundError: 'git'` — the tool diffs real git revisions, so its tests legitimately need a git binary; `test_migrate_markers.py`'s execute-path tests were next in line). Fixed at the root on #1371: the test Docker stage now installs git. No changes to this PR's own commits beyond the rebase.

Arsene12358:
Round 3 (head `18e70d29`): with git in the test image, 2211 tests passed and one remained — `test_real_repo_base_equals_head_is_all_unchanged`, the real-repo integration smoke test, fails in the container because `/workspace` is a source COPY without `.git` (`git rev-parse HEAD` → exit 128). That one test is now gated on the checkout actually being a git worktree; the 27 synthetic-repo changed_ops tests all run in CI. Verified locally both ways (28 passed in a real checkout; the probe returns skip in a git-less tree).

collector/collect.py
```diff
@@ -1320,6 +1351,139 @@ def _all_op_names() -> list[str]:
...
+        module = module_by_table.get(table)
+        if not full_names or module is None:
+            logger.warning(f"collection_meta: {table} has no registry mapping this run; skipping its provenance entry")
+            continue
+
+        case_ids: set[str] = set()
+        unresolved_failed = 0
+        for full_name in full_names:
+            checkpoint_path = checkpoint_root / f"{full_name}.json"
+            if not checkpoint_path.exists():
```
tianhaox, line 1433: **Confirmed fail-closed gap — worth fixing in this PR.** If *all* checkpoint JSONs for a table's ops are missing/unreadable, every iteration hits this `continue`: `case_ids` stays empty and `unresolved_failed` stays 0. The table is still added to `tables` with `case_plan_hash` computed over an empty list (line ~1450), and `derive_table_status` (provenance.py ~990) only checks `unresolved_failed_count > 0 or had_module_failure` → the table is written as `status: complete` with zero attestation.

Reachable in practice: crash between parquet write and finalize, checkpoint dir cleaned between runs, or `--resume` with a different checkpoint dir. And nothing downstream catches it — none of the six audit rules inspect `case_plan_hash` content, so an empty attestation passes CI green; the 5 writer unit tests all write checkpoints, so this branch is untested.

Suggestion: when `case_ids` is empty for a table that has parquet, either derive `status: partial` or refuse to write the sidecar (fail-closed), and add a missing-checkpoint test.

tools/perf_database/changed_ops.py
```diff
@@ -0,0 +1,655 @@
...
+    Raises `PreCollectorV3BaselineError` if `base_rev` predates Collector V3
+    provenance metadata (see module docstring "Exit codes") — the CLI (`main`)
+    maps this to exit code 3.
+    """
+    base_sha = _resolve_rev(repo_root, base_rev)
+    head_sha = _resolve_rev(repo_root, head_rev)
+    _check_v3_baseline(repo_root, base_sha)
+
+    base_ctx = _load_revision_context(repo_root, base_sha)
+    head_ctx = _load_revision_context(repo_root, head_sha)
```
tianhaox, line 576: **Minor robustness:** `_check_v3_baseline` runs for `base_sha` only; `_load_revision_context(repo_root, head_sha)` then does `git show head:<v3-file>` with `check=True`, so a pre-V3 HEAD dies with a raw `CalledProcessError` traceback instead of the documented exit code 3. Unreachable in CI (the PR branch has the V3 files by construction) but trivially reachable via CLI misuse (`--head <old-rev>`). One extra `_check_v3_baseline(repo_root, head_sha)` here closes it.

tianhaox (commented):
Reviewed the finalize sidecar writer, `changed_ops`, the audit rules, migration tools, and the loader fallback paths. **One confirmed fail-closed gap I'd fix before merge (inline on `collect.py`), plus two minor notes (inline).**

Also checked, all clean:
- The runtime-gate extension you flagged for sign-off (vllm/trtllm/XPU now gated): verified no behavior change for the existing sglang path; the new failure mode only fires for out-of-pin containers, as described.
- The six audit rules pass/fail predicates traced against the sidecar schema the finalize writer emits — field-for-field consistent.
- `migrate_markers.py`'s duplicated version-sort logic was flagged and then refuted as non-actionable: one-shot tool, output already committed, and the effective newest-first ordering is behaviorally identical to the loader's (verified on mixed parseable/unparseable version sets).

aic-core/src/aiconfigurator_core/sdk/operations/base.py
```diff
@@ -73,13 +75,37 @@ def _resolve_perf_data_path(perf_file: str) -> str:
 _KNOWN_BACKEND_DIRS = frozenset({"trtllm", "sglang", "vllm", "nccl", "oneccl"})
 
 
+def _version_dir_is_partial(version_dir: str) -> bool:
```
tianhaox, line 88: **Duplication with divergent error behavior (non-blocking, but worth a follow-up):** this partial-detection logic now exists in three places, and they already disagree on error handling — this copy swallows all exceptions and returns `False` (lenient), while `prediction_regression_gate/grid.py` and `sanity_check/create_charts.py` re-raise `YAMLError` as `ValueError` (strict, matching the loader). The circular-import constraint vs `perf_database` is real, but a dependency-free leaf module (e.g. `sdk/perf_markers.py`) would let all consumers share one implementation. When the sidecar schema evolves, the lenient/strict split is where a silent divergence bug will land.

collector/collect.py
Arsene12358: Confirmed and fixed in 665c4091, taking the fail-closed option: `_write_collector_provenance` now raises `RuntimeError` — naming the table, its producing ops, and the checkpoint root — when a finalized parquet table has zero checkpoint evidence (`case_ids` empty after the per-op loop). Writing `status: partial` would have claimed knowledge we don't have; a parquet table with no attestation is a collection-infra failure a human must see, per the collector's run-it-or-raise doctrine.

Two verification notes: (a) `ResumeCheckpoint.flush` is dirty-gated, so a checkpoint file with zero done+failed entries is mechanically unreachable — the one empty-check covers both the missing-file and unreadable-file paths (both from your list red-tested: `DID NOT RAISE` pre-fix, sidecar written `complete`); (b) partial op coverage is preserved — a multi-op table where some ops have checkpoints still finalizes on the real evidence, with the missing-op case noted in the PR record rather than silently changed.

tools/perf_database/changed_ops.py
Arsene12358: Fixed in 83abffc4 with exactly the one-liner: `_check_v3_baseline(repo_root, head_sha, role="head")` now runs beside the base check, with the message parameterized so head misuse names head. Red-tested: `--head <pre-V3 rev>` previously escaped as a raw `CalledProcessError` from `git show`; now exits 3 with the documented message. No CI-visible change (the PR head always carries the V3 files by construction).

aic-core/src/aiconfigurator_core/sdk/operations/base.py
Arsene12358: Agreed on the facts, and split the response in two:

1. **Done now (83abffc4, docs-only):** the lenient/strict split is partly intentional, so before consolidating anything it needed to be *stated*: `_version_dir_is_partial`'s docstring now carries the contract — resolver-side detection is deliberately lenient (it only skips candidates; a malformed sidecar must not make path resolution throw) because the admission layer in `perf_database` owns strictness (`_version_dir_state` → `_load_collection_meta_yaml`'s ValueError, strict-raises/non-strict-warns) — plus the full copy list, with back-pointers added at the grid.py and create_charts.py copies (same pattern as the `KNOWN_BACKEND_DIRS` canonical note).
2. **Deferred: the leaf-module consolidation.** Worth noting its ceiling: a shared `sdk/perf_markers.py` can only unify the two aic-core copies — grid.py and create_charts.py are standalone tools that can't import aic-core (the same constraint as the backend-dir set), so they stay textual copies either way, now cross-referenced. Given code freeze on 7/28 and this stack in final review, refactoring the admission-path helpers now is poor risk/reward; happy to file it as a tracked follow-up so it lands right after the release.

Arsene12358: One correction to my own reply above: "grid.py and create_charts.py can't import aic-core" is only accurate for grid.py (zero SDK imports, deliberately — see below). create_charts.py already imports the public SDK API (`get_database`), so its copy exists not because the import is impossible but because the predicate is a *private* discovery internal with no public surface — sharing it would mean promoting `_version_dir_is_partial`-style internals to public API. grid.py's copy is the deliberate one either way: the prediction gate enumerates coverage independently of the loader it validates, so a discovery bug can't corrupt the gate's expectations identically and mask itself. The follow-up scope is therefore: leaf module for the two aic-core copies (also dissolves the circular-import constraint), a public-vs-copy decision for create_charts, grid.py stays independent by design.

jasonqinzhou (approved):

jasonqinzhou:
**Cross-stack regression: #1371's documented migration verifier rejects the healthy V3 tree after this PR adds structured sidecars.**

At this PR's head (and at #1374), the following command fails:

```bash
python tools/perf_database/migrate_family_layout.py   --data-root aic-core/src/aiconfigurator_core/systems/data   --verify
```

`verify_tree()` walks every file in each version directory and treats `collection_meta.yaml` and `reuse.yaml` as performance tables, producing `table not in catalog` failures. The six-rule collector audit passes, so this is specifically a regression in the shipped verifier contract.

Minimal fix in `tools/perf_database/migrate_family_layout.py`:

1. define the structured sidecar names (`collection_meta.yaml`, `reuse.yaml`);
2. skip them before the `family_map` lookup and table-count increment in `verify_tree()`;
3. add a verifier fixture containing one parquet file plus both sidecars and assert that only the parquet contributes to the count manifest.

Please keep the legacy `SHARED_LAYER_REUSE.txt` / `INCOMPLETE.txt` handling unchanged.

Arsene12358:
Reproduced and fixed in b34d4e46, per your recipe: `verify_tree()` now skips the structured sidecar names (`collection_meta.yaml`, `reuse.yaml`, defined as a constant) before the family_map lookup and the count-manifest increment; legacy `SHARED_LAYER_REUSE.txt`/`INCOMPLETE.txt` handling is untouched. The fixture test (one parquet + both sidecars in a family-layout version dir) was red pre-fix with the exact "table not in catalog" failure you hit, and the real-tree `--verify` is green again at the stacked heads. `scan_legacy_tree` needed no change — it never iterates family-layout version dirs, and legacy dirs stay fail-closed.

Note on why our earlier verification missed this: the wave that hardened this tool ran `--verify` on the #1371 head, where the sidecars don't exist yet — the regression only manifests one PR up the stack. Cross-stack tool-contract checks now run at the stack top.

PR #1368, opened 2026-07-15
fix(webapp): use monospace font for debugging output textboxes

natoscott:
#### Overview:

The webapp debugging textboxes display CLI output (PrettyTable tables, Pareto frontier charts) that assumes a fixed-width font. Without monospace styling, columns are misaligned and ASCII charts are unreadable.

#### Details:

- Added `.debug-output textarea { font-family: monospace !important; }` CSS rule to `src/aiconfigurator/webapp/main.py`
- Applied `elem_classes=["debug-output"]` to all 5 existing debugging textboxes across the webapp tabs (static, agg, agg_pareto, disagg_pareto, disagg_pd_ratio)

#### Where should the reviewer start?

- `src/aiconfigurator/webapp/main.py` — the CSS rule (4 lines)
- Any of the 5 tab component files — identical one-line change each

#### Related Issues:

- None

## Summary by CodeRabbit

* **Style**
  * Improved the presentation of debugging output across estimation and configuration tabs.
  * Debugging text areas now use consistent monospace formatting for improved readability.

PR #1373, opened 2026-07-16
fix(webapp): move CSS to launch() for Gradio 6 and disable analytics

natoscott:
#### Overview:

Fixes Gradio 6.0 deprecation warning and suppresses noisy telemetry requests on webapp startup.

#### Details:

- Moved the `css` parameter from `gr.Blocks()` constructor to `demo.launch()`, as required by Gradio 6.0. Eliminates the `UserWarning: The parameters have been moved from the Blocks constructor to the launch() method` warning on every startup.
- Disabled Gradio telemetry by default via `os.environ.setdefault("GRADIO_ANALYTICS_ENABLED", "False")`. AIC is a self-hosted tool and the HuggingFace/Gradio analytics HTTP requests add startup noise without user benefit. Uses `setdefault` so users can re-enable with `GRADIO_ANALYTICS_ENABLED=True` if desired.

- `src/aiconfigurator/webapp/main.py` — single file, ~10 lines changed

#### Related Issues:

- None

## Summary by CodeRabbit

* **Enhancements**
  * Disabled Gradio analytics during application startup.
  * Improved consistency of the web app’s custom styling by ensuring the stylesheet is correctly applied when the server launches.

### DIFF
--- a/src/aiconfigurator/webapp/main.py
+++ b/src/aiconfigurator/webapp/main.py
@@ -51,6 +51,10 @@ def main(args):
     """
     Main function for the WebApp.
     """
+    import os
+
+    os.environ.setdefault("GRADIO_ANALYTICS_ENABLED", "False")
+
     try:
         import gradio as gr
     except ModuleNotFoundError as e:
@@ -97,9 +101,7 @@ def main(args):
     if not perf_database.get_supported_databases():
         raise SystemExit(perf_database.build_no_databases_message())
 
-    with gr.Blocks(
-        title="AIConfigurator for Disaggregated Serving Deployment",
-        css="""
+    _css = """
         .config-column {
             border-right: 5px solid #e0e0e0;
             padding-right: 20px;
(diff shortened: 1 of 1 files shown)


Arsene12358 (approved):
Verified against the pinned gradio 6.8.0 wheel (pyproject pins `gradio==6.8.0`): `Blocks(css=...)` pops the kwarg and emits exactly the quoted UserWarning (blocks.py:1094-1103), and `launch()` falls back to `self._deprecated_css` (blocks.py:2591), so the move is behavior-preserving — not just warning-silencing. The analytics claim also checks out: analytics.py:46 reads `GRADIO_ANALYTICS_ENABLED` with default "True" at Blocks construction time, and `setdefault` before `import gradio` preserves the user's ability to opt back in.

One nit, non-blocking: `import os` could sit at module top with the other imports; it only needs to precede `import gradio`, which is also inside `main()`.

Coordination note: this PR, #1368, and #1369 all touch the same region of `webapp/main.py` (the css block / launch call). Suggest merging this one first since it defines where CSS lives going forward.

tianhaox:
Though this is merged, we have a plan to deprecate current webapp design. Is current webapp meeting your demands? We can redesign a webapp v2 to simplify the journey. Current design is too complex for normal users.

PR #1375, opened 2026-07-16
feat: harden aic-core wheel and crate API contracts

jasonqinzhou:
## Summary

- define a small lazy `aiconfigurator_core.sdk` facade for the supported Python API
- ship `py.typed` plus a native-extension stub and verify both in built/installed wheels
- add a typed Rust `AicEngineBuilder` as the preferred API while preserving `build_aic_engine(...)` compatibility for 0.10
- route the builder, compatibility function, and internal `EngineConfig` path through one private `EngineBuildRequest` and compile implementation
- compile a separate workspace crate against only the public Rust exports
- enforce wheel/crate version and wire-schema contracts
- add a dedicated CI job for Python contracts and Rust tests with embedding both disabled and fully enforced
- document the wheel/crate API and compatibility policy
- merge the current `main` branch and resolve the speculative-decoding overlap using the current scalar `nextn_accepted` interface

## Compatibility and follow-up

`build_aic_engine(...)` remains available in this PR so the 0.10 release does not break the current Dynamo Mocker integration. `AicEngineBuilder` is the preferred API. Draft PR #1402 tracks the intentional removal of the flat function for 0.11, after Dynamo Mocker migrates to the builder.

## Validation

- Python facade, import identity, version/schema, and package-layer contract tests: 66 passed
- `cargo test --workspace --no-default-features`: 192 core + 3 external-consumer tests passed
- enforced `cargo test --workspace --all-features`: 197 core + embedded builder, compatibility, memory, and PyO3 tests + 3 external-consumer tests passed
- full unit suite: 2,154 passed, 10 skipped
- full end-to-end suite: 52 passed, 1 expected xfail
- embedding tests rerun with `AIC_REQUIRE_EMBEDDED_ROUND_TRIP=1` so skips are impossible
- real release wheels verified: 715 upper + 1,159 core payload files with no overlap
- fresh core-only wheel install verified, including a real MiniMax engine compile and positive prefill/decode latencies
- `cargo package -p aiconfigurator-core --allow-dirty` packaged and verified the publishable crate
- Ruff lint, formatting for all tracked Python files, Rustfmt on changed Rust sources, workflow YAML parse, and diff whitespace checks pass

## Summary by CodeRabbit

* **New Features**
  * Introduced `AicEngineBuilder` as the preferred Rust engine construction API and re-exported it from the Rust crate.
  * Added the stable `aiconfigurator_core.sdk` Python facade with lazy-loaded exports.
  * Added Python typing stubs and type markers for improved editor/type-checking support.
* **Documentation**
  * Expanded API/compatibility/versioning and embedded-Python guidance; updated examples to use the builder.
* **Tests**
  * Added cross-language public API/version/schema contract checks and strengthened installed-wheel validation (including embedded scenarios).
* **CI**
  * Added a dedicated workflow job to validate core public API contracts in multiple build modes.
* **Bug Fixes**
  * Updated a GPU bandwidth expectation in tests.


aic-core/rust/aiconfigurator-core/src/common/system_spec.rs
```diff
@@ -161,7 +161,7 @@ mod tests {
         let spec = SystemSpec::load(&systems_root().join("b200_sxm.yaml"))
             .expect("b200_sxm.yaml parse must succeed");
         assert_eq!(spec.data_dir, PathBuf::from("data/b200_sxm"));
-        assert_eq!(spec.gpu.mem_bw, 8_000_000_000_000.0);
+        assert_eq!(spec.gpu.mem_bw, 7_700_000_000_000.0);
```
simone-chen, line 164: Is this change needed by the PR? Seems unrelated?

aic-core/rust/aiconfigurator-core/src/operators/attention.rs
```diff
@@ -343,10 +343,13 @@ mod tests {
     #[test]
     fn mem_op_latency_uses_empirical_formula() {
         let db = b200_vllm_db();
-        // b200_sxm.yaml: mem_bw=8e12, scaling=0.8, constant=3e-6.
-        // For 1MB: latency = (1e6 / (8e12 * 0.8) + 3e-6) * 1000 = 0.156... + 0.003 = 0.159ms.
+        // Validate against the loaded system spec so fixture calibration
+        // updates do not leave a stale duplicate constant in this test.
```
simone-chen: Let's remove this block comment as the code is self-explanatory

aic-core/rust/aiconfigurator-core/src/py.rs
```diff
@@ -288,6 +288,210 @@ fn engine_spec_bincode_from_json(spec_json: &str) -> PyResult<Vec<u8>> {
...
 }
 
+/// Ergonomic builder for the Rust -> Python -> Rust compiled-engine entry point.
+///
+/// Only the model, system, and backend are required. Parallelism defaults to
+/// one, speculative decoding defaults to disabled, and all other options defer
+/// to Python's `compile_engine` defaults. [`build_aic_engine`] remains available
+/// for source compatibility with existing callers.
+#[derive(Clone, Debug)]
+pub struct AicEngineBuilder {
```
simone-chen: AicEngineBuilder mostly re-wraps the flat build_aic_engine(...) kwargs. Would a AicEngineBuildConfig struct + build_aic_engine(config) be easier to understand? Would like to hear @tedzhouhk's preference as well.

// Current (Option A)
```
AicEngineBuilder::new("Qwen/Qwen3-32B", "h200_sxm", BackendKind::Vllm)
    .tp_size(8)
    .attention_dp_size(4)
    .build()?;
```
// Proposed (Option B)
```
build_aic_engine(AicEngineBuildConfig {
    model_path: "Qwen/Qwen3-32B".into(),
    system: "h200_sxm".into(),
    backend: BackendKind::Vllm,
    tp_size: 8,
    attention_dp_size: 4,
    ..Default::default()
})?;
```

If we keep Option A, we should remove the pub function `build_aic_engine` and just keep the AicEngineBuilder::build instead:
```
impl AicEngineBuilder {
    pub fn build(self) -> Result<AicEngine, AicError> {
        let engine = compile_engine_from_flat(
            &self.model_path,
            &self.system,
            self.backend.as_str(),
            self.backend_version.as_deref(),
            self.tp_size,
            // ... rest from self
        )?;
        Ok(AicEngine::new(engine))
    }
}
```

aic-core/API.md
```diff
@@ -0,0 +1,132 @@
...
+
+`build_aic_engine` remains supported for existing callers that use the flat
+argument list. Both construction paths enter Python once to compile an engine
+specification. Calls on the returned `AicEngine` are pure Rust and do not
+re-enter Python.
+
+The supported root-level Rust surface is grouped as follows:
+
+- compiled engine: `AicEngineBuilder`, `AicEngine`, `build_aic_engine`,
+  `AicError`;
```
simone-chen: Might need to change here too - some duplication can be removed

aic-core/src/aiconfigurator_core/py.typed
simone-chen, line 1: Is it really needed? 👀

aic-core/rust/tests/public-api/Cargo.toml
simone-chen, line 1: I might be missing something, but why do we put `public-api-tests` under root instead of keeping it under tests?

aic-core/rust/aiconfigurator-core/src/operators/attention.rs
jasonqinzhou: Removed the redundant block comment in f9af38c6.

aic-core/rust/aiconfigurator-core/src/common/system_spec.rs
jasonqinzhou: Yes—this PR adds an enforced workspace Rust-test job, which exposed this stale test assertion. `aic-core/src/aiconfigurator_core/systems/b200_sxm.yaml` already declares `mem_bw: 7700000000000`, while the test still expected 8.0 TB/s. This only synchronizes the test with the existing fixture; it does not change the production system spec.

aic-core/src/aiconfigurator_core/py.typed
jasonqinzhou: Yes. `py.typed` is the PEP 561 marker that tells type checkers the installed package provides inline typing information; the marker is intentionally empty. This wheel also ships `_aiconfigurator_core.pyi`, and the release/install contract tests verify that both typing artifacts are present, so I’m keeping it.

aic-core/rust/tests/public-api/Cargo.toml
jasonqinzhou: Moved the standalone consumer crate under `aic-core/rust/tests/public-api/` in f9af38c6. It remains a separate workspace member because the contract test is meant to compile through a normal dependency on `aiconfigurator-core`, using only its public exports.

aic-core/rust/aiconfigurator-core/src/py.rs
jasonqinzhou: Addressed in 717c4b85 and clarified in 38e89bb4. `AicEngineBuilder` remains the canonical API because it models the three required values without inventing invalid `Default` values for model, system, or backend. The builder, the existing flat `build_aic_engine(...)` compatibility function, and the internal `EngineConfig` path all normalize into one private `EngineBuildRequest` and one Python compile implementation. The flat function stays source-compatible through the 0.10 release because Dynamo Mocker currently imports it. The 0.11 follow-up is now draft PR #1402, which removes the flat symbol entirely after Dynamo Mocker migrates to `AicEngineBuilder`; it does not merely deprecate it. The external consumer crate and enforced embedded round-trip continue to verify the 0.10 compatibility path here.

aic-core/API.md
jasonqinzhou: Addressed in 717c4b85 and clarified in 38e89bb4. `API.md` now states the Rust construction contract once: `AicEngineBuilder` is preferred, `build_aic_engine(...)` is retained only as the 0.10 compatibility adapter, and both share one private request/compile path. The duplicated setup and full usage example were removed from the contract document and linked to the crate README. Draft PR #1402 now removes the flat function for 0.11 after the Dynamo Mocker migration, rather than marking it deprecated.

jasonqinzhou:
Correction to the earlier follow-up note: draft PR #1402 now **removes** `build_aic_engine(...)` for 0.11 instead of marking it deprecated. It remains stacked on this PR and sequenced after the 0.10 release and the Dynamo Mocker migration to `AicEngineBuilder`. The #1402 implementation, title, description, tests, and source references have all been updated to reflect the breaking removal.

simone-chen (approved):

PR #1377, opened 2026-07-17
fix(tools): use fork context for support matrix parallelism on macOS

natoscott:
## Overview

Fix `generate_support_matrix.py` so Phase 1 parallel workers actually succeed on macOS, eliminating the ~60-100x slower sequential Phase 2 fallback.

## Details

macOS defaults to `spawn` for `multiprocessing` (since Python 3.8), which requires pickling all data passed to child processes. The `SupportMatrix` object contains `defaultdict(lambda)` entries from the perf database loader, which cannot be pickled. This caused every Phase 1 worker to crash with `BrokenExecutor`, silently falling through to sequential Phase 2 retry for all combinations.

The fix uses `multiprocessing.get_context("fork")` to create a fork-based process pool, so child processes inherit the parent globals via copy-on-write — matching the Linux behavior the code was originally written for.

**Before (macOS):** Phase 1 completes in ~44s but all 89 combinations fail → Phase 2 retries all 89 sequentially at ~40-100s each → ~1 hour per model.

**After (macOS):** Phase 1 succeeds, no Phase 2 needed. Full matrix generation with many models benefits from parallel workers across the pool.

Linux behavior is unchanged (already uses fork by default).

## Test plan

- [x] `ruff check tools/support_matrix/support_matrix.py` passes
- [x] `pytest tests/unit/support_matrix/ tests/unit/cli/test_support_matrix_version.py` — 94 tests pass
- [x] Manual: `generate_support_matrix.py --model zai-org/GLM-5 --no-save` completes with zero Phase 2 retries on macOS (previously 89/89 retried)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved support-matrix generation reliability by explicitly using fork-based multiprocessing for worker creation.
  * Added an explicit runtime check to ensure worker processing starts only after the necessary initialization is in place.

### DIFF
--- a/tools/support_matrix/support_matrix.py
+++ b/tools/support_matrix/support_matrix.py
@@ -12,6 +12,7 @@ the model/system/backend/version support matrix for AIConfigurator.
 import csv
 import json
 import logging
+import multiprocessing
 import os
 import shlex
 import traceback
@@ -606,9 +607,13 @@ def _compare_pareto_dfs(
 
 
 # Per-process SupportMatrix instance for ProcessPoolExecutor workers.
-# Set in the parent before forking; children inherit it via copy-on-write.
+# Set in the parent before forking; children inherit via copy-on-write.
+# We explicitly use a "fork" mp context (see _run_parallel_combinations) so
+# this works on macOS too, where the default start method is "spawn".
 _worker_matrix: "SupportMatrix | None" = None
 _worker_modes_to_test: tuple[str, ...] | None = None
+
+_fork_ctx = multiprocessing.get_context("fork")
 
 
 def _process_combination_worker(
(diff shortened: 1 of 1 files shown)


natoscott:
@coderabbitai Good catch. The current `except Exception` catch-all in `_run_parallel_combinations` does queue infrastructure crashes (OOM, segfaults) for Phase 2 sequential retry, where they'll just fail the same way.

However, this is pre-existing behavior unrelated to the macOS fork-context fix in this PR. The fix here is a one-liner to make `ProcessPoolExecutor` work on macOS at all — changing the retry classification logic is a separate behavioral change that should be tested against the support matrix CI.

Suggest as a follow-up: replace the `retry_combos.add(combo)` in the `except Exception` branch with a direct error result entry, so infra crashes are recorded immediately instead of retried.

natoscott:
@coderabbit let's leave it an open question for the maintainers at this stage.  I'm happy to look into it if its generally agreed to be worthwhile.

jasonqinzhou (approved):

PR #1380, opened 2026-07-17
feat(sdk): analytical queueing model — distribution-level TTFT/ITL estimates

tianhaox:
## What

Adds `sdk/queueing` (shipped in `aiconfigurator_core.sdk.queueing`, compat alias `aiconfigurator.sdk.queueing`): an analytical queueing correction for continuous batching, **percentile SLA semantics** resolved by a two-stage sweep funnel, a **disagg P/D tandem model** with computed KV-transfer contention, and the **DES oracle + validation gate** that anchor all of it, running in CI. Every term traces to a scheduling-loop mechanism or a queueing-theory identity; there are **no fitted constants** (provenance table in `docs/design/queueing_model.md`).

Fifteen additive summary columns across `ColumnsAgg` / `ColumnsStatic` / `ColumnsDisagg` (legacy `ttft` / `tpot` untouched):

- `ttft_steady_{mean,p50,p75,p90,p95,p99,p999}` — steady-state TTFT distribution (SLA semantics)
- `ttft_steady_p99_{lo,hi}` — the cohort bracket: structural bounds on the steady distribution's support, used by the sweep funnel
- `ttft_transient_{mean,max}` — initial-burst admission-staircase envelope (cold start / synchronized burst)
- `itl_{mean,p50,p99}` — inter-token-latency distribution
- `queueing_tier` — precision class of each row's numbers ("screening" | "quantitative" | "static" | "composed")

## Why

The agg TTFT estimate used an empirical queuing factor whose blended mean underestimated dynamic TTFT by ~30% on a reference workload and depends on benchmark length N (same deployment: 456ms at 200 requests, 223ms at 1000) — neither a steady-state nor a percentile semantic; as an SLA gate it both falsely accepts and falsely rejects. Disagg had it worse: the composed tier reports the prefill worker's single static-batch latency as TTFT, invisible to closed-loop queueing — on a rate-matched 5P3D deployment at C=252 the composed value is 183.8ms while the tandem p50 is 1185.7ms (prefill-pool queueing dominates).

## How

**One model, two precision tiers** — screening (O(1) closed form on quantities `run_agg` already computes: effective-isl chunk counts, encoder/dispatch additive stages; zero extra perf-DB queries) and quantitative (deterministic limit-cycle evaluation of the same model; timing delegates to the SDK's own phase runners, with a `mixed_pass_ms` hook onto the mix-step runner so fused passes pay the shared non-attention cost once).

**Backend calendars**: vLLM fused (validated), TRT-LLM fused + `GUARANTEED_NO_EVICT` cap, SGLang mixed-chunk by default — matching AIC's own deployment rule — with the alternating (dedicated prefill batch) structure available.

**Percentile SLA semantics, one rule on every surface**: targets become (value, percentile) pairs on the steady distributions. Setting any `--*-percentile` flag or an `--itl` target (CLI) — or the corresponding YAML field — activates percentile filtering; unset metrics use p50 (ttft/tpot/request-latency) and p99 (itl). The only explicit switch is `--sla-refine` / `sla_refine`: a precision upgrade that never changes elimination semantics.

**Two-stage sweep funnel** (`sdk/queueing/refine.py`) on the sweep path: (1) *wide-keep* on the bracket lower bound — screening bias can never falsely reject; (2) *lazy quantitative resolution* of straddlers in throughput order, all constraints enforced at their percentiles, unresolved rows kept with tier visible. The legacy path (`pareto_analysis` / `find_best_*`, webapp) keeps legacy semantics. Report boundary upgrades the rows a human reads to quantitative tier.

**Disagg tandem model** (`sdk/queueing/disagg.py`): serving-flow semantics — the prefill pool emits the first (TTFT) token; the KV handoff is a flow on a max-min-fair per-NIC transfer fabric (fan-in/fan-out contention computed from concurrent flows; bandwidths from the system spec `node.inter_node_bw` × efficiency), landing in the first ITL gap; decode workers pay no prefill compute (KV-connector convention). Rate matching becomes an output: pool imbalance surfaces as prefill queueing or decode saturation. Heterogeneous P/D pools are priced by their own timing models. The tandem system is multi-stable in cohort phase; `evaluate_disagg_mixed` reports a deterministic phase mixture (no RNG).

**DES oracle in-tree** (`tools/queueing_oracle/`, stdlib-only): a discrete-event simulation of vLLM v1 iteration-level scheduling (unified token budget, running-set-first, chunked prefill, put-back-no-preempt admission, LIFO preemption, BlockPool-style KV accounting with prefix reuse, pull-model KV hold until the transfer completes) plus agg and disagg drivers with the same transfer fabric. `validate_formula.py` is the gate: identical timing on both sides isolates scheduling semantics; it runs in CI as a unit test (~1s), so upstream scheduler drift is a failing check.

## Validation (recorded in `docs/design/queueing_model.md` §5)

- **Gate**: 13 gated families — 9 agg (isl 512–8192, osl 16–512, C 1–128, chunked on/off, prefix) + 4 disagg (1P1D, fan-in 2P1D, bandwidth-tight, 2P2D). Evaluator within 10–15% tolerances, most metrics 0.0%; C=1 and the disagg families essentially exact, including the fan-in transfer spike in `itl_p99` reproduced to 0.0%.
- **End-to-end** (Llama-3.1-8B / h200_sxm / vLLM 0.24.0): legacy −30% vs closed form −6.9% vs evaluator −2.5%; at `--ttft 210` the funnel rescued 16 configs — including the top-throughput one — that the legacy scalar would have falsely rejected. Funnel cost: ~2x sweep time at loose SLAs, ~7x boundary-tight.
- Failure modes (incl. speculative decoding, DP prefill cadence, disagg phase multi-stability) documented with detectors in §6.

## Testing

- `tests/unit/sdk/queueing/` — 48 unit tests (structural assertions, funnel semantics, disagg tandem, transfer-fabric sharing math, oracle gate + disagg serving-flow/KV-hold checks)
- full unit suite: 2364 passed, no regressions; new CLI flags covered by arg-parsing tests; report headers unit-tested
- E2E: `aic default` agg + disagg runs with `--sla-refine` on the real perf DB; tier distribution verified in saved artifacts


tianhaox:
@coderabbitai Review findings addressed in b0e4c6baf (and d983a7ef4 for the import-contract gate). Disposition summary:

**Fixed** — closed-form `itl_* = 0` for `osl <= 1` (matches run_agg's `tpot = 0` convention); `WorkloadSpec`/`EngineSpec`/`DisaggSpec` construction-time validation (incl. `guaranteed_no_evict` now failing loudly without `kv_capacity_tokens`); validate_formula fails closed on non-finite residuals, evaluator `_Timing` mirrors the DES-side clamps, success line names the gated tier; vllm_sim `--no-enable-prefix-caching` also disables hashed-block sharing. Regression tests added for each.

**Not applied, by design** — 'use pure prefill latency for queueing TTFT': under the fused-pass model a request's prefill chunks share the iteration with running decodes, so `chunks × t_mix` is the modeled (and DES-validated) semantics, not a bug. The open-loop `/10.0` helper was removed from the PR entirely (open loop is documented out of scope). The disagg findings (context carry, `decode_eng`, queue-wait accounting, per-worker batching) were superseded by the tandem-recursion rewrite in `disagg.py`, which tracks per-request chunk state, caps decode passes at `decode_eng.max_num_seqs`, and reports `prefill_start_ms − arrival_ms` waits. The calendar preemption/KV-block findings target a variant that never shipped in this PR — the in-tree calendar has no preemption or block accounting (KV pressure is a documented scope boundary, design doc §6.5).

The remaining threads were already marked addressed in earlier commits (e3ebd31…d70dd62, 0b7eb21, ae02524).

tianhaox:
Sharing measurements from a disagg investigation (context: #1396) that this model would be well placed to explain — but, as currently scoped, would not predict. Offering them as a possible validation case for the disagg tandem, plus three concrete gaps.

**Setup.** 8×B200, Kimi-K2.5-NVFP4, **dummy weights** (`--load-format dummy`), dynamo `1.3.0-dev.1` (vLLM 0.22.0), NixlConnector. Disagg 4 prefill GPUs (dp4 + EP) / 4 decode GPUs (dp4, tp1). ISL 8150, OSL 1024, fp8 KV, `max_model_len` 9216 → 9174 tokens ≈ 0.30 GiB KV per sequence, 286 MB per handoff. Caveat: dummy weights collapse MoE routing, so treat expert-parallel balance as unrepresentative; memory footprint and step timing are representative.

### Finding 1 — decode-side KV pool bounds concurrent transfers, and that is what sets throughput

In vLLM's NixlConnector the consumer pulls: `get_num_new_matched_tokens` "pull **all** prompt blocks from remote asynchronously", and `update_state_after_alloc` allocates those blocks *before* the transfer runs. So a request in `WAITING_FOR_REMOTE_KVS` occupies full-prompt KV **out of the same pool as the running batch**. That makes

```
staging headroom (seqs) = floor(kv_pool_tokens / seq_len) - max_num_seqs
```

a hard cap on in-flight handoffs. Sweeping decode `max_num_seqs` at `gpu_memory_utilization` 0.92 (note the pool itself shrinks as `max_num_seqs` grows, since NIXL/graph buffers scale with it):

| decode `max_num_seqs` | KV pool (tok/rank) | capacity (seq) | headroom | req/s | tok/s/GPU | mean TPOT | median ITL |
|---|---|---|---|---|---|---|---|
| 18 | 267,104 | 29.1 | **+11.1** | 3.71 | 475.0 | 16.62 ms | 14.09 ms |
| 24 | 258,784 | 28.2 | **+4.2** | 1.58 | 202.4 | 45.22 ms | 16.59 ms |
| 29 | 251,776 | 27.4 | **−1.6** | 1.25 | 160.5 | 54.09 ms | 17.71 ms |

**Controlled test** — same `max_num_seqs` 24, same concurrency, *only* the pool enlarged (`--gpu-memory-utilization` 0.92 → 0.94, pool 258,784 → 367,776 tok, headroom +4.2 → +16.1):

| | req/s | tok/s/GPU | mean TPOT |
|---|---|---|---|
| headroom +4.2 | 1.58 | 202.4 | 45.22 ms |
| headroom +16.1 | **2.66** | **339.9** | **15.84 ms** |

Headroom is causal. Note also that at low headroom **mean TPOT 45.2 ms vs median ITL 16.6 ms** — the loss is a stall/preempt long tail, not slower steps.

Two things in the design doc keep this out of reach today:
- the calendar models admission by `max_num_seqs` + token budget (plus `guaranteed_no_evict` for TRT-LLM), with no decode-side KV-pool cap on concurrent transfers;
- "vLLM preempts-and-recomputes … this calendar does NOT reproduce" — which is exactly the mechanism generating that 45 ms tail.

### Finding 2 — with a pull-based connector the prefill slot is held until the *transfer* completes, not until prefill completes

`prefill_inflight_cap` (κ) releases the prefill slot at prefill completion, and the handoff is placed in the first ITL gap. Under NixlConnector the producer must hold the request until the consumer has pulled it. Measured, with an identical prefill worker across all runs:

- conc-1 median TTFT = **320 ms** (prefill + transfer, no queueing) — the transfer path is *not* slow in isolation;
- prefill worker peak `Avg prompt throughput` is identical across runs (17170 / 14716 / 14717 tok/s per rank) while the **median** falls 8992 → 5723 → 4905 — i.e. it idles, it does not slow down;
- slot time = 4 slots / req_s → **1.08 s** (headroom +11.1) / **1.50 s** (+16.1) / **2.53 s** (+4.2), against ~0.32 s of actual prefill compute. The slot spends **70–87% of its life waiting for the pull.**

Direct test — raise prefill `max_num_seqs` 1 → 4 so a request awaiting its pull no longer blocks the scheduling slot (decode fixed at 24 / util 0.94; prefill compute is unchanged, one prompt per pass either way, since 2×8150 > the 9692 token budget):

| | conc 96 | conc 144 |
|---|---|---|
| prefill `max_num_seqs` 1 | 2.39 req/s (306.4 tok/s/GPU) | 2.66 (339.9) |
| prefill `max_num_seqs` 4 | **3.98 req/s (509.2 tok/s/GPU)** | 1.84 (235.6) |

**+67% at conc 96** from a prefill-side knob that the model treats as non-binding. With κ=1 the model would put prefill capacity at 4 / 0.32 s = 12.5 req/s and never flag it — over-predicting in the same direction as the current estimator.

The conc-144 regression is the interesting part: the two limiters are **coupled**. More prefill supply only helps while decode staging headroom lasts; past that, the extra in-flight requests re-flood the pool and throughput falls below the κ=1 baseline. A model with both effects would reproduce this non-monotonicity; a model with neither predicts monotone improvement.

### Finding 3 — a NIC-bandwidth fabric is the wrong cost model for intra-node handoffs

`_TransferFabric` shares nominal per-worker NIC bandwidth (`bw_efficiency` 0.8). Here both workers are on one node and NIXL selects the **UCX** backend (the TCP endpoints in the logs are only the metadata side channel). At NIC bandwidth a 286 MB handoff is ~7 ms — negligible, and nothing in the sweep would ever bind. The real cost is not bandwidth at all: an `osl=4` probe (decode work made negligible) sustained **6.95 req/s at concurrency 32** on the same deployment, per-slot 575 ms vs 320 ms idle. So the handoff path has ample capacity; what throttles the system is *occupancy* — pool blocks held during staging, and prefill slots held until the pull lands.

### Suggested minimal changes

1. Charge `WAITING_FOR_REMOTE_KVS` blocks against the decode pool and derive a concurrent-transfer cap `floor(pool_tokens / seq_len) − max_num_seqs`; serialize the tandem as it approaches 0. This alone reproduces the 3.71 / 1.58 / 1.25 ordering.
2. For pull-based connectors (NixlConnector), release the prefill slot at **transfer completion**, not prefill completion — i.e. κ should gate on the handoff, not the forward pass.
3. Route intra-node handoffs over NVLink/UCX-shm rather than the NIC fabric, and consider expressing the handoff cost as slot-occupancy rather than pure bandwidth.

Separately, this matters for config *picking*, not just latency reporting: the current sweep sizes the decode pool as though the running batch were its only occupant, so a recommended `max_num_seqs` can land at zero or negative staging headroom (bs29 above), which is the worst point measured.

Happy to re-run any of these, share the raw logs, or test a patched branch.

tianhaox:
**Correction to my comment above, plus a direct measurement of the mechanism.**

I attributed the mean-TPOT-45.2 ms / median-ITL-16.6 ms tail to vLLM's preempt-and-recompute. **That is wrong — there were zero preemptions in every run.** vLLM V1 appends `Preemptions: %d` to the periodic scheduler line whenever the counter is non-zero (`v1/metrics/loggers.py:244`), and no run emitted it. So the point I made about `GUARANTEED_NO_EVICT` being out of scope does not apply here; please disregard that part.

The same log line carries `Deferred: %d reqs`, which counts requests parked in `WAITING_FOR_REMOTE_KVS` (`v1/core/sched/scheduler.py:766` → `step_skipped_waiting`). That is a direct readout of in-flight handoffs, and it confirms the staging-headroom mechanism much more sharply than my indirect argument did:

| decode config | staging headroom | `Deferred` med / max | GPU KV usage med / max | req/s |
|---|---|---|---|---|
| bs18, util 0.92 | +11.1 | **7.5 / 14** | 57.2% / 97.1% | 3.71 |
| bs24, util 0.92 | +4.2 | **3.0 / 7** | 75.4% / **99.6%** | 1.58 |
| bs24, util 0.94 | +16.1 | 0 / 0 | 20.4% / 57.2% | 2.66 |

Headroom +11.1 → +4.2 halves the staging pipeline depth (7.5 → 3.0 concurrent transfers) while the pool saturates at 99.6%. Nothing is evicted; the **transfer pipeline is starved of admission**. (The third row has zero `Deferred` because it is supply-limited from the prefill side instead — the second limiter in my previous comment.)

This makes the modelling ask simpler and more mechanical than what I wrote before. There is no eviction dynamic to reproduce, so the calendar does not need vLLM's preemption semantics at all. It needs one state variable:

```
staging_depth <= floor(kv_pool_tokens / seq_len) - max_num_seqs
```

with the tandem's transfer stage serialized as that bound approaches 0. `Deferred` gives a free validation signal for it — a correct model should reproduce 7.5 and 3.0 for the two rows above.

For reference on the engine side: the scheduler's `load_kv_async` admission path (`scheduler.py:642` and `:763`) has **no cap at all** on concurrent remote-KV loads and no reserve, and there is no corresponding knob in `config/` or `arg_utils.py`. So today the only defence is static sizing — keeping `floor(pool / max_model_len) − max_num_seqs` comfortably positive — which is precisely a config-time decision, i.e. this project's job rather than vLLM's.

PR #1384, opened 2026-07-20
feat: add fpm based modeling

liyuanzhe1991:
## Overview:

Adds the AIC-FPM modeling Python vertical slice: an opt-in `forward_model=fpm` mode that predicts whole-model forward-pass latency directly from collected `fpm_forward_perf` data (schema v6, per-DP-rank iteration totals under the `balanced_v1` partition policy) instead of composing per-op estimates. The default (`op_level`) is unchanged everywhere.

Measured accuracy sits at the noise floor of the benchmark itself — overall holdout MAPE 2.33% on 509 off-grid points against median-of-5 truths (see **Validation** below).

## Details:

- `ModelConfig.forward_model = op_level|fpm` (default unchanged), validated and applied centrally in `get_model`: each phase op list becomes exactly one `FPMForwardOp`. Encoder/multimodal models and MTP (`nextn > 0`) are rejected explicitly — whole-model curves carry neither the MTP-head cost nor acceptance amortization (MTP support lands separately by applying `mtp_scale_factor` to the decode op).
- `operations/fpm_forward.py`: strict parquet/sidecar loader (sha256 digest, duplicate row keys, version and coordinate validation), per-cell tables with row-derived domains, exact-match-then-ScatteredSites resolution via `perf_interp`, hard out-of-domain errors (FPM never extrapolates), zero-energy convention, exact-model-path-only cell selection (no cross-model fallback) across the schema-v6 backend identity columns, plus a fail-closed guard for pinned vLLM backend/EPLB identities that the standard deployment path cannot yet reproduce. The whole-model SOL roofline is derived from the original op-level model queried in `DatabaseMode.SOL`, so it inherits the real physics (MoE activation, per-op quantization) instead of a hand-rolled approximation.
- `perf_interp`: new opt-in `ScatteredSites.own_curve_coverage_fallback` (default `False` — stock behavior preserved) so degenerate sites backed by a few stray sweep points defer to neighbour-site transfer instead of anchoring far extrapolation.
- Explicit FPM mixed-step branch: marginal-decode composition (prefill pass plus decode's KV marginal via pass-baseline subtraction; generation-only steps keep the full decode pass). Rust engine step is forced off for FPM models until the Rust op variant lands.
- `forward_model` threaded through TaskV2, v1-compat, CLI (`--forward-model`, documented in `docs/cli_user_guide.md`) and cli/api estimate/default/exp paths; AFD mode rejects fpm.
- `tests/unit/sdk/test_fpm_forward.py`: 53 tests incl. loader validation failures (digest/row-count/version mismatches, non-finite latency), interpolation goldens, DP local-batch semantics, all rejection paths, and static/mixed/genonly integration through a real `PerfDatabase`.

### Validation (GLM-5.2-NVFP4, 4×B200, vLLM 0.25.1)

**Protocol.** Predictions are frozen from the collected database through the real SDK path (`forward_model=fpm`, dp4: tp1 · attention-DP 4 · moe-EP 4), then compared against 509 off-grid holdout points in five strata; each truth is the **median of 5 independent measurements** (latency = max wall_time across DP ranks, matching the `balanced_v1` row convention). Leave-one-out over every collected row serves as the interpolation upper bound.

**Holdout (509 points, single-sample DB):**

### Table 1 · MiniMax-M2.7 (gcfix campaign, 10 topologies) — APE %, med / P95 / max

| Topology | LOO prefill† | LOO decode | Hold-out ×5 prefill (~423 pts) | Hold-out ×5 decode (~90 pts) |
|---|---|---|---|---|
| single | 4.40 / 15.97 / 249† | 1.76 / 11.71 / 27.4 | 3.91 / 8.35 / 23.4 | 3.82 / 15.86 / 22.6 |
| dep2 | 3.73 / 18.78 / 269† | 1.28 / 5.32 / 9.7 | 1.49 / 9.23 / 13.1 | 2.21 / 6.43 / 9.7 |
| dep4 | 1.93 / 12.10 / 268† | 0.89 / 3.47 / 14.6 | 1.31 / 6.77 / 12.2 | 2.55 / 6.11 / 8.4 |
| dep8 | 2.00 / 12.57 / 212† | 0.98 / 3.63 / 15.9 | 2.23 / 9.87 / 18.6 | 1.27 / 7.09 / 14.8 |
| pure_tp2 | 2.15 / 20.88 / 341† | 1.61 / 5.58 / 12.1 | 3.27 / 9.57 / 13.9 | 1.64 / 6.91 / 11.5 |
| pure_tp4 | 1.69 / 23.92 / 364† | 1.51 / 17.62 / 36.6 | 2.05 / 8.04 / 12.2 | 3.09 / 11.39 / 16.1 |
| pure_tp8 | 1.50 / 19.96 / 414† | 1.58 / 11.70 / 30.0 | 5.53 / 10.40 / 17.4 | 2.86 / 9.06 / 24.6 |
| tep2 | 2.67 / 20.09 / 350† | 1.56 / 7.27 / 13.8 | 2.95 / 10.51 / 18.4 | 1.73 / 5.54 / 11.2 |
| tep4 | 2.55 / 17.23 / 400† | 1.47 / 13.57 / 27.9 | 2.71 / 8.22 / 12.1 | 2.44 / 9.60 / 14.3 |
| tep8 | 1.73 / 18.65 / 387† | 1.94 / 18.60 / 39.3 | 6.19 / 11.35 / 18.8 | 1.67 / 9.04 / 33.9 |
| **Overall** | — | — | **2.77 / 9.68 / 23.4** (4,229) | **2.18 / 9.51 / 33.9** (929) |

### Table 2 · GLM-5.2 (reference, 6 topologies) — APE %, med / P95 / max

| GLM topology | LOO prefill† | LOO decode | Hold-out ×5 prefill | Hold-out ×5 decode |
|---|---|---|---|---|
| dp4 (=dep4) | 1.83 / 21.92 / 107† | 0.72 / 5.73 / 17.5 | 1.24 / 7.01 / 13.5 | 1.17 / 3.39 / 5.2 |
| dp8 (=dep8) | 1.67 / 18.70 / 79† | 0.79 / 4.12 / 8.5 | 2.00 / 9.52 / 13.9 | 0.75 / 2.62 / 6.1 |
| tp8ep8 (=tep8) | 1.12 / 55.60 / 215† | 0.49 / 4.82 / 54.5 | 3.44 / 10.91 / 17.2 | 1.06 / 2.64 / 4.9 |
| tp8tp8 (=pure8) | 0.88 / 55.55 / 209† | 0.59 / 5.35 / 53.4 | 5.07 / 9.32 / 16.9 | 0.81 / 2.50 / 5.9 |
| tp4ep4 (=tep4) | 0.79 / 9.50 / 243† | 0.70 / 4.50 / 31.0 | 5.63 / 12.28 / 17.4 | 1.74 / 4.98 / 13.3 |
| tp4tp4 (=pure4) | 0.77 / 12.38 / 249† | 0.71 / 4.42 / 28.9 | 3.18 / 13.49 / 18.9 | 1.20 / 4.80 / 6.8 |
| **Overall (6 topologies)** | — | — | **3.02 / 11.40 / 18.9** (2,571) | **1.08 / 3.93 / 13.3** (540) |

The truth-side 5-repeat spread has median 2.24%, i.e. prediction error is at the noise floor of the measurement itself.

**Leave-one-out:** prefill 4.94% MAPE / max 121% (4,097 points), decode 2.34% / max 98.6% (1,364 points). The same protocol across six parallel topologies (dp4 / dp8 / tp8ep8 / tp8tp8 / tp4tp4 / tp4ep4) lands at 1.96–5.19% holdout MAPE.

**Reproduce (prediction, user-facing):** place `fpm_forward_perf.parquet` + its `.metadata.json` under `<root>/data/b200_sxm/vllm/0.25.1/` next to `<root>/b200_sxm.yaml`, then:

```bash
aiconfigurator cli estimate \
  --estimate-mode static_gen \
  --model nvidia/GLM-5.2-NVFP4 \
  --system b200_sxm --backend vllm --perf-db-version 0.25.1 \
  --systems-paths <root> --forward-model fpm \
  --tp 1 --dp 4 --moe-ep-size 4 \
  --gemm-quant-mode nvfp4 --moe-quant-mode nvfp4 \
  --fmha-quant-mode fp8 --kvcache-quant-mode fp8 --comm-quant-mode half \
  --batch-size 8 --isl 32768 --osl 2
# → Generation lat. 40.377 ms (TPOT); --estimate-mode static_ctx gives the prefill side
```

The freeze/compare/LOO harness lives in the internal validation campaign (not part of this PR); the protocol is: freeze per-point predictions through `FPMForwardOp.query` at dp4, compare to the median-of-5 truths per stratum, and LOO = remove each collected row, predict it from the rest, skipping sole owners of a global axis extreme.

No perf data files ship with this PR; databases are published separately. Note for release notes: saved `exp_config.yaml` now carries `forward_model: op_level`, which older releases' strict `from_yaml` will reject (inherent to any new Task field).

## Where should the reviewer start?

- `aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py` — data contract in the module docstring, loader, cell selection, interp configs
- `aic-core/src/aiconfigurator_core/sdk/models/__init__.py` — `_apply_forward_model_fpm` rewrite + rejections
- `aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py` — `_get_fpm_mix_step_latency`, Rust gates
- `aic-core/src/aiconfigurator_core/sdk/perf_interp/config.py`, `engine.py` — `own_curve_coverage_fallback`
- `docs/fpm/aic-fpm-modeling-plan.md` — design decisions D1–D4

## Related Issues:

🚫 This PR is NOT linked to an issue — confirmed.

## Summary by CodeRabbit

* **New Features**
  * Added configurable forward-pass modeling with `op_level` and whole-model `fpm` modes.
  * Added CLI and task configuration support through the `--forward-model` option.
  * Added validated profiling-data support for FPM latency estimates.
  * Added FPM-aware prefill, decode, mixed-step, and generation-only performance estimation.
* **Bug Fixes**
  * Improved interpolation fallback when collected performance curves do not cover a requested value.
* **Documentation**
  * Documented forward-model modes, data requirements, defaults, and limitations, including unsupported AFD estimates.


docs/cli_user_guide.md
```diff
@@ -17,6 +17,7 @@ These flags are shared across modes (a few are sweep-only, as noted):
 - `--systems-paths`: System search paths (comma-separated). Use `default` for the built-in systems path; the first match wins for an identical system/backend/version. (`default`, `exp`, `generate`, `estimate`)
 - `--deployment-target`: Generated-artifact platform — `dynamo-j2` (default), `dynamo-python`, `llm-d-helm`, `llm-d-kustomize`, or `fpm`. See [Deployment Target Selection](#deployment-target-selection). (`default`, `exp`, `generate`, `estimate`)
 - `--engine-step-backend`: Experimental static-latency backend — `python` (default) or `rust` (routes static step estimates through the Rust FPM estimator). (`default`, `exp`, `generate`, `estimate`)
+- `--forward-model`: Forward-pass modeling mode — `op_level` (default; granular per-op modeling) or `fpm` (predicts from collected whole-model forward-pass data; requires `fpm_forward_perf` data for the exact model/system/backend/version and never extrapolates outside the collected domain). Not supported in the `afd` estimate mode. (`default`, `exp`, `generate`, `estimate`)
```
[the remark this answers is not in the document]
liyuanzhe1991: The doc line is accurate as written. `FPMForwardOp._resolve` applies a per-cell domain gate **before** `perf_interp` is invoked: any query coordinate outside the collected per-axis domain raises `PerfDataNotAvailableError` (`aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py`, `_resolve` — the module docstring states "the domain gate runs BEFORE perf_interp"). The neighbor-site transfer and curve-end util-hold in `perf_interp/engine.py` therefore only ever operate on in-domain queries, i.e. interpolation between collected points — never extrapolation beyond the collected domain. Keeping the current wording.

docs/fpm/aic-fpm-modeling-plan.md
```diff
@@ -0,0 +1,264 @@
...
+
+- Add `forward_model: str = "op_level"` to **`ModelConfig`**
+  (`aic-core/src/aiconfigurator_core/sdk/config.py:11`). ModelConfig, not RuntimeConfig,
+  because the op-list rewrite happens inside `get_model`, which receives ModelConfig only.
+  Values: `op_level | fpm`; anything else raises.
+- Thread it exactly like `engine_step_backend`:
+  TaskV2 field + `build_model_config` (`src/aiconfigurator/sdk/task_v2.py:354, 1074`),
+  v1-compat passthrough (`task_v1_compat.py:83`), CLI flag `--forward-model`
+  (`src/aiconfigurator/cli/main.py`, `cli/api.py`), webapp bridges, `pareto_analysis.py`.
+  Generator is **not** touched in V1 (module boundary).
```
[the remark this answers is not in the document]
liyuanzhe1991: `forward_model` is a modeling-mode switch, not a deployment/engine parameter: it never flows into generator output. The bridge (`generator/module_bridge.py`) reads named Task attributes and named result-row columns only, none of which change with this flag, so generated deployment configs are byte-identical with the parameter set or omitted — there is nothing for the backend-schema validator or `--generator-help` to surface. The guideline you cite targets parameters that reach rendered configs; keeping the generator untouched here is the repo's module-boundary rule, not a skipped validation. The plan doc now states this explicitly (see the updated M0 section).

liyuanzhe1991:
@coderabbitai review please

tests/unit/sdk/test_opspec_coverage.py
```diff
@@ -39,6 +39,9 @@
     # Dead class: no model instantiates it (Mamba2Kernel is the live op and
     # converts). Remove the class or this entry together.
     "Mamba2": "dead code — never instantiated; Mamba2Kernel is the live op",
+    # forward_model="fpm" is forced onto the Python engine step in
+    # base_backend (no Rust op variant yet); the Rust port lands separately.
+    "FPMForwardOp": "FPM forces the Python route; Rust port is a separate PR",
```
simone-chen, line 44: This line might not be detected automatically during the rebase when merging the Rust PR. Worth asking the agent to do a scan before merging https://github.com/ai-dynamo/aiconfigurator/pull/1461

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
```diff
@@ -0,0 +1,601 @@
...
+            available = sorted({(cell["model_path"], *cell["match_identity"]) for cell in cells.values()})
+            raise PerfDataNotAvailableError(
+                f"No FPM cell matches this model identity: "
+                f"{dict(zip(_CELL_MATCH_COLUMNS, self._match_identity, strict=True))}. "
+                f"Collected cell identities (model_path first): {available[:8]}"
+            )
+        exact = [cell for cell in matches if cell["model_path"] == self._model_path]
+        if exact:
+            matches = exact
+        distinct_paths = sorted({cell["model_path"] for cell in matches})
```
jasonqinzhou: **[P1] Do not fall back to an unrelated model's whole-model curve**

When `model_path` has no exact match, this path accepts the sole collected path as long as the quantization and parallelism fields match. Those fields contain no architecture, layer count, hidden size, or other model fingerprint, so a database containing only one model will silently answer a request for any other model with the same topology. I reproduced `_select_cell` returning `collected/other-model` for `requested/model`. Because these are whole-model latencies, please require an exact path or gate aliases on an immutable verified model fingerprint/explicit alias mapping; otherwise raise `PerfDataNotAvailableError`.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
```diff
@@ -0,0 +1,601 @@
...
+
+    # ------------------------------------------------------------------
+    # Cell selection (open decision D1: exact model_path, else unique fallback)
+    # ------------------------------------------------------------------
+
+    def _select_cell(self, cells: dict) -> dict:
+        matches = [
+            cell
+            for cell in cells.values()
+            if cell["match_identity"] == self._match_identity and cell["backend_axis"] == _SUPPORTED_BACKEND_AXIS
```
jasonqinzhou: **[P1] Include backend policy in the requested cell identity**

This filter checks only `backend_axis` plus `_match_identity`; the latter omits `enable_wideep`, `enable_eplb`, `moe_backend`, and `attention_backend`, and no requested `backend_policy` is derived from the model config. A policy-changing config therefore has the same identity as baseline, and a single arbitrary policy is accepted (the ambiguity check only fires when several exist). That silently prices WideEP/MegaMoE/FA3-style execution with a different whole-model curve. Please enforce the V1 `baseline_auto` contract and reject incompatible model-config options, or extend the schema/match key so the requested policy is matched explicitly.

src/aiconfigurator/cli/main.py
```diff
@@ -156,6 +156,14 @@ def _build_common_cli_experiments_parser() -> argparse.ArgumentParser:
         "crosses the FFI); use 'python' as the escape hatch or 'rust' to force the "
         "compiled engine.",
     )
+    common_parser.add_argument(
+        "--forward-model",
```
jasonqinzhou, line 160: **[P1] Do not silently drop `--forward-model` on recommend paths**

This common parent parser is also attached to `recommend`, so `recommend --forward-model fpm` is accepted. The same happens for `default` with only a load target, which routes through `_run_recommend`. However, `_run_recommend` does not pass `args.forward_model`, `cli_recommend` has no such parameter, and its `build_default_tasks` call therefore uses `op_level`. A direct probe parsed `fpm` and observed no forwarded value. That makes an explicit FPM procurement-sizing request silently run the wrong model. Please thread the argument through `cli_recommend`/`build_default_tasks` with regression coverage, or reject it on these paths.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
```diff
@@ -0,0 +1,601 @@
...
+    try:
+        import pyarrow.parquet as pq
+    except ImportError as exc:
+        raise RuntimeError(
+            "Loading fpm_forward perf data requires the 'pyarrow' package. "
+            "Install aiconfigurator with its declared runtime dependencies."
+        ) from exc
+
+    metadata_path = os.path.splitext(primary_path)[0] + ".metadata.json"
+    metadata = _validate_sidecar(metadata_path, primary_path)
```
jasonqinzhou: **[P2] Validate that the pair belongs to the requested database**

The sidecar carries `system`, `backend`, `backend_version`, and `measurement_policy`, but `_validate_sidecar` checks none of those fields; `_validate_row` checks only the version and also ignores each row's `system`/`backend`. Since those values are not part of the in-memory cell key, a copied or mispublished pair can be served as the current database even though its metadata names different hardware or a different backend. The SHA proves only that the two files match each other. Please pass the expected database identity into the loader and reject sidecar or row mismatches before building cells.

jasonqinzhou (changes requested):

## What this PR does

This PR adds an opt-in forward_model=fpm path that replaces granular forward-pass operations with whole-model FPM operations backed by validated parquet data, with SDK/CLI wiring, interpolation, Python mixed-step composition, documentation, and focused tests. The new head merges current main, including AFD CLI work and the Rust-default engine step, while preserving FPM's forced-Python execution path.

The merge resolution carefully preserves the FPM-specific Python routing alongside the new Rust-default and AFD surfaces, and the dedicated 41-test FPM suite still passes at the refreshed head.

## Why this score

The incremental merge is sound around the changed Rust-routing and CLI surfaces, but it does not address any of the four prior correctness findings. Three P1 paths can still silently select whole-model data for the wrong model or backend policy, or ignore an explicit FPM request on recommend/load-target execution; the loader also still omits system/backend identity validation. Those unchanged issues keep the score at 5.0 and require changes before approval.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
```diff
@@ -0,0 +1,678 @@
...
```
jasonqinzhou: **[P2] Validate the sidecar identity, not only the rows**

The new `_validate_row(..., expected_system, expected_backend)` check correctly prevents a misplaced parquet from being served, but `_validate_sidecar` still accepts metadata that names a different `system`, `backend`, or `backend_version` (and any `measurement_policy`). I wrote a pair whose rows matched `h200_sxm/vllm/test-fpm-version` while its sidecar named three foreign values, and the current loader still returned the 8.0 ms decode row. Since this module describes the sidecar as the writer's commit record and treats pair inconsistencies as structural errors, please pass the expected database identity into `_validate_sidecar` and reject contradictory metadata; ideally pin the supported measurement policy there as well.

docs/fpm/aic-fpm-modeling-plan.md
```diff
@@ -0,0 +1,273 @@
...
+forced off, and `tests/unit/sdk/test_fpm_forward.py` (33 tests). D1 shipped as
+exact-model-path-then-unique-fallback; D2 shipped as ScatteredSites (per-phase configs,
+one-line swap to Grid for the LOO bake-off); D3 shipped as zero-energy; D4 deferred to M2.
+The mixed step ships the marginal-decode composition (§M3), replacing the original plan's
+plain prefill+decode sum. M2 (Rust op + schema bump + parity) and real-data LOO
+qualification remain.
+
+## 5. Open decisions (need owner sign-off before implementation)
+
+- **D1 — model identity matching. RESOLVED (shipped).** Exact `model_path` match wins;
```
jasonqinzhou: **[P2] Keep D1 aligned with the new exact-only model rule**

This delta intentionally removes the unique collected-path fallback and now raises whenever `model_path` is not exact, but the shipped D1 text still says an unmatched path may select a unique collected model. That leaves the implementation plan promising behavior the loader no longer supports. Please update this decision (and the D1 sentence in the implementation-status paragraph above) to say that FPM requires the exact collected model path and never substitutes another model's curve.

jasonqinzhou (commented):
**COMMENT · 8.4/10 · high confidence**

## What this PR does

This PR adds an opt-in forward_model=fpm path that replaces granular forward-pass operations with whole-model FPM operations backed by validated parquet data, with SDK/CLI wiring, interpolation, Python mixed-step composition, documentation, and focused tests. The new head specifically makes backend-policy matching fail closed, requires exact model paths, preserves FPM on recommend/load-target execution, and pins loaded rows to the requested database identity.

The two follow-up commits address the three prior blockers directly and add focused regression coverage for exact-model selection, policy mismatches, recommend forwarding, row identity, and coordinate collisions rather than weakening the FPM data contract.

## Why this score

The current head earns 8.4/10 with a COMMENT recommendation: all three prior P1 correctness findings are fixed, the focused 69-test FPM/CLI suite passes, and all 19 live GitHub check contexts are green. One prior P2 is only partially fixed because row identity is now validated but contradictory sidecar identity is still accepted, and the exact-model semantic change leaves the shipped D1 documentation describing the removed unique-path fallback. Both remaining findings are actionable but non-blocking.

jasonqinzhou (commented):
Two documentation/design follow-ups from the current FPM plan review.

docs/fpm/aic-fpm-modeling-plan.md
```diff
@@ -0,0 +1,273 @@
+# AIC-FPM Modeling Plan (Task 2) — realigned to the as-built collector
+
+> Revision: 2026-07-19
+>
+> Supersedes: Task 2 sections (M0–M4, §5.6–§5.7) of `aic-fpm-integration-plan.md` (2026-07-13)
```
jasonqinzhou: Could we remove or reframe this dangling reference? `aic-fpm-integration-plan.md` is not present on `main`, this branch, or `feature/fpm-collector`, and I could not find it in repository history, so reviewers cannot inspect the plan being superseded. Suggested wording:

```md
> Historical context: An internal Task 2 draft dated 2026-07-13 assumed a
> different Task 1 collector contract. That draft is not part of this
> repository; this document is authoritative, and Section 1 records the
> relevant assumptions that were superseded.
```

It would also help to rename Section 1 to “Differences from the earlier internal draft.”

docs/fpm/aic-fpm-modeling-plan.md
```diff
@@ -0,0 +1,273 @@
...
+   (`base_backend.py:440-441` zeroes energy dicts) rather than a new hard-error path. The
+   original plan predates this precedent; consistency wins. Documented as a known limitation.
+
+## 2. Data contract consumed (fixed by Task 1 — do not change here)
+
+Location (post-restructure the canonical tree is `aic-core/src/aiconfigurator_core/systems/data/`;
+`src/aiconfigurator/systems` is a symlink to it):
+
+```text
+systems/data/<system>/<backend>/<version>/
```
jasonqinzhou, line 50: Can we align FPM data ownership with Collector V3 instead of making this a permanent legacy-layout/exact-version island? Collector V3 uses `<system>/<family>/<backend>/<version>/` with `collection_meta.yaml` and optional `reuse.yaml`; FPM currently bypasses that source-resolution and provenance model.

I suggest treating `fpm_forward` as a managed family and documenting a focused follow-up with stricter reuse semantics appropriate for whole-model data:

- exact requested backend/version data first;
- explicit same-backend reuse of one complete parquet/metadata pair through `reuse.yaml`;
- no implicit nearest-earlier per-shape fill and no cross-backend reuse;
- end-to-end FPM validation before approving a reuse declaration.

V1 can remain exact-only, but the document should state that boundary explicitly and avoid presenting the current location as an immutable contract. Reusing `_build_op_sources()` directly would be unsafe because it merges shapes from several sources, whereas the FPM parquet plus metadata sidecar is one atomic campaign artifact.

jasonqinzhou (commented):
One documentation-clarity suggestion.

aic-core/src/aiconfigurator_core/sdk/perf_interp/config.py
```diff
@@ -147,6 +147,15 @@ class ScatteredSites:
...
     k_tail: int = 3
+    #: "A collected site answers alone" assumes every site owns a representative
+    #: curve — true for GEMM's real (n,k) shapes, violated by tables whose axis
+    #: rules emit stray coordinates backed by a few sweep points (FPM's
+    #: max-batch / capacity-endpoint orphans). When True, an own curve that
+    #: does NOT cover the query coordinate defers to neighbour-site transfer
+    #: (the own site is excluded from candidates) instead of holding its own
+    #: tail. In-range own-site behavior and exact hits are unchanged. Default
+    #: False preserves stock behavior exactly.
+    own_curve_coverage_fallback: bool = False
```
jasonqinzhou: Could we define `site` before using it? In this resolver it means the fixed-coordinate tuple that owns a one-dimensional curve, but that specialized meaning is not obvious to readers unfamiliar with scattered-site interpolation. A simpler rewrite could be:

```suggestion
    #: A "site" is one fixed combination of ``site_axes`` values; it owns a
    #: one-dimensional curve over ``curve_axis``. Normally, when a query matches
    #: a collected site, only that site's curve is used. Some tables, including
    #: FPM, contain orphan sites with only a few points, so the matched curve may
    #: not cover the requested curve coordinate. When True, such a curve is
    #: ignored and the query is estimated from neighbouring sites instead. Exact
    #: hits and queries within the matched curve's range are unchanged. Default
    #: False preserves existing behavior.
    own_curve_coverage_fallback: bool = False
```

docs/cli_user_guide.md
```diff
@@ -17,6 +17,7 @@ These flags are shared across modes (a few are sweep-only, as noted):
...
 - `--deployment-target`: Generated-artifact platform — `dynamo-j2` (default), `dynamo-python`, `llm-d-helm`, `llm-d-kustomize`, or `fpm`. See [Deployment Target Selection](#deployment-target-selection). (`default`, `exp`, `generate`, `estimate`)
 - `--engine-step-backend`: Engine-step latency backend — unset defaults to the compiled Rust engine (databases with measured power data delegate to the Python step until energy crosses the FFI); `python` is the escape hatch, `rust` forces the compiled engine. Accepted by all modes but inert in `generate`, which performs no latency estimation. (`default`, `exp`, `generate`, `estimate`)
+- `--forward-model`: Forward-pass modeling mode — `op_level` (default; granular per-op modeling) or `fpm` (predicts from collected whole-model forward-pass data; requires `fpm_forward_perf` data for the exact model/system/backend/version and never extrapolates outside the collected domain). `fpm` always runs on the Python engine step regardless of `--engine-step-backend` (it compiles to no Rust op variant yet). Not supported in the `afd` estimate mode. (`default`, `exp`, `generate`, `estimate`)
```
jasonqinzhou: Could we make the Python-only boundary more explicit here? “Always runs on the Python engine step regardless of `--engine-step-backend`” is easy to read as something about the serving backend, and it also hides that an explicit `--engine-step-backend rust` currently has no effect. Suggested wording:

```md
When `--forward-model fpm` is selected, AIC evaluates step latency using the Python implementation. `--engine-step-backend rust` does not take effect because the compiled engine does not yet have an FPM operation; Rust support is tracked separately in PR #1461.
```

jasonqinzhou (commented):
One CLI documentation clarification.

jasonqinzhou (approved):

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
liyuanzhe1991: Fixed in 52cbe552: the unique-path fallback is removed — selection now requires the exact collected `model_path` plus the full identity tuple, and a miss raises `PerfDataNotAvailableError` listing the collected identities. The design doc's D1 is updated to match in 229036bf.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
liyuanzhe1991: Fixed, in two steps. ab8363c4 first derived the requested backend-policy identity from the config's backend fields and matched it against the collected cells. We then fixed this on the producer side too: the collector (re-cut from #1365 into the #1473/#1474/#1475 chain) now publishes `fpm_forward_perf` schema v6, where `moe_backend` / `attention_backend` / `enable_wideep` / `enable_eplb` are explicit identity columns (producer half: 1aa9f9d8 + 056f87c8 on #1475), and 7550e244 adopts it here — the four knobs join the cell match identity exactly like tp/quant, so a policy-changing config is a plain no-matching-cell data miss, and data collected under any knob combination answers exactly the configs that declare it.

src/aiconfigurator/cli/main.py
liyuanzhe1991: Fixed in 52cbe552: `_run_recommend` passes `forward_model` to `cli_recommend`, which forwards it into `build_default_tasks`; a unit test asserts the kwarg reaches task building.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
liyuanzhe1991: Fixed in 52cbe552: each row's `system`/`backend` is pinned to the resolved database identity (a misplaced pair fails loudly), unreadable parquet raises the documented ValueError, and same-coordinate rows differing only in `cell_id`/`weight_quantization` are rejected instead of silently last-winning. 229036bf extends the same pinning to the sidecar commit record per your follow-up.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
liyuanzhe1991: Fixed in 229036bf: `_validate_sidecar` now takes the expected database identity and rejects metadata that contradicts `system`/`backend`/`backend_version`, and pins `measurement_policy` to the collector's published regime (`dynamo_native_single_sample_v1`). Your repro — foreign sidecar values over matching rows — is now a parametrized test. The Rust loader on #1461 mirrors the same gate (c72bc9fa).

docs/fpm/aic-fpm-modeling-plan.md
liyuanzhe1991: Fixed in 229036bf: D1 and the implementation-status paragraph now both state exact-collected-path-only — FPM never substitutes another model's curve; the interim unique-path fallback is recorded as removed in review.

docs/fpm/aic-fpm-modeling-plan.md
liyuanzhe1991: Fixed in 229036bf with your suggested wording, and Section 1 is renamed to "Differences from the earlier internal draft".

docs/fpm/aic-fpm-modeling-plan.md
liyuanzhe1991: Agreed — documented in 229036bf: the location section now states the V1 exact-version-only boundary explicitly (including why reusing `_build_op_sources()` would be unsafe for an atomic whole-model pair) and records the Collector-V3 managed-family follow-up with the strict reuse semantics you outlined (exact first; explicit same-backend reuse of one complete pair via `reuse.yaml`; no implicit nearest-earlier per-shape fill; no cross-backend reuse; end-to-end FPM validation before approving a reuse declaration).

aic-core/src/aiconfigurator_core/sdk/perf_interp/config.py
liyuanzhe1991: Applied in 229036bf — your "site" definition now leads the docstring, with the orphan-site rationale kept after it.

docs/cli_user_guide.md
liyuanzhe1991: Applied in 229036bf with your suggested wording, inline in the `--forward-model` bullet.

tests/unit/sdk/test_opspec_coverage.py
liyuanzhe1991: Right — and the guard is built so this cannot slip through silently: the EXEMPT list in `test_opspec_coverage.py` carries a staleness assertion (an entry naming an op that HAS a conversion branch fails the test), so whichever of this PR / #1461 merges second will hit a hard CI failure on exactly this line until it is removed. That one-line removal is queued as part of the second merge.

liyuanzhe1991:
@jasonqinzhou thanks for the thorough review — all four findings and the follow-up comments are addressed:

- **Findings 1/3/4** (exact-model matching, recommend wiring, loader identity pinning): 52cbe552, with the sidecar commit-record gate from your follow-up in 229036bf.
- **Finding 2** (backend policy in the requested identity) was fixed on **both sides of the data contract**. The collector has been reworked meanwhile — #1365 was closed and re-cut into a three-PR chain: #1473 (shared Generator/collector FPM contract table), #1474 (generator FPM target), #1475 (collector workflow). As part of that, the collector side also changed: `fpm_forward_perf` **schema v6** (#1475: 1aa9f9d8, 056f87c8) replaces the `backend_axis`/`backend_policy` label pair with four explicit identity columns (`moe_backend`/`attention_backend` as "auto"-or-pinned strings, `enable_wideep`/`enable_eplb` as real booleans). 7550e244 adopts v6 here: the knobs join the cell match identity like tp/quant, so a policy-changing config is a plain query-time data miss and future knob-pinned collections take effect with no modeling-side code change.
- **Docs/comment follow-ups** (D1 text, dangling reference, Collector-V3 ownership boundary, "site" definition, CLI engine-step wording): 229036bf.

Validation at head: unit + cross-package 3316 green, FPM suite 53 green, ruff clean; the Rust twin (#1461) carries the mirrored loader gates with engine-step parity 335 green. CI is 19/19 on both.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
```diff
@@ -0,0 +1,700 @@
...
+            _norm_identity(model_config.fmha_quant_mode),
+            _norm_identity(model_config.comm_quant_mode),
+            _norm_identity(model_config.kvcache_quant_mode),
+            _norm_identity(model_config.tp_size),
+            _norm_identity(model_config.pp_size),
+            _norm_identity(model_config.attention_dp_size),
+            _norm_identity(model_config.moe_tp_size if model_config.moe_tp_size is not None else 1),
+            _norm_identity(model_config.moe_ep_size if model_config.moe_ep_size is not None else 1),
+            _norm_identity(model_config.cp_size),
+            _norm_backend_request(getattr(model_config, "moe_backend", None)),
```
jasonqinzhou, line 521: **[P1] Keep FPM backend identity aligned with emitted deployments**

The paired schema-v6 collector currently supports explicit identities only for vLLM: `moe_backend` is applied through `--kernel-config` and EPLB through `--enable-eplb`. This request key instead reuses `ModelConfig.moe_backend`, whose common CLI contract is SGLang-only: `build_default_tasks` clears it for every non-SGLang backend, and `backend_config_mapping.yaml` maps it to `null` for vLLM (there is no generator mapping for `enable_eplb` either). I revalidated this at the current head: a pinned `flashinfer_cutlass` FPM row returned 8.0 ms, while the common vLLM task builder produced `moe_backend=None`. Normal CLI users therefore cannot select that collected cell, while direct SDK/YAML users can price a pinned runtime whose generated deployment omits the pin. Please introduce vLLM-specific structured fields that flow through task construction and deployment generation (or reject these cells until that path exists), with an end-to-end test from a producer row through FPM selection to the generated engine arguments.

tests/unit/sdk/test_fpm_forward.py
```diff
@@ -0,0 +1,600 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for forward_model="fpm": loader, FPMForwardOp, centralized
+model rewrite, and the explicit mixed-step branch.
+
+Synthetic parquet/metadata pairs are written directly from the documented
+``aic_fpm_forward_perf`` schema (v5) — deliberately NOT via collector code, so
```
jasonqinzhou: **[P2] Keep the contract-test description on schema v6**

This fixture now writes `schema_version: 6` and includes the four explicit backend identity columns, but its module docstring still says it exercises schema v5. Since the text specifically claims this suite doubles as the producer/consumer contract test, please name the format it actually validates.

```suggestion
``aic_fpm_forward_perf`` schema (v6) — deliberately NOT via collector code, so
```

jasonqinzhou (commented):
**[P2] Refresh the PR overview to the current FPM contract**

The PR description still presents `fpm_forward_perf` as schema v5 and says cell selection is exact-path-first with a unique-identity fallback. The current code intentionally rejects schema v5, uses the four schema-v6 backend identity columns, and requires the exact collected model path with no fallback. Please update the overview and implementation bullets so reviewers, release-note consumers, and future data producers see the contract that will actually merge.

jasonqinzhou:
Addressed the PR-overview follow-up in b941d029: the description now names schema v6, documents exact-model-path-only selection with no cross-model fallback, records the fail-closed vLLM deployment-identity boundary, and updates the focused FPM test count. The branch is also refreshed onto current main in fedd8e5d.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
jasonqinzhou: Fixed in b941d029. I took the fail-closed option: before loading/selecting a cell, FPM now rejects vLLM requests with pinned `moe_backend`, pinned `attention_backend`, or `enable_eplb=True`, because the standard Task-to-generator path cannot reproduce those engine settings yet. Schema-v6 rows remain producer-valid, while automatic backend selection with EPLB disabled continues to work. A parametrized producer-row-to-selection test covers all three rejected identities, and the updated focused FPM/CLI/CLI-workflow suite passes 118 tests.

tests/unit/sdk/test_fpm_forward.py
jasonqinzhou: Fixed in b941d029: the contract-test module docstring now names schema v6. The focused FPM/CLI/CLI-workflow suite passes 118 tests.

PR #1392, opened 2026-07-22
feat(sdk): Nemotron-3-Ultra-550B H100 support + NVFP4 software dequant

natoscott:
## Overview

Add H100 support for Nemotron-3-Ultra-550B: expand MoE search space to 16 GPUs and add NVFP4 software dequant fallback for SM >= 80 GPUs.

## Details

- **Expand MoE search space to 16 GPUs** so Nemotron-3-Ultra-550B-A55B-BF16 fits on 16xH100 (per HuggingFace recommended config). At TP=8 the per-rank weight memory is ~130 GB (exceeds H100 80 GB); at TP=16 it drops to ~66 GB.
- **Add NVFP4 software dequant fallback** for non-Blackwell GPUs (SM >= 80). NVFP4 models can run on Ampere/Hopper via Marlin FP4 kernels or software casting -- previously blocked as HW_INCOMPATIBLE. New `nvfp4_wo` quant mode models FP4 weight memory with BF16 compute speed.
- **GEMM uses the quant-transfer ladder** (from #1455): `(0.5625, 1)` util-level line added to `_GEMM_QUANT_UTIL_LEVEL` (Python + Rust). HYBRID/EMPIRICAL estimates via XPROFILE borrowing from bfloat16; SILICON rejects until real data exists. No normalize, no validation_aliases -- follows the recipe in `docs/design/how_to_add_a_gemm_quant_mode.md`.
- **MoE uses direct normalization** (`nvfp4_wo -> bfloat16` in `_normalize_moe_quant_mode_for_table`). This gives exact 1:1 lookup since nvfp4_wo runs the identical BF16 compute kernel. The alternative (removing normalization and adding a `(0.5625, 1)` MoE util-level line) would produce a ~43% rescaling via XPROFILE (`0.30 / 0.53`), which is less accurate for a mode that is literally the same kernel.
- **Fix NemotronH n_groups_per_gpu** to use `max(..., 1)` at TP > n_groups for correct weight estimates.

### Reviewer decision: MoE normalization vs transfer ladder

GEMM follows the #1455 transfer ladder. MoE does not -- it normalizes `nvfp4_wo -> bfloat16` at the query level instead. The trade-off:

| Approach | GEMM | MoE |
|---|---|---|
| Transfer ladder (XPROFILE) | Used. `(0.5625, 1)` level line in table. | Would need a level line too. Rescales at `e(nvfp4_wo)/e(bf16)` -- but `(0.5625, 1)` is not in `_MOE_QUANT_UTIL_LEVEL`, so it would hit the default (0.30), giving a 0.57x rescaling vs bf16. |
| Direct normalization | Removed per recipe. | Used. Exact 1:1 lookup -- no rescaling error. |

Happy to switch MoE to the ladder if you prefer consistency over accuracy here. Would require adding `(0.5625, 1): 0.45` to `_MOE_QUANT_UTIL_LEVEL` and its Rust mirror, and removing the normalize functions in `moe.py`, `moe.rs`, and `wideep_moe.rs`.

## Test plan

- [x] `cargo test` -- 286 passed
- [x] `pytest tests/unit/sdk/models/test_nvfp4_fallback.py` -- 5 passed
- [x] `pytest tests/unit/sdk/database/test_gemm_quant_transfer.py` -- 6 passed (including `test_level_table_covers_every_gemm_quant_profile`)
- [x] `pytest tests/unit/sdk/task_v2/test_task_config.py` -- nvfp4 remap tests pass
- [x] `ruff check` -- clean


tianhaox:
looks like the general idea is introducing a new nvfp4_wo quant mode to save GPU memory/bw but leveraging existing bf16 data. Do you consider adding this mode directly by adding the collector's support?
another question is are you expecting this mode to work with all models or just Nemontron-3-ultra on H100? The required effort will be impacted significantly.

otherwise, i would suggest considering supporting it only in Hybrid mode by borrowing quant. what you need to do is only adding a quant mode to the enumeration. This will be a clean approach, you don't need to normalize the quant, just use it and let Hybrid mode borrow the util and calulate the latency based on that for you. ( this approach may need to fix some bugs since the quant-borrowing method is not implemented for gemm yet)

natoscott:
> looks like the general idea is introducing a new nvfp4_wo quant mode to save GPU memory/bw but leveraging existing bf16 data. Do you consider adding this mode directly by adding the collector's support? another question is are you expecting this mode to work with all models or just Nemontron-3-ultra on H100? The required effort will be impacted significantly.
> 
> otherwise, i would suggest considering supporting it only in Hybrid mode by borrowing quant. what you need to do is only adding a quant mode to the enumeration. This will be a clean approach, you don't need to normalize the quant, just use it and let Hybrid mode borrow the util and calculate the latency based on that for you. ( this approach may need to fix some bugs since the quant-borrowing method is not implemented for gemm yet)

@tianhaox Good points.  The issue is general, not model-specific - any NVFP4 checkpoint on non-Blackwell hardware will software-dequant to BF16 at runtime I believe.  We already have 10+ NVFP4 models in DefaultHFModels: DeepSeek-V3.1-NVFP4, Kimi-K2.5-NVFP4, GLM-5/5.1/5.2-NVFP4, Qwen3-235B-A22B-NVFP4, MiniMax-M2.5/M2.7-NVFP4, Nemotron Super and Ultra - all can potentially benefit from this.

That suggestion to use the HYBRID quant-borrowing does seem cleaner.  However, digging into it further I found XQUANT can't support nvfp4_wo - its (memory=0.5625, compute=1) profile is unique, so there's no same-profile quant to borrow from. XPROFILE could work for MoE (already implements the transfer cascade) by adding nvfp4_wo to _MOE_QUANT_UTIL_LEVEL at the bfloat16 level - giving a 1:1 ratio since the compute kernel is identical.

However GEMM doesn't implement any transfer tiers (XSHAPE/XQUANT/XPROFILE), so supporting nvfp4_wo via XPROFILE there would require building the full GEMM transfer cascade first - significantly more work than the current 2-line normalize function, and a change that should probably be designed and validated independently?

For this PR, the normalization (nvfp4_wo -> bfloat16 for table lookup) is semantically exact - same compute kernel, same perf data (I updated the GEMM and MoE docstring to cover this).  I do agree the GEMM transfer cascade would be a useful follow-up benefiting future quant modes.

tianhaox:
why I prefer using cross quant to borrow in hybrid mode only is becasue 1. it's actually emprical mode becasue we never collect the exact kernels. Once we have the real data path, we can have direct hit in default silicon mode. 2. current implementation will create a reference fallback path which is difficult to maintain in future. I think the cross-quant effort is something that I missed before. Can we move forward with some work split, i will create the PR to add the xquant capability for GEMM to make the mechanism complete. And you can rebase to that PR. Does it make sense? 

#1455 is ready for review.

natoscott:
> Can we move forward with some work split, i will create the PR to add the xquant capability for GEMM to make the mechanism complete. And you can rebase to that PR. Does it make sense?

@tianhaox sounds good to me.

tianhaox:
> > Can we move forward with some work split, i will create the PR to add the xquant capability for GEMM to make the mechanism complete. And you can rebase to that PR. Does it make sense?
> 
> @tianhaox sounds good to me.

Great! Thanks for the feedback. #1455 has been merged.

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -995,23 +996,25 @@ def _resolve_quant_modes(self) -> None:
...
+        # NVFP4 software fallback: on non-Blackwell systems, remap nvfp4 to
+        # nvfp4_wo (FP4 weight memory, BF16 compute speed) so the perf model
+        # queries BF16 data instead of native FP4 data.
+        for role in roles:
+            system = self._role_attr(role, "system_name")
+            if not is_blackwell_system(system):
+                if self._role_attr(role, "gemm_quant_mode") == common.GEMMQuantMode.nvfp4:
+                    self._set_role_attr(role, "gemm_quant_mode", common.GEMMQuantMode.nvfp4_wo)
+                if self._role_attr(role, "moe_quant_mode") == common.MoEQuantMode.nvfp4:
+                    self._set_role_attr(role, "moe_quant_mode", common.MoEQuantMode.nvfp4_wo)
```
jasonqinzhou, line 1023: **[P1] Make MoE validation match the direct table alias without changing GEMM’s transfer-ladder contract.** `_check_role_against_db()` still compares the raw `nvfp4_wo` name with each support list. On this exact head, Nemotron Ultra/H100/TensorRT-LLM 1.3.0rc20 resolves both modes to `nvfp4_wo`, but default HYBRID and EMPIRICAL validation fail on MoE before estimation. Please add an op-scoped `nvfp4_wo -> bfloat16` validation alias only for paths that actually normalize their table lookup (standard MoE), while preserving GEMM’s documented data-less/SILICON-rejecting transfer-ladder behavior. Add an end-to-end `Task.validate()`/run regression for the intended H100 configuration.

aic-core/rust/aiconfigurator-core/src/perf_database/moe.rs
```diff
@@ -165,7 +165,8 @@ impl MoeTable {
     ) -> Result<f64, AicError> {
         let loaded = self.load()?;
         let grids = &loaded.default;
-        let quant_name = quant.name();
+        let lookup_quant = normalize_moe_quant(quant);
```
jasonqinzhou, line 169: **[P1] Carry this normalization through Rust EMPIRICAL standard-MoE lookup.** `MoeTable::query()` now aliases the direct SILICON lookup to BF16, but `MoeOp::slice_points()` and sibling/reference enumeration still pass the raw `self.quant_mode.name()`. On this exact head, a MiniMax-M2.5 H200/vLLM 0.19.0 mixed-step probe with BF16 GEMM and `nvfp4_wo` MoE returns `36.083148 ms` in Python while Rust raises `EmpiricalNotImplementedError`. Please centralize the table quant name across direct, own-slice, and sibling/reference access while preserving the raw mode for SOL and FP4 weight accounting, and add an exact Python/Rust parity regression.

aic-core/src/aiconfigurator_core/sdk/operations/moe.py
```diff
@@ -2043,6 +2063,8 @@ def _normalize_quant_mode_for_table(
         """
         if quant_mode == common.MoEQuantMode.fp8_block:
             return common.MoEQuantMode.fp8
+        if quant_mode == common.MoEQuantMode.nvfp4_wo:
```
jasonqinzhou, line 2071: **[P1] Handle Python TensorRT-LLM WideEP compute as consistently as dispatch.** This addition normalizes `TrtLLMWideEPMoEDispatch`, and Rust WideEP compute is normalized, but Python `TrtLLMWideEPMoE._query_compute_table()` still indexes the raw `nvfp4_wo` mode in both SILICON and EMPIRICAL paths. The exact-head probe raises `PerfDataNotAvailableError` for the raw SILICON slice and `EmpiricalNotImplementedError` with EMPIRICAL/transfers off. If this runtime combination is supported, apply the BF16 table alias consistently to compute; otherwise reject it before model construction. Please cover the chosen contract end to end.

tools/support_matrix/support_matrix.py
```diff
@@ -380,6 +381,11 @@ def _gpu_supports_datatype(system: str, system_spec: dict, datatype: str) -> boo
         return "fp8_tc_flops" in gpu_spec or (sm_version is not None and sm_version >= 89)
     if datatype == "FP4":
         return "fp4_tc_flops" in gpu_spec or (sm_version is not None and sm_version >= 100)
+    if datatype == "FP4_SWDEQUANT":
+        # nvfp4 software dequant (Marlin FP4 / casting) on SM >= 80
+        if sm_version is not None and sm_version >= 80:
```
jasonqinzhou: **[P1] Gate software-dequant support by system, backend, and version rather than SM alone.** This currently admits every SM80+ backend/version, and `get_hardware_incompatibility()` has no backend-version input. That overstates runtime capability: vLLM documents Hopper dequant/QDQ emulation as a v0.21.0 addition (https://github.com/vllm-project/vllm/releases/tag/v0.21.0), while TensorRT-LLM’s hardware matrix does not list Hopper NVFP4 support (https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/features/quantization.md). Please centralize an explicit capability predicate across Task resolution, `resolve_nvfp4_for_system()`, and support-matrix preflight, with old-version and unsupported-backend rejection tests.

aic-core/src/aiconfigurator_core/sdk/models/nemotron_h.py
```diff
@@ -149,7 +149,7 @@ def _build_context_ops(self) -> None:
                         head_dim=cfg.mamba_head_dim,
                         d_state=cfg.ssm_state_size,
                         d_conv=cfg.conv_kernel,
-                        n_groups=cfg.n_groups,
+                        n_groups=n_groups_per_gpu,
```
jasonqinzhou: **[P2] Pin the TP=16 Mamba group clamp with a committed regression.** The implementation now passes the clamped per-GPU value into all four context/generation kernels, and an exact construction with `n_groups=8`, `tp_size=16` produces `_n_groups == 1`. No committed test exercises that boundary, however. Please add context and generation assertions covering the constructed `Mamba2Kernel` operations so a later refactor cannot silently restore the zero-group estimate.

jasonqinzhou (commented):
Exact-head follow-up review at `18baa59209256b1689500b326b1102b4eeb15d10`: four blocking correctness/capability issues and one regression-coverage gap remain. I am intentionally not requesting direct GEMM `nvfp4_wo -> bfloat16` table normalization: the merged GEMM transfer-ladder design requires HYBRID/EMPIRICAL borrowing while SILICON continues to reject data-less modes.

tianhaox:
Hi Nathan, please feel free to let us know when this PR is ready for review. Thanks

natoscott:
@tianhaox @jasonqinzhou — this is ready for re-review. All five items from Jason's review are addressed in the latest head (`e7ea64af`):

**P1 — MoE validation alias:** Added op-scoped `nvfp4_wo -> bfloat16` alias in `_check_role_against_db` for MoE/WideEP ops only. GEMM continues to use the transfer ladder (no alias). Covered by `test_validate_nvfp4_wo_moe_alias_admitted_in_hybrid` (HYBRID passes, SILICON rejects on GEMM).

**P1 — Rust EMPIRICAL MoE normalization:** `normalize_moe_quant` is now `pub(crate)` and applied in `MoeOp::slice_points()` and `collect_candidates()` (sibling enumeration). Parity oracle test added: `moe_nvfp4_wo_normalizes_to_bfloat16_matches_python_oracle` pins 4.144185638427734 ms on h200/vllm/0.19.0.

**P1 — Python WideEP compute normalization:** `TrtLLMWideEPMoE._query_compute_table()` now normalizes `nvfp4_wo -> bfloat16` via `table_quant_mode` in both SILICON and EMPIRICAL paths (cache key updated to match).

**P1 — Backend+version gate:** Centralized `check_nvfp4_swdequant()` predicate used across Task resolution, `resolve_nvfp4_for_system()`, and support-matrix preflight. Gates: vLLM >= 0.21.0, SGLang >= 0.5.3; TRT-LLM rejected on non-Blackwell. Tests cover each combination including old-version rejection.

**P2 — Mamba clamp regression:** `test_nemotron_h_mamba.py` pins TP=16/n_groups=8 -> n_groups_per_gpu=1 across both context and generation Mamba2Kernel ops.

CI is all green except the e2e timeout (GitHub Actions cancellation at 92%, same infrastructure issue as main).

tianhaox:
@natoscott @jasonqinzhou stepping back from spot-fixes for a moment — I'd like to propose a simpler overall shape for this PR, and check whether the direction makes sense to you both.

**How I think about it:** AIC's quant support separates two questions.

**(A) How fast / how much memory** — pure physics, derived from the quant's `(memory, compute)` profile. The transfer ladder exists exactly for quants without collected data: latency = SOL(own profile) / borrowed util. It's version-free and quant-agnostic; a new quant costs one enum entry plus one util-level line.

**(B) Whether a given framework build can load a checkpoint** — deployability, which belongs to the support matrix.

`nvfp4_wo` fits (A) cleanly: it's a weight-only profile, and its validity comes from physics rather than from any framework implementation — same as every other quant we model. On non-Blackwell, every runtime path computes at BF16 rate (hence `compute=1`), and they differ only in weight traffic: Marlin reads ~0.56x with fused in-kernel dequant, while the QDQ emulation path moves ~2.5x per forward. So BF16 table data isn't exact for either path (pessimistic vs Marlin, optimistic vs emulation) — the ladder with `memory=9/16` in SOL models the Marlin-class behavior, which I think is the right thing to represent. Memory is 9/16 on all paths, so that half is settled.

**Concretely, three changes — all net deletions:**

1. **MoE through the ladder, like GEMM.** Add `(0.5625, 1)` at the bfloat16 level to `_MOE_QUANT_UTIL_LEVEL` / Rust `MOE_QUANT_UTIL_LEVEL`; remove `_normalize_moe_quant_mode_for_table`, `normalize_moe_quant`, `normalize_wideep_moe_quant` and their call-site threading. Validation then admits `nvfp4_wo` via the existing profile-reachability path, so the op-scoped alias can go too. Beyond consistency, this matters for accuracy: the motivating NVFP4 models are all MoE, so expert weights dominate decode traffic — right now the dominant term is the one *not* modeling the memory-bound win. (The normalize gaps found across review rounds also suggest aliasing at every query site is easy to leak; the ladder removes that class of bug.)
2. **Make the remap unconditional on non-Blackwell.** No FP4 tensor cores is a hardware fact; the same H100 resolving different quants for `version=0.19` vs `0.22` is hard to justify. Drop `check_nvfp4_swdequant` from `resolve_nvfp4_for_system` and `_resolve_quant_modes`.
3. **Drop the version constants from the modeling layer.** The support matrix keeps one backend fact (TRT-LLM can't load NVFP4 on non-Blackwell); a version floor there is optional and low-stakes. No need to research which versions select which kernels — the perf model doesn't depend on it.

@jasonqinzhou on your review: the Mamba regression pin is great — keep it. Your gate ask was scoped to the support matrix and I agree with it at that scope; (2)/(3) just narrow the predicate back there. Items 1–3 were correct audits of the aliasing approach; with change (1) they become unnecessary rather than fixed — which is really the argument for it.

Happy to discuss — if you're both onboard, I believe the resulting diff ends up quite small. WDYT?

natoscott:
@tianhaox I like the cut of your jib.  That seems a much cleaner approach.  I've posted an implementation over in PR #1524 which looks good to me, what do you think?

tianhaox:
Thanks Nathan, replied to the new PR.

natoscott:
Closing in favour of #1524

PR #1399, opened 2026-07-22
feat(data): add vLLM 0.24.0 native DeepSeek-V4 w4a8_mxfp4_mxfp8 MoE silicon data for Blackwell

tianhaox:
## Summary

Data-only PR: adds `w4a8_mxfp4_mxfp8` MoE silicon data for the **native (FP4 routed-expert) DeepSeek-V4 checkpoints** on vLLM 0.24.0, covering four Blackwell systems, plus the corresponding support-matrix refresh.

- `moe_perf.parquet` for **b200_sxm / b300_sxm / gb200 / gb300** `vllm/0.24.0`:
  - deepseek-ai/DeepSeek-V4-Flash (4096/2048/topk6/e256): 1134 rows per system
  - deepseek-ai/DeepSeek-V4-Pro (7168/3072/topk6/e384): 1053 rows per system
  - kernel_source: `vllm_mxfp4moe_flashinfer_trtllm_mxfp4_mxfp8_trtllmmxfp4expertsmodular` (flashinfer trtllm-gen MXFP4xMXFP8, the kernel vLLM 0.24.0 dispatches for this artifact on SM100)
- Support matrix: 16 rows flip **HYBRID_PASS -> PASS** (Flash/Pro x 4 systems x agg/disagg), regenerated locally with `generate_support_matrix.py` filtered runs.

## Provenance

Collected with the collector changes from #1397 (branch head `818a6feb`) via aic-auto-collector pipelines (8/8 collect jobs green, `Total errors: 0`, parquet finalize confirmed in every trace). This PR intentionally carries **only data + support matrix**; the collector-side enablement (case gating, quant-mode interval, tests) stays in #1397.

## Verification

- Per-artifact acceptance: label/kernel/shape uniform, 27 token points, full tp/wideep-ep grid, sane latencies.
- Cross-check vs SGLang 0.5.14 b200 same-kernel rows (`sglang_mxfp4_flashinfer_trtllm_moe`): vLLM/SGLang latency ratio median 1.09-1.10 on overlapping grid points.
- Merge integrity: per-system row counts exactly base + new (e.g. b200_sxm 58326 + 2187 = 60513).
- End-to-end: `Task.validate()` now passes for both native checkpoints on all four systems with `moe=w4a8_mxfp4_mxfp8` (previously: unsupported-moe-quant-mode error).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1400, opened 2026-07-22
fix: tighten DeepSeek-V4 validation coverage

jasonqinzhou:
## Summary

- cap vLLM's native DeepSeek-V4 `w4a8_mxfp4_mxfp8` collector case at the SM100 capability family (`100 <= SM < 110`)
- keep missing native DeepSeek-V4 data on the supported vLLM 0.24.0 Blackwell matrix combinations as `FAIL`, while preserving framework-incompatible handling elsewhere
- make the MoE sanity chart select shapes that actually exist in the loaded table, preserving the historical preferred shape when available and plotting both DeepSeek-V4 Flash and Pro profiles otherwise

## Context

Follow-up to #1397 and #1399. Both changes are sound overall; these guards close the boundary and validation blind spots found during review.

## Validation

- `ruff check` and `ruff format --check` on changed Python files
- 86 relevant unit tests passed
- b200/vLLM 0.24.0 `validate_database.ipynb` smoke test passed
- actual merged vLLM 0.24.0 data selected both Flash and Pro profiles on `b200_sxm`, `b300_sxm`, `gb200`, and `gb300`
- notebook JSON validated with `jq empty`

## Summary by CodeRabbit

- **Support Matrix**
  - Refined DeepSeek-V4 vLLM 0.24.0+ compatibility handling by using system GPU/SM capability details to classify failures accurately.
  - Added/expanded unit coverage for native Blackwell gap failures, Hopper preflight incompatibility, and SM120 non-native behavior.

- **Compatibility**
  - Tightened native `w4a8_mxfp4_mxfp8` MoE gating cutoff from SM109 to SM110+ (now disallowed starting at SM110).

- **Charts and Validation**
  - Enhanced MoE chart profile selection with preferred-profile handling, optional predicate filtering, and result limiting.
  - Improved chart generation to iterate over candidate profiles, standardize labeling, and improve portability.


tianhaox (commented):
Reviewed all three changes (collector SM cap, support-matrix carve-out narrowing, MoE chart profile selection). Direction is right and the test coverage is appreciated; a few boundary issues below, roughly in severity order.

**1. `tools/support_matrix/support_matrix.py:262` — the `version == "0.24.0"` + hardcoded system set will silently mask regressions on the next version bump.**
Versions are auto-discovered from `get_supported_databases()`, so when vLLM 0.25.0 data lands this condition stops matching and a genuine data-path regression on Blackwell gets classified FRAMEWORK_INCOMPATIBLE and rescued by HYBRID — the exact failure class this PR wants to keep as FAIL, now failing *open*. This guard has already churned three times (pin 0.19.0 → version-agnostic → pin 0.24.0; the deleted comment documented the earlier lesson). Suggested fix, cheap enough to do here:
- Replace the system-name frozenset with the same SM predicate the collector yaml uses: pass `system_spec` in (it's already in scope in `run_single_test`, and `_gpu_supports_datatype` already reads `system_spec["gpu"]["sm_version"]`) and check `100 <= sm_version < 110`.
- Replace `version == "0.24.0"` with a parsed `version >= (0, 24)` comparison ("native data exists from 0.24.0 onward"). That flips the staleness direction: if a future version's data is broken it fails loudly (FAIL) instead of being silently whitelisted.
- Add tests for the staleness direction the current tests miss: `(b200_sxm, "0.25.0")` → FAIL, `(b200_sxm, "0.19.0")` → FRAMEWORK_INCOMPATIBLE.
Note that introspecting the *loaded* database ("is the label present?") does not work here — the error fires precisely when the label is absent, so it's circular; an independent expectation (SM interval + minimum version) is needed.

**2. `collector/cases/base_ops/moe.yaml:157-166` — the analogous SGLang `w4a8_mxfp4_mxfp8` entry was left at `max_sm_exclusive: 120` with the same SM100-family-only rationale.**
Its own comment cites flashinfer_mxfp4 accepting only SM90 (cutlass) / SM100 (trtllm-gen), its backends map lists only 100/103 (`moe.yaml:116`), and `tests/unit/collector/sglang/test_collect_moe_population.py:382` says "SM100/103-only" — yet the gate still admits SM110-119. Either tighten it to 110 with the same reasoning as the vLLM entry, or add a comment saying why SGLang deliberately stays at 120. Also `collector/cases/models/DeepseekV4ForCausalLM_cases.yaml:47` still says "min_sm 100 / max_sm_exclusive 120" about the vLLM base entry this PR changed — now factually stale.

**3. `tools/sanity_check/moe_chart_profiles.py:75` + notebook — divisibility filtering happens *after* profile selection, so selected slots can be silently discarded.**
`select_moe_chart_profiles` returns `[preferred]` exclusively (or truncates to top-2) before the notebook's `(inter_size // tp) % (256 // weight_bits)` guard runs. Concrete trigger: `w4a16_mxfp4`'s preferred shape has inter_size=2880, and 2880/tp fails the %64 check at every tp>1 — so at (tp=4, ep=1) / (tp=8, ep=1) on a vLLM/SGLang database the selector returns the preferred alone, the guard skips it, and the quant mode is silently unplotted while the new fallback machinery never engages. (The empty column predates this PR, but the refactor's stated goal is selecting shapes that actually plot.) Fix: apply the divisibility predicate before/inside selection (e.g. pass a `predicate` into `select_moe_chart_profiles`), so filtered-out shapes don't consume the `max_profiles` slots.

**4. Module boundary (noting for the record; waiving for this PR).**
`.claude/rules/collector/layer_permissions.md` says a collector task may modify `collector/` + `tests/unit/collector/` and nothing else, with bundling allowed only for approved data-contract changes. This PR carries the collector yaml change together with `tools/support_matrix/` and `tools/sanity_check/` changes in one commit. Waived here since the substance is support-matrix classification and the yaml edit aligns an existing bound rather than changing collection behavior — but future PRs should split these.

**5. `tests/unit/support_matrix/test_framework_incompatibility.py:88` — the mHC test fakes the error as a plain `ValueError`, but production raises `PerfDataNotAvailableError` (a RuntimeError subclass).**
The real exception makes the attempt rescuable via `has_perf_data_not_available_cause`, exercising the HYBRID-retry branch the fake never reaches. The end state is still FAIL (the HYBRID retry reads the same missing table and also fails), so no behavior bug — but raising the real exception type in the fake would make the "must remain FAIL" assertion actually cover the production path.

**6. `tools/sanity_check/moe_chart_profiles.py:50-73` — the defensive layers are dead code against the only data source, and internally inconsistent.**
`load_moe_data` builds `_moe_data` as a uniform nested dict with int keys cast at load time (single writer; `LoadedOpData` freezing preserves types), so the four `isinstance(Mapping)` guards, `_mapping_at`, and the `int(...)` key casts can never fire. Worse, if a non-numeric key ever did appear, `int(key)` would crash with ValueError — the "defense" converts unexpected data into a crash rather than a skip. Plain nested `.items()` loops with `.get(moe_tp_size, {}).get(moe_ep_size, {})` are simpler and no less safe.

**7. Minor: cross-panel color stability in `visualize_moe` is lost.**
Old code keyed color on the quant_mode's enumerate index (stable across all panels); new code resets `color_id` per panel and increments only on a successful plot, so the same quant mode changes color between panels and the same color denotes different modes. Per-panel legends keep it correct, and the new scheme rightly avoids two same-colored profiles within one panel — a stable `(quant_mode, profile)` → color mapping would get both.

Also verified and *not* flagging: the preferred-profile short-circuit ignoring token coverage is documented + test-asserted intent, and single-point-overflow extrapolation in `query_moe` handles it without exceptions; the "mHC data not loaded" case does remain FAIL in production despite being HYBRID-rescuable (the retry fails on the same missing table).

jasonqinzhou:
@tianhaox Thanks for the detailed review. I addressed all actionable points in `3f69d57`:

- Switched the vLLM carve-out to `>=0.24.0` plus the SM100–109 predicate, with 0.19/0.25 boundary coverage.
- Tightened the SGLang gate to `<110` and corrected the stale comment.
- Moved divisibility filtering into profile selection.
- Exercised the real perf-data exception and accurate per-system hardware specs.
- Removed the unnecessary defensive traversal layers.
- Made `(quant_mode, profile)` colors stable across panels.

I left the module-boundary point unchanged as waived for this PR. The relevant local suites and formatting checks pass. Could you take another look when you have time?

tianhaox (commented):
Second-round review: verified all seven round-1 points are correctly addressed in `3f69d57` — version/SM predicate with fail-closed staleness direction and boundary tests, both SGLang gates tightened to `<110` with citations updated, divisibility predicate applied before preferred/max_profiles selection, real `PerfDataNotAvailableError` exercised (the `attempts == 4` assertion matches agg+disagg × SILICON+HYBRID), defensive layers removed, and `(quant_mode, profile)`-keyed colors stable across panels. Thanks for the thorough follow-up.

One issue introduced by the fix commit, then this is good to go:

**`tools/support_matrix/support_matrix.py` — the new predicate dereferences `system_spec` unguarded, turning a recorded FAIL into an `AttributeError` crash when `system_spec` is `None`.**

The new signature declares `system_spec: dict`, but the real call chain allows `None`: `run_single_test` defaults to `system_spec: dict | None = None`, and when `get_database(...)` returns `None` it explicitly keeps `system_spec = None` (the `else None` at the database-lookup fallback shows this is a supported path; hardware preflight is skipped for the same reason). `_run_mode` then raises on the missing database, and inside `_attempt`'s except-handler classification the new code runs

```python
sm_version = (system_spec.get("gpu") or {}).get("sm_version")
```

which raises `AttributeError: 'NoneType' object has no attribute 'get'` out of the except handler — the old code recorded `STATUS_FAIL` with the raw error on this path; the new code crashes the whole `run_single_test` call. One-line fix:

```python
sm_version = ((system_spec or {}).get("gpu") or {}).get("sm_version")
```

plus updating the parameter annotation to `dict | None`. A small regression test pinning `system_spec=None` → still classified (FRAMEWORK_INCOMPATIBLE, not a crash) would keep this from regressing. (CodeRabbit's outside-diff comment points at the same gap.)

tianhaox (approved):

PR #1406, opened 2026-07-23
docs: clarify forward-pass API selection

jasonqinzhou:
## What changed

- add a short decision guide to `aic-core/API.md`
- recommend `RustForwardPassPerfModel.best_available(...)` / `ForwardPassPerfModel::best_available(...)` for adaptive forward-pass modeling
- distinguish strict-native `from_native(...)` from `AicEngineBuilder`
- show how to inspect the selected source and fallback warning through `diagnostics()`
- align Rust perf-database tests with the current family-first data layout

## Why

The public API contract listed the adaptive forward-pass model and compiled engine surfaces, but did not explain which one users should choose. The newly enforced core API-contract job also exposed test fixtures that still referenced the legacy perf-data layout even though the loader already supports the family-first layout.

## User impact

Users can now select the adaptive model for native AIC estimates, online correction, and eligible regression fallback, while reserving `AicEngineBuilder` for strict native prefill/decode/mixed-step engine calls. The public API-contract test now validates the current perf-data layout.

## Validation

- `git diff --check`
- `cargo test --workspace --no-default-features` — 198 core tests and 3 external public-API consumer tests passed

## Summary by CodeRabbit

- **Documentation**
  - Added guidance on selecting a forward-pass API, including how diagnostics report the active source and when regression fallback is used.
  - Clarified the role of native-only forward-pass loading versus adaptive “best available” selection.
  - Included a Python example demonstrating diagnostics and estimating forward-pass time with fallback.

- **Tests**
  - Updated performance database tests to verify resolved op-source file paths instead of only legacy data-root layout checks.

- **Chores**
  - Corrected test documentation/comments and adjusted GEMM test dataset path construction for the expected NCCL/GEMM directory layout.

### DIFF
--- a/aic-core/API.md
+++ b/aic-core/API.md
@@ -72,8 +72,13 @@ replacement for `best_available(...)`.
 ```python
 from aiconfigurator_core.sdk import RustForwardPassPerfModel
 
+# Engine-config and per-rank FPM dictionary setup is omitted here.
 model = RustForwardPassPerfModel.best_available(config)
 diagnostics = model.diagnostics()
+print(diagnostics["source"])
+if diagnostics["last_warning"] is not None:
+    print(diagnostics["last_warning"])
+
 estimate_ms = model.estimate_forward_pass_time_ms(metrics_by_rank)
 ```
 
--- a/aic-core/API.md
+++ b/aic-core/API.md
@@ -55,19 +55,22 @@ Python modules carry their own annotations.
 For adaptive forward-pass modeling, use
 `RustForwardPassPerfModel.best_available(...)` from Python or
 `ForwardPassPerfModel::best_available(...)` from Rust. This path uses the
-native AIC estimate when the engine configuration is supported, learns online
-correction factors from FPM observations, and falls back to regression for an
-eligible native-unsupported configuration. Check `diagnostics()` to determine
-whether the active source is `aic`, `aic_with_correction`, or
-`fallback_regression`, and to inspect any fallback warning.
+native AIC estimate when the native estimator can be built, learns online
+correction factors from FPM observations, and falls back to regression for
+eligible native build or data-availability failures. These include unsupported
+models and missing or unreadable model, system, or performance data. Check
+`diagnostics()` to determine whether the active source is `aic`,
+`aic_with_correction`, or `fallback_regression`, and to inspect any fallback
+warning.
 
 Use `from_native(...)` instead when native AIC support is required and an
-unsupported configuration should fail rather than fall back.
+unsupported configuration or native data failure should surface rather than
+fall back.
 
 `AicEngineBuilder` serves a different purpose: it constructs the strict native
-Rust engine for direct prefill, decode, and mixed-step latency calls. It does
-not provide regression fallback or online correction, so it is not a
-replacement for `best_available(...)`.
+Rust engine for direct public prefill and decode latency calls. It does not
+provide regression fallback or online correction, so it is not a replacement
+for `best_available(...)`.
 
 ```python
 from aiconfigurator_core.sdk import RustForwardPassPerfModel
@@ -80,6 +83,12 @@ if diagnostics["last_warning"] is not None:
     print(diagnostics["last_warning"])
 
 estimate_ms = model.estimate_forward_pass_time_ms(metrics_by_rank)
+if estimate_ms is None:
+    # Regression fallback starts without observations for each workload kind.
+    # Supply observed FPM iterations with positive wall_time until the configured
+    # min_observations threshold is reached, then retry the estimate.
+    model.tune_with_fpms(observed_iterations)  # Observed-iteration setup omitted.
+    estimate_ms = model.estimate_forward_pass_time_ms(metrics_by_rank)
 ```
 
 ## Stable Rust facade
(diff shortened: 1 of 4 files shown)


aic-core/API.md
```diff
@@ -50,6 +50,33 @@ extension contract:
...
+not provide regression fallback or online correction, so it is not a
+replacement for `best_available(...)`.
+
+```python
+from aiconfigurator_core.sdk import RustForwardPassPerfModel
+
+model = RustForwardPassPerfModel.best_available(config)
+diagnostics = model.diagnostics()
+estimate_ms = model.estimate_forward_pass_time_ms(metrics_by_rank)
+```
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in `09708501`. The example now explicitly marks the engine-config and per-rank FPM setup as omitted, and it demonstrates inspecting both `diagnostics["source"]` and `diagnostics["last_warning"]` before estimating the iteration.

aic-core/API.md
```diff
@@ -50,6 +50,38 @@ extension contract:
...
+eligible native-unsupported configuration. Check `diagnostics()` to determine
+whether the active source is `aic`, `aic_with_correction`, or
+`fallback_regression`, and to inspect any fallback warning.
+
+Use `from_native(...)` instead when native AIC support is required and an
+unsupported configuration should fail rather than fall back.
+
+`AicEngineBuilder` serves a different purpose: it constructs the strict native
+Rust engine for direct prefill, decode, and mixed-step latency calls. It does
+not provide regression fallback or online correction, so it is not a
```
tedzhouhk: [P1] Do not advertise mixed-step as a public Rust builder API

`AicEngineBuilder::build()` returns the public `AicEngine`, whose inherent Rust API exposes only `prefill_latency_ms` and `decode_latency_ms`. `mixed_step_latency` is currently a private method inside `#[pymethods]`, so downstream Rust crates cannot make the direct mixed-step call documented here. Either limit this guidance to the public prefill/decode calls or expose a public Rust mixed-step method first.

aic-core/API.md
```diff
@@ -50,6 +50,38 @@ extension contract:
...
+
+# Engine-config and per-rank FPM dictionary setup is omitted here.
+model = RustForwardPassPerfModel.best_available(config)
+diagnostics = model.diagnostics()
+print(diagnostics["source"])
+if diagnostics["last_warning"] is not None:
+    print(diagnostics["last_warning"])
+
+estimate_ms = model.estimate_forward_pass_time_ms(metrics_by_rank)
+```
```
tedzhouhk, line 92: [P1] Document that regression fallback is initially unable to estimate

`best_available()` can return a regression model with zero observations. For a non-empty workload, `estimate_forward_pass_time_ms()` then returns `None` until the matching prefill/decode/mixed store reaches `min_observations` through `tune_with_fpms()`. As written, this example implies `estimate_ms` is immediately available after fallback. Please show the readiness/tuning step or explicitly handle and explain the `None` result.

aic-core/API.md
```diff
@@ -50,6 +50,38 @@ extension contract:
 The wheel includes `py.typed` and a stub for that native extension. The SDK
 Python modules carry their own annotations.
 
+## Choosing a forward-pass API
+
+For adaptive forward-pass modeling, use
+`RustForwardPassPerfModel.best_available(...)` from Python or
+`ForwardPassPerfModel::best_available(...)` from Rust. This path uses the
+native AIC estimate when the engine configuration is supported, learns online
+correction factors from FPM observations, and falls back to regression for an
```
tedzhouhk: [P1] Describe the full set of conditions that trigger regression fallback

`can_fallback_to_regression()` catches `DataRoot`, `ModelConfig`, `PerfDatabase`, `Io`, and `Parquet` errors in addition to `UnsupportedModel`. Consequently, even a supported engine configuration can enter regression when native data is missing or unreadable. Please describe this as native build/data unavailability, or narrow the implementation if fallback is intended only for unsupported configurations.

tedzhouhk (commented):
Three P1 API-contract issues need to be resolved before this guidance is reliable.

tedzhouhk (approved):

aic-core/API.md
jasonqinzhou: Fixed in `cb9c6611`. The Python example now handles an initial `None` estimate, explains that regression fallback starts without observations for each workload kind, calls `tune_with_fpms()` with observed FPMs containing positive `wall_time` until `min_observations` is reached, and then retries the estimate. All 16 PR checks pass, including the aic-core public API contract.

aic-core/API.md
jasonqinzhou: Fixed in `cb9c6611`. I removed mixed-step from the `AicEngineBuilder` guidance; it now documents only the public Rust prefill and decode latency calls exposed by `AicEngine`. I also verified the corrected contract with the external public-API tests. All 16 PR checks pass.

aic-core/API.md
jasonqinzhou: Fixed in `cb9c6611`. The guidance now describes regression fallback as covering eligible native build or data-availability failures, explicitly including unsupported models and missing or unreadable model, system, or performance data. The strict `from_native(...)` guidance now also says native data failures surface instead of falling back. All 16 PR checks pass.

PR #1408, opened 2026-07-23
chore: promote B200, H200, and GB200 power overlays with coverage gating

kaim-eng:
## Summary

- promote 65 collected latency-and-power parquet overlays for B200, H200, and GB200
- place every overlay in the Collector V3 family-first layout
- add 41 honest legacy-provenance sidecars for artifacts that predate full Collector V3 hashes
- replace four `SHARED_LAYER_REUSE.txt` markers with direct silicon data
- add first-party GB200 TRT-LLM `1.3.0rc17` coverage
- remove six physically invalid zero-latency B200 vLLM compute-scale rows
- report power as unavailable below 90% latency-weighted positive-energy coverage while preserving latency, throughput, and memory results
- preserve power analysis when Rust supplies static latency, and avoid energy access in latency-only Rust and Python execution

## Promoted datasets

- B200 vLLM `0.22.0`
- B200 TRT-LLM `1.3.0rc15`
- H200 vLLM `0.22.0`
- GB200 vLLM `0.22.0`
- GB200 TRT-LLM `1.3.0rc17`

H200 TRT-LLM `1.3.0rc15` remains intentionally excluded because FP8 KV generation attention aborted on SM90 during collection.

## Power behavior

Power coverage is measured per estimate as the fraction of modeled latency backed by positive energy data. When coverage is below 90%, `power_w` is `None` and the CLI prints `unavailable (<coverage>% coverage)`; non-power estimates remain valid.

Coverage is command-dependent. Representative dense probes show that some promoted cohorts already support power, while other configurations remain partial. Additional collection and cohort-level dense/MoE regression tests are tracked in #1439.

## Known limitations and follow-ups

- The promoted artifacts lack full Collector V3 collection hashes, so their sidecars use `provenance: legacy` rather than inventing collector identity.
- MoE measurements collected before #1358 may be inflated and require recollection or stronger annotation. This accepted data-quality follow-up is tracked in #1414.
- The 90% gate establishes completeness, not calibration accuracy.
- Passing the gate does not imply that every configuration in a cohort has complete energy coverage.

## Validation

- all 65 changed parquets are in the family-first tree
- 1,537,004 rows scanned with no nonpositive latency or power values
- Collector V3 audit R1-R6 passed
- family-layout structural verification passed
- Ruff and formatting passed
- full local unit suite: 2,645 passed, 10 skipped
- latest focused SDK/CLI regression suite: 129 passed
- GitHub unit, e2e, Rust/Python parity, collector-data, parquet-diff, sanity-chart, wheel, public-API, prediction-regression, DCO, and CodeRabbit checks passed
- corrected stock-vs-promoted support-matrix jobs completed successfully: B200 `3182297`, H200 `3182299`, GB200 `3182300`
- no existing `PASS` row regressed and no support rows were removed
- 52 existing model-mode rows improved support status
- 118 rows were added for GB200 TRT-LLM `1.3.0rc17`
- 22 additional differences were traceback-text-only and did not change status or source

## Test plan

- [x] Validate each promoted database is discoverable and not marker-only
- [x] Verify family-first placement and sidecar coverage
- [x] Remove nonpositive latency/power samples
- [x] Exercise available and unavailable power-coverage outcomes
- [x] Verify latency-only execution skips energy access in Rust and Python paths
- [x] Generate B200, H200, and GB200 support matrices from real Git LFS baseline data
- [x] Compare stock and promoted matrices for status/source regressions

## Summary by CodeRabbit

* **Performance**
  * Latency-only static breakdowns now skip energy computation for faster results.
* **Improvements**
  * Static energy estimates for context and generation phases are now correctly computed and reported for the Rust-based execution path (with scaling applied), while latency-only mode avoids energy output.
* **Tests**
  * Updated existing routing tests to validate non-zero energy values where appropriate.
  * Added a regression test to confirm latency-only mode does not invoke the Python context/generation phase runners in the Rust path.


tianhaox:
Ran a cross-round latency audit on the promoted datasets (same-config best-kernel joins vs vllm 0.19.0/0.24.0, sglang 0.5.14, trtllm rc10 baselines). Overall the data looks good — power stress introduces no global latency shift (only shapes at >95% TDP move, ~+20% median on GB200 GEMM, +4–5% on MoE), and the promoted H200 context_attention data is *correct*: the old 0.19.0 rows were physically impossible (bs=8 × isl=16k × 64 heads recorded at 0.09 ms; the shape is ~70 PFLOP → the new 62–92 ms matches theory).

Two things to address before merge:

1. **MoE rows look inflated by the pre-#1358 collector bug.** vllm 0.19.0 and 0.22.0 MoE agree (median ratio 1.003, both collected pre-fix), but vs the post-fix 0.24.0 round the promoted 0.22.0 sets are systematically slower: b200 median **1.60×**, gb200 **1.33×**. b200 trtllm rc15 MoE (bf16/fp8) shows the same signature (+27–42% vs rc10, flat across token counts, uncorrelated with power). Were these MoE sweeps run before or after the #1358 fix? If before, they should be re-collected or annotated — a 1.6× MoE inflation on B200 directly skews MoE model selection.

2. **`b200_sxm/vllm/0.22.0/computescale_perf.parquet` has 6 rows with `latency = 0.0`** (fp8, k=32, m=3329–32768). Please drop them — zero latencies poison interpolation.

Non-blocking observations (fp8_block GEMM slow path affecting both vllm rounds, 0.24 attention baseline shift) are consolidated in #1414.

jasonqinzhou (changes requested):

## What this PR does

This PR promotes 65 latency-and-power parquet overlays for validated B200, H200, and GB200 backend versions and adds a Rust/Python SDK bridge so Rust static latency results retain Python-derived energy estimates.

The latest commit directly addresses the prior SDK feedback by fixing the combined `static` mode, clearing Ruff, and adding regression coverage for both scaled energy and the latency-only Rust fast path.

## Why this score

Score 5.0/10, REQUEST_CHANGES: the SDK correction is now focused and well-tested, but the promoted payload still bypasses the repository's family-first provenance contract, and six confirmed zero-latency B200 compute-scale rows remain in the data. The zero-latency issue already has a current human GitHub comment, so this packet recommends only the still-unposted layout feedback.

### P1 — Move the overlays into the family-first data tree

**[P1] Move these overlays into the family-first data tree.** The latest SDK-and-test delta leaves the complete data subtree unchanged: all 65 parquets are still added under the deprecated `<system>/<backend>/<version>` depth, with none at their catalog-owned `<system>/<family>/<backend>/<version>` paths. On this exact head, `migrate_family_layout.py --verify` rejects all 65 tables and the five flat backend directories, while `audit_collector_data.py` reports green because it walks only family-first version directories. The runtime therefore reaches these measurements only through the transition fallback, which strict provenance intentionally skips; because the data tree is byte-for-byte identical to the prior reviewed head, the revalidated prior mapping still has 52 promoted tables declared through older `reuse.yaml` donors and 13 GB200 TRT-LLM tables without target `collection_meta.yaml` coverage. Please relocate the tables through the catalog mapping, add the relevant sidecars, and reconcile each reuse declaration before rerunning the migration verifier, collector audit, and changed-operation evidence.

jasonqinzhou:
**[P1] Remove or recollect the six zero-latency compute-scale samples.** `b200_sxm/vllm/0.22.0/computescale_perf.parquet` still contains `latency == 0.0` for FP8, `k=32`, at `m=3329, 4352, 4353, 8192, 16384, 32768`. These physically impossible values can poison interpolation and make affected work appear free. Please remove or recollect them and rerun the parquet/data-quality validation.

jasonqinzhou:
**[P1] Do not expose partial energy as a complete per-GPU power estimate.** I tried the promoted datasets end to end on this exact head with a representative static SILICON estimate (`Qwen/Qwen3-32B-FP8`, TP=4, BS=8, ISL=2048, OSL=128). All added parquet rows carry positive power, but latency-weighted energy coverage is only **23.08%** for B200 TRT-LLM `1.3.0rc15`, **87.70%** for H200 vLLM `0.22.0`, and **74.22%** for GB200 TRT-LLM `1.3.0rc17`; only B200 and GB200 vLLM clear the existing 90% threshold. B200 TRT-LLM nevertheless reports **89.95 W**, effectively treating the remaining 76.92% of modeled latency as zero energy. Moving the files into the family-first tree will not fill these missing shapes/operations.

Please either fill the missing energy coverage or report power as partial/unavailable when `InferenceSummary.has_sufficient_power_data()` fails, and add representative dense and MoE end-to-end power-coverage tests for each promoted cohort.

jasonqinzhou (approved):
Re-reviewed exact head 0802f59f9c95c04155028decd734801daaf49ad0. The prior P1 findings are resolved: all overlays are in the family-first tree with declared provenance, the six zero-latency rows are removed, incomplete energy coverage is reported as unavailable, and latency-only execution skips energy work across Rust and Python paths. Both CodeRabbit threads are resolved and the required CI checks are green. The documented pre-#1358 MoE recollection and remaining power-coverage work are accepted follow-ups in #1414 and #1439.

PR #1410, opened 2026-07-24
fix(sdk): model speculative progress in agg scheduling

tianhaox:
## Summary

- add a structured `InferenceSession.run_mixed` API that returns raw mixed-step latency, energy, component, per-op, and token-count details
- distinguish logical decode requests from target-model query tokens and align the Python mixed-step path with the existing Rust `(nextn + 1)` compute semantics
- feed expected speculative decode progress into aggregate scheduling so iteration counts, mixed/gen-only composition, TPOT, throughput, and power are recomputed before service metrics are produced
- expose the same mixed-step component breakdown from the Rust engine and preserve the upper-layer projection fallback for summaries without scheduler metadata

## Why

#1405 correctly moved `nextn_accepted` above `aic-core`, but the aggregate path then applied acceptance as a final scalar projection. That is only equivalent when all work scales linearly with decode progress. Aggregate execution also includes fixed prefill and mixed-step work, so the projection can be optimistic.

The mixed-step Python path also counted `num_decode_requests` as shared target work directly, while the Rust path counted `num_decode_requests * (nextn + 1)`. This made Python and Rust disagree for MTP workloads.

This change keeps the facts separate:

- `nextn` determines current-iteration compute: each logical decode request queries `nextn + 1` target tokens
- `1 + nextn_accepted` determines expected committed-token progress
- the aggregate scheduler uses that progress to compute the mean-field decode iteration count

## Review follow-ups

A review pass over the initial revision surfaced and fixed:

- **multimodal regression**: `run_agg` now hands `run_mixed` a config carrying the image-augmented effective isl. `run_mixed` derives isl from `runtime_config`, which is text-only, so the mixed step disagreed with the genonly step for VLM workloads (regression-tested)
- **sweep dedup alignment**: the `gen_tokens` dedup key in `sweep.py` now uses the same `decode_iterations = 1 + (osl - 1) / progress` boundary as `run_agg`, so speculative sweeps no longer skip points the scheduler treats as distinct; behavior is bit-identical when no profile is active
- **projection guard hardening**: when `run_agg` recorded a `decode_tokens_per_iteration`, the scalar projection is always skipped (scheduler values are authoritative); a mismatch now logs a warning instead of silently stacking two speedups. The guard also runs before the deep frame copy
- **rust cleanup**: dropped the `get_mix_step_ops` transliteration wrapper; the single caller composes `get_mix_step_breakdown().total_ms()` directly
- clarified that `step_estimates["mixed"]` is the raw pre-`mix_efficiency` estimate and that the legacy `find_best_agg_result_under_constraints` sweep is not speculation-aware

## Validation

- `ruff check .` / `ruff format --check .`
- `cargo test -p aiconfigurator-core --lib` (198 passed) + `cargo clippy` (no new warnings)
- `pytest -m unit tests/unit tests/cross_package` (2627 passed, 39 skipped, incl. 4 new tests)
- Rust/Python engine-step parity: `parity_tests/test_engine_step_parity.py` + `test_compile_engine_parity.py` (278 passed against a freshly rebuilt extension)
- `tests/e2e/cli/test_cli_estimate_static.py::test_agg_estimate_responds_to_common_nextn`

For a QWEN3-32B/H100/TRT-LLM aggregate case with `nextn=1` and `nextn_accepted=0.85`, post-hoc scalar projection gives approximately `8.514 ms TPOT / 451.65 tok/s`, while scheduler-aware recomputation gives `8.608 ms / 432.91 tok/s`.

## Summary by CodeRabbit

* **New Features**
  * Added mixed-step estimation with total latency plus shared, context-attention, and decode-attention breakdowns.
  * Exposed mixed-step estimation through the Python engine and inference session APIs.
  * Added structured step estimates containing latency, energy, token counts, and per-operation details.
* **Improvements**
  * Improved speculative-decoding scheduling and throughput calculations using configurable decode progress.
  * Added validation for invalid decode-progress values and empty or invalid mixed steps.
  * Prevented speculative progress from being applied twice during aggregate projections.


tests/unit/sdk/test_rust_engine_step.py
```diff
@@ -98,6 +98,9 @@ def mixed_step_latency(self, *args):
             mixed_calls.append(args)
             return 8.5
 
+        def mixed_step_breakdown(self, *args):
+            return 8.5, 5.0, 2.0, 1.5
```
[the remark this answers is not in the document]
tianhaox: Fixed in d595a2a22 — the fake now records its calls and the test asserts `breakdown_calls == [(384, 7, 256, 256, 128)]`, matching the existing mixed/decode call assertions.

tests/unit/sdk/sweep/test_sweep.py
```diff
@@ -163,6 +163,49 @@ def _record(*, runtime_config, **_kwargs):
...
+        return points
+
+    baseline = _run(None)
+    inactive = _run(SpeculativeDecodingProfile(0.0))
+    speculative = _run(SpeculativeDecodingProfile(1.0))  # progress = 2.0
+
+    assert inactive == baseline
+    assert speculative != baseline
+
+
```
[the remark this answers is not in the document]
tianhaox: Strengthened in 413d6f325. Hard-coding the full 61-point set would couple the test to the batch schedule, so instead it now pins a hand-computed boundary point directly: at isl=ctx=1024, b=4 is evaluated under the baseline boundary (balance = 4/4, not > 1) but deduped into b=3's capped-gen group under progress=2 (4//1.6 == 3//1.2 == 2), plus a uniqueness assertion over the evaluated set. The inactive-profile equality check is preserved.

aic-core/src/aiconfigurator_core/_aiconfigurator_core.pyi
```diff
@@ -21,6 +21,16 @@ class AicEngine:
...
+    def mixed_step_breakdown(
+        self,
+        ctx_tokens: int,
+        gen_tokens: int,
+        isl: int,
+        osl: int,
+        prefix: int = 0,
+        seq_imbalance_correction_scale: float = 1.0,
+        gen_seq_imbalance_correction_scale: float = 1.0,
+    ) -> tuple[float, float, float, float]: ...
```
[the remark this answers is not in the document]
tianhaox: Done in b46a9c66a — `mixed_step_latency` / `decode_step_latency` now carry their imbalance-correction scale kwargs (defaults 1.0, matching the pyo3 signatures) and `last_provenance() -> str | None` is declared, so the stub matches the full `AicEngine` binding surface.

src/aiconfigurator/sdk/speculative.py
```diff
@@ -93,8 +96,24 @@ def project_summary(
         if frame is None or frame.empty:
             return projected
 
-        frame = frame.copy(deep=True)
         progress = self.tokens_per_iteration
+        if role == "agg":
+            step_estimates = projected.get_step_estimates()
+            scheduling = step_estimates.get("scheduling", {}) if step_estimates else {}
+            applied_progress = scheduling.get("decode_tokens_per_iteration")
+            if applied_progress is not None:
```
jasonqinzhou, line 104: [P1] Preserve the post-hoc aggregate projection path

`run_agg` records `decode_tokens_per_iteration=1.0` even when its caller omits that kwarg, so this presence check treats every aggregate summary as scheduler-projected. The existing direct flow `summary = session.run_agg(...); SpeculativeDecodingProfile(1.0).project_summary(summary, role="agg")` therefore warns about 1.0 versus 2.0 and silently leaves TPOT/throughput at their baseline values; a current-head probe kept TPOT at 7.0 ms instead of projecting it to 3.5 ms. Please preserve whether progress was explicitly supplied and only bypass projection when the scheduler actually applied an explicit value. Explicit mismatches should remain authoritative (or fail), and the omitted-kwarg flow needs a regression test.

src/aiconfigurator/sdk/sweep.py
```diff
@@ -304,7 +312,7 @@ def _sweep_one_parallel_agg(
                 break
 
             # Skip equivalent gen_tokens slices to avoid recomputing the same point.
-            balance_score = isl * b / ctx_tokens / osl
+            balance_score = isl * b / ctx_tokens / decode_iterations
```
jasonqinzhou: [P1] Do not deduplicate distinct speculative sweep points

This generic key does not represent the backend's actual mixed-step schedule. In the added test's own case (`isl == ctx_tokens == 1024`, `osl=4`, progress 2, hence `decode_iterations=2.5`), batches 3 and 4 both floor to key 2, so the later duplicate check drops batch 4. But vLLM schedules `b - ceil(ctx/isl)` mixed decode requests, which is 2 versus 3 for those points; their latency, memory, and feasibility are not equivalent. The current-head probe reproduced keys `{3: 2, 4: 2}` versus vLLM request counts `{3: 2, 4: 3}`. Please disable this heuristic for speculative/fractional schedules or derive an equivalence key from the backend's actual scheduling shape, then cover the vLLM case with a real scheduler assertion.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -934,43 +936,79 @@ def _get_mix_step_latency(
...
-        """
+    def run_mixed(
+        self,
+        model: BaseModel,
+        database: PerfDatabase,
+        runtime_config: RuntimeConfig,
+        step: MixedStepInput,
+    ) -> StepEstimate:
+        """Estimate one scheduled mixed prefill/decode forward pass."""
+        isl = int(runtime_config.isl or 0)
```
jasonqinzhou, line 974: [P1] Account for visual tokens in direct mixed-step calls

The new public `InferenceSession.run_mixed` forwards a normal `RuntimeConfig` unchanged, but that config's `isl` is text-only. `run_static` and `run_agg` add `_visual_context_tokens`; only the internal `run_agg` caller compensates by overwriting a copied config. A direct VLM call with text ISL 8, 16 visual tokens, and `MixedStepInput(context_tokens=24, num_decode_requests=1)` consequently queried context attention as `(batch=3, s=8)` and decode attention at `s=11`, instead of the effective-ISL shapes `(batch=1, s=24)` and `s=27`. Please centralize effective-ISL handling for `run_mixed` and adjust the internal override to avoid double-counting; add a direct multimodal API regression test.

jasonqinzhou (changes requested):

## What this PR does

This review covers the implementation and verification changes in PR #1410.

## Why this score

The head update fully fixes the prior unselected stub-surface P2, but the three previously reported P1 correctness issues remain unchanged and reproducible at the current head.

src/aiconfigurator/sdk/speculative.py
tianhaox: Fixed in 20bfbce74. `run_agg` now records `decode_tokens_per_iteration` in the scheduling metadata only when the caller explicitly supplied the kwarg (`None`/omitted → no marker), so the legacy flow `session.run_agg(...)` + `project_summary(role="agg")` projects again. Explicit mismatches remain authoritative with a warning. Explicit-1.0 and omitted calls no longer share an agg-cache entry (they schedule identically but differ in projection eligibility). Regression tests: `test_run_agg_records_progress_only_when_explicitly_supplied` and `test_aggregate_projection_applies_when_scheduler_saw_no_explicit_progress`; an end-to-end probe of your scenario now gives raw tpot 17.739 ms → projected 9.589 ms (= raw / 1.85).

src/aiconfigurator/sdk/sweep.py
tianhaox: Fixed in 20bfbce74 by taking the first option: the capped-gen dedup heuristic is disabled entirely when a speculative profile is active, so every guard-passing point is evaluated (run_agg's `_agg_cache` keeps repeat cost low). Non-speculative sweeps keep the legacy osl-based key bit-for-bit. I did not attempt a backend-accurate equivalence key — the generic key mis-groups vLLM's `b - ceil(ctx/isl)` schedule exactly as you showed, and any shared key would need per-backend derivation, which isn't worth the evaluation savings. Test `test_sweep_agg_disables_gen_dedup_for_speculative_schedules` pins a hand-computed dropped point ((6, 1024): baseline dedups it into b=5's group, speculative evaluates it) plus baseline ⊂ speculative and inactive == baseline.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
tianhaox: Fixed in 20bfbce74. Effective-isl handling is centralized in `run_mixed` itself: it adds `_visual_context_tokens(model, runtime_config)` to the config's text-only isl (the internal pass configs carry no image fields, so the contribution applies exactly once), and `run_agg` now passes the raw runtime_config unchanged — no pre-adjusted copy, no double-counting. Direct multimodal API regression test added (`test_run_mixed_derives_effective_multimodal_isl_for_direct_calls`): with text isl 8 + 16 visual tokens it asserts context attention is queried as (batch=1, s=24) and decode attention at kv length 24 + osl//2, i.e. your exact expected shapes; `test_run_agg_does_not_double_count_visual_tokens_in_run_mixed` locks the no-double-count contract.

Arsene12358 (approved):

jasonqinzhou (approved):
**APPROVE · 9.2/10 · high confidence**

## What this PR does

This PR adds a structured mixed-step SDK surface and makes aggregate speculative-decoding estimates scheduler-aware, including Python/Rust parity, service-metric projection, sweep behavior, and multimodal sequence-length handling.

The latest update responds directly to each reported correctness gap with narrowly scoped ownership fixes, cache-state separation, and regression tests that pin both the legacy and speculative paths.

## Why this score

The current head is approval-quality: all three prior P1 findings are fixed and independently revalidated, the changed test surface passes, and no new actionable issue was found in the five-file delta. The remaining uncertainty is limited to GitHub e2e, Rust/Python parity, and snapshot checks that were still running when this packet was written.

PR #1413, opened 2026-07-25
fix(collector): B60 vllm gemm -- cover LM-head n-dims broken by strict perf_interp

mandy-li:
PR #1303 (perf_interp engine v2) made the perf-DB lookup strict: shapes
missing from B60 gemm_perf.parquet now error out instead of extrapolating 
from neighbors. This broke projection for Llama-3.1-8B on B60 at low GPU counts

## Summary by CodeRabbit

* **New Features**
  * Expanded XPU vLLM GEMM shape coverage by adding explicit LM-head output feature sizes, now reaching up to 65,536.

* **Bug Fixes**
  * Improved XPU GEMM memory management by estimating peak memory needs, automatically tuning execution repetition, and ensuring reliable cleanup to reduce memory buildup.

* **Tests**
  * Updated GEMM case-expansion expectations and added unit coverage for peak memory-footprint calculations (including FP8 variants) and for confirming cleanup behavior after each run.


tianhaox:
thanks for raising this issue. I'm okay to proceed.
Meanwhile, do you think the util-hold extrapolation works for this case on B60, actually it's mostly about the logits gemm where the vocab size can be super large. If so, I will do a fix to the interp v2 engine. Created an issue to track, https://github.com/ai-dynamo/aiconfigurator/issues/1415

mandy-li:
Thanks for the quick turnaround on this, and for filing #1415.

Yes, util-hold with SOL(query)/SOL(boundary) should work well for the LM-head GEMM case on B60. The dominant shape there is m = decode concurrency (small), k = 4096, n = vocab (128256 for Llama-3.1). It's firmly memory-bandwidth bound, and along the n-axis at fixed (m, k) the memory SOL is essentially linear in n, so SOL(query)/SOL(boundary) ≈ n_query / n_boundary — a reasonable extrapolation of the kernel's actual behavior. Compute-bound prefill GEMMs behave the same way at fixed (m, k), so the same anchor should generalize.

One thing worth calling out: the two fixes are complementary, not either/or.

Your engine fix (#1415) removes the hard-fail so out-of-grid LM-head shapes project gracefully — critical for backends where nobody sweeps the full vocab tier.
The boundary anchors in this PR (16384 / 32768 / 51200 / 65536 for vllm_xpu) give the extrapolation a closer boundary to hold from. Extrapolating LM-head at n=128256 from a real n=65536 measurement is a lot more trustworthy than extrapolating from n=12288 (a ~10× jump vs a 2× jump).
So the accuracy story for B60 is:

- with just #1415 → projection succeeds, extrapolated from the previous ceiling (n=12288);
- with just this PR → projection succeeds via the added anchors, but other backends may still crash on LM-head;
- with both → engine no longer hard-fails anywhere, and where boundary data exists the util-hold has a short lever arm.

Happy to help validate on B60 once #1415 lands — I can rerun the same CLI and diff projection numbers before/after to sanity-check the util-hold delta from the new anchors vs the old ceiling.

tianhaox:
Thanks for the feedback. #1419 is almost ready for review. let's merge this PR first.

tianhaox (approved):

PR #1418, opened 2026-07-26
fix(sdk): resolve tc_flops by compute dtype with strict missing-key error

### LINKED ISSUE #1398: MoE Operation uses bf16 compute flops instead of chosen data type
Reference:

https://github.com/ai-dynamo/aiconfigurator/blob/41530cd5dea0097992d6ca97c47ea05475c318af/aic-core/src/aiconfigurator_core/sdk/operations/moe.py#L322

AIConfigurator system YAML allows us to define FLOP efficiencies for few data types. This operator implementation simply ignores these numbers. While for a number of NVidia systems I've checked that relationship holds true, it makes the code unsound if that wasn't true.

Haven't checked other operators.

### LINKED ISSUE #1417: b60 (Intel XPU) system YAML: confirm and complete per-dtype tc_flops entries
## Context

While fixing #1398 (operators using bf16 flops scaled by a hardcoded factor instead of the per-dtype `*_tc_flops` defined in the system YAML), we are moving SOL math to strictly resolve FLOPS from the system spec by compute dtype:

- quant mode → compute dtype (`bfloat16` / `int8` / `fp8` / `fp4`) → `<dtype>_tc_flops` YAML key
- **A missing key will become a hard error** ("platform does not support this quant, or the YAML entry is missing"), instead of silently falling back to `bfloat16_tc_flops * compute_factor`.

## Problem

`aic-core/src/aiconfigurator_core/systems/b60.yaml` currently defines only:

```yaml
float32_tc_flops: 12280000000000   # 12.28 TFLOPS
bfloat16_tc_flops: 98304000000000  # 98.3 TFLOPS
int8_tc_flops: 197000000000000     # 197 TOPS
```

There are no `fp8_tc_flops` / `fp4_tc_flops` entries. Once the strict resolution lands, any fp8- or fp4-based quant mode on `b60` will raise an error instead of producing an extrapolated (fictional) number.

## Ask (for the b60/XPU maintainers, cc PR #246 / #694 authors)

1. Confirm whether Arc Pro B60 (Battlemage XMX) hardware supports fp8 (and fp4) matrix compute.
2. If supported: add the corresponding `fp8_tc_flops` (and `fp4_tc_flops`) entries with real dense-throughput numbers.
3. If not supported: no YAML change needed — the new hard error is the intended behavior; please just confirm here so we can close.

## References

- #1398 (MoE ignores per-dtype flops)
- `sdk/operations/gemm.py` `_get_quant_tc_flops` (the resolver being promoted to a shared helper)

tianhaox:
#### Overview:

SOL math resolved tensor-core FLOPS as `bfloat16_tc_flops * quant.compute`, ignoring the per-dtype `*_tc_flops` entries the system YAMLs already define (#1398). This PR makes every SOL path resolve FLOPS by the quant mode's actual compute dtype, and turns a missing YAML entry into a hard error instead of a bf16-scaled extrapolation.

#### Details:

**Why it matters now:** the fixed 1/2/4 ratio assumption is already broken in-repo. On b300/gb300, `fp4_tc_flops` is 14/15 PFLOPS (not bf16×4 = 9/10), so nvfp4 MoE/attention/MLA SOL was over-estimated ~1.56x; `int8_tc_flops` is 153.5/165 TFLOPS (~29x below fp8), so int8-pipeline modes keyed by compute factor would read the wrong entry entirely.

- `QuantMapping` (Python namedtuple + Rust struct) gains a `compute_dtype` field naming the MMA pipeline (`bfloat16`/`int8`/`fp8`/`fp4`; `None` for memory-only KV-cache/comm modes). `sq` (w8int8) now maps to `int8_tc_flops` — previously it borrowed `fp8_tc_flops` via its compute factor; weight-only modes (`int8_wo`, `int4_wo`, `w4a16_*`) map to `bfloat16_tc_flops`.
- New shared resolver `common.get_quant_tc_flops` (Rust: `perf_database::gemm::quant_tc_flops`) replaces `GEMM._get_quant_tc_flops` and every bare `bf16 * compute` site in moe/attention/mla (+ wideep variants). dsa/dsv4/msa/gemm callers repointed to it.
- A missing `*_tc_flops` entry raises `MissingSystemFlopsError` (`ValueError` subclass, reported as an expected CLI error; Rust `AicError::MissingSystemFlops`, mapped across the FFI to the same Python class). The platform either lacks hardware for that dtype or the YAML entry must be added — a fictional throughput is never produced.
- Load-time SOL clamps (`_correct_sol` and the Rust grid clamps) skip a quant slice whose dtype has no YAML entry rather than failing the whole database load, so measured silicon data stays usable (b60 shipped fp8 data with no `fp8_tc_flops`).
- b60: `fp8_tc_flops` added from its own collected data — the vllm gemm tables (7581 shapes × 2 versions) top out at 99% of the bf16 peak for fp8 (and MoE agrees), so fp8 evidently runs on the bf16 pipeline; recorded at the bf16 rate, not int8's 2x (#1417).
- Rust sol formulas now take pre-resolved FLOPS so the hot sol closures stay `-> f64`; resolution happens once per query with `?` propagation. Python-parity oracle tests confirm float-precision parity is preserved.

**Expected numeric changes:** b300/gb300 nvfp4 SOL for moe/attention/mla gets ~1.56x faster; all other shipped systems are unchanged (their ratios happen to match the old scaling, verified by the full unit suites passing with per-dtype fixture entries at the historical ratios).

- `aic-core/src/aiconfigurator_core/sdk/common.py` (`QuantMapping.compute_dtype`, `get_quant_tc_flops`)
- `aic-core/rust/aiconfigurator-core/src/perf_database/gemm.rs` (`quant_tc_flops`)
- `aic-core/src/aiconfigurator_core/sdk/operations/gemm.py` `_correct_sol` (load-time clamp skip semantics)

#### Testing

- `cargo test --workspace --all-features`: 292 passed (includes Python-parity oracle tests)
- `pytest tests/unit/sdk`: 1444 passed (`test_moe_dispatch.py` needs torch, unavailable locally; collector failures are pre-existing on main)
- New tests: per-mode dtype→key mapping, strict missing-key / memory-only errors, b300 fp4-entry-wins assertion (both languages)

#### Review hardening (commit 91823faab)

An adversarial multi-agent review pass over the change surfaced and fixed:

- **Eager resolution parity**: Python resolved flops lazily inside `get_sol` closures (never invoked on a SILICON exact hit) while Rust resolves at query entry — the engines diverged in the "silicon data exists but YAML lacks the entry" case. Python now resolves eagerly at every query-table entry; a regression test asserts SILICON exact hits cannot bypass the check.
- **Non-positive entries rejected**: an explicit `fp8_tc_flops: 0` placeholder would have made SOL = inf and let load-time clamps overwrite measured data; both resolvers now treat non-positive values like missing entries.
- **kv→FMHA derivation centralized** in one Python helper (`generation_attn_flops`, mirroring the Rust helper) instead of four inline copies; WideEP MLA's bf16 terms now use the strict resolver (MissingSystemFlopsError instead of a bare KeyError).
- **Divergent-ratio op-level tests**: the shared fixtures pin the historical 1/2/4 ratios (where old and new formulas coincide), so `test_quant_dtype_flops.py` deliberately breaks the ratio and asserts MoE / generation-attention SOL follow the per-dtype YAML entry.
- Rust `quant_tc_flops` moved to `common/system_spec.rs` (common-layer utility), re-exported from `perf_database::gemm`.

#### User-visible behavior changes (intentional, from the strict policy)

- **b300/gb300**: nvfp4 SOL for moe/attention/mla gets ~1.56x faster (fp4 = 14/15 PFLOPS, not bf16×4); `sq` SOL gets ~29x slower (int8 = 153.5/165 TFLOPS — Blackwell Ultra gutted int8 TC throughput; previously sq wrongly borrowed the fp8 rate via its compute factor). No collected sq data exists on those systems, so no silicon baselines move.
- **Ampere (a100/a30)**: models with intrinsic fp8 op groups (DSA / MSA / DSV4 indexer) now raise `MissingSystemFlopsError` instead of producing a bf16×2-extrapolated estimate — Ampere has no fp8 hardware, so the old number was fictional.
- **Out-of-tree system YAMLs** (`local_systems/`): configurations that relied on the bf16-scaled fallback (e.g. sq without `int8_tc_flops`, or any quant whose dtype entry is absent) now fail with an actionable error naming the exact YAML key to add. All 13 in-repo YAMLs already define the required entries.
- **Known limitation**: `compute_dtype` is a single per-mode value; if a future architecture executes a mode on a different pipeline (e.g. smoothquant dequantizing to fp8 on Blackwell), a per-architecture override will be needed before collecting that data — otherwise the SOL clamp could floor measured latencies against the wrong pipeline's rate.

- Closes #1398
- Closes #1417 (b60 fp8_tc_flops added; no-native-fp8 confirmed at kernel level for both GEMM and MoE by Intel review)

## Summary by CodeRabbit

* **New Features**
  * Added per-quantization compute dtype support (BF16/INT8/FP8/FP4) and dtype-specific `*_tc_flops` resolution for roofline/SOL calculations.
  * FLOPs-aware SOL propagation is applied to attention, GEMM, MoE, MLA, MSA, DSA, and DSV4.

* **Bug Fixes**
  * Roofline/SOL calculations now strictly validate required `*_tc_flops` and raise a new typed `MissingSystemFlops` error for missing/non-positive/unsupported values.
  * Load-time SOL clamping now skips only the affected quant entries.

* **Tests**
  * Expanded unit tests for per-dtype `*_tc_flops` selection and eager validation/rejection behavior.


tianhaox:
@coderabbitai full review

aic-core/src/aiconfigurator_core/sdk/operations/mla.py
```diff
@@ -1062,6 +1074,11 @@ def _query_wideep_generation_mla_table(
         database_mode: common.DatabaseMode | None = None,
     ):
         """Query WideEP generation MLA table. Verbatim port of the legacy body."""
+        # Strict eager resolution (parity with the Rust engine, which resolves
+        # flops with `?` at query entry): reject a missing *_tc_flops entry up
+        # front — a SILICON exact hit never invokes the get_sol closure.
+        common.get_quant_tc_flops(database.system_spec, fmha_quant_mode)
+        common.get_quant_tc_flops(database.system_spec, common.FMHAQuantMode.bfloat16)
```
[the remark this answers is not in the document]
tianhaox: Fixed in 6c40b4280 — good catch. The eager check now resolves the DERIVED dtype via `generation_attn_flops(system_spec, kvcache_quant_mode)` (matching the Rust engine, which derives from the kv dtype before resolving, see `perf_database/wideep_mla.rs`). Both requested regressions added in `test_quant_dtype_flops.py::test_wideep_generation_mla_validates_derived_dtype_not_label`: fp8 KV + bf16 label with missing `fp8_tc_flops` raises; bf16 KV + fp8 label with missing `fp8_tc_flops` is not falsely rejected.

aic-core/src/aiconfigurator_core/sdk/common.py
```diff
@@ -1089,26 +1098,63 @@ class FMHAQuantMode(Enum):
...
+    """
+    compute_dtype = quant_mode.value.compute_dtype
+    if compute_dtype is None:
+        raise MissingSystemFlopsError(f"quant mode '{quant_mode.value.name}' is memory-only and has no compute FLOPS")
+    key = _COMPUTE_DTYPE_TO_FLOPS_KEY[compute_dtype]
+    # A non-positive value is treated the same as a missing entry: it can only
+    # be a placeholder or a typo, and letting it through turns every SOL into
+    # inf and load-time clamps into silent data corruption.
+    value = system_spec["gpu"].get(key)
+    if value is None or value <= 0:
```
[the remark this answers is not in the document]
tianhaox: Fixed in 6c40b4280: `not value > 0` per the suggestion, matching the Rust resolver's `filter(|v| *v > 0.0)`; NaN regression added next to the zero-placeholder test.

aic-core/rust/aiconfigurator-core/src/perf_database/dsa.rs
```diff
@@ -631,18 +636,41 @@ fn build_generation_nodes(grids: &DsaGrids) -> NodeCache {
...
+/// Python (`fmha_mode = FMHAQuantMode.bfloat16`).
+pub(crate) fn dsa_generation_sol_flops(
+    spec: &SystemSpec,
+    gemm_quant: GemmQuantMode,
+) -> Result<DsaSolFlops, AicError> {
+    Ok(DsaSolFlops {
+        gemm: quant_tc_flops(spec, gemm_quant.mapping())?,
+        indexer_fp8: quant_tc_flops(spec, FmhaQuantMode::Fp8.mapping())?,
+        attn: quant_tc_flops(spec, FmhaQuantMode::Bfloat16.mapping())?,
+    })
```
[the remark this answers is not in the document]
tianhaox: Declining the skip_indexer gate as unreachable, with reasoning: DSA silicon data ships only for h100/h200/b200/b300/gb200/gb300 — every one of which defines `fp8_tc_flops` — and the DSA model families (DeepSeek-V3.2, GLM-5.2) require the fp8 indexer kernels that Ampere-class hardware does not have, so an fp8-less platform can never model them. More decisively: a GLM-5.2 model always contains full (non-skip) DSA layers alongside the reuse layers in the same run, so gating the skip-indexer variant alone cannot make such a platform modelable — the full layers would raise on the very same query batch. The strict up-front error with the exact YAML key is the intended signal here (documented under 'User-visible behavior changes' in the PR body). Happy to revisit if a skip-indexer-only modeling use case ever materializes.

aic-core/rust/aiconfigurator-core/src/perf_database/dsv4.rs
```diff
@@ -978,17 +985,28 @@ fn resolve_head_key<T>(by_native: &BTreeMap<u32, T>, local_heads: u32) -> Option
...
+    gemm_quant: GemmQuantMode,
+    fmha_quant: FmhaQuantMode,
+) -> Result<Dsv4SolFlops, AicError> {
+    Ok(Dsv4SolFlops {
+        gemm: quant_tc_flops(spec, gemm_quant.mapping())?,
+        bf16: quant_tc_flops(spec, GemmQuantMode::Bfloat16.mapping())?,
+        fp8: quant_tc_flops(spec, GemmQuantMode::Fp8.mapping())?,
+        attn: quant_tc_flops(spec, fmha_quant.mapping())?,
+    })
 }
```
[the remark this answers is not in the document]
tianhaox: Declining for the same reachability reason as the DSA thread: DSV4 is architecturally Blackwell-only (the MegaMoE query hard-rejects SM < 100, and DSV4 silicon data ships only for Blackwell systems), and every system that can model DSV4 defines `fp8_tc_flops`. The a30/a100 specs without the entry cannot reach `dsv4_sol_flops` through any supported model/system combination, and DSV4's per-layer indexer is intrinsically fp8 — a platform without fp8 compute cannot model the architecture regardless of the HCA/CSA mix. Threading `compress_ratio` through the flops resolution would add a Python/Rust parity surface for a case that cannot occur; documented as intended strictness in the PR body.

aic-core/rust/aiconfigurator-core/src/py.rs
```diff
@@ -94,7 +98,12 @@ fn aic_to_py(e: AicError) -> PyErr {
     let sdk_class: Option<(&'static GILOnceCell<Py<PyType>>, &str)> = if e.is_missing_perf_data() {
         Some((&PERF_DATA_NOT_AVAILABLE_ERROR, "PerfDataNotAvailableError"))
     } else if matches!(e, AicError::EmpiricalNotImplemented(_)) {
-        Some((&EMPIRICAL_NOT_IMPLEMENTED_ERROR, "EmpiricalNotImplementedError"))
+        Some((
+            &EMPIRICAL_NOT_IMPLEMENTED_ERROR,
+            "EmpiricalNotImplementedError",
+        ))
+    } else if matches!(e, AicError::MissingSystemFlops(_)) {
+        Some((&MISSING_SYSTEM_FLOPS_ERROR, "MissingSystemFlopsError"))
```
jasonqinzhou, line 109: [P1] Resolve typed errors from the standalone core namespace

`sdk_error_type` imports `aiconfigurator.sdk.errors`, but the `aiconfigurator-core` wheel intentionally ships only `aiconfigurator_core`; the installed-package verifier explicitly rejects `aiconfigurator` in a core-only install. This new branch therefore falls back to plain `ValueError` instead of `MissingSystemFlopsError` for standalone-core users, defeating exact-class catches and the advertised FFI mapping. The added unit test masks this because it defines SDK availability with the same legacy import. Please load `aiconfigurator_core.sdk.errors` here (the full install's legacy alias preserves class identity) and cover the core-only wheel path.

aic-core/rust/aiconfigurator-core/src/perf_database/gemm.rs
```diff
@@ -125,7 +125,10 @@ impl GemmTable {
             ))
         })?;
         let spec = &self.system_spec;
-        let sol = move |c: &[f64]| gemm_sol_latency_ms(spec, lookup_quant, c[0], c[1], c[2]);
+        let tc_flops = quant_tc_flops(spec, lookup_quant.mapping())?;
```
jasonqinzhou: [P1] Resolve dtype FLOPS before any perf-data lookup

This runs only after `load_gemm` and the quant-table lookup. For H200 + NVFP4, Rust SILICON therefore raises `PerfDataNotAvailableError` first, while HYBRID reaches the fallback resolver and raises `MissingSystemFlopsError`; Python now resolves at `_query_gemm_table` entry in every mode. The FFI tests at `parity_tests/test_engine_step_parity.py:1342-1383` currently codify this contradictory split. The strict missing-key contract and Rust/Python parity remain mode-dependent. Please move resolution ahead of table loading/selection, audit the analogous Rust table entrypoints, and update the SILICON parity expectation.

aic-core/src/aiconfigurator_core/sdk/common.py
```diff
@@ -1089,26 +1098,65 @@ class FMHAQuantMode(Enum):
...
+    if compute_dtype is None:
+        raise MissingSystemFlopsError(f"quant mode '{quant_mode.value.name}' is memory-only and has no compute FLOPS")
+    key = _COMPUTE_DTYPE_TO_FLOPS_KEY[compute_dtype]
+    # A non-positive value is treated the same as a missing entry: it can only
+    # be a placeholder or a typo, and letting it through turns every SOL into
+    # inf and load-time clamps into silent data corruption.
+    value = system_spec["gpu"].get(key)
+    # `not value > 0` (rather than `value <= 0`) also rejects NaN, matching
+    # the Rust resolver's `filter(|v| *v > 0.0)`.
+    if value is None or not value > 0:
```
jasonqinzhou: [P2] Reject non-finite FLOPS values

`not value > 0` still admits positive infinity. PyYAML parses `.inf` as `float('inf')`, this guard accepts it, and `ops / tc_flops` becomes zero, silently collapsing compute-bound SOL to the memory roof. Rust's `filter(|v| *v > 0.0)` has the same behavior. Please require a positive, finite value in both resolvers (`math.isfinite` / `is_finite`) and add a `+inf` parity test.

jasonqinzhou (changes requested):

## What this PR does

This review covers the implementation and verification changes in PR #1418.

## Why this score

The dtype mappings and SOL propagation are largely consistent, but two P1 typed-error contracts remain broken: SILICON can fail before strict FLOPS validation, and core-only wheels cannot resolve the new canonical FFI exception. Positive infinity also bypasses both strict validators.

aic-core/src/aiconfigurator_core/systems/b60.yaml
```diff
@@ -10,6 +10,7 @@ gpu:
   float32_tc_flops: 12280000000000 # 12.28TFLOPS
   bfloat16_tc_flops: 98304000000000 # 98.3TFLOPS
   int8_tc_flops: 197000000000000 # 197TOPS
+  fp8_tc_flops: 98304000000000 # same rate as bf16, NOT int8's 2x: collected vllm gemm data (0.12.0/0.20.0, 7581 shapes each) tops out at ~97.5 TFLOPS for fp8 = 99% of the bf16 peak, indicating fp8 runs on the bf16 pipeline (no native fp8 MMA)
```
Spycsh: Can we simplify the comments as below
```
# B60 has no native FP8 matrix-compute support, so this path falls back to BF16.
```

aic-core/src/aiconfigurator_core/sdk/operations/gemm.py
```diff
@@ -433,9 +423,13 @@ def _query_gemm_table(
...
         """Query GEMM table — preserves PR #721 exact-hit → 1D → 3D fast path."""
+        # Strict eager resolution (parity with the Rust engine, which resolves
+        # flops with `?` at query entry): reject a missing *_tc_flops entry up
+        # front — a SILICON exact hit never invokes the get_sol closure.
+        common.get_quant_tc_flops(database.system_spec, quant_mode)
 
         def get_sol(m_v: int, n_v: int, k_v: int, qm: common.GEMMQuantMode) -> tuple[float, float, float]:
-            tc_flops = cls._get_quant_tc_flops(database.system_spec, qm)
+            tc_flops = common.get_quant_tc_flops(database.system_spec, qm)
             sol_math = 2 * m_v * n_v * k_v / tc_flops * 1000
```
Spycsh, line 481: Do we currently distinguish W8A16-like kernels in the SOL memory model?

For B60/XPU, the measured vLLM FP8 GEMM path appears to be W8A16 rather than native W8A8. I confirmed at runtime:
```
  quant_method.fp8_linear = XPUW8A16FP8LinearKernel
  weight dtype = torch.float8_e4m3fn
  input dtype = torch.bfloat16
```
So the current generic memory formula based only on quant_mode.value.memory may be too optimistic for this path. It treats all three matrices as if they had the quant mode byte width, but W8A16 only reduces the weight matrix traffic.

A more accurate first-order memory model could be:
```
  bf16:
    bytes = 2*m*k + 2*n*k + 2*m*n

  W8A16:
    bytes = 2*m*k + 1*n*k + 2*m*n + scale/dequant overhead

  W8A8:
    bytes = 1*m*k + 1*n*k + 2*m*n + scale/quant overhead
```

aic-core/rust/aiconfigurator-core/src/py.rs
tianhaox: Fixed in 66a01e1e8 — good catch, and it turned out to be a pre-existing gap for the other two typed errors as well: `sdk_error_type` has imported `aiconfigurator.sdk.errors` since the original typed-FFI work, so core-only installs were already degrading PerfDataNotAvailableError / EmpiricalNotImplementedError to plain ValueError. The helper now imports `aiconfigurator_core.sdk.errors` (always installed next to the extension; the upper module is a pure `_alias_module` of it, so full installs see the same class objects and existing `except` clauses are unaffected). The pure-Rust-context fallback to ValueError is preserved.

aic-core/rust/aiconfigurator-core/src/perf_database/gemm.rs
tianhaox: Fixed in 66a01e1e8: `GemmTable::query` now resolves dtype flops before `load_gemm` and the quant-table lookup, matching Python's query-entry resolution in every mode — H200+NVFP4 classifies as MissingSystemFlopsError under both SILICON and HYBRID on both engines. The silicon-data-gap FFI test kept its purpose but moved to a vehicle whose flops resolve fine (MiMo-V2-Flash: head_dim=192 vs collected {128, 256} attention on b200), so the PerfDataNotAvailableError contract stays pinned by a case that can still reach the data lookup. Full parity suite re-run locally: 303 passed.

aic-core/src/aiconfigurator_core/sdk/common.py
tianhaox: Fixed in 66a01e1e8: both resolvers now require a positive AND finite value (`math.isfinite` in Python, `is_finite()` in Rust), with +inf regressions added next to the NaN ones in both languages. Agreed on the failure mode — `.inf` would have silently collapsed compute-bound SOL onto the memory roof, which is exactly the class of quiet corruption this PR is trying to eliminate.

aic-core/src/aiconfigurator_core/systems/b60.yaml
tianhaox: Simplified in 66a01e1e8, with two details kept on purpose: (1) "executes at the BF16 rate" rather than "falls back to BF16" — the mode still runs and still reads fp8 weights, it just has no faster MMA to run on, so "rate" describes what the number in this line actually is; (2) your runtime W8A16 finding from the other thread is folded in, since it is the strongest one-line justification for why this value equals bfloat16_tc_flops. Current wording:

```yaml
fp8_tc_flops: 98304000000000 # B60 has no native FP8 matrix compute: vLLM's FP8 GEMM runs W8A16 on the BF16 pipeline, so FP8 executes at the BF16 rate (collected data ceiling ~99% of BF16 peak)
```

Happy to trim further if you'd still prefer the shorter form.

aic-core/src/aiconfigurator_core/sdk/operations/gemm.py
tianhaox: Short answer: no — the SOL memory model does not distinguish W8A16 from W8A8 today, and your runtime confirmation is really useful on two fronts.

**Compute side (this PR's scope): your finding closes the loop.** `XPUW8A16FP8LinearKernel` with fp8 weights x bf16 activations means the MMA genuinely runs on the BF16 pipeline — which independently confirms the empirical ceiling we derived (~99% of the BF16 peak across 7581 shapes) and settles the open question on #1417. The per-system YAML absorbs this cleanly: the global `fp8` mode resolves whatever the system's `fp8_tc_flops` says, and on b60 that now equals the BF16 rate. No per-arch special-casing needed for compute.

**Memory side (your actual question): there is a real, bounded fidelity gap.** The generic formula applies `quant_mode.value.memory` (1 byte/elem for fp8) uniformly to all three GEMM terms — `m*n + m*k + n*k`. For W8A16 that is right for the weights (`n*k` @ 1B) but understates the BF16 activations (`m*k` input, `m*n` output) by 2x. Where it matters:

- **Decode-shaped GEMMs (small m)**: `m*(n+k) << n*k`, sol_mem is weight-dominated, so the error is negligible exactly where the op is memory-bound.
- **Large m**: the activation share grows, but the kernel crosses into compute-bound territory where `max(sol_math, sol_mem)` masks the sol_mem error.
- **Worst case**: shapes near the roofline knee, where the understated activation traffic can put the knee in the wrong place.

Note the uniform factor is not an oversight for NVIDIA systems — native W8A8 fp8 really does move 1-byte activations — so this is an XPU-kernel-specific deviation, same shape as the compute question you already caught. Two mitigations exist today: HYBRID/EMPIRICAL modes calibrate util against measured b60 latencies, absorbing most of this in practice; only SOL-only estimates carry the distortion undiluted.

**The structural fix** would be splitting `QuantMapping.memory` into separate weight/activation byte factors (per mode, possibly per system) and threading them through every op's mem formula in both Python and Rust — a real cross-cutting modeling change, out of scope for this PR. If b60 SOL fidelity matters for your use cases I'm happy to file a follow-up issue with this analysis as the starting point.

One question back: do you know whether the b60 MoE fp8 path (`moe_perf.parquet` fp8 rows) is also W8A16? The MoE SOL uses the same uniform memory factor, and the collected fp8:bf16 MoE ceiling ratio (~1:1) suggests the same kernel structure, but a runtime confirmation like your GEMM one would settle it.

tianhaox: BTW, why we choose w8a16 kernel and record it as w8a8? if it's designed like this, maybe we should record it as w8a16 in the perf db in future.

aic-core/src/aiconfigurator_core/systems/b60.yaml
tianhaox: On reflection — your suggested wording adopted verbatim in the latest commit. It's your system's YAML and the shorter line reads better; the W8A16 kernel detail and the empirical derivation are preserved in this PR's discussion threads and commit history, so nothing is lost. Thanks for the review!

aic-core/src/aiconfigurator_core/sdk/operations/gemm.py
Spycsh: > Short answer: no — the SOL memory model does not distinguish W8A16 from W8A8 today, and your runtime confirmation is really useful on two fronts.
> 
> **Compute side (this PR's scope): your finding closes the loop.** `XPUW8A16FP8LinearKernel` with fp8 weights x bf16 activations means the MMA genuinely runs on the BF16 pipeline — which independently confirms the empirical ceiling we derived (~99% of the BF16 peak across 7581 shapes) and settles the open question on #1417. The per-system YAML absorbs this cleanly: the global `fp8` mode resolves whatever the system's `fp8_tc_flops` says, and on b60 that now equals the BF16 rate. No per-arch special-casing needed for compute.
> 
> **Memory side (your actual question): there is a real, bounded fidelity gap.** The generic formula applies `quant_mode.value.memory` (1 byte/elem for fp8) uniformly to all three GEMM terms — `m*n + m*k + n*k`. For W8A16 that is right for the weights (`n*k` @ 1B) but understates the BF16 activations (`m*k` input, `m*n` output) by 2x. Where it matters:
> 
> * **Decode-shaped GEMMs (small m)**: `m*(n+k) << n*k`, sol_mem is weight-dominated, so the error is negligible exactly where the op is memory-bound.
> * **Large m**: the activation share grows, but the kernel crosses into compute-bound territory where `max(sol_math, sol_mem)` masks the sol_mem error.
> * **Worst case**: shapes near the roofline knee, where the understated activation traffic can put the knee in the wrong place.
> 
> Note the uniform factor is not an oversight for NVIDIA systems — native W8A8 fp8 really does move 1-byte activations — so this is an XPU-kernel-specific deviation, same shape as the compute question you already caught. Two mitigations exist today: HYBRID/EMPIRICAL modes calibrate util against measured b60 latencies, absorbing most of this in practice; only SOL-only estimates carry the distortion undiluted.
> 
> **The structural fix** would be splitting `QuantMapping.memory` into separate weight/activation byte factors (per mode, possibly per system) and threading them through every op's mem formula in both Python and Rust — a real cross-cutting modeling change, out of scope for this PR. If b60 SOL fidelity matters for your use cases I'm happy to file a follow-up issue with this analysis as the starting point.
> 
> One question back: do you know whether the b60 MoE fp8 path (`moe_perf.parquet` fp8 rows) is also W8A16? The MoE SOL uses the same uniform memory factor, and the collected fp8:bf16 MoE ceiling ratio (~1:1) suggests the same kernel structure, but a runtime confirmation like your GEMM one would settle it.

I agree that we can keep the current SOL modeling for now. The SOL correction is only triggered in relatively rare cases, so the impact of this memory-modeling gap is quite limited in practice.

There are still a few outstanding issues in the latest MoE kernel collector. I'll be sending follow-up PRs to address those, and figure out the precision used in both the collector and the real runtime paths.

Spycsh: > BTW, why we choose w8a16 kernel and record it as w8a8? if it's designed like this, maybe we should record it as w8a16 in the perf db in future.

Currently w8a16 is chosen on the default path https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/kernels/linear/__init__.py#L352, and I think for `gemm` AIC does not distinguish between w8a8 and w8a16, both record as fp8. Just checked with the xpu kernel team that the reason for default choice is that w8a16 has better perf than w8a8 on b60 (which does not support native fp8). I think we can keep the current default behavior (choose w8a16 and record it as fp8).

Spycsh: > One question back: do you know whether the b60 MoE fp8 path (moe_perf.parquet fp8 rows) is also W8A16? The MoE SOL uses the same uniform memory factor, and the collected fp8:bf16 MoE ceiling ratio (~1:1) suggests the same kernel structure, but a runtime confirmation like your GEMM one would settle it.


I've validated it with Qwen/Qwen3-30B-A3B on B60/XPU. The MoE FP8 path selects Fp8MoeBackend.XPU with XPUExpertsFp8, and the runtime XpuFusedMoe.apply() inputs show:

```
  hidden_states.dtype = torch.bfloat16
  w1/w2.dtype = torch.float8_e4m3fn
  a1q_scale = None
```

So although FusedMoEQuantConfig reports use_fp8_w8a8=True, the actual XPU MoE kernel is not receiving pre-quantized FP8 activations or activation scales. In fused_moe_interface.py, a1q_scale=None means act_quant=False, remapped_scales=None, and GEMM1 is called with ptr_A as BF16 and ptr_A_scale=None, while weights are FP8 with weight scales.

So yes, the B60 MoE FP8 path is effectively W8A16-style at runtime, consistent with the observed fp8:bf16 MoE ceiling ratio being close to 1:1.

tianhaox: Thanks for the thorough runtime validation — that settles all three open questions on this thread: (1) memory model stays as-is (agreed, and your rare-trigger point matches our roofline-knee analysis); (2) recording W8A16 under the fp8 label stays (makes sense given AIC's gemm label doesn't distinguish activation width and W8A16 is simply the faster kernel on b60); (3) MoE confirmed W8A16 with kernel-level evidence, consistent with the ~1:1 fp8:bf16 ceiling in the collected data. With both GEMM and MoE confirmed, the b60 `fp8_tc_flops = bf16 rate` entry in this PR now has runtime evidence behind it end-to-end, so I've upgraded this PR to close #1417 on merge. Looking forward to the MoE collector follow-up PRs — happy to review the precision-audit side when they land.

aic-core/src/aiconfigurator_core/sdk/operations/attention.py
```diff
@@ -210,6 +210,20 @@ def _cache_key(database: PerfDatabase) -> tuple:
     )
 
 
+def generation_attn_flops(system_spec: dict, kvcache_quant_mode: common.KVCacheQuantMode) -> float:
```
Arsene12358, line 230: **[P1] This derivation strands 2,534 shipped a100 silicon rows: fp8 KV -> fp8 pipeline is wrong on pre-sm89 hardware.**

a100_sxm ships measured generation-MLA data with `kv_cache_dtype=fp8` (`systems/data/a100_sxm/mla/trtllm/1.0.0/generation_mla_perf.parquet`, 2,534 rows, reused into 1.3.0rc15; the context file's 1,218 fp8-kv rows are unaffected since context treats KV as memory-side only). Ampere has no fp8 tensor cores — the decode kernel dequantizes KV and issues the MMA on the bf16 pipeline, which is exactly why that data could be collected in the first place. With the eager query-entry check, every one of those queries now raises:

```
db.query_generation_mla(1, 65, 64, kv=fp8, mode=SILICON)   # a100_sxm / trtllm / 1.0.0
main    -> 0.0158187 ms  (silicon exact-hit region; bf16-kv control 0.0158160 ms)
this PR -> MissingSystemFlopsError: quant mode 'fp8' needs 'fp8_tc_flops' ...
```

This is the PR body's "Known limitation" (per-architecture pipeline override) arriving early, on data already in-tree — the same physics as the b60 `fp8_tc_flops = bf16 rate` entry this PR adds, minus the fix. Suggested shape:

```python
fp8_kv = kvcache_quant_mode == common.KVCacheQuantMode.fp8
fmha_mode = (
    common.FMHAQuantMode.fp8
    if fp8_kv and system_spec["gpu"]["sm_version"] >= 89
    else common.FMHAQuantMode.bfloat16
)
return common.get_quant_tc_flops(system_spec, fmha_mode)
```

with the identical gate mirrored in Rust `perf_database/attention.rs::generation_attn_flops` and the derivation copies in `wideep_mla.rs` (~:219) / `dsv4.rs` (~:377) — DSA generation already hardcodes bf16 in both engines. Please add regressions both ways: a100 fp8-kv generation MLA returns the silicon value; an sm>=89 system still resolves `fp8_tc_flops` strictly (missing entry still errors).

Deliberately NOT the alternative fix: adding `fp8_tc_flops` to a100_sxm.yaml would flip `_gpu_supports_datatype(..., "FP8")` in `tools/support_matrix/support_matrix.py` (presence-keyed) and silently un-gate every FP8 checkpoint on Ampere.

One open question you're better placed to answer: whether trtllm's fp8-kv decode on sm89/90 truly runs the fp8 MMA — if some sm>=89 backend also dequantizes to bf16, the gate belongs on the collected data's semantics (like the b60 thread concluded) rather than a bare sm threshold; either way the a100 case is unambiguous.

aic-core/rust/aiconfigurator-core/src/perf_database/attention.rs
```diff
@@ -164,9 +165,18 @@ impl AttentionTable {
         // kv-head/window/head-size setup; c = [n, full_s, b].
         let spec = &self.system_spec;
         let n_kv_lookup = key.n_kv_lookup;
+        let attn_flops = quant_tc_flops(spec, fmha_quant.mapping())?;
```
Arsene12358: **[P1] 13 sibling entrypoints still resolve after the fallible lookup** — this file :168/:220/:271, `mla.rs:190/:218/:268/:295/:328`, `wideep_mla.rs:162-163/:223-224`, `dsa.rs:352/:428`, `dsv4.rs:311/:386`.

`GemmTable::query` got the hoist (`gemm.rs:123`, resolve before `load_gemm`), but here `load_context()?` and the `by_keys` miss (`missing_key` -> `AicError::PerfDatabase` -> `PerfDataNotAvailableError` across the FFI) still fire before `quant_tc_flops`. In SILICON mode that classifies a missing-dtype system as a data miss whenever the slice is also uncollected — the typical pairing — while Python classifies it `MissingSystemFlopsError` at `_query_*_table` entry. That splits `has_perf_data_not_available_cause` between the engines for the same query (on main they agreed), which is the same contradictory-split shape as the original gemm finding.

Reproduced through the real FFI: b200_sxm spec with `fp8_tc_flops` removed, context attention fmha=fp8 with an uncollected head_size, SILICON -> Python raises `MissingSystemFlopsError` (attention.py:336), Rust raises `PerfDataNotAvailableError` (this file :158); HYBRID converges because the ladder's empirical helper resolves. Reachable with shipped YAMLs via nvfp4 MLA-BMM / DSA / DSV4 on h200_sxm (no `fp4_tc_flops`).

The fix is the same mechanical hoist in all 13 — resolve the flops (or build the `*_sol_flops` struct) before `load_*()` and the key lookup:

```rust
// Resolve flops BEFORE any perf-data lookup: a missing dtype entry must
// classify as MissingSystemFlops on both engines, in every mode.
let attn_flops = quant_tc_flops(&self.system_spec, fmha_quant.mapping())?;
let grids = self.load_context()?;
let node = grids.by_keys.get(&key).ok_or_else(|| missing_key(...))?;
```

Suggest also pinning the missing quadrant in the parity suite: SILICON x missing-dtype on one of these lazy families (e.g. nvfp4 MLA-BMM on h200), asserting `MissingSystemFlopsError` from both engines. The current pins — SILICON with all dtypes present (test_engine_step_parity.py:1342) and HYBRID missing-dtype on an eager op (:1364) — are exactly why this quadrant slipped through. After the hoist, the "parity with the Rust engine ..." comments in the 15 Python query methods become true as written.

Arsene12358 (changes requested):
**REQUEST_CHANGES — two blockers, both narrow and mechanically fixable; everything else verified clean.**

I ran an independent verification pass against head 68c1b13b: before/after probes on shipped data (a main worktree vs this branch, same venv and parquet), a sweep of all 906 shipped parquets' quant columns cross-referenced against the 14 system YAMLs' `*_tc_flops` keys, a dual-engine trace of resolution order across every Python `_query_*_table` and Rust table entrypoint, the prediction-gate artifact from run 30360649109 (124,864 rows, categories decoded from drift_report.csv), and the local suites (pytest unit/sdk 1452 passed, cargo 279 passed). The numeric claims all check out: b300 `sq` lands 29.4x below fp8 and `nvfp4` 3.1x above it, b60 fp8 SOL is exactly the bf16 value, h200 `sq` is numerically unchanged, b300/gb300 really have zero collected sq rows, the gate shows zero OK->not-OK transitions, all 9,612 status changes are `DATA_MISS -> MissingSystemFlopsError` (already-failing combos now failing with the actionable error), and the b300/gb300 drift is all-faster nvfp4 at the documented magnitude. API compat is safe: zero 3-arg `QuantMapping` constructions or 3-field `.value` unpacking anywhere in the repo, docs, tools, or collector, and the upper-package alias re-exports the new error with class identity. The two blockers below are the residue.

**[P1] Blocker 1 — shipped-data regression: a100_sxm generation-MLA with fp8 KV-cache.** `systems/data/a100_sxm/mla/trtllm/1.0.0/generation_mla_perf.parquet` carries 2,534 measured rows with `kv_cache_dtype=fp8` (reused into 1.3.0rc15 via reuse.yaml), and this branch hard-errors every query against them:

```
db.query_generation_mla(1, 65, 64, kv=fp8, mode=SILICON)   # a100_sxm / trtllm / 1.0.0
main @ eed87c1c    -> 0.0158187 ms   (measured silicon; bf16-kv control 0.0158160 ms)
pr-1418 @ 68c1b13b -> MissingSystemFlopsError: quant mode 'fp8' needs 'fp8_tc_flops' ...
```

This sits outside the documented intentional-failure set (Ampere fp8-*intrinsic* families — MLA decode is none of them, and fp8 KV is storage, not fp8 compute). It is the b60 W8A16 situation on another SKU — fp8-quantized operand, MMA physically on the bf16 pipeline — without the b60 treatment. Reachable through the public SDK (above) and the CLI `kvcache_quant_mode` override; the support-matrix daily run cannot see it because DeepSeek FP8 checkpoints are already HW_INCOMPATIBLE on a100 for unrelated (GEMM) reasons. Repro details and a suggested fix are in the inline comment on `generation_attn_flops`.

**[P1] Blocker 2 — the "audit the analogous Rust table entrypoints" half of the earlier gemm finding is incomplete: 13 Rust SILICON paths still resolve flops after the fallible data lookup.** `GemmTable::query` got the fix (verified: resolution now precedes `load_gemm`), but attention x3 (`perf_database/attention.rs:168/:220/:271`), MLA x5 (`mla.rs:190/:218/:268/:295/:328`), wideep-MLA x2 (`wideep_mla.rs:162/:223`), DSA x2 (`dsa.rs:352/:428`), and DSV4 x2 (`dsv4.rs:311/:386`) keep the old order. Consequence in SILICON mode (the Rust default): when the dtype gap coincides with a data gap — the normal pairing, since no hardware means no collected data — Rust raises `PerfDataNotAvailableError` where Python raises `MissingSystemFlopsError`, so `has_perf_data_not_available_cause` classifies the same combo differently per engine; on main the engines agreed. Reproduced through the real FFI (details in the inline comment on attention.rs), and reachable with shipped YAMLs via e.g. nvfp4 MLA-BMM/DSA/DSV4 on h200. The comment stamped into the 15 Python query methods ("parity with the Rust engine, which resolves flops with `?` at query entry") is currently false for those 13 counterparts.

**Notes, non-blocking:**
- PR body says "all 13 in-repo YAMLs" (there are 14) and "all other shipped systems are unchanged" — the gate artifact contradicts the latter for b60 (418 drifts + 327 `DATA_MISS -> OK` gains; good changes worth stating, they are the point of #1417) and l40s (39 rows at <=0.1%). Worth updating so the merge record matches the measured blast radius.
- The support-matrix preflight keys FP8 capability on key *presence* (`tools/support_matrix/support_matrix.py:360-404`) while the new resolver requires positive+finite — a placeholder `fp8_tc_flops: 0` passes preflight and fails every query. Worth aligning the predicate.
- The clamp-skip comments citing "e.g. b60 fp8" as a live example (gemm.py `_correct_sol`, gemm.rs `clamp_gemm_grids_to_sol`) became historical the moment this PR added the b60 key.
- Cheap hardening skipped from earlier rounds: EMPIRICAL/SOL_FULL in the rejection parametrization (I probed both — behavior is already correct, this just pins it), a `check(AicError::MissingSystemFlops(..), "MissingSystemFlopsError")` case in `aic_to_py_maps_typed_errors_to_sdk_classes`, and Rust NaN/zero/negative resolver cases (+inf exists; the earlier reply's "added next to the NaN ones in both languages" overstates — Rust never had a NaN case).

Spycsh (approved):

aic-core/src/aiconfigurator_core/sdk/operations/attention.py
tianhaox: Fixed in e89feb022, exactly along the lines you sketched. The derivation now lives in one gated helper per language — Python `attention.generation_attn_mode` / Rust `perf_database::attention::generation_attn_mode` — with `fp8 KV -> fp8` only when `sm_version >= 89`; pre-89 and specs without `sm_version` (XPU) derive bf16. All the derivation copies (attention/mla generation get_sol, `_correct_sol`, the eager entry checks, wideep-MLA, DSV4 generation in both engines, operators/dsv4.rs) route through it. Agreed on NOT adding `fp8_tc_flops` to a100_sxm.yaml — good catch on the support-matrix presence-keying; the commit message records that constraint so nobody 'fixes' it that way later.

Regressions added per your list: `test_pre_sm89_fp8_kv_generation_returns_shipped_silicon` (real a100/trtllm/1.0.0 data, your exact repro shape — fp8-kv and bf16-kv exact hits agree), `test_sm89_plus_fp8_kv_still_requires_fp8_entry`, and a pre-89 SOL test asserting the bf16 rate is priced (not fp8, not an error).

On your open question: to my knowledge trtllm's fp8-kv decode on sm90 genuinely runs the fp8 MMA (the XQA fp8 path), so the sm gate matches reality for the shipped NVIDIA systems. I agree the *principled* home for this is the collected data's semantics rather than a bare SM threshold — b60 reached the same conclusion for its W8A16 fp8 label — but that needs a per-mode/per-system pipeline annotation (same shape as the deferred weight/activation memory split), and every shipped configuration is correctly described by the gate today. Happy to fold both into one 'per-system kernel-semantics annotation' follow-up if a case shows up that the gate misdescribes.

aic-core/rust/aiconfigurator-core/src/perf_database/attention.rs
tianhaox: Fixed in e89feb022: all thirteen entrypoints now resolve flops (or build their `*_sol_flops` struct) before `load_*()` and the key lookup — attention.rs ctx/gen/enc, mla.rs ctx/gen/bmm/ctx-module/gen-module, wideep_mla.rs ctx/gen, dsa.rs ctx/gen, dsv4.rs ctx/gen — with the same hoist comment GemmTable::query carries. The wideep and DSV4 generation paths also switched from their inline kv->fmha derivations to the shared sm-gated `generation_attn_mode`, so the P1 gate applies uniformly.

Your suggested quadrant pin is in: `test_silicon_missing_dtype_on_lazy_family_matches_python` (nvfp4 MLA-BMM on h200, SILICON) asserts `MissingSystemFlopsError` — not a perf-data miss — from the Rust FFI and the Python query surface alike. Full parity suite local: 305 passed; cargo --workspace --all-features: 292.

Arsene12358 (approved):
**APPROVE — both blockers verified fixed at 66ff990ba (fix commit e89feb022); independent re-verification below.**

**Blocker 1 (a100 fp8-KV decode).** Re-ran my exact repro on this head: `query_generation_mla(1, 65, 64, kv=fp8, SILICON)` on a100_sxm/trtllm/1.0.0 returns **0.0158187 ms** again (bf16-kv control 0.0158160 ms) — identical to main, so the 2,534 shipped rows are queryable once more. Edge probes: b60 (no `sm_version`) derives the bf16 pipeline and keeps working; h200 with `fp8_tc_flops` deleted still raises `MissingSystemFlopsError` under fp8 KV, so sm>=89 stays strict. The two `generation_attn_mode` implementations are semantically identical including the missing-`sm_version` -> bf16 case (`(gpu.get("sm_version") or -1) >= 89` vs `is_some_and(|sm| sm >= 89)`), and every inline derivation copy now routes through the helper (dsv4.py, mla.py WideEP, operators/dsv4.rs, perf_database dsv4/wideep_mla) — one benign exception noted below. Agreed on keeping a100's YAML untouched, and the commit-message record of the support-matrix presence-keying constraint is exactly the breadcrumb that prevents the next person from "fixing" it that way.

**Blocker 2 (resolve-after-load).** Verified in source at this head that every listed entrypoint resolves flops (or builds its `*_sol_flops` struct) before `load_*()` and the key lookup: attention.rs :153/:206/:266, mla.rs :183/:215/:248/:296/:333-334, wideep_mla.rs :152-153/:212-213, dsa.rs :334/:412, dsv4.rs :305 (and the generation path derives via the shared rule before its loads). Ran the new quadrant pin through a locally rebuilt FFI: `test_silicon_missing_dtype_on_lazy_family_matches_python` passes — both engines classify nvfp4 MLA-BMM on h200 SILICON as `MissingSystemFlopsError` — alongside the three neighboring typed-error pins. Local suites at this head: unit sdk 1521 passed (plus the known torch-only `test_moe_dispatch` collection error, covered in CI), cargo workspace green, and CI is green including the parity job and the prediction gate.

**One residual, non-blocking:** `operators/wideep_mla.rs::wideep_generation_mla_empirical` (the HYBRID util-transfer copy) still derives fp8 FMHA from fp8 KV *without* the sm gate, while its Python counterpart closes over the gated mode and the silicon-path table query uses `generation_attn_mode`. It is value-identical on every shipped system (all wideep data is sm>=90) and reachable only as an error-class skew on platforms that ship no wideep data at all — but it is the last inline copy of the derivation, and the one-liner (`let fmha_quant = generation_attn_mode(spec, kv_quant);`) would make "the gate applies uniformly" literally true. Fine to fold into the kernel-semantics follow-up or any next touch of this file.

Thanks for the fast, thorough turnaround.

tianhaox:
@Arsene12358 thanks for the thorough independent re-verification and the approve. Your residual is fixed in c21f38439: `wideep_generation_mla_empirical` now derives via the shared sm-gated `generation_attn_mode` like its Python counterpart and the silicon path — that was indeed the last inline copy of the derivation rule. (Also merged upstream/main twice along the way: the #1431 rank-local heads signature and the #1455 quant-transfer ladder; the #1455 ladder-oracle's nvfp4-on-h200 rows now assert the strict MissingSystemFlopsError instead — h200 has no fp4 MMA, so that vehicle's old xprofile estimate was anchored on a fictional bf16*4 SOL; sq/int4_wo keep pinning the xquant/xprofile relations, and weight-only fp4 modes like the #1392 nvfp4_wo plan will declare the bf16 pipeline and keep using the ladder. Post-merge suites: cargo 300, sdk 1531, parity 309.)

aic-core/rust/aiconfigurator-core/parity_tests/test_engine_step_parity.py
```diff
@@ -1374,30 +1377,77 @@ def test_silicon_data_gap_raises_typed_perf_data_miss(
...
         # and NOT be classified as a plain perf-data miss.
         _prepare_rust_core(monkeypatch)
         case = EngineStepParityCase(
             model_path="nvidia/MiniMax-M2.5-NVFP4",
             system_name="h200_sxm",
             database_mode="HYBRID",
-            transfer_policy="balanced",
         )
-        with pytest.raises(errors.EmpiricalNotImplementedError) as excinfo:
+        with pytest.raises(errors.MissingSystemFlopsError) as excinfo:
```
jasonqinzhou, line 1400: **[P1] Keep transfer-aware validation aligned with strict FLOPS resolution**

This assertion captures the runtime contract: H200 + NVFP4 must raise `MissingSystemFlopsError` before the HYBRID ladder because the system has no `fp4_tc_flops`. The new shipped-data CI guard does not change `Task._check_role_against_db`, which still pre-admits any GEMM/MoE mode that is XQUANT/XPROFILE-reachable without checking whether that mode's compute dtype can resolve. On this exact head I again reproduced both H200/trtllm/1.3.0rc10 GEMM and MoE `nvfp4` tasks passing `Task.validate()`, then their first HYBRID query immediately raising this error. That breaks the validator's fail-early contract and can admit impossible sweep configurations. Please require strict FLOPS resolution before direct or transfer admission for GEMM/MoE/WideEP modes, update `test_validate_gemm_quant_transfer_reachable_in_hybrid` to reject H200 NVFP4, and use a quant whose dtype FLOPS exist for the positive XPROFILE coverage.

tests/unit/sdk/database/test_quant_dtype_flops.py
```diff
@@ -0,0 +1,249 @@
...
+    fp8_tc_flops). This sweep moves the failure to CI: data implies key.
+
+    Labels are resolved through the quant enums' own ``compute_dtype`` (so new
+    modes are covered automatically); memory-only labels (kv/comm) carry
+    ``None`` and are skipped, as are loader-remapped raw labels listed in
+    ``_EXTRA_LABELS``.
+    """
+
+    _QUANT_COLUMNS = frozenset(
+        {"gemm_dtype", "moe_dtype", "fmha_dtype", "mla_dtype", "quant_mode", "gemm_quant_mode", "fmha_quant_mode"}
```
jasonqinzhou: **[P2] Include the shipped attention and MLA-BMM dtype columns**

This set misses the actual `attn_dtype` and `bmm_dtype` schemas. The current tree has `attn_dtype` in 203 parquet files and `bmm_dtype` in 50; their loaders convert those values to `FMHAQuantMode` and `GEMMQuantMode`, and the corresponding context-attention and MLA-BMM query entrypoints now call `get_quant_tc_flops` strictly. Because the guard scans neither column (while `fmha_dtype` is not present in the shipped schemas), landing an `fp8` attention or BMM row on a system such as A100 would leave this test green even though every matching query would hard-fail for the missing `fp8_tc_flops`. Please add those two column-to-enum mappings and cover each with a missing-key regression so the test protects every strict table family it claims to audit.

tests/unit/sdk/database/test_quant_dtype_flops.py
```diff
@@ -0,0 +1,249 @@
...
+            system = spec_path.stem
+            data_dir = systems_root / "data" / system
+            if not data_dir.is_dir():
+                continue
+            gpu = (yaml.safe_load(spec_path.read_text()) or {}).get("gpu") or {}
+            needed: dict[str, set[str]] = {}
+            for parquet_path in data_dir.rglob("*.parquet"):
+                try:
+                    schema_names = set(pq.read_schema(parquet_path).names)
+                except Exception:
```
jasonqinzhou: **[P2] Fail the guard when shipped parquet metadata is unreadable**

Catching every schema-read failure and continuing silently removes that shipped file from the invariant, so a truncated/corrupt parquet (or a reader incompatibility) can hide the exact missing-YAML-key mismatch this test is meant to catch. All 1,106 current parquet files are checked in directly and `read_schema` succeeds for every one in the targeted audit, so there is no expected legacy/LFS case to suppress here. Please let the exception fail with the file path, or append the read failure to `problems`, rather than treating unreadable shipped data as if it had no compute dtype requirements.

jasonqinzhou (changes requested):

## What this PR does

This PR makes Python and Rust SOL calculations resolve tensor-core throughput from each quantization mode's actual compute dtype and raises a typed error when the required system FLOPS entry is absent or invalid. The latest commit adds a repository-wide CI scan intended to ensure that every compute dtype present in shipped silicon parquet data has a usable system-YAML FLOPS key.

The new guard is a useful defense-in-depth response to the b60 data/YAML mismatch: it derives labels through the shared quant enums, aggregates failures by system and dtype, and successfully audits more than a thousand shipped parquet files without adding production-path complexity.

## Why this score

6.5/10, REQUEST_CHANGES. The new shipped-data guard passes, the focused Python tests and formatting checks pass, and all 20 exact-head CI checks are successful, but it does not repair the previous P1: transfer-aware validation still admits H200 NVFP4 GEMM and MoE tasks that strict query-entry FLOPS resolution immediately rejects. The added guard also omits the real `attn_dtype` and `bmm_dtype` schema columns used by strict attention and MLA-BMM queries, and it silently ignores unreadable parquet files, leaving two false-negative paths in the CI safety net.

jasonqinzhou (approved):

aic-core/rust/aiconfigurator-core/parity_tests/test_engine_step_parity.py
tianhaox: Fixed in 3ca1ebd10 — this was a real fail-early break, thanks for reproducing it end-to-end. `_check_role_against_db._check` now performs the same strict `get_quant_tc_flops` resolution the query entrypoints do, at the top of the check — before data-presence, alias, XQUANT and XPROFILE admission — so H200+NVFP4 tasks fail `validate()` with `MissingSystemFlopsError` in every mode and policy. Memory-only kv modes skip (compute_dtype None), keeping the sm-gated generation derivation a query-time concern. `test_validate_gemm_quant_transfer_reachable_in_hybrid` now asserts exactly that for nvfp4, and the positive XPROFILE coverage moved to `int4_wo` on h200/**vllm**/0.19.0 — no h200/vllm version collects it (h200/trtllm was unusable as a vehicle: 1.2.0rc5 carries int4_wo and shared-layer siblings feed it forward), its (0.5, 1) profile is level-listed, and its bf16 pipeline resolves. The level-table refusal test switched to the same vehicle for the same reason.

tests/unit/sdk/database/test_quant_dtype_flops.py
tianhaox: Fixed in 3ca1ebd10: `attn_dtype` and `bmm_dtype` are now swept (and the never-shipped `fmha_dtype` name dropped from the set). The audit still finds zero gaps across all 253 additional files. Missing-key regressions added for both families: `test_context_attention_missing_fp8_entry_rejected` (prefill path — no kv-derivation gate) and `test_mla_bmm_missing_fp4_entry_rejected`.

tests/unit/sdk/database/test_quant_dtype_flops.py
tianhaox: Fixed in 3ca1ebd10: an unreadable shipped parquet now appends a `{system}: unreadable shipped parquet {path}: {exc}` entry to `problems` and fails the assertion, instead of silently dropping the file from the invariant.

PR #1419, opened 2026-07-26
fix(sdk): scale-up frontier fallback for the perf_interp site distance gate

### LINKED ISSUE #1415: perf_interp v2 site gate: Gemma-4 tp1 LM-head GEMM misses on every GPU database (silently dropped from Pareto search)
## Summary

`perf_interp` v2's `ScatteredSites` gate (`max_site_distance = 2.0` octaves, joint log2 distance over `(n, k)`) makes **Gemma-4-26B-A4B's tp1 LM-head GEMM a structured miss on every GPU database in the repo** (66/66 system×backend×version combos checked). The failure is invisible today: the Pareto search swallows per-combination exceptions, so the support matrix still reports PASS while tp1 configurations are silently excluded from recommendations.

This is the same root cause as the B60/Llama-3.1-8B breakage fixed by the data patch in #1413, but it is not B60-specific and no data patch exists for it.

## Details

Gemma-4-26B-A4B has the largest vocab in `model_configs/`: `vocab_size = 262144`, `hidden_size = 2816`. The model queries its logits GEMM as `(n = vocab/tp, k = hidden)` (`sdk/models/gemma4.py:309,448`).

At tp1 the query site is `(262144, 2816)`:

- n axis: `log2(262144 / 65536) = 2.0` exactly (all GPU gemm tables cap n at 65536) — would pass on its own since the gate is `<=`.
- k axis: `hidden = 2816` is not a collected k grid point; nearest is 2560, adding `log2(2816/2560) = 0.1375` octaves.
- Joint distance: `sqrt(2.0^2 + 0.1375^2) = 2.0047 > 2.0` → `raise _miss(..., "no site within max_site_distance")` (`sdk/perf_interp/engine.py` scattered resolve; mirrored in Rust `perf_database/perf_interp.rs`).

Verified against every `systems/data/*/gemm/*/*/gemm_perf.parquet`: tp1 distance is ~2.004 on all GPU databases (n_max = 65536 everywhere); tp2 = 1.01, tp4 = 0.13 are fine. `b60/vllm/0.12.0` (still n_max = 12288) misses at tp1/tp2/tp4.

Sweep of all 68 model configs (tp1, H200 vLLM 0.24.0 sites):

| model | tp1 joint distance | status |
|---|---|---|
| google/gemma-4-26B-A4B (262144 / 2816) | **2.0039** | **MISS — all GPU databases** |
| Qwen3.5 family (248320) | 1.9218 | passes, brushing the gate |
| all other (vocab, hidden) combos | < 1.8 | fine |

Any future model with vocab > 262144 will miss at tp1 on every system.

## Why the support matrix doesn't catch it

`sdk/pareto_analysis.py` wraps each parallel config in `except Exception: ... "skip this combination"; continue`. The tp1 candidates raise `InterpolationDataNotAvailableError`, get dropped, and the search returns results from tp>=2 configs → support-matrix rows stay PASS. For a 26B A4B MoE, tp1 × N replicas is plausibly the actual tokens/s/gpu optimum, so recommendations are silently degraded rather than failing loudly. (The B60/Llama case in #1413 was visible only because at low GPU counts tp1/tp2 were the *only* candidates, so the result set came back empty.)

## Suggested fixes

1. **Mechanism (root fix):** make the frontier direction-aware in `ScatteredSites`. When the query site lies beyond the collected frontier in the scale-up direction (`n_q >= n_i`, `k_q >= k_i` for frontier sites), fall back to boundary util-hold (median util of the K log-nearest covering frontier sites, then `SOL(query)/util`) instead of a miss — the same mechanism already trusted unbounded along the m curve axis and for Grid ops' out-of-range. Keep the 2-octave gate for interior holes and the scale-down direction, where util genuinely doesn't transfer. Validate with a frontier-holdout LOO (hold out the largest-n sites), not the interior-hole LOO used to tune 2.0. Changes: Python `sdk/perf_interp/engine.py` + Rust `perf_database/perf_interp.rs` (both gate sites).
2. **Short-term data patch:** extend `output_feature_sizes` with vocab-derived grid points (adding n = 131072 alone brings gemma tp1 to ~1.0 octaves). Deriving the n axis from the supported-model vocab list (`vocab/tp` for tp in 1..8) prevents the next big-vocab model from re-tripping this.
3. **Guardrail:** a CI test that enumerates every supported model's LM-head `(vocab/tp, hidden)` shapes and asserts each system database resolves them (or fails the gate loudly), so coverage gaps surface at merge time instead of as silently-degraded recommendations.

## Repro

```python
import pandas as pd, glob, math
V, H = 262144, 2816  # google/gemma-4-26B-A4B
for p in sorted(glob.glob('aic-core/src/aiconfigurator_core/systems/data/*/gemm/*/*/gemm_perf.parquet')):
    df = pd.read_parquet(p, columns=['n', 'k'])
    sites = set(zip(df.n, df.k))
    d = min(math.sqrt(math.log2(V/sn)**2 + math.log2(H/sk)**2) for sn, sk in sites)
    print(f"{p.split('/data/')[1]:<40} tp1 dist={d:.4f} {'MISS' if d > 2.0 else 'ok'}")
```

Related: #1303 (perf_interp v2), #1413 (B60 LM-head data patch for Llama-3.1-8B).

tianhaox:
## Summary

Root fix for ai-dynamo/aiconfigurator#1415: the `ScatteredSites` `max_site_distance` gate (2.0 octaves joint log2 over the site axes) made Gemma-4-26B-A4B's tp1 LM-head GEMM `(n=262144, k=2816)` a structured miss on every GPU database (distance \~2.005; every gemm table caps n at 65536), and the Pareto search silently dropped the tp1 candidates while the support matrix stayed PASS.

This PR makes the gate **direction-aware** in both engines (Python `aic-core/src/aiconfigurator_core/sdk/perf_interp/engine.py` and Rust `aic-core/rust/aiconfigurator-core/src/perf_database/perf_interp.rs`, mirrored implementations):

* **Waiver condition** (both must hold when the gate finds no neighbour): the query sits strictly above the collected maximum on at least one site axis (the query is beyond the scale-up frontier; this implies no site dominates it), and it is not below the collected minimum on any axis. Dominated interior holes, incomparable notches inside the bounding box, and scale-down / mixed-direction queries keep the miss.
* **Waiver action**: the gate is skipped and the query falls through to the *existing* nearest-site IDW util transfer (coverage-filtered, `SOL(query)` re-applies the scaling) — the same unbounded trust the engine already gives along the m curve axis and past the Grid staircase. A median-anchored variant was A/B'd on the frontier-holdout LOO and measured equivalent (within 0.6% median), so the simpler reuse of the transfer path wins.
* **Unchanged**: interior holes and scale-down queries still raise `InterpolationDataNotAvailableError`; the existing mixed-direction `(16384, 512)` test keeps its miss semantics.

## Validation

* **Real tables**: Gemma tp1 now resolves on h200_sxm/vllm/0.24.0, h100_sxm/trtllm/1.3.0rc15, b200_sxm/sglang/0.5.10 with latency \~4.0x the collected n=65536 site — exactly the linear-n physics (tp2/tp4 answers unchanged).
* **Frontier-holdout LOO** (as suggested in the issue): dropping all sites with n > 8192 forces every measured n=65536 point through the waived-gate path (≥3 octaves, harsher than gemma's 2.005) → median error 8.9–15.1% (p90 51–112%) across the three databases above. A milder n > 16384 cut leaves the eval points exactly at the 2.0 gate boundary, where they resolve through the gated path at 5.9–9.3% median — the waived path at gemma's just-past-the-gate distance anchors on those same sites, so its accuracy there is comparable.
* **Tests**: 2 new Python + 2 mirrored Rust unit tests (gemma-shaped scale-up fallback; interior-hole still misses). Full SDK unit suite (1443) passes; `cargo test --workspace --all-features` (287) passes; Rust engine-step parity tests pass.

## Not in this PR (issue suggestions 2 & 3)

* Data patch extending `output_feature_sizes` with vocab-derived n grid points (would bring gemma tp1 back from extrapolation to interpolation accuracy).
* CI guardrail enumerating every supported model's LM-head `(vocab/tp, hidden)` shapes against each database.

Fixes ai-dynamo/aiconfigurator#1415. Related: ai-dynamo/aiconfigurator#1303, ai-dynamo/aiconfigurator#1413.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

## Summary by CodeRabbit

* **Enhancements**
  * Improved scattered-sites performance interpolation so that when `ScatteredSites.max_site_distance` would block all candidates for scale-up queries beyond the scale-up frontier, the resolver waives the distance gate and falls back to frontier boundary utilization-hold behavior instead of returning an immediate miss.
  * Interior-hole and comparable “incomparable notch” cases beyond the gate still raise structured misses when no valid site data applies.
* **Documentation**
  * Clarified `max_site_distance` “structured miss” vs “frontier fallback” behavior, including the scale-up-frontier exception.
* **Tests**
  * Added unit coverage for frontier util holding and for continued miss behavior in interior-gap scenarios.

## Summary by CodeRabbit

- **New Features**
  - Improved performance interpolation for scale-up queries beyond the collected frontier.
  - Eligible frontier queries can now resolve using nearest-site utilization transfer instead of failing immediately.

- **Bug Fixes**
  - Preserved structured misses for interior holes, scale-down queries, multi-axis overflows, and other unsupported gaps.
  - Improved candidate eligibility checks for frontier resolution.

- **Documentation**
  - Clarified frontier-waiver behavior, miss conditions, and validation results.

### DIFF
--- a/aic-core/rust/aiconfigurator-core/src/perf_database/perf_interp.rs
+++ b/aic-core/rust/aiconfigurator-core/src/perf_database/perf_interp.rs
@@ -621,29 +621,29 @@ impl SiteIndex {
     }
 
     /// True iff the query site exceeds the collected frontier in the scale-up
-    /// direction ONLY: no collected site weakly dominates it (>= on every site
-    /// axis), and it is not below the collected minimum on any axis. Interior
-    /// holes (dominated) and scale-down / mixed-direction queries keep the
-    /// distance-gate miss — util genuinely doesn't transfer downward, but at
+    /// direction ONLY: strictly above the collected maximum on at least one
+    /// site axis, and not below the collected minimum on any axis. Everything
+    /// else — dominated interior holes, incomparable notches inside the
+    /// bounding box, scale-down / mixed-direction queries — keeps the
+    /// distance-gate miss: util genuinely doesn't transfer downward, but at
     /// the top of the collected range boundary efficiency is the best signal
     /// we have (the same reasoning as the unbounded m-curve / grid hold).
     fn beyond_frontier_scale_up(&self, q_site: &[f64]) -> bool {
-        let dominated = self
-            .site_logs
-            .iter()
-            .any(|(key, _)| key.iter().zip(q_site).all(|(&s, &q)| s as f64 >= q));
-        if dominated {
-            return false;
+        let mut above_some_max = false;
+        for (a, &q) in q_site.iter().enumerate() {
+            let (mut lo, mut hi) = (u32::MAX, 0u32);
+            for (key, _) in &self.site_logs {
+                lo = lo.min(key[a]);
+                hi = hi.max(key[a]);
+            }
+            if q < lo as f64 {
+                return false;
+            }
+            if q > hi as f64 {
+                above_some_max = true;
+            }
         }
-        (0..q_site.len()).all(|a| {
-            let min_a = self
-                .site_logs
-                .iter()
-                .map(|(key, _)| key[a])
-                .min()
-                .expect("beyond_frontier_scale_up on empty index");
-            q_site[a] >= min_a as f64
-        })
+        above_some_max
     }
 
     /// Evaluate one site's curve at coordinate `q`.
@@ -1016,6 +1016,22 @@ mod tests {
         }
         let cfg = gemm_cfg(&gemm_lat);
         assert!(query(&cfg, &t, &[32.0, 8192.0, 8192.0]).is_err());
+    }
+
+    #[test]
+    fn gemm_incomparable_notch_inside_bounding_box_still_misses() {
+        // (4096, 4096) sits in the notch between (256, 65536) and (65536,
+        // 256): no site dominates it, yet it exceeds the collected maximum on
+        // NO axis — an interior hole in the Pareto staircase, not a scale-up
+        // frontier query. The waiver must not fire; the gate stands.
+        let mut t = Node::branch();
+        for m in [16u32, 32, 64] {
+            for &(n, k) in &[(256u32, 65536u32), (65536, 256)] {
+                t.insert(&[m, n, k], gemm_lat(&[m as f64, n as f64, k as f64]));
+            }
+        }
+        let cfg = gemm_cfg(&gemm_lat);
+        assert!(query(&cfg, &t, &[32.0, 4096.0, 4096.0]).is_err());
     }
 
     #[test]
--- a/aic-core/rust/aiconfigurator-core/src/perf_database/perf_interp.rs
+++ b/aic-core/rust/aiconfigurator-core/src/perf_database/perf_interp.rs
@@ -14,8 +14,8 @@
 //!                             the dropped axis). ScatteredSites: site curve
 //!                             eval; unknown site -> nearest-site transfer in
 //!                             util space (log2 IDW, curve-coverage filter,
-//!                             distance gate; a site beyond the scale-up
-//!                             frontier falls back to boundary util-hold).
+//!                             distance gate; the gate is waived for a site
+//!                             beyond the scale-up frontier).
 //! 3. beyond the range      -> hold the boundary util (k_tail-median anchor),
 //!                             latency = SOL(query) / util
 //! 4. nothing to anchor on  -> Err (structured miss; never fabricate)
@@ -568,18 +568,19 @@ impl SiteIndex {
         let cmp =
             |a: &(f64, usize), b: &(f64, usize)| a.0.total_cmp(&b.0).then_with(|| a.1.cmp(&b.1));
         if let Some(gate) = max_site_distance {
-            if !ranked.iter().any(|&(d, _)| d <= *gate) {
-                // Direction-aware frontier fallback: a query site beyond the
-                // collected frontier in the scale-up direction holds the
-                // boundary util instead of missing (big-vocab LM heads at low
-                // tp). Interior holes and scale-down queries keep the miss.
+            if ranked.iter().any(|&(d, _)| d <= *gate) {
+                ranked.retain(|&(d, _)| d <= *gate);
+            } else {
+                // The gate is waived for a query site beyond the collected
+                // frontier in the scale-up direction (big-vocab LM heads at
+                // low tp): the ungated nearest (near-frontier) sites anchor
+                // the transfer below and SOL(query) carries the growth.
+                // Interior holes and scale-down queries keep the miss.
                 let q_site: Vec<f64> = site_axes.iter().map(|&p| coords[p]).collect();
-                if self.beyond_frontier_scale_up(&q_site) {
-                    return self.hold_frontier(cfg, ranked, coords, q);
+                if !self.beyond_frontier_scale_up(&q_site) {
+                    return Err(miss(cfg, coords, "no site within max_site_distance"));
                 }
-                return Err(miss(cfg, coords, "no site within max_site_distance"));
             }
-            ranked.retain(|&(d, _)| d <= *gate);
         }
         let k = (*nn_sites).min(ranked.len());
         if k < ranked.len() {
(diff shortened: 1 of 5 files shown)


aic-core/rust/aiconfigurator-core/src/perf_database/perf_interp.rs
```diff
@@ -610,6 +620,32 @@ impl SiteIndex {
...
+            let min_a = self
+                .site_logs
+                .iter()
+                .map(|(key, _)| key[a])
+                .min()
+                .expect("beyond_frontier_scale_up on empty index");
+            q_site[a] >= min_a as f64
+        })
+    }
+
```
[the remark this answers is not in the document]
tianhaox: Fixed in 6c90159d1: the predicate now requires the query to sit strictly above the collected maximum on at least one site axis (which implies no site dominates it, so the dominance scan is dropped) while keeping the per-axis minimum check — in both engines, with a mirrored incomparable-notch regression test. All motivating big-vocab LM-head cases exceed the collected n maximum and keep the waiver; real-database results are unchanged.

aic-core/src/aiconfigurator_core/sdk/perf_interp/engine.py
```diff
@@ -252,6 +254,21 @@ def _eval_curve(cfg: OpInterpConfig, curve: list, q, n_axes, curve_pos, site_pos
...
+    axis, and not below the collected minimum on any axis. Everything else —
+    dominated interior holes, incomparable notches inside the bounding box,
+    scale-down / mixed-direction queries — keeps the distance-gate miss: util
+    genuinely doesn't transfer downward, but at the top of the collected range
+    boundary efficiency is the best signal we have (the same reasoning as the
+    unbounded m-curve / Grid out-of-range hold)."""
+    axes = range(len(site_key))
+    if any(site_key[a] < min(s[a] for s in site_keys) for a in axes):
+        return False
+    return any(site_key[a] > max(s[a] for s in site_keys) for a in axes)
```
jasonqinzhou: The predicate waives the entire joint 2-octave safety gate whenever any axis exceeds its maximum. On this exact head, [Prediction Regression Gate run 30196326703](https://github.com/ai-dynamo/aiconfigurator/actions/runs/30196326703) reports 27 unrelated `rtx_pro_6000_server` / `Qwen3-32B-FP8` rows changing from `DATA_MISS` to `OK`, ranging from 15,775.6 ms to 250,773,178.6 ms. That database's `fp8_block` GEMM sites cover only `n,k=32..128`; a real logits query `(m=1,n=151936,k=5120)` is 12.011 octaves away, yet current HEAD returns 24,506.7 ms while `origin/main` correctly misses and the full FP8 table predicts 0.543 ms. These values are labeled valid silicon and can silently enter support/Pareto results. Please make the waiver opt-in or per-axis for the intended frontier overflow, retain a distance/coverage guard on the other site axes (or reject sparse simultaneous multi-axis overflow), mirror the rule in Rust, and add this real-table regression.

aic-core/src/aiconfigurator_core/sdk/perf_interp/engine.py
```diff
@@ -277,9 +294,15 @@ def dist(i: int) -> float:
 
     ranked = sorted(candidates, key=dist)
     if res.max_site_distance is not None:
-        ranked = [i for i in ranked if dist(i) <= res.max_site_distance]
-        if not ranked:
+        gated = [i for i in ranked if dist(i) <= res.max_site_distance]
+        if gated:
+            ranked = gated
+        elif not _beyond_frontier_scale_up(site_keys, site_key):
```
jasonqinzhou: `_beyond_frontier_scale_up` is evaluated over every collected site, but `ranked` may already contain only curve-covering candidates. The waiver can therefore pass even when every anchor actually used is mixed-direction or scale-down. For example, with site `(1,1)` covering only `m=1..2`, site `(64,64)` covering `m=16..64`, and query `(m=32,n=1024,k=1)`, both sites are outside the gate and only `(64,64)` is coverage-eligible. The global predicate passes because `n` exceeds the global maximum and `k` equals the global minimum, then transfers utilization downward from `k=64` to `k=1`; current HEAD returns latency `327680` instead of the documented mixed-direction miss. Please compute admissibility from the coverage-eligible candidates, restrict waived anchors to the scale-up direction, mirror the change in Rust, and add the paired regression.

aic-core/src/aiconfigurator_core/sdk/perf_interp/config.py
```diff
@@ -123,7 +125,14 @@ class ScatteredSites:
     #: How many nearest sites to blend (inverse-distance, util space) when the
     #: query site is not collected. 1 = single nearest.
     nn_sites: int = 4
-    #: Log-space distance gate: no site within this radius -> genuine miss.
+    #: Log-space distance gate: no site within this radius -> genuine miss for
```
jasonqinzhou, line 128: Please update `docs/design/perf_interp_design.md` with this public behavior change. Its ScatteredSites section still describes a "2.0-octave miss gate," while this change makes scale-up extrapolation unbounded. The accuracy table also omits the new frontier-holdout evidence reported in the PR, including p90 errors up to 112%. `ScatteredSites` is exported through both the core package and legacy compatibility namespace, so callers currently receive contradictory safety-contract documentation.

jasonqinzhou (changes requested):

## What this PR does

This review covers the implementation and verification changes in PR #1419.

## Why this score

The intended Gemma frontier fix is unsafe at the current predicate boundary: it accepts gross sparse-table extrapolations and can authorize a waiver from sites that are not eligible to anchor the query.

aic-core/src/aiconfigurator_core/sdk/perf_interp/engine.py
tianhaox: Fixed in 1fd0b86c1. The waiver now requires **exactly one** site axis to exceed the collected maximum — simultaneous multi-axis overflow (the fp8_block stub case) stays a structured miss — and the non-overflow axes must still pass the 2-octave gate (residual log2 distance), so the anchors are genuine frontier neighbours. Verified against the cited case on real tables: `(1, 151936, 5120)` under `fp8_block` now misses on every stub table (sglang 0.5.10/0.5.12/0.5.14, vllm 0.22.0/0.24.0) while the full `fp8` tables answer 0.54–0.81 ms; vllm 0.19.0's fp8_block has real coverage and resolves through the normal gated path. All motivating cases (Gemma tp1, big-vocab LM heads) are single-axis overflow with in-gate hidden dims and are byte-identical. Paired synthetic regression added in both engines (`test_sparse_stub_multi_axis_overflow_still_misses`).

aic-core/src/aiconfigurator_core/sdk/perf_interp/engine.py
tianhaox: Fixed in 1fd0b86c1. Admissibility is now computed over the coverage-eligible candidates (the same set the anchors come from) and `_frontier_waiver_anchors` returns the admissible anchor set directly, so predicate and anchors can no longer diverge. In your example the only coverage-eligible candidate is (64, 64), whose minimum puts k=1 below the candidates' range → miss (and even under the global minimum, the k=64→1 residual distance of 6 octaves fails the residual gate). The waived transfer is scale-up along the single overflow axis only. Mirrored in Rust; paired regression added in both engines (`test_waiver_admissibility_computed_over_coverage_candidates`).

aic-core/src/aiconfigurator_core/sdk/perf_interp/config.py
tianhaox: Done in 1fd0b86c1. `docs/design/perf_interp_design.md` now documents the waiver semantics in the ScatteredSites resolver row (single-axis overflow over coverage-eligible candidates, residual gate on the remaining axes, multi-axis overflow stays a miss) and adds the frontier-holdout evidence to the accuracy table: median 8.9–15.1% / p90 51–112% at the ≥3-octave holdout, with the tail attribution (skinny k≤256 shapes, systematically positive signed error — the safe direction), the B60 ground-truth numbers for the Llama LM-head shape (3.6–4.3% bf16 / 10–11% fp8), and the median-vs-IDW anchoring A/B.

Arsene12358 (approved):

jasonqinzhou (approved):
**APPROVE · 9.3/10 · high confidence**

## What this PR does

This PR lets the scattered-site GEMM interpolator answer single-axis scale-up queries beyond the collected frontier, primarily big-vocabulary LM-head shapes, while preserving structured misses for unsafe sparse-table and mixed-direction cases. The behavior is implemented in both Python and Rust with mirrored regression coverage and an updated interpolation design contract.

The follow-up turns concrete regression evidence into a focused boundary rule in both engines, adds paired tests for both reported failure modes, and documents the waiver's measured accuracy instead of leaving the safety contract implicit.

## Why this score

9.3/10, APPROVE. The current delta fixes both prior P1 findings and the P2 documentation gap: simultaneous multi-axis overflow is rejected, waiver eligibility and anchors now share the coverage-filtered candidate set with a residual gate on non-overflow axes, and the public design document describes the exception and error evidence. Targeted Python, Rust, and checked-in RTX Pro table checks all pass, and no actionable current-head finding remains.

PR #1420, opened 2026-07-26
feat(tools): perf-data sanity checks — flag suspicious measurements for review

tianhaox:
## What

A multi-reference **sanity checker** for the op-centric perf database. Different backends legitimately differ — this tool does NOT demand consistency. It flags **suspicious measurements** by checking each table against several independent reference frames, and asks a human to double-check them; nothing is blocked by default.

The four failure classes it hunts (and the reference frame each uses):

1. **Collector harness/config wrong** (eager mode, missing torch.compile): `systematic_offset` — the same backend pair shows a uniform median offset on *every* covered system; `region_deviation` — a whole shape region deviates from how the pair normally compares. Each labeled `harness_config` / `kernel_choice` / `unknown` by normalizing both sides' kernel_source via `kernel_source_backends.yaml`.
2. **Kernel backend selection wrong**: the attribution above, plus `kernel_choice_cost` — within one framework's table, the median price of each kernel vs the per-shape best.
3. **Collection environment broken**: `machine_op_deviation` — a (system, backend, op) off relative to the same machine's own level on its other ops (hardware factor removed; hint, ungated); `spike_violation` / `mono_violation` — points that break the latency curve's own shape, no reference framework needed.
4. **Impossibly fast data** (measured something else than defined): `below_sol` — gemm latency under the physical bound from the gpu spec; `nonpositive_latency` — collector corruption (delta/calibration tables exempt); cross-framework `faster_than_expected` suspects.

Every report leads with a **Triage** section: all detectors' evidence merged per (backend, op), ranked by strength — the 15 lines a human reads first; per-detector tables are the appendix.

## Verdict policy

Findings are **hints for double-checking, not bans**. `--fail-on-anomalies` exists for future gating and only trips when a kind's tally exceeds per-kind allowances (`--gate-threshold KIND=N`; e.g. 100 pair outliers, 20 region deviations) — a gate should fire on a pattern, not a stray point. It is **not enabled in CI**: the informational job posts the triage list to the job summary and uploads the full report as an artifact.

## CI

- `collector-check.yml`: new never-blocking "Perf data sanity" job on data PRs (full sweep ~4 min, full LFS pull)
- daily manifest run archives the same report

## Naming

`audit_*` perf-database tools renamed to `check_*`; `collector-audit.yml` → `collector-check.yml` (job `audit` → `check`). ⚠️ If branch protection lists the old required check name ("Collector Data Audit / Audit collector data"), update it when this merges.

## Sample findings (real, current tree — validated against known issues)

- vllm MLA modules: uniform 1.40-1.55x offset vs trtllm on all 7 systems — matches the independently-discovered missing-torch.compile collection config
- rtx_pro_6000_server sglang dsv4 modules: 40-88x off that machine's own level on a 15-op baseline (machine fingerprint)
- vllm gdn: 2.75-4.27x offset attributed `harness_config` (same kernel families as the references)
- trtllm/vllm fp8_block gemm corners: corroborated suspects up to 184x, cross-confirmed by 17k mono-drop curve violations


.github/workflows/collector-audit.yml
```diff
@@ -87,3 +87,64 @@ jobs:
...
+
+      - name: Pull LFS content for changed systems only
+        run: |
+          SYSTEMS=$(git diff --name-only "origin/${BASE_REF}" -- "${DATA_ROOT}" | awk -F/ '{print $6}' | sort -u)
+          if [ -z "${SYSTEMS}" ]; then
+            echo "No perf data changed; skipping LFS pull."
+            exit 0
+          fi
+          for sys in ${SYSTEMS}; do
+            git lfs pull --include="${DATA_ROOT}/${sys}/**"
```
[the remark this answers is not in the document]
tianhaox: Resolved by removal: the informational cross-backend CI job (and its LFS pull) was dropped from this PR in ca18eb61f — CI wiring is deferred to a follow-up, so collector-check.yml no longer touches LFS.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,798 @@
...
+    """Sortable key for backend version strings like '1.3.0rc10' or '0.5.6.post2'."""
+    parts = re.findall(r"(\d+|[a-zA-Z]+)", version)
+    key = []
+    for p in parts:
+        if p.isdigit():
+            key.append((1, int(p)))
+        else:
+            # Pre-release tags (rc, a, b) sort before the release; post after.
+            key.append((0 if p.lower() in ("rc", "a", "b", "alpha", "beta", "dev") else 2, p.lower()))
+    return tuple(key)
```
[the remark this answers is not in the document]
tianhaox: Fixed in b9ef0a4b5: `_version_key` now appends a rank-1 terminator element so a release sorts above its own pre-release (`1.0.0 > 1.0.0rc4`, `0.5.10 > 0.5.10rc0`) while post-releases stay above (`0.5.6.post2 > 0.5.6`). Verified against the 11 release/rc coexisting version dirs in the current tree.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,798 @@
...
+        text=True,
+        check=True,
+    ).stdout
+    root_parts = Path(data_root).parts
+    slices: set[tuple[str, str]] = set()
+    for line in out.splitlines():
+        parts = Path(line).parts
+        if parts[: len(root_parts)] != root_parts:
+            continue
+        rel = parts[len(root_parts) :]
```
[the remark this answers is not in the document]
tianhaox: Resolved by removal in b9ef0a4b5: `--changed-since` was dropped entirely — an adversarial review found four silent false-green/false-red failure modes in its git plumbing (this path-prefix issue, two-dot diff semantics, reuse-only version dirs, non-latest-version changes) and the full sweep is only ~3 min, so the mode was premature. It will be reintroduced with repo-root-relative paths and merge-base semantics when CI wiring lands.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,798 @@
...
+        type=float,
+        default=0.03,
+        help="Latency (same unit as the table, typically ms) below which monotonicity noise is ignored.",
+    )
+    parser.add_argument(
+        "--min-bucket-points",
+        type=int,
+        default=5,
+        help="Minimum joined points for a bucket/series to serve as a local baseline.",
+    )
```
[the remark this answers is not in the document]
tianhaox: Fixed in 7b871cacc: argparse-level validation now rejects `--anomaly-factor`/`--gap-factor` <= 1, `--gap-factor` above `--anomaly-factor`, and `--mono-tolerance` outside (0, 1].

tools/perf_database/check_kernel_source.py
```diff
@@ -15,17 +15,17 @@
   - `shared_fallback`  : kernel_source == "default" (framework-implicit,
                          low-fidelity placeholder)
 
-Rows whose kernel_source is blank/`<unknown>` are skipped during audit (and
+Rows whose kernel_source is blank/`<unknown>` are skipped during the scan (and
 logged) — they have no name to key inheritance on, and the current corpus has
 zero such rows.
```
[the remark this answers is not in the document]
tianhaox: Fixed in 7b871cacc: `_scan_one_file` now skips the literal `<unknown>` the same as blank (counted in `rows_unnamed_kernel_source`), matching the docstring; regression test added in tests/unit/tools/test_check_kernel_source_scan.py.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,948 @@
...
+
+
+def _load_op_table(path: Path, backend: str, version: str) -> tuple[OpTable | None, dict[str, int]]:
+    """Returns (table, nonpositive-latency row counts by kernel_source).
+
+    Nonpositive rows are excluded from the returned frame but reported so the
+    caller can emit them as Layer-1 anomalies instead of silently cleaning
+    them up.
+    """
+    df = _read_table(path)
```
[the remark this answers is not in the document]
tianhaox: Fixed, and went one step further than skip-and-warn: an unreadable table is itself the most extreme form of the corruption this tool hunts, so the read failure is caught per table (the sweep continues), emitted as a Layer-1 `unreadable_table` finding (triage list + report section), and gated with zero allowance in `--fail-on-anomalies` mode. Verified with a truncated parquet in a synthetic tree: sweep completes, finding is triage item #1, gate exits 1.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,1664 @@
...
+    latency_col = _pick_latency_column(df.columns)
+    if latency_col is None:
+        logger.warning("No latency column in %s; skipping", path)
+        return None, {}, []
+    if "kernel_source" not in df.columns:
+        df["kernel_source"] = "<unknown>"
+    nonpositive = df[df[latency_col].notna() & (df[latency_col] <= 0)]
+    npos_by_ks = {str(k): int(v) for k, v in nonpositive["kernel_source"].value_counts().items()}
+    shape_cols = [c for c in df.columns if c not in _META_COLUMNS and c not in _LATENCY_COLUMNS_PRIORITY]
```
jasonqinzhou: `_META_COLUMNS` contains `op_name`, but this checker cannot safely treat it as metadata because several tables store multiple logical operations in one file and the SDK indexes them by `op_name`. On the current H200 MLA-BMM latest tables, this reduction shrinks both SGLang and TRT-LLM from 848 rows to 448, and 115 of the resulting 448 joined shapes compare `mla_gen_pre` on one side with `mla_gen_post` on the other. Those ratios are not comparisons of the same operation. Please partition by `op_name` (or retain it in the comparison identity) before selecting the minimum latency, and add a regression test with two operation names sharing the same numeric shape.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,1664 @@
...
+    if kind == "region_deviation":
+        return f"{kind}|{a['system']}|{a['op_file']}|{_backends_of(a['pair'])}|{a['bucket']}"
+    if kind in ("mono_violation", "spike_violation"):
+        at = a.get("sweep_at", f"{a.get('sweep_from')}>{a.get('sweep_to')}")
+        return (
+            f"{kind}|{a['system']}|{a['op_file']}|{a['backend']}|{a['kernel_source']}|"
+            f"{a['sweep_col']}|{_sig(a['fixed_shape'])}|{at}"
+        )
+    # Group-level kinds: nonpositive_latency, below_sol, unreadable_table.
+    return f"{kind}|{a['system']}|{a['op_file']}|{a['backend']}|{a.get('gemm_dtype', '')}"
```
jasonqinzhou, line 580: The ratchet suppresses this entire group-level finding by key, including any newly added rows or points. For example, after baselining one nonpositive row, growing the same table to 500 nonpositive rows still produces the identical key and the gate reports zero new findings; `below_sol` behaves the same way within a dtype. That violates the stated contract that a PR should fail for problems it introduces. Please preserve per-point identities or baseline and subtract the aggregate counts instead of dropping the whole finding, and add a CLI regression test for a count increasing inside an existing group.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,1664 @@
...
+    within-framework kernel-choice costs).
+
+    Nonpositive rows are excluded from the returned frame but reported so the
+    caller can emit them as Layer-1 anomalies instead of silently cleaning
+    them up. Kernel-choice costs quantify, for tables that measured several
+    kernel_sources on the same shapes, how much slower each kernel's median is
+    than the per-shape best — the price of picking that backend.
+    """
+    df = _read_table(path)
+    latency_col = _pick_latency_column(df.columns)
```
jasonqinzhou: Selecting the first recognized latency column is not valid for the current WideEP schemas. `wideep_deepep_ll_perf.parquet` contains combine and dispatch times, and the runtime consumer sums both, but this checker selects only `combine_avg_t_us`; a corrupt dispatch measurement is therefore invisible. `wideep_deepep_normal_perf.parquet` uses four timing components that the consumer sums, but none is recognized here, so both real H100/H200 tables are warning-only skips and disappear from the artifacts. Please derive the same effective latency as the consumer for supported multi-component schemas, or explicitly report unsupported schemas with deliberate gate semantics, and cover both paths with tests.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,1664 @@
...
+        return []
+
+
+def _normalize_kernels(kmap: list[dict], framework: str, op_file: str, kernel_sources) -> set[str]:
+    out: set[str] = set()
+    for entry in kernel_sources:
+        for ks in str(entry).split(","):
+            ks = ks.strip()
+            if ks:
+                out.add(translate(kmap, framework, op_file, {}, ks) or ks.lower())
```
jasonqinzhou: The translation registry's conditional entries cannot match here because `op_file` includes the `.parquet` suffix while the registry expects stems such as `gemm_perf`, and the empty axis map omits conditions such as `gemm_dtype`. A current SGLang FP8-block GEMM therefore normalizes to generic `sgl_kernel` instead of `deepgemm`, and SGLang MLA-BMM `default` remains opaque instead of mapping to `cublas`. That can make the reported `kernel_choice` versus `harness_config` attribution wrong or unknown. Please carry the relevant row/fact axes with each kernel source and pass the operation stem to `translate()`.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,1664 @@
...
+    """
+    df = _read_table(path)
+    latency_col = _pick_latency_column(df.columns)
+    if latency_col is None:
+        logger.warning("No latency column in %s; skipping", path)
+        return None, {}, []
+    if "kernel_source" not in df.columns:
+        df["kernel_source"] = "<unknown>"
+    nonpositive = df[df[latency_col].notna() & (df[latency_col] <= 0)]
+    npos_by_ks = {str(k): int(v) for k, v in nonpositive["kernel_source"].value_counts().items()}
```
jasonqinzhou, line 284: `value_counts()` drops null groups by default, so a nonpositive row with `kernel_source = NaN` is omitted from `npos_by_ks`. A synthetic one-row table with latency 0 and a null kernel source returns an empty count, then the table is discarded and no anomaly is emitted. Please normalize missing labels to `<unknown>` before grouping or use `dropna=False`, and add a regression case so corrupt latency rows cannot disappear because their label is absent.

tools/perf_database/check_cross_backend.py
```diff
@@ -0,0 +1,1664 @@
...
+            continue
+        b_label, _, a_label = s["pair"].partition(" vs ")
+        groups[(s["op_file"], a_label.split("/")[0], b_label.split("/")[0])].append(s)
+
+    log_thresh = math.log(offset_factor)
+    offsets: list[dict] = []
+    for (op_file, backend_a, backend_b), entries in groups.items():
+        if len(entries) < min_systems:
+            continue
+        logs = sorted(math.log(e["median_ratio"]) for e in entries)
```
jasonqinzhou, line 918: This detector checks only that every system is beyond the threshold in the same direction; it never checks that the multiplier is near-constant. Synthetic ratios of 1.2x, 4x, and 20x are therefore reported as one "uniform" 4x systematic offset even though that pattern does not support the harness/configuration conclusion described in the report. Please add a dispersion or consistency criterion across systems, or rename the detector and weaken the attribution claim to match the behavior.

jasonqinzhou (changes requested):

The score is 5.0/10 because the direction is valuable and the implementation has broad verification, but three P1 correctness gaps currently make key comparisons and the ratchet unreliable: op_name variants can be merged, new corruption can be suppressed inside baselined groups, and multi-component WideEP latency is mismeasured or skipped. The three P2 findings further affect attribution and anomaly coverage, so changes are requested before this is review-ready.

tools/perf_database/check_cross_backend.py
tianhaox: Fixed in cfcdd13f3: op_name is no longer treated as metadata for this checker (`_SHAPE_EXCLUDED_META = _META_COLUMNS - {"op_name"}`) — it partitions the min-reduction, the join, the buckets, and the cluster keys. Verified on the H200 mla_bmm latest tables: 848 rows retained (was 448), and mla_gen_pre/post can no longer be joined against each other. Regression test added: `test_op_name_partitions_the_min_reduction`.

tools/perf_database/check_cross_backend.py
tianhaox: Fixed in cfcdd13f3: baselines are now count-aware (schema 2). Group-level kinds (nonpositive rows, below-SOL points) store their COUNT per key; `evaluate_gate` suppresses only up to the baselined count, so a group growing from 1 to 500 rows contributes 499 new findings. Point-level kinds keep key-membership suppression. The gate logic was extracted into testable `snapshot_baseline`/`evaluate_gate` functions; regression tests cover count growth inside a known group, unchanged-group suppression, and version-bump key stability (`test_ratchet_flags_count_growth_inside_a_known_group`, `test_ratchet_point_findings_suppressed_by_key`).

tools/perf_database/check_cross_backend.py
tianhaox: Fixed in cfcdd13f3: `_resolve_latency` now supports multi-component schemas summed like the consumer — wideep_deepep_ll audits combine+dispatch summed, and wideep_deepep_normal's four notify/transmit components are recognized (both real H100/H200 tables now enter the audit; verified 437 rows loaded). Derived `*_bandwidth_gbps` columns are excluded from the shape key. Tables with unrecognized schemas raise `_SchemaUnsupportedError` and are reported as explicit `unsupported_schema` entries in the report (deliberately informational, not gated — the report tells you to register the schema instead of silently skipping). Tests: `test_component_latency_is_summed_like_the_consumer`, `test_unsupported_schema_raises_explicitly`.

tools/perf_database/check_cross_backend.py
tianhaox: Fixed in cfcdd13f3: `_normalize_kernels` now passes the op STEM to `translate()` and carries the cluster's categorical axes (gemm_dtype etc., and now op_name) as the axis map, so conditional registry entries match — verified: sglang gemm fp8_block → `deepgemm`, sglang mla_bmm `default` → `cublas`. Where axes are unavailable (pair-level offset aggregation), labels whose only registry entries are conditional resolve to an `unresolved:` marker that attribution treats as opaque → `unknown`, rather than guessing.

tools/perf_database/check_cross_backend.py
tianhaox: Fixed in cfcdd13f3: kernel_source is normalized with `fillna("<unknown>")` before the nonpositive counting, so a corrupt latency row can no longer vanish because its label is absent. Regression test: `test_nan_kernel_source_nonpositive_rows_are_counted` (one-row table, latency 0, empty label → counted under `<unknown>`).

tools/perf_database/check_cross_backend.py
tianhaox: Fixed in cfcdd13f3: added a dispersion criterion — a systematic offset now also requires the per-system medians to sit within `--offset-dispersion` (default 2.0x) of each other; same-direction 1.2x/4x/20x no longer reports as one uniform offset (regression test `test_systematic_offset_requires_near_constant_multiplier`). The emitted finding carries the observed `spread`. On the full tree this culled 18 → 10 offsets, all with spread ≤ 1.9 — notably the vllm MLA-module offsets survive at spread 1.14–1.27, which is the uniformity the harness/configuration conclusion actually needs.

jasonqinzhou (approved):
**APPROVE · 9.2/10 · high confidence**

The current head addresses all six prior findings with focused regression coverage. The comparison identity, count-aware ratchet, consumer-equivalent WideEP latency, conditional kernel attribution, null-label handling, and systematic-offset dispersion behavior all now match their intended contracts. No remaining blocking or actionable P2 finding was identified.

Issue #1396, opened 2026-07-22
AIConfigurator's prediction fidelity is far from actual GPU benchmarking on Kimi-K2.5-NVFP4


wanzhenchn:
### AIC Throughput-vs-Speed Summary 

Experiment: `Kimi-K2.5-NVFP4_b200_sxm_vllm_isl8192_osl1024_ttft1000000000_tpot1000000000_439084`

Notes:

- top1 has been removed from this derived summary per request; raw benchmark artifacts are unchanged.
- Throughput is `tok/s/GPU` on `8*B200`: AIC uses `tokens/s/gpu`; measured uses AI-Perf output token throughput divided by 8 GPUs.
- Speed is `tok/s/user`: AIC uses `tokens/s/user`; measured uses AI-Perf `output_token_throughput_per_user.avg`.
- top3 has no benchmark JSON; top4 failed during deployment readiness and therefore has no measured speed/throughput point.

### Throughput-vs-Speed Points

| Rank | Status         | Conc | AIC speed tok/s/user | AIC throughput tok/s/GPU | Measured speed tok/s/user | Measured throughput tok/s/GPU |
| ---- | -------------- | ---- | -------------------- | ------------------------ | ------------------------- | ----------------------------- |
| top2 | succeeded      | 384  | 21.87                | 883.70                   | 44.59                     | 192.44                        |
| top3 | missing_result | 216  | 22.54                | 560.42                   | n/a                       | n/a                           |
| top4 | failed         | 216  | 22.49                | 559.24                   | n/a                       | n/a                           |
| top5 | succeeded      | 116  | 22.67                | 302.76                   | 36.53                     | 228.66                        |
| top6 | succeeded      | 12   | 51.40                | 71.01                    | 82.72                     | 109.25                        |

## Eval scripts

<details>
<summary> kimi_b200_aic_vs_gpu.py</summary>

```python
#!/usr/bin/env python3
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0

"""P/D disagg AIC-vs-GPU evaluation for nvidia/Kimi-K2.5-NVFP4 on 8x B200.

All build and test entry points for this experiment live in ~/workspace.

Workflow:
  python3 kimi_b200_aic_vs_gpu.py docker-build
  docker run --rm --network=host --gpus all -v "$PWD/results:/workspace/results" \
[...]

Issue #1208, opened 2026-06-05
Consolidate KV-cache OOM check onto the shared memory.kv_cache_budget_bytes


simone-chen:
Follow-up to #1159.

`sdk/memory.py` introduced `kv_cache_budget_bytes(*, capacity, non_kv, fraction, of_free, reserved=0.0, tolerance=0.0)` as the single KV-cache budget formula, used by `estimate_kv_cache`.

`InferenceSummary._check_and_set_kv_cache_oom` (`sdk/inference_summary.py`) still carries its own copy of the same budget math — `(capacity - non_kv) * fraction * (1 - reserved) * (1 - tolerance)`. The two should share `memory.kv_cache_budget_bytes` so they cannot drift.

#### Scope

- Point `_check_and_set_kv_cache_oom` at `memory.kv_cache_budget_bytes` (OfFree, passing `reserved` + `tolerance`).
- Reconcile the semantic differences surfaced during #1159: `_native_estimate` currently omits the `reserved` (block-allocator overhead) term and applies tolerance as a separate `tolerance_adjusted` step, whereas the OOM check folds `reserved`/`tolerance` into the fraction multiplicatively. Decide whether the estimate should also honor `reserved`, and how OfTotal should treat it.
- Keep units consistent across callers (the helper is unit-agnostic; the OOM check works in GiB, the estimate in bytes).

#### Why deferred

Split out of #1159 to keep that PR focused on the Mocker-facing capacity API. The OOM path feeds the config sweep (`inference_session`, `pareto_analysis`), so changing it is a separate concern with its own regression surface.

PR #1423, opened 2026-07-26
fix: enforce vLLM KV-cache budget in disagg sweep; run module collectors with engine-fused ops (#1396)

tianhaox:
## Summary

Two fixes for the prediction-fidelity gap reported in #1396 (AIC over-predicts Kimi-K2.5-NVFP4 vLLM disagg throughput 4.6x on 8xB200), verified end-to-end on 8xB200 with dummy weights (same memory footprint and kernel timing as real weights).

### 1. `fix(sdk)`: enforce vLLM `gpu_memory_utilization` KV budget in the disagg sweep

The fraction-based KV-cache OOM check never ran for vLLM: only `trtllm_backend` overrides `get_default_free_gpu_memory_fraction()`, so the static (disagg component) path packed memory to 100% of device capacity. On Kimi-K2.5-NVFP4 / 8xB200 / isl8192 osl1024 the search selected a decode worker at bs=96/rank (KV 28.95 GiB/GPU) while a real vLLM 0.22 deployment admits only ~29 seqs per 4-rank worker (17 under dynamo+NIXL) — issue #1396's top1–top5 rows are all memory-infeasible configs, and the reporter's measured effective concurrency (192.44×8/44.59 ≈ 34.5 vs configured 384) matches the KV admission limit.

- `VLLMBackend` now supplies `gpu_memory_utilization=0.9` defaults with **of_total** semantics (the fraction caps TOTAL device memory; TRT-LLM's `free_gpu_memory_fraction` caps the free pool after non-KV — applying the trtllm formula to vLLM would compute a ~26 GiB budget instead of the correct ~8 GiB).
- Budget math is delegated to `memory.kv_cache_budget_bytes`, closing the in-code `TODO(#1208)`.
- `run_static` passes backend defaults into the OOM check, and both disagg worker-candidate enumerators (`sweep._get_disagg_worker_candidates` — the one `cli exp` actually uses — and `DisaggInferenceSession.get_worker_candidates`) exclude KV-infeasible batch sizes; `run_disagg` propagates the flag.

**Validation** (B200, dummy weights, same yaml as #1396): top1 moves from bs96/conc384 (predicted 871.6 tok/s/gpu; measured 213.7 end-to-end) to bs26/conc104 (predicted 366.4) — within 10% of the measured KV admission limit (29 seqs/worker standalone) and 12% of the measured standalone decode ceiling (2616 tok/s).

### 2. `fix(collector)`: run vLLM module benchmarks with engine-fused ops

Module-level collectors execute vLLM modeling code eagerly with no IR-op priority installed, so `rms_norm`/`fused_add_rms_norm` dispatch to native implementations (eager mean/pow/rsqrt/mul chains) and CustomOps (rotary) run native decompositions — a batch-independent ~50–65 µs/layer that a serving engine never pays (inductor fuses these under torch.compile; eager regions use the `vllm_c` CUDA kernels). Every collector run prints the smoking gun: `Priority not set for op rms_norm, using native implementation`. This inflates small/medium-batch module datapoints ~25–35% and is the main source of #1396's opposite-direction error at low concurrency (top6 measured 60% faster than predicted).

- `create_vllm_config` enables custom CUDA ops (`custom_ops=["all"]`).
- New `enable_engine_fused_ops()` installs the `vllm_c` implementations as the process-lifetime IR-op default (`IrOp.set_priority` is a scoped context manager; `set_default` is the permanent API) and **raises** where the registry/provider is missing — a silent fallback to the native chain would benchmark a kernel path serving never runs (per `collector/layer_permissions.md`).
- Wired into `collect_mla_module` (MLA/DSA) and `collect_dsv4_attn`, the two module collectors that instantiate norms/rotary. Kernel-level collectors (`collect_gemm`, `collect_attn`, `collect_mhc_module`, `collect_gdn`) and `collect_moe` (verified by kernel-level profiling: ~4 µs eager glue) are unaffected.

**Measured** (MLA generation module, heads64/kv8192/fp8, vLLM 0.22.0 and 0.24.0 containers, B200): bs1 109.5→78.6 µs, bs16 143.7→95.2 µs, bs64 198.6→147.9 µs; the bs16 point moves from +37% vs the torch-profiler-measured in-engine per-layer cost to within ~10%.

## Test plan

- `pytest tests/unit/sdk` (1445 passed incl. 4 new budget-semantics tests; `tests/unit/sdk/database/test_moe_dispatch.py` collection error pre-exists on main)
- `pytest tests/unit/collector` (back to main baseline: 3 pre-existing GLM-naming failures)
- `ruff check` clean on changed files; `ruff format --check` clean (2 offenders pre-exist on main); `git diff --check` clean
- Runtime smoke inside `vllm/vllm-openai:v0.24.0` (exact provenance of the shipped 0.24.0 tables): quick point runs at 97.5 µs with no native-impl warning
- Full `cli exp` sweep on the #1396 yaml before/after (results above)

## Notes for reviewers / follow-ups (not in this PR)

- The two commits are module-isolated (sdk vs collector) per `.claude/rules`; happy to split into two PRs if preferred.
- Affected perf tables (`mla_*_module_perf`, `dsa_*`, dsv4 attention) need **re-collection** for the small/medium-batch region to benefit from fix 2; shipped parquet is unchanged by this PR.
- NIXL/dynamo KV-buffer reservation is not yet modeled (real decode admission drops 29→17 seqs/worker under dynamo).
- Weights for checkpoint-level BF16-ignored layers (`quantization_config.ignore`: all `self_attn` + `lm_head` in the nvidia Kimi-K2.5-NVFP4 checkpoint) are still estimated as NVFP4: −4.8 GiB/GPU underestimate (147.18 predicted vs 151.95 measured).
- SGLang's `mem_fraction_static` has the same unwired default (of_total semantics ready to reuse).
- Small-token MoE distribution assumption (`power_law_1.01`) still shows ~+25% vs the issue's real-weights top6 anchor; needs a real-weights routing histogram to close.

## Results on the #1396 workload (Kimi-K2.5-NVFP4, 8xB200, vLLM, isl 8192 / osl 1024)

| | top-1 decode config | predicted tok/s/gpu |
|---|---|---|
| before (main) | dp4, bs 96/rank — KV-infeasible | 883.24 |
| after (this PR) | dp4 etp4, bs 18/rank | 310.53 |

Measured on this workload: 192.44 tok/s/gpu (issue author, real weights) / 213.7 (our dummy-weights reproduction). Prediction error: **4.1x -> ~1.45x**. The residual is not kernel speed: in the reproduction, clean decode steps measure median ITL 16.6 ms (faster than predicted) while mean TPOT is 33.4 ms — the gap sits in disagg runtime effects AIC does not model (KV-transfer stalls and P/D scheduling bubbles, not yet isolated from each other).

Fixes the memory-infeasibility half of #1396.

## Follow-on loader fixes (added after the first review round)

Validating the re-collected data surfaced three loader defects; each is a separate commit with its own tests, Python and Rust in parity:

1. **Shared-layer first-wins** — 11 module-level loaders used direct assignment, so nearest-earlier sibling-version rows (loaded after the primary) silently overrode the requested version's own data, violating the design §6.1 contract ("a donor can never shadow the primary"). E.g. b200/vllm/0.24.0 MLA-module queries returned 0.19.0 values verbatim, and one h200 pin had captured a 1.2.0rc5 row shadowing 1.3.0rc10's own value. Fixed with the standard skip-on-conflict guard; the Rust sites that had deliberately mirrored the last-wins behavior flipped to `or_insert` / `Node::insert_first_wins`, and the tests that pinned last-wins were re-minted.
2. **DSV4 head identity** — the DSV4 kind-module loaders keyed on the `num_heads` column alone and dropped `tp_size`, collapsing a model identity and its whole tp sweep (30-50% latency spread) into one row-order-resolved bucket; a Pro tp8 query actually read Flash tp8 rows. Tables now key `[native][local]` and queries resolve the model's native count (identity) before the rank-local count (per-rank shape).
3. **num_heads semantics sniffer** — shipped DSV4 files disagree on the column's meaning (sglang 0.5.10 = native, collected pre-#1131; vllm 0.24.0 = rank-local, post-#1131 — see #1429 for the collector-side unification). The loaders sniff the semantics per (model, version) from the data itself, so both file generations land in identical `[native][local]` buckets; verified against the real b200 files.

DSV4 prediction values move as a result (Pro queries now read Pro's own, correctly-labeled rows); the two Rust b200/sglang oracles were re-minted from the fixed Python engine. Collector-side convention unification and sglang re-collection are tracked in #1429.

## Summary by CodeRabbit

* **New Features**
  * Added KV-cache-aware OOM budgeting with backend-specific defaults and configurable “free vs total” KV-budget semantics (including new `free_gpu_memory_fraction` wiring across disaggregated prediction/sweeps).
  * Enabled fused/optimized engine execution for vLLM during collection/benchmark runs.
  * Improved per-checkpoint attention quantization exclusion handling for DeepSeek-family models.
* **Bug Fixes**
  * Disaggregated candidate filtering now rejects configurations failing either model memory OOM or KV-cache budget OOM.
  * Static-path OOM checks now consistently factor KV-budget constraints.
* **Tests**
  * Added/expanded unit tests for KV-budget semantics and vLLM/MLA + DeepSeek quantization behaviors.


aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -779,7 +779,11 @@ def _run_generation(bs: int, bw: int, effective_isl: int, eff_osl: int, strd: in
         summary.set_context_power_avg(context_power_avg)
         summary.set_generation_power_avg(generation_power_avg)
         summary.set_e2e_power_avg(e2e_power_avg)
-        summary.set_memory_and_check_oom(memory, database.system_spec["gpu"]["mem_capacity"])
+        summary.set_memory_and_check_oom(
+            memory,
+            database.system_spec["gpu"]["mem_capacity"],
+            **self._static_oom_check_kwargs(database.system_spec["gpu"]["mem_capacity"]),
```
jasonqinzhou: [P1] Thread the configured and versioned memory fraction through disagg

This new static check always resolves `_static_oom_check_kwargs(...)` from a generic backend default. The public `Task.free_gpu_memory_fraction` is populated for disagg, and the estimate API accepts the same value, but `sweep_disagg_kwargs`, `sweep_disagg`, `predict_disagg_worker`/`run_static`, `run_disagg`, and `_run_disagg_estimate` never forward it. A deployment requested at 0.80 is therefore filtered at vLLM's 0.92 default (or SGLang's derived default), so the new candidate gate can still select a batch the requested deployment rejects. The fallback also needs the resolved backend version: supported vLLM 0.14/0.19 use 0.90, while 0.22/0.24 use 0.92. Please thread the explicit value and backend version through every disagg/static path, fall back only when the user omitted it, and cover both Task and estimate APIs.

aic-core/src/aiconfigurator_core/sdk/backends/sglang_backend.py
```diff
@@ -9,6 +9,39 @@
...
+    capacity_mb = mem_capacity_bytes / (1 << 20)
+    if capacity_mb < 20 * 1024 or capacity_mb < 35 * 1024:
+        chunked_prefill = 2048
+    elif capacity_mb < 60 * 1024:
+        chunked_prefill = 4096
+    elif capacity_mb < 160 * 1024:
+        chunked_prefill = 8192
+    else:
+        chunked_prefill = 16384
+    reserved_mb = 512 + 1.5 * chunked_prefill
```
jasonqinzhou, line 50: [P1] Include SGLang's configuration-dependent reserves

The omitted terms are much larger than the stated 0.5-1.5 GiB bias. The supported SGLang v0.5.14 startup reserve (`python/sglang/srt/server_args.py:3276-3314@v0.5.14`) adds decode-graph memory plus `max_bs * dp_size * 3`, and another `* 1.5` when `max_bs > 300`, for DP attention. AIC enables DP attention for MoE deployments with DP > 1 and sweeps DP up to 8 normally (64 in WideEP). On B200 with the upstream default `max_bs=512` and DP=8, those DP terms alone are 18,432 MiB; this capacity-only helper returns 0.864 while the resolved runtime default is roughly 0.76 after the omitted terms. The new filter can therefore retain workers SGLang cannot allocate. Please derive the fraction from role/TP/PP/DP/graph/DeepEP configuration and backend version (or use a defensible conservative reserve) and add DP-attention decode regressions.

collector/vllm/utils.py
```diff
@@ -206,6 +206,13 @@ def create_vllm_config(
     device_config = DeviceConfig()
     load_config = LoadConfig()
     compilation_config = CompilationConfig()
+    # Module-level benchmarks run eagerly (no torch.compile), so without this
+    # every CustomOp (rotary, activations, ...) falls back to its native
+    # decomposed implementation — a batch-independent per-call overhead the
+    # serving engine does not pay (inductor fuses these under compilation).
+    # Enable the CUDA custom ops so eager module forwards match engine
+    # execution. See enable_engine_fused_ops() for the IR-op counterpart.
+    compilation_config.custom_ops = ["all"]
```
jasonqinzhou, line 215: [P1] Benchmark the serving dispatch instead of an eager surrogate

For the canonical DeepSeek-V3 MLA and GLM/DeepSeek-V3.2 DSA cases, this changes the timed path away from vLLM 0.24's serving path. `CompilationConfig()` resolves to VLLM_COMPILE/Inductor, for which CUDA's `get_default_ir_op_priority()` selects native IR ops (`vllm/platforms/cuda.py:636-653@v0.24.0`) and disabled CustomOps are fused by Inductor; this collector stays eager and instead forces all CustomOps plus `vllm_c`. The helper itself acknowledges exact parity would require `torch.compile`. Under the collector dispatch rule, a latency-matched substitute is not sufficient because it can change kernel families and shape behavior while producing plausible rows. Please execute the module through vLLM's actual compile/worker dispatch (or record an explicitly eager serving mode separately) and assert the provider that ran. Also, `set_default(["vllm_c"])` is not fail-closed: `_filter_priority_impls()` appends native when the provider does not cover all calls (`vllm/ir/op.py:352-374@v0.24.0`), contrary to the stated guarantee.

aic-core/src/aiconfigurator_core/sdk/models/helpers.py
```diff
@@ -36,6 +36,44 @@
...
+def attention_modules_excluded_from_quant(raw_config: dict) -> bool:
+    """Whether the checkpoint keeps self-attention projections unquantized (bf16).
+
+    Matches either full projection names (e.g. ``self_attn.q_a_proj``) or the
+    layer-prefixed globs the ModelOpt exporter emits (``model.layers.N.self_attn*``).
+    Per-checkpoint fact, not a family constant: the NVFP4 releases of
+    DeepSeek-R1/V3.2, Kimi-K2.5 and GLM-5 exclude attention (and lm_head),
+    while DeepSeek's native FP8 checkpoints quantize attention too.
+    """
+    return any("self_attn" in str(pattern) for pattern in quant_exclude_patterns(raw_config))
```
jasonqinzhou: [P1] Preserve per-projection attention quantization

`any("self_attn" ...)` does not establish that the whole attention block is unquantized. Both bundled NVIDIA DeepSeek V3.1/V3.2 NVFP4 exclusion lists contain q/kv projections (and V3.2 indexer entries) but no `o_proj`; the published V3.2 shard metadata likewise has BF16 q/kv tensors while `self_attn.o_proj.weight` is quantized with NVFP4 scales. This boolean then assigns one BF16 mode to every attention projection. On the existing V3.1 fallback alone, `context_proj_gemm` becomes 13.34 GiB instead of approximately 3.75 GiB and uses the wrong perf key; the V3.2 DSA module key has the same mixed-checkpoint ambiguity. Please model exclusions per projection (including a mixed module/perf identity where needed) and add partial-exclusion tests rather than asserting that any match means all attention is BF16.

aic-core/src/aiconfigurator_core/sdk/models/deepseek_v32.py
```diff
@@ -126,13 +112,16 @@ def create(cls, model_info: dict, model_config, backend_name: str) -> BaseModel:
...
-            model_info.get("raw_config", {})
-        ):
+        # Checkpoint-driven, not architecture-gated: nvidia/DeepSeek-V3.2-NVFP4
+        # excludes every self_attn* (incl. the indexer) exactly like the GLM-5
+        # NVFP4 releases, and vLLM honors ModelOpt exclude_modules wildcards for
+        # any architecture (hf_quant_config.json is read in
+        # transformers_utils/config.py:726; excluded prefixes fall back to the
+        # unquantized path via ModelOptNvFp4Config.is_layer_excluded /
+        # is_layer_skipped, modelopt.py:150-161 @0.24.0).
+        if _dsa_attention_modules_excluded_from_quant(model_info.get("raw_config", {})):
```
jasonqinzhou: [P1] Count DSA projection weights in the new memory gate

Changing the DSA compute mode does not change its memory contribution: both `ContextDSAModule` and `GenerationDSAModule` initialize `_weights = 0.0`, and `BaseBackend._get_memory_usage()` sums those operation weights for the model total. A current-head construction of `nvidia/DeepSeek-V3.2-NVFP4` therefore reports `context_attention = 0 GiB`, omitting roughly 13 GiB/rank at TP1 for the 61 layers' q/kv/indexer/output projections. Because this PR now uses that total to size the KV budget, the omission directly admits infeasible batches. Please implement checkpoint-aware DSA `get_weights()` accounting (including the mixed projection dtypes) and add an end-to-end model-memory regression.

aic-core/src/aiconfigurator_core/model_configs/nvidia--DeepSeek-V3.2-NVFP4_config.json
```diff
@@ -0,0 +1,56 @@
+{
```
jasonqinzhou, line 1: [P2] Register the new bundled model config

These two cached files are never selected because `nvidia/DeepSeek-V3.2-NVFP4` was not added to `common.DefaultHFModels`. `_load_model_config_from_model_path()` consequently takes the Hugging Face download branch even though the packaged config exists; a current-head probe that makes `_download_hf_config` raise fails with `network path reached`. Please register the model ID and add a regression that loads both config and `hf_quant_config` with downloads disabled.

aic-core/src/aiconfigurator_core/sdk/inference_summary.py
```diff
@@ -109,6 +109,7 @@ def set_memory_and_check_oom(
         free_gpu_memory_fraction: float | None = None,
         kv_cache_reserved_fraction: float = 0.0,
         kv_cache_tolerance: float = 0.0,
+        fraction_of_free: bool = True,
```
jasonqinzhou, line 117: [P2] Carry the fraction semantics into the detail report

`fraction_of_free` affects the feasibility check here but is not stored on the summary. `estimate_detail_report.py` therefore always derives the displayed maximum batch as `(capacity - non_kv) * fraction / kv_per_seq`, which is TRT-LLM's of-free formula. For vLLM/SGLang it must be `(capacity * fraction - non_kv) / kv_per_seq`: with capacity 100, non-KV 50, fraction 0.9, and one unit per sequence, the report prints 45 although the correct bound is 40. Please persist the semantic flag and reuse `kv_cache_budget_bytes` for the report so the user-facing recommendation cannot disagree with the gate.

.claude/skills/aic-silicon-align/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+   -8/+2.3/-1.4 GiB netting to a small total delta).
+4. **Standalone per-role benchmarks** (prefill worker, decode worker) at
+   KV-feasible batch sizes; compare against AIC's per-role rows.
+5. **Deployment-faithful e2e**: exact images, generated scripts, patches,
+   and flags the reporter used. Only this rung may be compared to the
+   reported end-to-end numbers.
+
+## What dummy weights do and don't preserve
+
+- Preserve: memory footprint (identical allocation), all data-independent
```
jasonqinzhou: [P2] Make dummy-weight parity conditional on the loader

This blanket guarantee is unsafe: dummy loaders do not universally preserve checkpoint allocation or quantized-kernel behavior. This tree already documents vLLM FP8-block paths that may need real checkpoint layouts/scales (`collector/vllm/collect_dsv4_attn.py`), TRT-LLM builds that allocate BF16 dummy weights while attaching FP8 scales (`collector/trtllm/collect_mla_module.py`), and routed top-k timing differences with dummy data. Please require verifying parameter dtype/shape/layout/scales, post-load transforms, and actual framework dispatch against the real checkpoint before claiming parity, and require a real-weight validation whenever any of those differ.

jasonqinzhou (changes requested):

Five confirmed P1 correctness issues require changes: disagg drops configured/versioned memory fractions, SGLang omits large configuration-dependent reserves, the vLLM collector times a different dispatch path, partial attention exclusions are collapsed to BF16, and DeepSeek V3.2 attention weights are absent from memory accounting. Three additional P2 issues affect offline loading, detail reporting, and the new validation runbook.

tianhaox:
Thanks for the thorough review — all five P1s were real. Addressed in a02dabf5..0a8f5782:

- **Per-projection exclusions (helpers)**: `attention_projection_exclusions()` parses q/kv/o/indexer groups; V3.1/V3.2-NVFP4 now keep o_proj at NVFP4 (your 13.34-vs-3.75 GiB example is unit-tested); mixed checkpoints key module perf rows on o_proj (largest projection). (a02dabf5)
- **DSA weights**: `dsa_block_weights_bytes()` from DSA_MODEL_DIMS with per-group dtypes, wired through all DSV32 variants. V3.2-NVFP4 tp1 = 13.15 GiB/rank (unit test); whole-model dummy-load anchor on 4xB300: AIC 99.53 vs loaded 93.51 GiB/GPU, residual is the known MoE dense-layer item. (a02dabf5)
- **Fraction threading**: Task.free_gpu_memory_fraction flows through sweep_disagg -> predict -> run_static; explicit value wins; vLLM default version-keyed (0.90 @0.14/0.19, 0.92 @0.22+, verified per tag). (8d560bac)
- **SGLang reserves**: derivation now includes tp*pp slack, decode-graph max_bs*2, and DP-attention max_bs*dp*3 (+x1.5 >300) per server_args.py@0.5.16 — DP=8/180GiB resolves 0.758, matching your ~0.76. Worker parallel config reaches the check. DeepEP reserves remain out (moe-backend-specific, documented). (8d560bac)
- **Collector dispatch**: provider assertion added, and we root-caused why plain torch.compile is not an alternative. Outside a serving worker, every norm path (incl. `forward_native`) routes through the IR torch custom op — `_ENABLE_TORCH_WRAP=True` module default (vllm/ir/op.py:50 @0.22.0) — which is opaque to inductor: compiling through it fuses nothing (B200 glue microbench with kernel-level verification: native-eager 74.1us, compile-through-wrap 64.6us with the aten chain intact, vllm_c-eager 21.0us). The engine escapes via worker-init `ir_enable_torch_wrap` handling (worker_base.py:94) plus custom inductor lowerings that fuse norms into neighboring GEMMs (serving trace: `triton_red_fused_..._fused_qkv_a_proj_rms_norm`). Replicating the engine trick (`set_default_torch_wrap(False)` + compile) fuses fully but overshoots at module level: engine 107us/layer vs vllm_c-eager 95-97 (-10%) vs wrap-off-compile 87.7 (-18%) — the engine's fused glue isn't free (norm+quant fusion, paged rope). So eager+vllm_c is the closest surrogate and avoids per-shape compile cost during sweeps; full account in the `enable_engine_fused_ops` docstring. (339b1b2b + docstring update)
- **P2s**: DefaultHFModels registration + offline regression; fraction_of_free persisted and the detail report now uses kv_cache_budget_bytes; skill parity claim made loader-conditional. (a9d554c5)

Also picked up coderabbit's round-2 findings (0a8f5782): `run_single_disagg` now threads the fraction and rejects `check_kv_cache_oom()` like the sweep path (single-point evaluation could previously accept a KV-infeasible config — the exact #1396 failure mode), and `predict_disagg_worker` omits the fraction kwarg when unset so injected Predictor implementations keep working.

Not addressed (deliberate): coderabbit's fixture-call-count test — the provider assertion hard-fails at runtime if the fused ops aren't applied, which covers the regression more directly than a stub-count unit test; and hoisting the IR-op imports to module scope — they are function-scoped by design so kernel-level collectors keep importing `utils` on vLLM builds without the IR registry while module-level benchmarks fail loudly at `_init_cuda()` (documented in-code). Happy to revisit either if you disagree.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
tianhaox: Fixed in 8d560bac: Task.free_gpu_memory_fraction now flows through sweep_disagg -> predict -> run_static, an explicit value wins, and the vLLM default is version-keyed (0.90 @0.14/0.19, 0.92 @0.22+, verified per tag). Also threaded through run_single_disagg with check_kv_cache_oom rejection in 3041c598.

aic-core/src/aiconfigurator_core/sdk/backends/sglang_backend.py
tianhaox: Fixed in 8d560bac: the derivation now includes tp*pp slack, decode-graph max_bs*2, and DP-attention max_bs*dp*3 (+x1.5 above 300) per server_args.py@0.5.16 — DP=8/180GiB resolves 0.758, matching your ~0.76. Worker parallel config reaches the check. DeepEP reserves stay out (moe-backend-specific, documented).

collector/vllm/utils.py
tianhaox: Addressed in 339b1b2b + 12168f5f: provider assertion added, and we root-caused why plain torch.compile cannot reproduce serving here — outside a worker every norm path goes through the IR torch custom op (_ENABLE_TORCH_WRAP=True, vllm/ir/op.py:50), opaque to inductor; the engine escapes via worker-init wrap handling plus custom inductor lowerings. Module-level anchors: vllm_c-eager 95-97us vs in-engine 107us (-10%) vs wrap-off+compile 87.7us (-18%), so eager+vllm_c stays. Both b200 MLA tables were re-collected with the fixed collector (5ce28de2, 01f5b807); full account in the enable_engine_fused_ops docstring.

aic-core/src/aiconfigurator_core/sdk/models/helpers.py
tianhaox: Fixed in a02dabf5: attention_projection_exclusions() parses q/kv/o/indexer groups; V3.1/V3.2-NVFP4 keep o_proj at NVFP4 (your 13.34-vs-3.75 GiB example is unit-tested) and mixed checkpoints key module perf rows on o_proj.

aic-core/src/aiconfigurator_core/sdk/models/deepseek_v32.py
tianhaox: Fixed in a02dabf5: dsa_block_weights_bytes() computes per-group projection weights from DSA_MODEL_DIMS and is wired through all DSV32 variants; V3.2-NVFP4 tp1 = 13.15 GiB/rank is unit-tested, and a 4xB300 dummy-load anchor gives AIC 99.53 vs 93.51 GiB/GPU loaded (residual is the known MoE dense-layer item).

aic-core/src/aiconfigurator_core/model_configs/nvidia--DeepSeek-V3.2-NVFP4_config.json
tianhaox: Fixed in a9d554c5: registered in DefaultHFModels with bundled configs and an offline regression test.

aic-core/src/aiconfigurator_core/sdk/inference_summary.py
tianhaox: Fixed in a9d554c5: fraction_of_free is persisted on the summary and the detail report computes the bound via kv_cache_budget_bytes with the same semantics.

.claude/skills/aic-silicon-align/SKILL.md
tianhaox: Fixed in a9d554c5: the parity claim is now loader-conditional with explicit verification steps (dtypes/shapes/scales after load, post-load transforms, kernel dispatch).

collector/vllm/utils.py
```diff
@@ -491,6 +498,41 @@ def create_kv_cache_and_block_mappings(
...
+
+    Raises on vLLM builds without the IR-op registry or the ``vllm_c``
+    provider — a silent fallback to the native chain would benchmark a kernel
+    path serving never runs and poison the database.
+    """
+    import vllm.ir.ops as ir_ops
+    import vllm.kernels.vllm_c  # noqa: F401 — import registers the 'vllm_c' impls
+
+    ir_ops.rms_norm.set_default(["vllm_c"])
+    ir_ops.fused_add_rms_norm.set_default(["vllm_c"])
```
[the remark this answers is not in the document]
tianhaox: Keeping the imports function-scoped by design: kernel-level collectors must keep importing utils on vLLM builds without the IR registry, while module-level runs fail loudly at _init_cuda() — documented in-code next to the imports. The set_default citation (vllm/ir/op.py @0.24.0) is in the docstring.

tests/unit/collector/test_vllm_dsv4_attn.py
```diff
@@ -82,6 +82,7 @@ def context(*_args, **_kwargs):
         BatchSpec=SimpleNamespace,
         create_common_attn_metadata=lambda *_args, **_kwargs: None,
         create_vllm_config=lambda *_args, **_kwargs: None,
+        enable_engine_fused_ops=lambda *_args, **_kwargs: None,
```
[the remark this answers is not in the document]
tianhaox: Covered by the runtime provider assertion inside enable_engine_fused_ops (hard-fails when dispatch does not resolve to vllm_c), which guards this regression more directly than a stub-count unit test.

jasonqinzhou:
[P1] Preserve the explicit KV fraction in `cli_estimate` disagg mode

Rechecked at `d21367d6434aebaab7766ecafb623adc815cadd1`.

The Task/sweep and `run_single_disagg` paths now thread and check the KV budget, but the public single-point path still discards it: the disagg branch calls `_run_disagg_estimate` without `free_gpu_memory_fraction`, and that helper has no such parameter. Consequently, `cli_estimate(..., mode="disagg", free_gpu_memory_fraction=0.8)` evaluates both roles with backend defaults and can return a configuration the caller's requested budget cannot admit.

Please thread the explicit value through both worker evaluations and add a non-default regression test; use the version-derived default only when the argument is omitted.

collector/vllm/utils.py
```diff
@@ -491,6 +498,70 @@ def create_kv_cache_and_block_mappings(
...
+    provider — a silent fallback to the native chain would benchmark a kernel
+    path serving never runs and poison the database.
+    """
+    # Imports stay function-scoped on purpose: kernel-level collectors that
+    # never touch this helper must keep importing utils on vLLM builds
+    # without the IR-op registry; module-level benchmarks fail HERE, loudly.
+    import vllm.ir.ops as ir_ops
+    import vllm.kernels.vllm_c  # noqa: F401 — import registers the 'vllm_c' impls
+
+    ir_ops.rms_norm.set_default(["vllm_c"])
```
jasonqinzhou, line 553: [P1] Do not publish surrogate dispatch as canonical serving data

The provider assertion makes this pin fail closed, but it only proves that `vllm_c` ran. The comment now confirms that the serving path is compiled/fused and that eager `vllm_c` is a closer surrogate, not the framework-selected serving dispatch. That still conflicts with the collector rule requiring the actual framework-selected path and forbidding a substituted backend based on latency proximity.

Because this head also ships recollected B200 vLLM 0.24 MLA parquets from this path, please collect through the actual worker/compile dispatch and record/assert what executed, or obtain explicit approval for a separately identified surrogate data contract. These rows should not be published under the canonical serving identity as-is.

aic-core/src/aiconfigurator_core/sdk/models/deepseek.py
```diff
@@ -118,12 +143,45 @@ def __init__(self, topk: int, num_experts: int, moe_inter_size: int, *args, back
...
+        attn_kv_gemm_quant_mode = _attn_mode("kv")
+        attn_o_gemm_quant_mode = _attn_mode("o")
+        # downscale GEMM fuses q_a + kv_a: BF16 only when both groups are excluded.
+        attn_downscale_gemm_quant_mode = common.GEMMQuantMode.bfloat16 if {"q", "kv"} <= excl else gemm_quant_mode
+        # Module perf rows are keyed by ONE gemm_type; when the checkpoint
+        # mixes dtypes across projections no row matches exactly, so key on
+        # o_proj's dtype — the largest projection by bytes and FLOPs (heads x
+        # v_head_dim x hidden ~ 58% of block weights at 64 heads).
+        attn_modes = {attn_q_gemm_quant_mode, attn_kv_gemm_quant_mode, attn_o_gemm_quant_mode}
+        attn_gemm_quant_mode = attn_modes.pop() if len(attn_modes) == 1 else attn_o_gemm_quant_mode
```
jasonqinzhou: [P1] Keep the mixed projection identity through module and WideEP paths

The granular fallback now correctly models q/kv BF16 plus o-proj NVFP4, but the primary MLA module query collapses that checkpoint to an all-NVFP4 row by selecting `attn_o_gemm_quant_mode`. When that row exists, `FallbackOp` bypasses the precise per-projection path and uses a profile for kernels the checkpoint does not run.

The same exclusions are not passed into either WideEP constructor, so their projection operations and weight accounting still use global NVFP4. In the probed TRT-LLM WideEP configuration, that undercounted q/kv weights by roughly 5.7 GiB per rank.

Please use the granular path unless an exact mixed module identity exists, and thread the per-projection modes into both WideEP implementations.

tianhaox: Fixed in e17f5128, both halves: mixed checkpoints now bypass the module row entirely and emit the granular per-projection ops directly (a single-gemm_type table can never carry the mixed identity; uniform checkpoints keep the FallbackOp(module, granular) composition), and both WideEP constructors now receive the exclusions - TRT-LLM WideEP's downscale/q_b/kv_b/o GEMMs follow per-projection dtypes for rows and weight byte-widths, with the absorbed-BMM dtype following kv; SGLang WideEP switches its granular qkv_a/downscale GEMMs (its q_b/kv_b/o live inside the WideEP MLA module rows, whose query carries no gemm axis). The old mixed-key pin now asserts the granular bypass, plus new WideEP per-projection dtype assertions.

collector/vllm/utils.py
tianhaox: You're right that eager+vllm_c is a validated surrogate, not the framework-selected serving dispatch - the docstring says as much, and the strict reading of the dispatch rule holds. Two mitigating facts for deciding what to do with the rows: (1) the rows they replace were collected with NO IR-op priority installed at all (the native mean/pow/rsqrt chain - a path serving never runs anywhere), so reverting to the previous parquets restores a worse violation of the same rule, with +34% error against the in-engine anchor vs -9% for these; (2) the deviation is asserted (provider check) and documented, and the exact-parity path - collecting through the worker's VLLM_COMPILE dispatch - is real collector engineering we agreed to track as follow-up.

Your call on the contract: (a) keep these rows with the surrogate deviation recorded explicitly in the data dir's collection_meta.yaml (norm_dispatch: vllm_c_eager surrogate, compile-dispatch harness tracked) until the harness lands and the tables are re-collected through it; or (b) drop the two parquets from this PR and land them together with the harness. I'd prefer (a) since the shipped alternative is strictly less faithful, but I'll do (b) if you want the canonical identity kept clean.

tianhaox: Decision (as the data owner): keeping the rows, with the deviation now recorded explicitly - option (a). e31879d1 annotates both tables' collection_meta.yaml with `norm_dispatch: vllm_c_eager_surrogate` plus the rationale and the re-collect-through-VLLM_COMPILE follow-up. The reasoning: the compiled serving dispatch is not reproducible in a standalone collector (IR torch-op wrapping is opaque to inductor outside a worker - verified kernel-by-kernel), the surrogate is provider-asserted and anchored at -9% vs an in-engine trace, and the rows it replaces ran the never-in-serving native chain at +34%. Resolving on that basis; happy to reopen if you want the stricter option (b) instead.

Arsene12358 (approved):

tianhaox:
Good catch - fixed in cb57bf26. The explicit fraction now flows cli_estimate -> _run_disagg_estimate -> DisaggInferenceSession.run_disagg -> both per-role run_static evaluations, with the version-derived default applying only when the argument is omitted. Added the non-default regression test you asked for: a deliberately tiny fraction flips a feasible h200/trtllm disagg point to the KV-budget OOM while the default path keeps passing.

jasonqinzhou (approved):
Re-reviewed the exact current head `71837996130788e7a96546a5e7e28da941d013be`. All previously identified P0/P1/P2 findings are addressed. The explicit disaggregated KV fraction now reaches both role-specific evaluations with a non-default regression test; the mixed projection/WideEP paths are corrected; the documented collector surrogate is an accepted, explicitly labeled residual risk. CI and focused validation are green. Approving.

PR #1430, opened 2026-07-28
feat(fpm): add Grove orchestration and preserve ComputeDomain for multinode collection

Ethan-ES:
> [!IMPORTANT]
> This PR depends on #1365 and is based on its current head (`0b90d2aac`).
> It should not be merged before #1365. Once #1365 is merged, this branch will
> be rebased onto the updated `main`.

## Overview

This PR closes the multinode ComputeDomain gap identified in #1365 and adds
optional Grove orchestration for FPM collection on GB200/GB300-class systems.

Existing behavior remains unchanged by default:

- Single-node FPM deployments still generate a `Pod`.
- Multinode deployments still generate a `LeaderWorkerSet` by default.
- Grove is enabled explicitly through `K8sConfig.fpm_orchestrator: grove` or
  `--fpm-orchestrator grove`.
- Other Generator deployment targets are unaffected.

## Details

- Generate a Grove `PodCliqueSet` for explicitly selected multinode FPM
  deployments.
- Preserve generated `ComputeDomain` and IMEX resource claims instead of
  stripping them during FPM lowering.
- Emit `ComputeDomain + PodCliqueSet` as a multi-document
  `k8s_deploy.yaml`.
- Support LWS, Grove, and explicitly supplied rank/master discovery in
  generated `run.sh`.
- Extend the FPM Collector to:
  - parse and validate multi-document manifests;
  - recognize `PodCliqueSet` workloads;
  - discover all Grove workload Pods;
  - apply and verify every generated resource;
  - clean up and verify deletion of every owned resource;
  - fail closed on unsupported or ambiguous workload shapes.

## Validation

- Generator unit tests: `233 passed, 4 skipped`
- Collector unit tests: `625 passed, 5 skipped`
- Ruff, formatting, and `git diff --check`: passed
- B200 TP2 prefill smoke:
  - native schema v2 preflight passed;
  - 4/4 FPM points collected;
  - result marked complete, valid, and usable.
- GB200 TP1 decode smoke:
  - 4/4 FPM points collected;
  - result marked complete, valid, and usable.
- GB200 multinode Grove validation:
  - `ComputeDomain` and `PodCliqueSet` accepted by the live controllers;
  - two workload Pods created and discovered by the Collector;
  - cleanup completed with no leaked resources.
  - Full multinode data collection was blocked by the lack of two
    simultaneously available GB200 nodes, rather than a manifest or Collector
    failure.

## Summary by CodeRabbit

* **New Features**
  * Added “FPM forward” whole-model forward-pass campaign support (plan-only preview, multinode orchestration, and resolved execution).
  * Introduced model capability resolution and memory feasibility filtering to plan the runnable FPM collection set.
* **Documentation**
  * Updated FPM V1 deployment and generator guidance with precise artifact emission and clarified LeaderWorkerSet vs Grove PodCliqueSet orchestration behavior.
  * Expanded the CLI user guide with stricter runtime/collector validation criteria.
* **Bug Fixes**
  * Tightened benchmark result schema expectations for newer native collection flows.
* **Tests**
  * Added extensive unit coverage for planning, orchestration, native artifact validation, database writing, and CLI isolation.


collector/fpm_forward/runtime/preflight.py
```diff
@@ -0,0 +1,58 @@
...
+    audit = {
+        "schema_version": 1,
+        "runtime_contract": "dynamo_pr11509_native_schema_v2",
+        "benchmark_point_fields": sorted(fields),
+        "missing_fields": missing_fields,
+        "missing_methods": missing_methods,
+        "status": "passed" if not missing_fields and not missing_methods else "failed",
+    }
+    Path("/results/runtime-preflight.json").write_text(json.dumps(audit, indent=2, sort_keys=True) + "\n")
+    if missing_fields or missing_methods:
```
jasonqinzhou, line 49: **[P1] Make the default runtime satisfy this mandatory contract.** With no image/version override, the Generator currently emits `nvcr.io/nvidia/ai-dynamo/vllm-runtime:1.2.0`. I checked the upstream `v1.2.0` tagged scheduler: it lacks three required `BenchmarkPoint` fields and ten of the twelve methods checked here because PR 11509 merged after that release. The wrapper therefore allocates a Pod and then deterministically fails this preflight on the normal Collector path; the current Dynamo version matrix does not offer a compatible release. Please pin a known-compatible image/digest or require one explicitly, reject incompatible `--dynamo-version` choices before cluster allocation, and add a regression test for the default execution target.

collector/fpm_forward/database.py
```diff
@@ -0,0 +1,319 @@
...
+    for part in path.split("."):
+        if not isinstance(value, dict) or part not in value:
+            raise KeyError(path)
+        value = value[part]
+    return value
+
+
+def _validate_backend_markers(cell: FPMCell, cell_dir: Path) -> None:
+    expected = cell.backend_policy.expected_markers
+    if not expected:
```
jasonqinzhou, line 65: **[P1] Require runtime-resolved evidence for every formal cell.** The only production `BackendPolicy` currently sets `expected_markers={}`, so this returns without reading the emitted `resolved-config*.json`. Native artifact validation proves provenance, ranks, grid identity, and timing, but the formal row below copies the planned TP/PP/DP, MoE topology, KV dtype, quantization, and backend policy without proving that the engine actually resolved those values. A normalized, ignored, or fallback runtime setting can therefore be published under the wrong physical key. Please validate the full resolved execution contract unconditionally and add a test where one runtime-resolved value differs from the plan and aggregation fails.

collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1421 @@
...
+        resource.collect(pods, require_benchmark=False)
+    except Exception as error:
+        logger.warning("FPM salvage for %s did not capture a complete artifact set: %s", cell_id, error)
+
+
+# Statuses whose salvaged artifacts may be acknowledged without a rerun. A
+# cell that failed, was interrupted, or lost its host process can leave a
+# complete, attempt-matched artifact set behind; "cleanup_failed" is excluded
+# because only a rerun re-drives the verified deletion of leaked resources.
+FPM_RECOVERABLE_STATUSES = frozenset({"failed", "interrupted", "running"})
```
jasonqinzhou, line 1099: **[P1] Do not promote a persisted `running` attempt from artifacts alone.** `running` is durably written before cluster work, while the in-memory `passed` result is not checkpointed until the `finally` block after cleanup. A host loss after `collect()` can therefore leave complete local artifacts and a live GPU workload. Because this set includes `running`, resume validates only the artifacts, marks the cell passed, and skips it without re-driving deletion. Please reconstruct the saved manifest and perform verified cleanup before recovering `running` (or keep it quarantined/retryable), and add a regression test for this kill window.

collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1421 @@
...
+        {"name": "DYN_FPM_BENCHMARK_OUTPUT_PATH", "value": "/results/benchmark.json"},
+        {"name": "FPM_RUN_ID", "value": cell.cell_id},
+    ]
+    total_gpus = cell.topology.total_gpus
+    generated = {
+        "ServiceConfig": service,
+        "DynConfig": {"mode": "agg"},
+        "WorkerConfig": {"agg_workers": 1, "agg_gpus_per_worker": total_gpus},
+        "K8sConfig": {
+            "name_prefix": cell.cell_id,
```
jasonqinzhou, line 806: **[P1] Make the Kubernetes workload identity unique to the plan/attempt, or acquire and verify exclusive ownership before apply and delete.** `cell_id` omits the plan, Generator config, and attempt UUID, yet it is used directly as `name_prefix`. A fresh non-resume run with no checkpoint record can adopt a same-named resource left by a killed collector, and concurrent campaigns for the same modeled cell can successfully apply, select, clear, collect from, and delete each other's workload. The plan label is mutable under `kubectl apply` and cleanup has no ownership precondition, so it does not prevent the race. Please use attempt-scoped names/labels or a verified lease, require that identity before collection/deletion, and cover both stale-resource and concurrent-run collisions.

src/aiconfigurator/generator/builders/fpm_builder.py
```diff
@@ -662,8 +815,36 @@ def _lower_worker_to_resource(
         }
 
     if len(metadata["name"]) > 50:
-        raise ValueError("FPM LeaderWorkerSet names must be at most 50 characters")
+        raise ValueError("FPM multi-node resource names must be at most 50 characters")
     pod_labels = copy.deepcopy(metadata["labels"])
+    if orchestrator == "grove":
+        pod_labels.setdefault("kai.scheduler/queue", "dynamo")
```
jasonqinzhou: **[P1] Do not make `grove` silently select a site-specific scheduler and queue.** The docs promise Grove with its configured scheduler, but every generated clique is assigned to `kai-scheduler` and queue `dynamo`; clusters using another Grove scheduler integration, or KAI without that queue, leave these Pods Pending. A direct Generator caller can override `schedulerName`, but the Collector's normal `--fpm-orchestrator grove` path exposes no scheduler/queue controls and rejects user `fpm_resource_labels`, so it cannot override the queue label. Please leave these fields unset unless explicitly requested, or add first-class scheduler/queue configuration with validation and tests for default, KAI, and non-KAI Grove clusters.

jasonqinzhou (changes requested):

## What this PR does

This PR extends the stacked FPM collection work with optional Grove PodCliqueSet orchestration, preserves multinode ComputeDomain and IMEX resources, and teaches the Collector to apply, discover, collect from, and clean up multi-document LWS or Grove workloads.

The change shows substantial care in its fail-closed manifest validation, per-resource apply and deletion checks, controller-aware generated run script, result integrity checks, documentation, and broad unit coverage. Preserving the existing single-node and LWS behavior while adding Grove is useful, thoughtful work.

## Why this score

Score 4.0/10, REQUEST_CHANGES: the default generated runtime cannot pass the new mandatory FPM preflight, formal database rows are not bound to the runtime-resolved execution settings, and the checkpoint/resource lifecycle can recover leaked work or let stale and concurrent campaigns share one Kubernetes identity. Grove selection also silently hardcodes a site-specific KAI scheduler and queue. These issues can prevent the feature from running or can publish mislabeled measurements; current-head CodeRabbit feedback separately identifies a verified Grove rendezvous defect that is intentionally not duplicated here.

PR #1435, opened 2026-07-28
feat: Kimi-K3 support — hybrid KDA+MLA model, DSPARK, multi-arch sglang + vllm perf data (8 systems)

tianhaox:
Adds **Kimi-K3** (`KimiK3ForConditionalGeneration`) support: modeling, collectors, and perf data for sglang + vllm across eight systems. Campaign record: `docs/perf_database/kimi-k3-blackwell-bringup-ledger.md`.

### Method

- **Model**: 93-layer hybrid — 69 KDA linear-attention + 24 MLA layers, LatentMoE (896 experts, top-16, SiTU, MXFP4). Dual-pool memory: paged MLA latent KV + constant per-request KDA state.
- **Collection**: kernel-level, mirroring serving dispatch — the collector offers each case to the framework's own kernel-selection predicates and records whatever lane serving would run (`kernel_source` per row). Three KDA phases: context / generation / target-verify.
- **DSPARK speculative decoding**: `nextn` = block size; generation ops priced at verify width `nextn+1`, KDA kernels switch to measured verify tables, plus an explicit 5-layer draft model (block-parallel; per-backend geometry — RadixArk for sglang, Inferact MLA-style for vllm). Acceptance (`nextn_accepted`) is user input, never assumed.
- **Precision routing**: sglang+Blackwell → W4A8 (mxfp8 activations), Hopper/SM120 → W4A16 marlin, SM89 → triton; vllm → W4A16 marlin on every SM (CUTLASS whitelist rejects SiTU) — all mirroring serving.
- **Parity**: every op has a Rust twin; engine-step comparison green on all K3 silicon cells.

### Alignment

| Check | Result |
|---|---|
| bs1 DSPARK vs sglang Day-0 blog (GB300 TP8, accepted≈7) | 404 vs ~423 tok/s/user (**-4.5%**) |
| Per-op vs 8×B300 serving (nsys, CUDA-graph trace) | MoE within **3%** (uniform-routing bound); KDA op-level median \|err\| **1.6%** |
| Support matrix (b300/gb300, sglang + vllm, agg+disagg) | all silicon-**PASS** |
| Python vs Rust engine (10 silicon cells, rtol 0.05) | **identical** |

Reference (b300 TP8, isl/osl 1024, bs1): no-spec 13.94 ms TPOT; DSPARK n7 accepted 5/6/7 → 303 / 354 / 404 tok/s/user.

### Others
- **Shared-mechanism change** (collector meta-rule 2): `collector/helper.py log_perf` stale-lock breaking is now rename-based (single-winner) with a 30s wait window — declared here and in the bring-up ledger.

- **Data**: kda on all 8 systems (both backends); K3 MoE on 8 (sglang) / 6 (vllm, full EP); mla_bmm tables (new) on 6; MegaMoE module on b200. All quality-gated by content (row counts, classified failures, dup-key scans, digest-pinned metas).
- **Known gaps** (owner-decided, documented): sglang 0.5.14 marlin EP>1 IMA crash on 4 system families (upstream issue pending; surviving moe_tp slices ship); vllm MegaMoE W4A8 lane → #1462 (unblocked on vllm nightlies); kda_onorm ~1% double-count on the fused shard → #1463; vllm DSPARK draft is 5 layers vs sglang's — modeled per backend.
- Single-node capacity: K3 fits 8-GPU workers only on 288 GB parts (b300/gb300); PP/scale-out search-space enablement is an SDK follow-up.


aic-core/src/aiconfigurator_core/sdk/models/kimi_k3.py
```diff
@@ -0,0 +1,661 @@
...
+        gemm_q = self.config.gemm_quant_mode
+        kvcache_q = self.config.kvcache_quant_mode
+        fmha_q = self.config.fmha_quant_mode
+        counts = self._count_layer_types()
+        mla = self._mla_dims(cfg)
+
+        # KDA per-shard dims for kernel lookup (collector rows are per attention-TP shard)
+        nk_local = cfg.kda_num_heads // tp
+        hk = cfg.kda_head_dim
+        p_local = nk_local * hk  # per-rank projection width
```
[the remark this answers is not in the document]
tianhaox: Fixed in 02beeaab5: constructor now raises when kda_num_heads % tp_size != 0 (fail-loud instead of silently truncating to a shard with no collector rows).

aic-core/src/aiconfigurator_core/sdk/models/kimi_k3.py
```diff
@@ -0,0 +1,661 @@
...
+        workload_dist = (
+            self.config.workload_distribution + "_1.01"
+            if self.config.workload_distribution == "power_law"
+            else self.config.workload_distribution
+        )
+
+        latent = cfg.routed_expert_hidden_size or h
+        num_dense = cfg.first_k_dense_replace
+        num_moe = self._num_layers - num_dense
+        shared_inter = cfg.num_shared_experts * cfg.moe_inter_size * 2  # 2*3072=6144 total width
```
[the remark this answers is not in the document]
tianhaox: Confirmed against the checkpoint (num_shared_experts=2 x moe_intermediate_size=3072 = 6144 total) and fixed in 02beeaab5 — the stray *2 is removed; the gate/up factor is applied only at the GEMM sites. Shared-expert FFN cost was indeed ~2x inflated (~2-3% end-to-end pessimism).

aic-core/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -663,6 +664,39 @@ def _parse_hf_config_json(config: dict) -> dict:
...
             "qk_rope_head_dim": config.get("qk_rope_head_dim", 0),
         }
+    elif architecture == "KimiK3ForConditionalGeneration":
+        # Kimi-K3: hybrid KDA linear attention + MLA full attention with LatentMoE.
+        # linear_attn_config.kda_layers / full_attn_layers are 1-based layer ids.
+        linear_attn_cfg = config.get("linear_attn_config") or {}
+        kda_layer_ids = set(linear_attn_cfg.get("kda_layers") or [])
+        if not kda_layer_ids:
+            raise ValueError("Kimi-K3 config must define linear_attn_config.kda_layers")
+        layer_types = tuple("linear_attention" if (i + 1) in kda_layer_ids else "full_attention" for i in range(layers))
```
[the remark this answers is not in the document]
tianhaox: Fixed in 02beeaab5: out-of-range 1-based ids in linear_attn_config.kda_layers now raise with the offending ids listed.

aic-core/src/aiconfigurator_core/systems/data/h20_3e/moe/sglang/0.5.14/collection_meta.yaml
```diff
@@ -0,0 +1,12 @@
...
+runtime:
+  framework: sglang
+  version: 0.5.14
+tables:
+  moe_perf:
+    # Kimi-K3 w4a16_mxfp4 shapes, TP 1-32 x 3 distributions. EP>1 slices are
+    # absent (0.5.14 marlin moe kernel crashes with an illegal memory access
+    # there — recorded in the collection failure log); EP queries fall back
+    # to HYBRID empirical.
+    status: complete
```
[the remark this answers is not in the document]
tianhaox: Stale: the h20_3e data directories were removed from this PR (Blackwell-first scope; Hopper-class H20 needs recollection with the fixed int32 conv guard and will come in a follow-up).

collector/vllm/collect_kda.py
```diff
@@ -0,0 +1,694 @@
...
+                        results["power_stats"],
+                    )
+                ok += 1
+            except Exception as e:
+                err += 1
+                print(f"  Error at batch_size={batch_size}, seq_len={seq_len}: {e}")
+                continue
+            finally:
+                tensors.clear()
+                _cleanup("context")
```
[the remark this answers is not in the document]
tianhaox: Fixed in 02beeaab5: all three strict-completeness raises (context/generation/verify) now carry the first 8 failing cells with shape + exception type/message, so the classified failure record is traceable without worker stdout.

tests/unit/collector/sglang/test_collect_kda_contract.py
```diff
@@ -0,0 +1,42 @@
...
+    assert "int32 token-offset overflow" in source
+    assert "causal_conv1d_triton.py:373-379" in source
+
+
+def test_kda_case_phases_cover_context_generation_verify():
+    # The registry getter must emit all three phases for every declared shape;
+    # verify rows carry the draft-token width in the seq_len slot.
+    source = SOURCE_PATH.read_text(encoding="utf-8")
+    for phase in ("context", "generation", "verify"):
+        assert f'"{phase}"' in source
```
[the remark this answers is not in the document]
tianhaox: Fixed in 02beeaab5: both contract tests now assert phase coverage against collector.case_generator.get_common_kda_test_cases() (the shared spec generator both backend getters adapt, importable without torch) instead of grepping module source.

docs/perf_database/kimi-k3-blackwell-bringup-ledger.md
```diff
@@ -0,0 +1,186 @@
...
+- **verify (DSPARK)**: `fused_kda_decode_mtp_dspark`, the CuTeDSL fused
+  conv+chain-verify kernel (`_can_run_dspark_cutedsl_mtp`,
+  kda_backend.py:858-938: capability==10, cutlass importable, draft width
+  2..8, 128-dim symmetric bf16 heads, fp32 contiguous ssm state). Under spec
+  the kimi_k3 hook pins `--linear-attn-verify-backend nv_cutedsl`. Serving
+  folds gated RMSNorm into the kernel only on the TP8 12-head shard
+  (`_prepare_fused_decode` is compiled for seg = 12*128,
+  models/kimi_k3.py:1522-1560). → collector grew a dispatch branch mirroring
+  exactly this probe, one row per point, `kernel_source
+  "fused_kda_decode_mtp_dspark"`, per-shard onorm.
```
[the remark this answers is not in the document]
tianhaox: Fixed in 02beeaab5: onorm is now spelled out as the output gated RMSNorm (the kernel's argument naming) at the dispatch record.

collector/sglang/collect_dsv4_megamoe.py
```diff
@@ -69,6 +69,21 @@
...
+    # activation_clamp=_K3_MEGA_SITU_SENTINEL_CLAMP) @ kimi-k3 branch).
+    "kimi_k3": {
+        "model": "moonshotai/Kimi-K3",
+        "hidden_size": 3584,
+        "inter_size": 3072,
+        "routed_num_experts": 896,
+        "routed_topk": 16,
+        "routed_scaling_factor": 1.0,
+        "norm_topk_prob": True,
+    },
```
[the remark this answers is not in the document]
tianhaox: Fixed in 02beeaab5: the kimi_k3 model config now declares activation_clamp=0.03125; --activation-clamp defaults to None and resolves from the model config, and an explicit conflicting value raises instead of silently collecting the wrong kernel path.

collector/vllm/collect_kda.py
```diff
@@ -0,0 +1,694 @@
...
+
+    for batch_size in batch_size_list:
+        for seq_len in seq_len_list:
+            nt = batch_size * seq_len
+            tensors = {}
+            try:
+                if nt * proj >= 2**31:
+                    raise ValueError(
+                        f"causal_conv1d int32 token-offset overflow guard: total_tokens={nt} * proj={proj} >= 2**31"
+                    )
```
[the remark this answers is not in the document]
tianhaox: Partially adopted in 02beeaab5, with one deliberate rebuttal:

**Adopted**: the strict-completeness raise now carries the first failing cells (shape + exception), so the classified record is traceable — that part of the finding was real.

**Not adopted — pre-execution filtering.** `.claude/rules/collector/failure_handling.md` (which governs reviews too, per the repo guide) is explicit: the doctrine is *observe, don't predict* — there is deliberately no expected-failure layer, and "never hide failures with broad skips ... reduced coverage". The int32 cells are not memory-feasibility drops (the one sanctioned generation-time filter): they are a deterministic *kernel limit*, and the guard exists precisely to convert what was an IMA/context-corruption into a fast, clean, classified failure at the invocation site. That is the decision tree's default DONE path, and it is how every prior campaign shipped kernel-limit families (e.g. the SM100/103 DSA local-head<8 cells). Empirically the claim "the 96-head row can never complete" is also not what happens: all passing cells log their rows before the aggregate raise; all eight systems shipped full kda tables this way, each with the identical 21-cell classified spectrum documented in the collection metas.

Also note the OOM arithmetic in the comment doesn't materialize: the guard raises before allocation, and the observed failure signature on every system is the int32 guard, not OOM.

collector/vllm/collect_kda.py
```diff
@@ -0,0 +1,722 @@
...
+    q_w, k_w, v_w = conv_weight.split([proj] * 3, dim=0)
+    ok = err = 0
+    failures: list[str] = []
+
+    for batch_size in batch_size_list:
+        for seq_len in seq_len_list:
+            nt = batch_size * seq_len
+            tensors = {}
+            try:
+                if nt * proj >= 2**31:
```
Arsene12358, line 211: **[BLOCKER] The agreed `FIXME(kernel-limit)` + framework citation from the open review thread is still missing here.**

Your rebuttal to the pre-filtering suggestion was the rule-correct call (failure_handling.md: observe-don't-predict), and CodeRabbit verified it — but the one requirement it left ("Add an unverified FIXME(kernel-limit) comment at the guard. Include the pinned vLLM source file and line range that establishes the int32 offset limit") never landed. layer_permissions.md is explicit that guards need the framework source line ("Do not implement guards from unverified claims"), and the sglang twin already shows the target state (`collector/sglang/collect_kda.py:199-214`: `causal_conv1d_triton.py:373-379` + silicon evidence).

Two adjacent gaps at this site while you're here: (a) this bound is `nt * proj` — 3× looser than sglang's silicon-corrected `total_tokens * conv_channels` for what looks like the identical strided-view construction (lines 208-209 below) — the campaign's identical 21-cell spectrum suggests vLLM's kernel genuinely offsets per-block, but that reasoning lives nowhere in the code; (b) the contract test pins only the substring `"2**31"`, and its comment claims "Same int32 token-offset guard as the sglang kda collector", which is false as written (different multiplier).

Proposed shape (fill in the real vLLM citation):

```python
# FIXME(kernel-limit): vLLM's causal_conv1d computes token offsets per
# Q/K/V block (<file:line @ 0.1.dev19262>), so the bound is nt * proj —
# unlike sglang's Triton kernel, whose per-block views stride the full
# 3-block buffer (bound nt * 3*proj, causal_conv1d_triton.py:373-379).
# Silicon: identical 20-context + 1-generation guard spectrum on all
# eight systems; in-band cells pass (ledger 2026-08-01).
if nt * proj >= 2**31:
```

…and pin the exact expression in `tests/unit/collector/vllm/test_collect_kda_contract.py` so a future bound edit is a conscious act.

aic-core/src/aiconfigurator_core/systems/data/b200_sxm/kda/sglang/0.5.16/collection_meta.yaml
```diff
@@ -0,0 +1,21 @@
...
+schema_version: 1
+provenance: local
+runtime:
+  framework: sglang
+  version: 0.5.16
+tables:
+  kda_perf:
+    # 2026-07-28 base collection predates the kda_fused_decode collector
+    # branch. 2026-08-02: the GENERATION phase was recollected in the
+    # current kimi-k3 image (sha256:6d9594a4) and REPLACED wholesale (77
```
Arsene12358: **[BLOCKER] Truncated digest — this lane (and the MegaMoE table) cannot be re-pulled or verified.**

`sha256:6d9594a4` is 8 hex chars; the full digest appears nowhere in the repo — `collector/framework_manifest.yaml` pins the *first* branch build (`4b8a7542…`), and no commit message or ledger line carries this second one. Same truncation in `b200_sxm/moe/sglang/0.5.16/collection_meta.yaml:14` (the 64-row K3 MegaMoE lane). For a custom ephemeral build of the kimi-k3 branch this is the only provenance there is, and these are the only two lanes in the PR that can't be re-collected or verified from their recorded pins.

Fix: record the full 64-hex digest in both metas. While at it: (a) upgrade the two bare metas (`b300_sxm/kda/sglang/0.5.16`, `b200_sxm/kda/vllm/0.1.dev19262` — currently `provenance: local` with no structure) to the structured format their later siblings use (image_digest / collector_ref / case_plan_hash / rows); (b) consider anchoring the "@ kimi-k3 branch" citations in the case YAMLs to a digest or build id — the campaign itself proved a branch name is not re-verifiable (two builds, both reporting 0.5.16).

collector/cases/models/KimiK3ForConditionalGeneration_cases.yaml
```diff
@@ -0,0 +1,157 @@
...
+      sglang_moe_routing_method_type: DeepSeekV3
+      sglang_moe_routed_scaling_factor: 1.0
+      sglang_moe_renormalize: true
+      sglang_moe_has_correction_bias: true
+      sglang_moe_num_expert_group: 1
+      sglang_moe_topk_group: 1
+      sglang_moe_backends:
+        # kimi-k3 branch dispatch: Blackwell -> flashinfer_mxfp4
+        # (trtllm-gen SiTU), SM90 -> marlin W4A16 (cookbook §2 MoE runner).
+        w4a16_mxfp4: {90: marlin, 100: flashinfer_mxfp4, 103: flashinfer_mxfp4}
```
Arsene12358: **[MAJOR] The sglang stock-image lanes this map routes to are the same situ→silu approximation the vllm lane marks — but unmarked.**

K3's checkpoint activation is SiTU; the sglang K3 moe rows were collected on stock 0.5.14 (no SiTU), so the marlin lanes (h100 541 / h200 488 / rtx 551 rows) and the SM89 triton lane (l40s 2,997 rows) benchmark the silu epilogue under the K3 `w4a16_mxfp4` label with no marking. This PR set its own standard on the vllm side — bb6eb4b2: "a silu-kernel row labelled situ would be silently wrong data" → `_situ_as_silu` kernel_source suffix + hard-check, with the epilogue delta measured up to ~5.5% at small token counts (8555ac6a). The Blackwell trtllm-gen lanes are unaffected (same kernel, epilogue constants). case_authoring.md: mislabeled rows "poison the database".

Fix: suffix the sglang stock-lane rows the same way, or record the dated owner decision accepting the unmarked silu epilogue (yaml comment + ledger), mirroring the vllm precedent.

Related citation gap: this map has no 89/120 keys, so K3 on SM89/SM120 silently rides the base map's GPT-OSS-era citations (`moe.yaml:106-115`) — the l40s meta's "SM89 dispatch, no marlin/flashinfer lane" assertion has no source. A one-line kimi-k3-branch citation for K3's checkpoint class (or explicit 89/120 keys here) would make those lanes serving-true on paper, not just plausibly.

collector/sglang/collect_kda.py
```diff
@@ -0,0 +1,1038 @@
...
+            # fallback). No shard shapes are replicated here — covered() is
+            # the framework's dispatch decision.
+            #
+            # SM90+ only: the JIT kernel emits mbarrier.try_wait.parity, which
+            # ptxas rejects below sm_90 ("requires .target sm_90 or higher",
+            # L40S/SM89 2026-08-01 — covered() accepts the shapes and the
+            # compile failure then kills every batch cell of the 12-head
+            # shard). Pre-Hopper decode is the Triton packed pair, same as
+            # serving after its own JIT failure.
+            fused = _fused_decode_module() if get_sm_version() >= 90 else None
```
Arsene12358: **[MAJOR] The load-bearing serving claim in this gate has no citation.**

"Pre-Hopper decode is the Triton packed pair, same as serving after its own JIT failure" is the one claim that makes this SM≥90 gate a legal fallback replication rather than a manual pin (layer_permissions.md: "Manual pinning requires source proof… Replicating a fallback chain is legal only when the framework itself performs it in serving"). Every other pin in this file carries file:line (`kda_backend.py:119-157`, `:858-938`, `models/kimi_k3.py:1522-1560`); this one cites silicon evidence for the *collector-side* ptxas failure only. If serving on SM89 does not actually catch the deferred JIT failure and reach Triton-pair steady state, the l40s 12-head decode rows are invented-fallback data.

Fix: add the kimi-k3-branch file:line for the catch-and-fallback path (then this gate is exactly the sanctioned fallback replication), or let SM<90 12-head cells raise classified like everything else.

collector/cases/models/KimiK3ForConditionalGeneration_cases.yaml
```diff
@@ -0,0 +1,157 @@
...
+        # bmm_MxE4m3_MxE2m1MxE4m3 kernels). The base moe.yaml quant-mode
+        # axis gates this label to SM in [100, 120). Hopper keeps the
+        # bf16-activation marlin W4A16 lane above.
+        w4a8_mxfp4_mxfp8: {100: flashinfer_mxfp4, 103: flashinfer_mxfp4}
+      wideep: true
+      framework_quantization:
+        sglang:
+          allowed_modes: [w4a16_mxfp4, w4a8_mxfp4_mxfp8]
+        trtllm:
+          allowed_modes: [w4a16_mxfp4]
```
Arsene12358: **[MAJOR] This block makes K3-on-trtllm plan moe and expand to zero cases with no logged drop.**

`all_frameworks_op_cases.moe: cases: all` activates moe for trtllm, and this `allowed_modes: [w4a16_mxfp4]` intersects with the trtllm base allowlist (`moe.yaml:195-199`) — which does not include `moonshotai/Kimi-K3` — so every mode is rejected and `collect.py --backend trtllm --model-path moonshotai/Kimi-K3` hits the bare `continue` at `trtllm/collect_moe.py:256`: zero cases, no counting, no log. case_authoring.md: "A planned op that expands to zero cases must be explainable from logged drops… Zero cases with no logged reason is a population bug to fix — not a clean completion" — and this PR's own vllm getter implements the mandated logging citing that exact rule (`vllm/collect_moe.py:296-301`). K3 is the first model whose trtllm policy zeroes out (K2.5's int4_wo is allowlisted).

Fix options: (a) scope the moe activation to `framework_specific_op_cases.{sglang,vllm}` and drop this trtllm block; (b) add vllm-style fully-dropped logging to the trtllm (and sglang) getters; (c) if trtllm K3 is real intent, allowlist it in `moe_trtllm` with a serving citation. Note: a `frameworks: [sglang, vllm]` key on the moe row alone will NOT fix it — the trtllm getter calls `get_common_moe_test_cases()` without `backend`, so the row-scoping filter at `case_generator.py:1475` never fires.

tests/unit/collector/test_sglang_attention_0514.py
```diff
@@ -173,8 +173,11 @@ def test_sglang_sm90_full_structural_population_is_stable(monkeypatch):
     monkeypatch.delenv("COLLECTOR_MODEL_PATH", raising=False)
 
     for phase, sweep_getter, expected in (
-        ("context", get_attention_context_shape_sweeps, 147),
-        ("generation", get_attention_generation_shape_sweeps, 132),
+        # 151/137 since the Kimi-K3 declarations added 8 SM90 head configs
+        # (DSPARK draft GQA shards 64/32/16/8 q-heads @hd64 + MLA 96-family
+        # 96/48/24/12 @hd128, all fa3-routed; net +4 context / +5 generation).
+        ("context", get_attention_context_shape_sweeps, 151),
+        ("generation", get_attention_generation_shape_sweeps, 137),
```
Arsene12358, line 182: **[MAJOR] These new pins are correct — but this test never runs in CI.**

Verified locally: this file passes 17/17 with no marker filter, but CI runs `pytest tests/unit -m "unit or build"` (`build-test.yml:141`) and only 4 tests in this module carry marks — this function isn't one of them. So the only regression pin on the K3 DSPARK + MLA-96 attention population (and this "+4 context / +5 generation net of dedup" claim) is invisible to CI and will only ever be validated by hand-runs. layer_permissions.md meta-rule 4: "A test that only works with `-m unit` unmarked is invisible to CI."

The invisibility pre-dates this PR, but the PR changes the pinned values inside it. Fix is one line — module-level `pytestmark = pytest.mark.unit` (every test in this module is a pure YAML/AST population test, no GPU).

collector/helper.py
```diff
@@ -688,15 +688,30 @@ def log_perf(
...
+    for _ in range(300):
         try:
             fd = os.open(lock_file, os.O_CREAT | os.O_EXCL | os.O_WRONLY)
             os.close(fd)
             got_lock = True
             break
         except OSError:
+            try:
+                if time.time() - os.path.getmtime(lock_file) > stale_lock_seconds:
+                    print(f"Breaking stale lock for {perf_filename}")
```
Arsene12358, line 717: **[MAJOR] Two-waiter unlink race; also an undeclared shared-mechanism change.**

The break is the right idea and the H200 evidence (47 lost rows) is solid. But as written: waiter A stats a stale lock → unlinks → recreates via `os.open(O_CREAT|O_EXCL)`; waiter B, still holding its stat of the *old* lock, unlinks A's *fresh* lock on its next loop iteration and acquires — two writers interleaving appends inside the critical section, in exactly the multi-waiter crash-storm pile-up this change targets. Standard fixes: re-verify lock identity between stat and unlink (open + `fstat`, compare inode against the stat you decided on), or break via `os.rename(lock_file, lock_file + f".breaking-{os.getpid()}")` so only one breaker can win.

Separately: `log_perf` is shared executor mechanism for every collector, and this change is folded into a mixed commit (39a950b2: "verify sglang kda on SM89/SM120; gate fused decode to SM90+; survive perf-log lock storms") and absent from the PR description — layer_permissions.md meta-rule 2 wants mechanism changes explicitly surfaced for approval. Declaring it (PR body + a line in the ledger) plus the race fix keeps it; nothing here argues for reverting.

tests/unit/collector/vllm/test_collect_kda_contract.py
```diff
@@ -0,0 +1,38 @@
...
+    assert "int32 token-offset overflow" in source
+
+
+def test_kda_dispatch_mirrors_serving():
+    # The collector must dispatch prefill like serving (FlashKDA when
+    # supported, Triton fallback) and probe the fused decode kernel via the
+    # same predicate serving uses — never pin a kernel unconditionally.
+    source = SOURCE_PATH.read_text(encoding="utf-8")
+    assert "is_flashkda_supported" in source
+    assert "is_fused_kda_decode_supported" in source
```
Arsene12358: **[MAJOR] This test is satisfiable by the module docstring — the probes could be deleted and it would still pass.**

`collector/vllm/collect_kda.py`'s module docstring says "probed via is_fused_kda_decode_supported exactly like serving", so `assert "is_fused_kda_decode_supported" in source` holds with all probe code gone. Your fix wave converted the sglang sibling to AST (`sglang/test_collect_kda_contract.py:13-21`) and the phase test to a runtime assertion, but this one kept the raw greps. Validated replacement below — ruff check, ruff format, and pytest all green against head (both probe names are real `ast.Name` references in the collector):

```suggestion
def test_kda_dispatch_mirrors_serving():
    # The collector must dispatch prefill like serving (FlashKDA when
    # supported, Triton fallback) and probe the fused decode kernel via the
    # same predicate serving uses — never pin a kernel unconditionally.
    # AST name references (not substring greps), so docstrings/comments
    # cannot satisfy the contract — mirrors the sglang twin test.
    import ast

    tree = ast.parse(SOURCE_PATH.read_text(encoding="utf-8"), filename=str(SOURCE_PATH))
    referenced = {node.id for node in ast.walk(tree) if isinstance(node, ast.Name)}
    assert "is_flashkda_supported" in referenced
    assert "is_fused_kda_decode_supported" in referenced
```

Arsene12358 (changes requested):
**Verdict: REQUEST_CHANGES — the substance is excellent; what blocks merge is three mechanical, rule-mandated items.**

I reviewed head 2ab60b0 against main (eed87c1) in five layers — SDK modeling math, Rust parity, collector code vs `.claude/rules/collector/*`, case YAML + tests, and a row-level data audit — plus local verification. First, what I verified *holds*, because it frames everything below as a soft block:

- **Goal fulfilled, replicated end-to-end from the shipped artifacts**: `aiconfigurator cli estimate` (SILICON, isl/osl 1024, bs1) reproduces the PR-body reference predictions — b300 sglang TP8 no-spec 14.08 ms TPOT vs the claimed 13.94 (−1.0%); DSPARK nextn=7/accepted=6 → 2.88 ms vs 2.83 (−1.9%); b200 vllm 17.53 ms giving a vllm/sglang ratio of 0.83 vs the claimed 0.79 (directionally consistent; residuals look like the unstated moe-split knobs).
- **No regressions**: I extracted all 19 modified parquets at the merge base and diffed row-multisets against head — **0 rows removed, 0 pre-existing rows value-modified, on every file** (the rtx moe byte-shrink is re-encoding: +551/−0). All 23 new tables are duplicate-free; added rows collide with zero pre-existing keys. Shared-code diffs (`collect_moe.py`, `collect_dsv4_megamoe.py`, `operations/mamba.py` +347/−0, `moe.yaml`) are additive for existing models, and every arithmetically checkable case-count pin re-derives exactly and moves K3-only.
- **Suites**: `pytest -m unit` 2805 passed; `cargo test --all-features` 296 passed (incl. the 6 new kda tests); 23/23 CI checks green.
- The CodeRabbit round's SDK fixes are all real at head (TP-divisibility guard, kda_layers range validation, shared-expert 2× fix — I re-derived the SwiGLU accounting — evidence-carrying strict-completeness raises, the per-family manifest pin test, vllm MLA-as-attention cases).

### Blocker 1 — the new KDA lane ships with no Python↔Rust parity anchor (`.claude/rules/rust-core/parity.md` requires it in this PR)

parity.md: a latency-affecting change "MUST in the same PR … **Anchor it**: a Python-generated oracle in the Rust `#[cfg(test)]` module (1e-9 … the `operators/gemm.rs` pattern), and/or a case in `parity_tests/test_engine_step_parity.py` when a new config class becomes reachable", and it names the `rust-engine-step-parity` CI job as "the enforcement mechanism". Verified at head:

- `grep -ri "kimi.k3\|kda" aic-core/rust/aiconfigurator-core/parity_tests/` → zero hits. SMOKE_CASES has no K3 entry, so the parity job is blind to the entire K3 lane: the KdaKey loader, both fused-kernel reroutes, the verify 2-axis path, and the ×(nextn+1) / ÷draft_tokens composition.
- The existing Python-generated oracle (`state_space.rs:1054`, `state_space_queries_match_python_v2_engine`) gained no KDA case — it still covers GDN + Mamba2 only, despite real packaged kda tables to oracle against.
- The new Rust tests are self-referential: `kda_sol_byte_model_matches_python_formulas` computes `expected` from the same inline constants it tests, so it cannot catch a shared misreading of the Python spec; the fused tests assert sum identities. (This also makes the ledger's "unit tests assert this on both sides" line inaccurate for the *routing*.)

The PR body's support-matrix rtol=0.05 comparison shows parity holds *today*; nothing keeps it holding. To be fair: I walked the twins line-by-line and they match rule-for-rule (SOL byte models bit-for-bit for all 7 kernel sources, reroute precedence, banker's-rounding divide-back — genuinely nice work), which is exactly why this is cheap to close. Suggested fix, either or both:

```python
# parity_tests/test_engine_step_parity.py — data already ships in-repo;
# EngineStepParityCase defaults (tp8/ep8) are already the fused 12-head shard.
pytest.param(
    EngineStepParityCase(model_path="moonshotai/Kimi-K3", system_name="b300_sxm",
                         backend_name="sglang", backend_version="0.5.16"),
    id="kimi-k3-b300-sglang-0516-nospec",
),
pytest.param(
    EngineStepParityCase(model_path="moonshotai/Kimi-K3",
                         backend_name="vllm", backend_version="0.1.dev19262"),
    id="kimi-k3-b200-vllm-dev19262-nospec",
),
# + a nextn>0 case (a small `nextn` field on EngineStepParityCase) so the
# DSPARK verify reroute is crossed end-to-end — the b300 sglang dataset is
# fused-CuTeDSL-verify-only, so a spec case exercises the reroute by construction.
```

…and/or Python-generated `query_kda` oracle rows (context + generation + verify; exact-hit / interpolated / beyond-range) appended to `state_space_queries_match_python_v2_engine`, following the GDN block directly above it.

### Blocker 2 — the one unresolved review thread: the agreed `FIXME(kernel-limit)` citation never landed

Inline comment on `collector/vllm/collect_kda.py:199`. Your doctrine rebuttal on the int32 cells was the rule-correct call (and CodeRabbit verified it); the single requirement that exchange left — the citation comment at the guard — is still absent at head.

### Blocker 3 — two shipped lanes pin a truncated, unresolvable image digest

Inline comment on `aic-core/.../b200_sxm/kda/sglang/0.5.16/collection_meta.yaml:13`. `sha256:6d9594a4` (8 hex chars) appears in full nowhere in the repo; for a custom ephemeral build, that makes the b200 MegaMoE lane and the replaced b200 kda generation phase unverifiable and non-recollectable.

### Majors — inline comments attached (fix here, or convert to dated owner decisions / fast-follows)

- **M1** (`KimiK3…_cases.yaml:64`): the sglang stock-image K3 moe lanes embody the same situ→silu approximation the vllm lane marks `_situ_as_silu` — unmarked.
- **M2** (`collector/sglang/collect_kda.py:445`): the fused-decode SM90 gate's "same as serving after its own JIT failure" claim has no file:line.
- **M3** (`KimiK3…_cases.yaml:78`): K3-on-trtllm plans moe and expands to zero cases with no logged drop.
- **M4** (`tests/unit/collector/test_sglang_attention_0514.py:180`): the only pin on the K3 attention-grid expansion lives in a CI-invisible test this PR edits (verified: 17/17 pass locally, 4 run under CI's `-m "unit or build"`).
- **M5** (`collector/helper.py:708`): the stale-lock break has a two-waiter unlink race; it's also a shared-mechanism change undeclared at PR level.
- **M6** (`tests/unit/collector/vllm/test_collect_kda_contract.py:21`): the dispatch contract test is satisfiable by the module docstring — one-click suggestion attached (validated: ruff check/format + pytest green against head).

### Minors — no merge impact; several already in your ledger fast-follow list (⟲)

- vllm `"fused_kda_decode"` SOL byte model aliases to the recurrence-only model although the kernel folds conv+recurrence+norm (both twins agree, so not a parity break — but the beyond-grid util-hold anchor is ~4× lighter on vllm, and the two confusable names `fused_kda_decode` / `kda_fused_decode` carry no distinguishing comment).
- ⟲ Nearest-shard fallback is unscaled for row-less divisor TPs (tp=6/12/16 → ~25–100% silent mispricing; the divisibility guard only rejects non-divisors); its tie-break also differs latently between Python dict order and Rust BTreeMap order — unreachable with the shipped {96,48,24,12} shards, but Mamba2 documents its analogous divergence in-code and KDA/GDN don't.
- ⟲ vllm mla_bmm silently halves the YAML-declared dtype axis (fp8 dropped) with no drop log; the trtllm precedent at least carries the FIXME lifecycle note.
- `_cleanup()` raising from `finally` after a fatal CUDA error masks the accumulated per-cell failure evidence (defeats the 02beeaab traceability fix exactly when it matters) and abandons the remaining cells.
- b300 vllm kda meta says "4 classified cell failures" — grid arithmetic and the 7 sibling metas say 20+1; b200's vllm kda meta records no failure info at all.
- "computelab" / "dlcluster" (internal cluster names) appear in the rtx moe meta and the ledger of a public repo.
- Ledger staleness: the status table still says b200 sglang kda "987 rows, Triton generation" (shipped: 976 with the fused replacement); the removed linear_attn_module row still reads "done" (the legend defines done = packaged); h200 IMA count 2,587 vs the meta's 2,590 (survivors+crashes = 3078 backs the meta; repeated in the rtx meta).
- The PR body's "Known gaps" still lists the vllm DSPARK draft geometry as won't-fix — 1c6767b7 fixed it (real Inferact geometry at head).
- Smaller: FlashKDA probe hardcodes head_dim=128 with the case's dim in scope; `tests/unit/collector/vllm/` lacks `__init__.py` (basename-collision hazard with the sglang twin file); the estimate CLI still prints "MTP nextn:" for K3; kda rows carry no power column (K3 power estimates report 0% coverage); the situ delta comment says <~3% where 8555ac6a measured up to ~5.5%; `__compat__ = "sglang==0.5.16"` would accept a future *stock* 0.5.16 on manual runs.

### Questions (non-blocking)

1. Does sglang serving on SM<90 actually catch the fused-decode ptxas failure and fall back to the Triton pair (M2)? A branch file:line settles it.
2. l40s vllm kda: `chunk_kda_with_fused_gate` has 414 context rows vs 428 for `flashkda_fwd` on every other system — which guard ate the extra 14 cells? The documented spectrum is 20+1.
3. Is the l40s (SM89) triton moe lane the kimi-k3 branch's real dispatch for K3's checkpoint class, or stock/GPT-OSS fallback behavior that happened to run (M1 adjunct)?
4. b300 sglang 0.5.14 carries only the w4a8 K3 lane (no w4a16 lane, unlike b200/gb200/gb300) — intended? Relatedly, the b300 vllm K3 moe run (972 rows, 2026-08-02 per its meta) is absent from the ledger's wave narratives.

collector/vllm/collect_kda.py
tianhaox: Landed in c87695892. The FIXME(kernel-limit) is there, but with an honest twist the thread didn't anticipate: the pinned preview's source is not publicly addressable (commit b6bbf29dd exists in no public vllm repo, branch kimi-k3 404s), and the era file at v0.24.0 computes token offsets from **int64** strides (`causal_conv1d.py:39,47` `stride_x_token: tl.int64` + explicit `.to(tl.int64)` casts) — mainline vLLM establishes NO int32 token-offset limit, per-block or otherwise. So the source supports neither the per-block reasoning nor tightening to the 3-block bound; the `nt * proj` bound stays as a documented unverified conservative guard (in-band cells pass on silicon; the guarded 21-cell spectrum is the collected baseline), to be re-verified — possibly deleted — at the next preview bump. The contract test now AST-pins the exact expression (`nt * proj >= 2 ** 31` via ast.unparse of If tests) and the false "same guard as sglang" comment is gone.

aic-core/src/aiconfigurator_core/systems/data/b200_sxm/kda/sglang/0.5.16/collection_meta.yaml
tianhaox: Recovered — Docker Hub still serves it. The kimi-k3 tag was re-pushed 2026-07-29T10:16Z and its manifest-list digest matches the prefix: `sha256:6d9594a421be244f2af29d726158ebffe9c3c2b3f39b5b89affd8150a106e187` (branch commit c6ad1f26, build tag `kimi-k3-c6ad1f26-20260729`; amd64 image `e35551fd...`). Full digest recorded in both metas (b200 kda + b200 MegaMoE), and the two bare metas upgraded to the structured format in 2652bee19 — with honest notes where provenance was never captured: b300 kda sglang gets the harness-pinned first-build digest + rows/collected_at (collector/case-plan hashes not recorded by that pre-provenance-writer run; gb300's identical 976-row spectrum is the corroborating record), b200 kda vllm gets the manifest-pin digest `e90e2603...` + rows 1203 + the phase breakdown. New citations in this wave anchor to the build commit (`@ c6ad1f26`) rather than the branch name, per your (b)/(c).

collector/cases/models/KimiK3ForConditionalGeneration_cases.yaml
tianhaox: Checked the query-key question both ways, then relabeled (2652bee19). Python `load_moe_data` + Rust `perf_database/moe.rs` consume kernel_source only as exact-match load-time routing (`moe_torch_flow_min_latency` split + the two DSV4 quant remaps) — none match these names, and the slice key is quant/shape/dist, so the suffix is invisible to queries: h100 K3 tp8ep1 power_law_1.01 returns bit-identical 0.5942489624023437 before == after, at both 0.5.14-exact and 0.5.16-nearest-version resolution. Rows now carry `sglang_marlin_moe_situ_as_silu` (h100 541 / h200 488 / rtx 551) and `sglang_fused_moe_triton_situ_as_silu` (l40s 2997); counts unchanged; all four lane metas updated. One honest caveat: kernel_source IS a shared-layer inheritance key via op_kernel_source_manifest.yaml — a regen produces ~1400 lines of tree-wide staleness churn (the checked-in manifest also predates this PR's own vllm _situ_as_silu rows), so I deferred it as its own owner pass; interim state matches the vllm precedent (suffixed approximation rows are not cross-backend inheritable — conservative). Also landed (c87695892): explicit 89/120 keys on the K3 map with the branch citation — the K3 moe-runner override fires on SM100/103 only (`arg_groups/overrides.py:479-504 @ c6ad1f26`), so SM120 resolves to marlin like SM90 and SM89 to the Triton default, matching the collected kernel_sources; the l40s meta now cites this.

collector/sglang/collect_kda.py
tianhaox: You were right to demand the citation — fetching the branch at the build commit (c6ad1f26) shows serving has NO catch-and-fallback: the model stashes the fused args on every NVIDIA GPU (`srt/models/kimi_k3.py:1563-1614` — only HIP and weight-layout checks), the backend calls the kernel wherever covered() accepts with no try/except around the JIT load (`srt/layers/attention/linear/kda_backend.py:426-476`), and the JIT compiles lazily at first call (`kernels/ops/attention/kda_fused_decode.py:42-51`). So the gate's fallback claim was wrong and I took your alternative (c87695892): the SM>=90 gate is removed, the collector now hands every decode step to covered() with no SM predicate, and covered SM<90 cells raise classified — the same ptxas crash serving hits. Data implication, flagged in the ledger for owner decision: the shipped l40s kda 12-head generation Triton rows have no serving-truth backing (serving on SM89 crashes rather than reaching the Triton pair), and future l40s kda runs will fail generation strict-completeness on those cells — prune vs keep-as-interpolation-support is the owner's call.

collector/cases/models/KimiK3ForConditionalGeneration_cases.yaml
tianhaox: Option (a) landed (c87695892) with two hardenings your warning pointed at. Moving the activation to `framework_specific_op_cases.{sglang,vllm}` kills the flagrant case: a K3-scoped trtllm run plans no moe at all. But dropping the yaml trtllm block entirely would have OPENED a hole in cross-model sweeps: the trtllm getter runs with no model filter (and, per your note, no backend arg — so `frameworks:` can't fire), and with no model-level allowed_modes the ungated base modes (bfloat16/fp8/fp8_block/nvfp4) become allowed for K3. So the block stays as an explicit empty tombstone (`allowed_modes: []`, dated comment — same idiom as Nemotron's vllm rows), and the trtllm getter now logs fully-dropped models exactly like the vllm getter (case_authoring.md citation included). `test_kimi_k3_moe_is_planned_per_framework_and_never_for_trtllm` pins all three: moe in sglang/vllm plans, absent from the trtllm plan, and every trtllm mode policy-rejected for K3.

tests/unit/collector/test_sglang_attention_0514.py
tianhaox: Landed in c87695892 — module-level `pytestmark = pytest.mark.unit`, so all 17 tests (including the K3 DSPARK + MLA-96 population pins) are now visible to CI's `-m "unit or build"` gate.

collector/helper.py
tianhaox: Landed in c87695892 as the rename-based breaker: the breaker renames the stale lock to a pid-suffixed name and unlinks the renamed file — os.rename is atomic, so exactly one breaker wins; the loser's rename raises ENOENT and it retries against whatever fresh lock now exists (never unlinking it). 30s wait window and 60s stale threshold unchanged. Unit tests extended: `test_log_perf_breaks_stale_lock_via_rename_and_writes` and `test_log_perf_losing_breaker_never_unlinks_the_fresh_lock` (the latter fails against the retired unlink-based breaker). The shared-mechanism change is declared in the bring-up ledger per layer_permissions meta-rule 2; PR-body line to add: "collector/helper.py log_perf: stale-lock breaking is now rename-based (single-winner) with a 30s wait window — shared executor mechanism, declared per collector meta-rule 2."

tests/unit/collector/vllm/test_collect_kda_contract.py
tianhaox: Landed in c87695892, verbatim (plus the guard-expression AST pin from the other thread in the same file). Both probe names are asserted as ast.Name references, so the module docstring can no longer satisfy the contract.

collector/vllm/collect_kda.py
Arsene12358: Verified at ee35dd33 — resolving. The FIXME(kernel-limit) block (collect_kda.py:194-206) is a better outcome than the citation I asked for: it documents that no citation is *possible* (preview commit b6bbf29dd unaddressable; the v0.24.0-era causal_conv1d.py:39,47 computes offsets from int64 strides, so mainline vLLM establishes no int32 limit at all) and pins the guard as deliberately-conservative-unverified with the next-bump re-verify hook. The contract test now AST-pins the exact `nt * proj >= 2 ** 31` expression and the false "same guard" comment is gone. Exactly the honest state layer_permissions.md wants.

aic-core/src/aiconfigurator_core/systems/data/b200_sxm/kda/sglang/0.5.16/collection_meta.yaml
Arsene12358: Verified — resolving. Full digest in both metas (kda meta :17, moe/0.5.16 structured `image_digest:` field), and the two bare metas (b300 sglang kda, b200 vllm kda) are upgraded to the structured format with full image digests plus the pre-provenance-writer backfill annotation. Nice recovery via the Docker Hub manifest-list; recording the build↔digest mapping (c6ad1f26 = 6d9594a4…, 74968e5653 = 4b8a7542…) in the ledger findings closes the mutable-ref citation concern too.

collector/cases/models/KimiK3ForConditionalGeneration_cases.yaml
Arsene12358: Verified in-parquet at head — resolving. Suffix counts exact on all four systems (h100 541 + h200 488 + rtx 551 `sglang_marlin_moe_situ_as_silu`; l40s 2997 `sglang_fused_moe_triton_situ_as_silu`); I re-diffed h100 against main — 0 pre-existing rows touched, counts unchanged, so the relabel stayed strictly K3-scoped. The query-invariance argument checks out against both loaders (kernel_source is exact-match load-time routing only; none of these names participate). Explicit 89/120 keys with the overrides.py:479-504 @ c6ad1f26 citation landed, l40s meta cites it. And the manifest follow-through went beyond the deferral — the absence_load_bearing outcome is a genuinely good piece of engineering; discussed in my PR comment.

collector/sglang/collect_kda.py
Arsene12358: Verified — resolving, with respect: you chased the branch source instead of assuming the fallback existed, and the finding turned out load-bearing. At head: the SM gate is gone (every decode step goes to covered(), :448), the no-fallback citations are in the comment (:429-436), the 22 unbacked l40s h12 decode rows are pruned in-parquet (1096→1074 — I checked: exactly the Triton-pair generation rows, with the 54 dispatch-gated h12 verify rows correctly kept), and the ledger records the owner decision plus the honest future-runs-raise-classified state.

collector/cases/models/KimiK3ForConditionalGeneration_cases.yaml
Arsene12358: Verified — resolving. Scoping to `framework_specific_op_cases.{sglang,vllm}` plus the `allowed_modes: []` tombstone (:93) — the ungated-base-modes hole you closed by keeping the block is a sharper failure mode than the zero-expansion I flagged; good call not just deleting it. trtllm getter drop-logging with the case_authoring citation is in, and `test_kimi_k3_moe_is_planned_per_framework_and_never_for_trtllm` passes locally at head.

tests/unit/collector/test_sglang_attention_0514.py
Arsene12358: Verified — resolving. Module-level `pytestmark` at :17; all 17 population tests (the K3 pins included) now run under CI's `-m "unit or build"` filter — confirmed locally.

collector/helper.py
Arsene12358: Verified — resolving. Rename-based single-winner breaker (:715-716; the loser's ENOENT path never touches the fresh lock), both new lock tests pass locally including the loser-never-unlinks contract that fails against the retired unlink breaker, and the ledger declaration is in place. One dangling item from your reply: the PR-body line you drafted ("stale-lock breaking is now rename-based…") isn't in the description yet — worth adding on your next edit.

tests/unit/collector/vllm/test_collect_kda_contract.py
Arsene12358: Verified — resolving. Applied verbatim, and the guard-expression AST pin via `ast.unparse` is a nice extension of the pattern; both contract-test files pass locally at head.

Arsene12358:
**Fix-wave re-review complete (at ee35dd33): every inline item is verified fixed and now resolved. One review item remains open — Blocker 1 from the review body: the Python↔Rust parity anchor.**

It was the only finding without an inline thread (`parity_tests/` isn't in the diff, so there was nothing to anchor it to), and it looks like it fell through the notification cracks: `parity_tests/` still has no Kimi-K3 case, the Python-generated oracle in `state_space.rs` (`state_space_queries_match_python_v2_engine`) still covers GDN + Mamba2 only, and the ledger's review-wave section records no disposition for it. `.claude/rules/rust-core/parity.md` requires the anchor in the same PR.

This wave itself produced the strongest argument for it: the donor-injection regression you caught and fixed (b300 spec bs1 TPOT 3.296→3.336 ms) was — as your own commit message notes — **invisible to the Python-vs-Rust compare gate, because both engines consume the same serialized source chain and shift together**. A committed K3 parity case is a different kind of net: it freezes the K3 engine-step numbers themselves, so a silent shift on either side fails CI instead of sailing through. It would also put the new exact-head-first mla_bmm routing (e1d42504 — which properly shipped its Rust twin and tests on both sides) under the same committed umbrella.

Ready-to-paste from the review, still current:

```python
# parity_tests/test_engine_step_parity.py — EngineStepParityCase defaults
# (tp8/ep8) are already the fused 12-head shard.
pytest.param(
    EngineStepParityCase(model_path="moonshotai/Kimi-K3", system_name="b300_sxm",
                         backend_name="sglang", backend_version="0.5.16"),
    id="kimi-k3-b300-sglang-0516-nospec",
),
pytest.param(
    EngineStepParityCase(model_path="moonshotai/Kimi-K3",
                         backend_name="vllm", backend_version="0.1.dev19262"),
    id="kimi-k3-b200-vllm-dev19262-nospec",
),
# + a nextn>0 case (a small `nextn` field on EngineStepParityCase) so the
# DSPARK verify reroute — and the donor-absence contract it now depends on —
# is crossed end-to-end in CI.
```

…and/or Python-generated `query_kda` oracle rows (context + generation + verify; exact-hit / interpolated / beyond-range) appended to `state_space_queries_match_python_v2_engine`, following the GDN block above it.

Everything else is closed — as soon as this lands I'll flip to approve.

collector/helper.py
tianhaox: Thanks for the thorough re-verification across the whole round. The dangling item is closed: the PR description now carries the shared-mechanism declaration line ("Shared-mechanism change (collector meta-rule 2): collector/helper.py log_perf stale-lock breaking is now rename-based (single-winner) with a 30s wait window — declared here and in the bring-up ledger"). No other action items remain from this round on our side.

tianhaox:
Blocker 1 landed in 4197820c5 — both halves.

**parity_tests**: your two cases verbatim (`kimi-k3-b300-sglang-0516-nospec`, `kimi-k3-b200-vllm-dev19262-nospec` — the harness prices a fixed config, so b200 memory-fit doesn't apply) plus `kimi-k3-b300-sglang-0516-dspark-nextn7` via a new `nextn` field on `EngineStepParityCase` threaded into `ModelConfig`. nextn=7 → verify width 8, the fused CuTeDSL kernel's collected draft-width cap, so the verify reroute + conv fold-to-zero + the donor-absence contract are crossed end-to-end. 12/12 K3 parity tests green; the full parity suite is 321 passed.

**state_space.rs oracle**: an 11-point Python-generated `query_kda` block appended to `state_space_queries_match_python_v2_engine` following the GDN pattern — b300 sglang 0.5.16 at the 12-head shard, generated against a `shared_layer=False` view per the parity rule: context `chunk_kda` (exact / seq-interp / batch-interp / beyond-seq), generation `kda_fused_decode` (exact / interp / beyond-range), and `fused_kda_decode_mtp_dspark` verify (exact / batch-interp / draft-interp / beyond), all at 1e-9 with SOL closures replicating `KdaOp::sol_total_bytes_with`.

Your framing was right and is now recorded in the ledger's decision list: the donor-injection class shifts both engines through the shared serialized source chain, so only committed anchors that freeze the K3 numbers themselves turn a silent shift into a CI failure.

Gate at 4197820c5: cargo workspace 307 passed; parity suite 321 passed; full collector+sdk 2241 passed / 0 failed; spot anchors 3.296000 (b300 sglang spec bs1) / 4.398000 (b300 vllm) exact; ruff clean.

Arsene12358 (approved):
**Approve.** Blocker 1 verified at 4197820c — both halves, independently re-run:

- The three parity cases landed (my two verbatim, plus `kimi-k3-b300-sglang-0516-dspark-nextn7` via the new `nextn` field — crossing the fused CuTeDSL verify reroute, conv fold-to-zero, and the kda donor-absence contract end-to-end in CI). I ran them locally against a fresh extension build at schema v6: all selected K3 cases plus a pre-existing control pass (16/16).
- The 11-point Python-generated `query_kda` oracle in `state_space_queries_match_python_v2_engine` is in and green (cargo lib suite 300 passed locally), with documented generation provenance (`shared_layer=False` view, 1e-9) covering exactly the kernels the K3 routing depends on — including `kda_fused_decode`, whose absence-load-bearing Triton twin the manifest now protects.
- The `rust-engine-step-parity` CI job is green on this head — the K3 lane is now inside its enforcement scope, which was the point.

With that, every item from my 2026-08-04 review is closed: the fix wave (verified in-code and in-parquet at ee35dd33, thread-by-thread confirmations attached there) plus this anchor. No-regression status re-confirmed on the final data state — every modified table remains a strict superset of main, and the relabel/prune stayed K3-scoped.

Nice campaign. The donor-injection catch and the SM89 no-fallback investigation in particular left the shipped data strictly more honest than where the review started. (e2e was the one check still running at approval time; branch protection gates the merge on it regardless.)

PR #1437, opened 2026-07-28
fix(cli): GPU recommender follow-ups and nextn merge-skew repair

Arsene12358:
## Summary

Post-merge follow-ups for the GPU recommender (#1369): the minor findings deferred during that review, plus a repair for the recommend path itself, which a review pass on this branch found to be broken on main by merge skew.

Tracks Linear issues AIC-1613, AIC-1614, AIC-1616, and the materialization half of AIC-1615 (the `parallel_experiments` thread-safety audit remains open there).

## Commit 1 — deferred review minors (AIC-1613, AIC-1614, AIC-1615-part)

- **cli**: the no-results GPU hint now keys off the load target instead of the unreachable `mode == "recommend"` branch, so recommend users no longer get advice about `--total-gpus` — a flag recommend mode doesn't have.
- **cli**: default mode warns when `--total-gpus` and a load target are both given (the target is silently ignored on that path).
- **sdk (aic-core)**: `get_static_info()` materializes the deferred summary row via `get_summary_df()` instead of asserting on a DataFrame that `run_static` no longer builds eagerly. New regression test covers the `get_static_info()`-first call order (the ordering the demo CLI and the former webapp static flow relied on).
- **tests**: recommend monotonicity helper takes the minimum across all modes instead of the first non-empty one; parametrized non-positive (`0`/`-1`/`nan`/`inf`) load-target parse tests; disagg replica-scaling case for the module_bridge `total_gpus_needed` path.
- **docs**: README and CLI user guide state the exactly-one (mutually exclusive) load-target contract.

## Commit 2 — recommend path repair (AIC-1616)

An independent review pass on this branch found that `aiconfigurator cli recommend` and `cli_recommend()` raise `TypeError: build_default_tasks() got an unexpected keyword argument 'nextn_accept_rates'` on current main: #1369 was developed against the old MTP interface while #1383/#1405 renamed it to `nextn`/`nextn_accepted`, and the rebase reconciled `cli_default` but not the recommend path.

The fix mirrors `cli_default` exactly: `cli_recommend` takes `nextn: int | 'auto'` + `nextn_accepted` and runs the same resolve-auto + `normalize_speculative_decoding` fail-fast block; the recommend subparser gets the default-mode `--nextn`/`--nextn-accepted` definitions in place of the removed `--nextn-accept-rates`; the recommend dispatch runs `_resolve_and_validate_nextn`.

The same commit repairs `tests/integration/test_picking.py`, which the #1369 merge broke twice over (the recommend tests hit the TypeError; the default/load-match tests unpack the pre-#1369 5-tuple from `_execute_tasks`, now a 6-tuple). This suite is not in CI, which is how both breaks landed — AIC-1616 notes adding it (or a fast subset) to presubmit as a follow-up.

## Test plan

- [x] `tests/integration/test_picking.py`: 8/8 pass (6/8 fail on main)
- [x] `tests/unit/cli` + new/changed sdk and generator tests: 218 pass
- [x] Full `tests/unit`: 2829 passed; only pre-existing failures remain (3 GLM checkpoint-name asserts in `tests/unit/collector/sglang/test_collect_mla_module.py`, unrelated and failing on clean main)
- [x] `ruff check` / `ruff format --check` clean on all touched files

## Summary by CodeRabbit

* **New Features**
  * Added `--nextn auto` support for recommend-mode speculative decoding; MTP now uses `--nextn-accepted`.
  * Improved disaggregated GPU sizing/scaling behavior.
* **Bug Fixes**
  * Fixed lazy inference summary static-info generation to correctly materialize deferred data.
* **Documentation**
  * Updated CLI user guide and README to describe recommend mode as load-target procurement sizing and to clarify argument mutual exclusivity and default-mode behavior.
* **Validation**
  * Strengthened checks for positive, finite load-target values and `--nextn`/`--nextn-accepted` constraints.


tests/integration/test_picking.py
```diff
@@ -80,7 +80,7 @@ def test_agg_vs_disagg(self, tmp_path):
             ttft=2000,
             tpot=50,
         )
-        chosen, best_configs, _, _, best_latencies = _execute_tasks(
+        chosen, best_configs, _, _, best_latencies, _ = _execute_tasks(
```
jasonqinzhou, line 83: **[P2] Put these repaired contracts under a presubmit gate.** This module is still outside every CI test path: `.github/workflows/build-test.yml` selects only `tests/unit` and `tests/e2e`, so `tests/integration/test_picking.py` never runs. That leaves both merge-skew failures fixed here—the real `build_default_tasks` keyword contract and the `_execute_tasks` return shape—able to recur with green checks. Could we add a fast unit/build smoke using autospec or real signatures, or explicitly gate a small subset of this file? The full roughly 200-second suite does not need to move wholesale.

jasonqinzhou (commented):
**COMMENT · 8.6/10 · high confidence**

## What this PR does

This PR follows up on the GPU recommender merge by aligning recommend mode with the current nextn/nextn_accepted contract, repairing broken picking tests, tightening load-target validation and guidance, and restoring lazy static-summary consumers. It also adds focused disaggregated artifact-sizing and deferred-summary regression coverage.

The change carefully reconciles two independently evolved interfaces across the parser, public API, task builder, generator bridge, and tests; the real picking integration suite now passes end to end, and the deferred-summary fix reuses the existing lazy materialization boundary instead of adding a second path.

## Why this score

Score 8.6/10, COMMENT: the implementation is sound across the reviewed production paths, all targeted local verification passed, and no correctness defect was found. The one recommended P2 asks that a fast version of the repaired real-signature and return-shape contracts enter presubmit, because the changed integration module is currently outside every CI test path and the same merge skew could otherwise recur with green checks. Existing current-head CodeRabbit feedback is not duplicated.

jasonqinzhou (approved):

docs/cli_user_guide.md
```diff
@@ -366,7 +366,7 @@ aiconfigurator cli recommend --model-path Qwen/Qwen3-32B --system h200_sxm --bac
 **Required arguments:**
 - `--model-path` (alias `--model`): HuggingFace model path or local path containing `config.json`
 - `--system`: System name (GPU type)
-- One of:
+- Exactly one of the following (mutually exclusive):
   - `--target-request-rate`: Target system throughput in req/s
   - `--target-concurrency`: Target number of concurrent users
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in c601359d. The Recommend section now consistently says “load target” and documents `--nextn auto`, when `--nextn-accepted` is required, and its valid range.

tests/integration/test_picking.py
jasonqinzhou: Fixed in c601359d with fast presubmit coverage: the Recommend API test now autospecs the real `build_default_tasks` signature, and the unit workflow test explicitly asserts the six-value `_execute_tasks` result including outcomes. Recommend-specific parser/dispatch cases were added as well. The affected suite passes 134 tests locally.

jasonqinzhou (approved):
All reviewed P2 findings are addressed on the final head: Recommend MTP docs and parser/dispatch coverage are complete, the real builder/executor contracts are now guarded in presubmit, all review threads are resolved, and the full required CI matrix is green.

PR #1440, opened 2026-07-29
fix(cli): address post-merge review follow-ups from GPU recommender

natoscott:
## Overview

Address the 5 non-blocking CodeRabbit minors flagged by @Yiming992 during the GPU recommender PR (#1369) review.

## Details

- **min_gpus cross-mode minimum**: The integration test helper compared GPU counts across all modes (agg + disagg), which could produce spurious results. Now compares only the `chosen_exp` mode.
- **Doc wording**: Removed redundant "exactly one ... mutually exclusive" phrasing in README and user guide -- argparse enforces mutual exclusivity, so the docs now say just "mutually exclusive".
- **Silent load target drop**: When both `--total-gpus` and a load target were provided, the load target was silently dropped with a warning. Now raises a hard error so the user must choose one or the other.
- **Non-positive target parse tests**: Added default-mode test coverage mirroring the existing recommend-mode tests for `_positive_float` validation (0, -1, nan, inf).
- **Disagg bridge test**: Added the missing `test_zero_total_gpus_needed_falls_back` case to `TestRecommendGPUCountDisagg`.

Relates to: #1369

## Test plan

- [x] `pytest tests/unit/cli/test_argument_parsing.py` -- 79 passed (8 new)
- [x] `pytest tests/unit/generator/test_module_bridge_recommend_gpus.py` -- 7 passed (1 new)
- [x] `ruff check` -- all clean

## Summary by CodeRabbit

- **Bug Fixes**
  - The CLI now rejects combining `--total-gpus` with load-target options instead of silently ignoring the targets.
  - Improved GPU recommendation handling for disaggregated deployments, including zero-GPU fallback scenarios.
  - Recommendation calculations now use results from the selected experiment only.

- **Documentation**
  - Clarified that request-rate and concurrency targets are mutually exclusive in `recommend` mode.
  - Updated guidance for using load targets in `default` mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ aiconfigurator cli support --model-path Qwen/Qwen3-32B-FP8 --system h200_sxm
   `examples/aisimulate/spica`.
 - Use `exp` to run customized experiments defined in a YAML file.
 - Use `generate` to quickly create a naive configuration without a parameter sweep.
-- Use `recommend` to find the minimum GPU count and optimal deployment configuration needed to meet a performance target. This mode is designed as a procurement sizing tool -- specify exactly one load target (`--target-request-rate` or `--target-concurrency` — mutually exclusive) along with SLA constraints, and the system calculates the minimum GPUs required. You can also omit `--total-gpus` in default mode with a load target for the same behavior.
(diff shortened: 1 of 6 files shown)


Arsene12358:
Hi @natoscott, thanks again for contributing, but I already took initiative to address these in #1437 . Could you check whether that's sufficient? Sorry for not informing you.

natoscott:
@Arsene12358 ah, thank you!  That looks great.

PR #1441, opened 2026-07-30
feat(collector): Upgrade vLLM xpu collector to v0.26.0

Spycsh:
#### Overview:

Upgrade vLLM xpu collector to vLLM v0.26.0

#### Details:

- Upgrade code, data to v0.26.0
- fix kv layout issue on v0.26.0
- fix moe interface errors on v0.26.0

- closes GitHub issue: #xxx

## New Features
- Added support for collecting vLLM performance data on Intel XPU hardware.
- Expanded XPU benchmark coverage for GEMM, attention, and MoE workloads.
- Added XPU-specific runtime and provenance tracking.

## Bug Fixes
- Improved XPU attention KV-cache handling across supported backends.
- Updated XPU MoE benchmarking with the current fused implementation and weight preparation flow.

## Compatibility
- Added vLLM XPU runtime support, pinned to version 0.26.0.
- CUDA and XPU now use their respective runtime and registry configurations.


collector/collect.py
```diff
@@ -1177,11 +1180,20 @@ def collect_vllm(
...
         return None, None
 
-    from collector.framework_manifest import require_collector_runtime
+    from collector.framework_manifest import CollectorRuntime, require_collector_runtime
 
     requested_ops = set(ops if ops is not None else (case_plan.ops if case_plan is not None else []))
     wideep_ops = {entry.op for entry in _wideep_registry_for_backend("vllm")}
-    runtime = require_collector_runtime("vllm", version, requested_ops=requested_ops, wideep_ops=wideep_ops)
+    if is_xpu_backend:
+        runtime = CollectorRuntime(
```
jasonqinzhou: **[P1] Keep XPU behind an exact, reproducible runtime gate.** This branch builds `CollectorRuntime` from whatever vLLM happens to be installed and records `installed-vllm-xpu:<version>` as though it were the producing image. That makes version validation tautological, loses the actual image/digest identity, and bypasses `require_collector_runtime`'s requested-op check (`build_collections(REGISTRY_XPU, "vllm", "0.26.0", ["not_a_real_op"])` still returns an empty selection). It is especially unsafe here because the changed MoE path depends on the v0.26 XPU-kernel contract while all three XPU modules still declare `vllm>=0.11.0`. Please add a first-class manifest-backed XPU runtime pinned to vLLM 0.26.0 and its immutable XPU environment, route this path through the same requested-op gate, and emit the actual runtime identity.

collector/collect.py
```diff
@@ -1162,10 +1162,13 @@ def collect_vllm(
     """Collect performance data for vLLM"""
     from collector.version_resolver import build_collections
 
+    is_xpu_backend = False
     if _cuda_available():
         from collector.vllm.registry import REGISTRY
     elif _xpu_available():
         from collector.vllm.registry import REGISTRY_XPU as REGISTRY
+
```
jasonqinzhou, line 1272: **[P1] Integrate the active XPU registry instead of deleting its closures.** The follow-up removes the three XPU keys to satisfy the stale-entry test, but finalization still iterates collections from `REGISTRY_XPU` and calls `collector_hash(module, ...)`. Because `enumerate_registry_modules()` and `changed_ops.py` read only `REGISTRY`, `load_closures()` now passes while a direct hash of `collector.vllm.collect_gemm_xpu` raises `KeyError`; the full upgrade also still reports `changed: []`. Please make `REGISTRY_XPU` a supported provenance and changed-op input, restore authored closures for all three modules (including `utils_xpu.py` for GEMM and attention), and add a regression that finalizes an XPU collection. Deleting the keys only hides the enumeration gap and makes real XPU finalization fail closed.

collector/vllm/collect_attn_xpu.py
```diff
@@ -256,9 +256,10 @@ def run_attention_torch(
 
     if backend_name_str in {"FLASH_ATTN", "FLASHINFER", "TRITON_ATTN"}:
         # The collector helper populates cache as [2, num_blocks, ...] because
-        # that layout makes K/V insertion simple. vLLM V1 backends consume it as
-        # [num_blocks, 2, ...].
-        kv_cache = kv_cache.transpose(0, 1).contiguous()
+        # that layout makes K/V insertion simple. vLLM V1 FA-style backends
+        # consume packed K/V as [num_blocks, num_kv_heads, block_size,
+        # 2 * head_size], then transpose block/head dims and split K/V.
+        kv_cache = torch.cat([kv_cache[0], kv_cache[1]], dim=-1).transpose(1, 2).contiguous()
```
jasonqinzhou: **[P1] Preserve vLLM's selected physical KV-cache layout.** In vLLM 0.26 the logical cache is `[B, H, N, 2D]`, but XPU's default FlashAttention path uses physical NHD stride order `(0, 2, 1, 3)`. The concatenation here is already contiguous `[B, N, H, 2D]`; transposing it creates the correct logical view with NHD strides, but the trailing `.contiguous()` converts it to physical HND. Collection would therefore time different cache strides from serving. Please apply the selected backend's `get_kv_cache_stride_order()` the same way runtime and [vLLM's backend tests](https://github.com/vllm-project/vllm/blob/v0.26.0/tests/v1/attention/test_attention_backends.py#L495-L508) do, under the active vLLM config, and add a stride-focused regression.

jasonqinzhou: **[P1] Complete the split v0.26 cache-update contract.** Repacking the cache here is only half of the new path: `FlashAttentionBackend.forward_includes_kv_cache_update` is false, and `impl.forward()` only reads the cache. This collector still never calls `impl.do_kv_cache_update`, so context cases read wholly uninitialized current-token slots, generation omits the current token, and the recorded latency excludes cache insertion. Although the omission predates this patch, the stated v0.26 upgrade cannot produce trustworthy rows without it. Invoke the selected implementation's update inside the timed operation when required, as [vLLM's test](https://github.com/vllm-project/vllm/blob/v0.26.0/tests/v1/attention/test_attention_backends.py#L304-L310) and the CUDA collector do, and add an update-before-forward regression.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 4.8/10 · high confidence**

## What this PR does

This PR updates the Intel XPU collector toward vLLM 0.26.0 by repacking attention KV cache, migrating MoE execution to the stateful `XpuFusedMoe` interface, reusing output buffers, and bypassing the CUDA vLLM runtime pin for XPU environments.

The MoE migration thoughtfully preserves the v0.26 operator's weight, scale, output-buffer, and expert-parallel inputs, and the author has responded quickly to both Ruff and provenance-test feedback with narrowly scoped follow-up commits.

## Why this score

Score 4.8/10 with REQUEST_CHANGES. The latest commit clears the stale-closure unit failure by deleting the three XPU closure keys, but this leaves XPU collection finalization unable to compute a fail-closed collector hash and still makes the entire XPU upgrade invisible to changed-operation tracking. The prior runtime-reproducibility, attention stride-order, and split KV-cache-update blockers are unchanged, so current-head XPU rows are not yet reproducible or trustworthy.

collector/vllm/collect_attn_xpu.py
Spycsh: Thanks, fixed.

Spycsh: Fixed.

Spycsh:
https://github.com/ai-dynamo/aiconfigurator/commit/c56db49f  leads to strict kv cache budget check and causes some local test failure. I'm checking with it.

Spycsh:
> [c56db49f](https://github.com/ai-dynamo/aiconfigurator/commit/c56db49f) leads to strict kv cache budget check and causes some local test failure. I'm checking with it.

resolved with releasing `other_mem` in `b60.yaml`, data upstreamed v0.26.0 for reference. Validate with local matrix.

tianhaox:
**Review: vLLM XPU collector upgrade to 0.26.0**

I verified the core changes against the vLLM v0.26.0 sources. The attention/MoE adaptation is solid — two issues below worth fixing before merge, both quick.

**Verified correct (checked against vLLM v0.26.0):**
- The KV-cache rework matches `vllm/v1/attention/backends/flash_attn.py@v0.26.0` exactly: packed logical shape `(num_blocks, num_kv_heads, block_size, 2*head_size)`, NHD stride order `(0,2,1,3)`, and the split `forward_includes_kv_cache_update=False` / `do_kv_cache_update` contract. `_pack_v1_fa_kv_cache`'s permute→contiguous→permute correctly preserves the logical shape while applying the backend's physical stride order, and the stride-locking unit test pins it. FLASHINFER at v0.26.0 uses the same packed shape, so the unified branch is fine.
- Driving layout from the backend's own classmethods (`get_kv_cache_stride_order` / `get_required_kv_cache_layout`) instead of hand-coded transposes is the right call, and the `exit_stack.callback(set_kv_cache_layout, None)` fixes the old global-layout leak.
- The MoE migration to the stateful `XpuFusedMoe` (constructed once, outside the token loop) removes the fragile `inspect.getsource` kernel-version sniffing — net improvement.
- The `vllm_xpu` manifest entry is clean: digest-pinned image, `Version(...).public` strips `+xpu`, unknown-op requests fail closed.

**Issue 1 — the new `__compat__` pin test never runs.**
`tests/unit/collector/test_framework_manifest.py:68` — `test_active_vllm_xpu_collectors_are_exactly_pinned_to_manifest_version` is indented inside the CUDA test's body, so pytest never collects it and nothing calls it. This is load-bearing, not just hygiene: `changed_ops` does not track a standalone `vllm_xpu` manifest pin bump (no `collector/vllm_xpu/registry.py`, so that framework contributes no entries). The only chain that forces recollection on an XPU runtime bump is: manifest bump → this exact-pin test fails → collector modules get edited → `collector_code` fires under `vllm`. Fix: dedent to module scope. (CodeRabbit flagged this too; still present at head.)

**Issue 2 — the new oneCCL 2021.15.9 dataset is unreachable.**
oneCCL data is selected strictly via `b60.yaml`'s `misc.oneccl_version` (`aic-core/src/aiconfigurator_core/sdk/operations/communication.py:339`), which is still `'2021.17.2'` — and the 2021.17.2 dataset still exists. Comm ops never inherit sibling-version rows, so `systems/data/b60/comm/oneccl/2021.15.9/` is dead data as shipped. If the 0.26.0 XPU image ships oneCCL 2021.15.9, bump `oneccl_version` accordingly; otherwise the directory version looks mislabeled.

**Non-blocking notes:**
- Merging `REGISTRY_XPU` into the `vllm` framework in `changed_ops` is conservatively correct but over-flags both ways (a CUDA-only pin bump marks b60 XPU data stale; an XPU-only edit marks CUDA families changed). Worth a one-line note in the module docstring.
- The new provenance-writer test asserts `runtime.framework: vllm_xpu` in `collection_meta.yaml`, while the shipped (legacy-format) meta files say `vllm` — worth confirming downstream loaders are indifferent to that field before the first real finalize run lands.
- `getattr(backend_cls, "forward_includes_kv_cache_update", False)` inverts vLLM's base-class default (`True`). At the pinned 0.26.0 the attribute always resolves via inheritance so it never matters — but since `__compat__` is exact, a plain attribute access (fail-loud if missing) would be safer.

Spycsh:
@jasonqinzhou @tianhaox , there is a CI issue here https://github.com/ai-dynamo/aiconfigurator/actions/runs/32215067044/job/95954910421?pr=1441

The CI failure is caused by test-order pollution in the unit test fixtures. In the failing CI run, `test_comprehensive_router_survives_scoped_stub_patch` runs on the same xdist worker shortly before `test_supported_quant_modes_structure`. That first test activates the stub_perf_db scoped fetch_table_view monkeypatch, and when the shared comprehensive_perf_db singleton is built or rewarmed under that patch, the stub fetcher masks the module-level comprehensive router. As a result, the synthetic test_system database only receives the few tables defined by the stub override, while the other comprehensive tables resolve to None and get cached as empty/unloaded data. Later, supported_quant_mode reads one of those cached empty tables, so one expected op returns [], and the assertion len(supported[op]) > 0 fails.
The [fix](https://github.com/ai-dynamo/aiconfigurator/pull/1441/commits/a2671b5e150646fa615043a7aab3252ef20b94a8) keeps scoped stub patches from intercepting the comprehensive fixture: when the patched fetcher sees database.system == "test_system" and _COMPREHENSIVE_OVERRIDES is available, it returns the comprehensive synthetic table instead of the stub override. This preserves the intended isolation: stub_perf_db still gets its minimal stub data, while comprehensive_perf_db always gets the full comprehensive synthetic data regardless of fixture ordering.

To reproduce without the fix

```
pytest -q tests/unit/sdk/database/test_data_loaders.py::test_comprehensive_router_survives_scoped_stub_patch tests/unit/sdk/database/test_edge_cases.py::TestSupportedQuantModes::test_supported_quant_modes_structure -vv -s
```

Spycsh:
@jasonqinzhou @tianhaox pls review or merge this pr at your convenience.

tianhaox:
all cleared from my side. Please resolve the merge conflict.

tianhaox (approved):

Spycsh:
@tianhaox thanks for review. Conflicts are resolved.

jasonqinzhou (approved):

PR #1442, opened 2026-07-30
feat(core-sdk): unify large-EP MoE perf schemas, ops, and coverage

Arsene12358:
#### Overview

First PR of the model-agnostic large-scale EP refactor (Linear project "AIC refactor: model-agnostic large-scale EP", issue AIC-1438). It introduces the unified MoE communication and expert-compute data model in the core SDK: `moe_a2a_perf` / `moe_expert_compute_perf` schemas, lossless adapters for all six shipped legacy parquet families, the `MOE_A2A_BACKENDS` registry, `MoEAllToAll` / `MoEExpertCompute` ops, `query_moe_a2a` / `query_moe_expert_compute` database forwards, and coverage-probe APIs.

This PR is **aic-core only and nothing consumes the new surface yet**. Model wiring, coverage-driven config search, and `enable_wideep` deprecation land in follow-up AIC-1439. Modeling and collection methodology remain unchanged; the refactor is gated on numerical equivalence with the existing paths.

#### Details

- **Unified schemas:** `moe_a2a_perf` is keyed by `[comm_backend][phase][comm_dtype][ep_size][node_num][hidden][topk][num_experts][sms][tokens]`; `moe_expert_compute_perf` is keyed by `[kernel_source][quant][distribution][inference_phase][topk][num_experts][num_slots][hidden][inter][moe_tp][moe_ep][tokens]`.
- **Lossless legacy adapters:** SGLang `wideep_deepep_normal` / `_ll`, TRT-LLM `trtllm_alltoall`, SGLang `wideep_context/generation_moe`, and TRT-LLM `wideep_moe`. Existing query paths coexist until the wiring follow-up.
- **Communication registry:** `deepep_ht`, `deepep_ll`, `nvlink_two_sided`, and `nvlink_one_sided`, with framework, inference-phase, communication-phase, and feasibility metadata.
- **Ops and queries:** `MoEAllToAll` and `MoEExpertCompute`, backed by `query_moe_a2a` and `query_moe_expert_compute`, with exact/alias/sole-available dtype resolution and phase-scoped distribution fallback.
- **Coverage APIs:** `moe_a2a_coverage(...)` requires dispatch/combine intersection per backend; `moe_expert_compute_coverage(...)` exposes the expert-compute support surface for the follow-up candidate enumerator.
- **Current-main compatibility:** updated to `main@90f820f6`; native FPM now owns its OpSpec coverage. DeepEP-HT frontier queries resolve the summed dispatch+combine curve before preserving phase shares, matching `main`'s nonlinear joint-log Grid hold.
- **Data hardening:** required latency rejects null, NaN, and infinities; optional power keeps null/NaN as unmeasured (`0.0`) and rejects measured infinities.
- **Review cleanup:** renamed the expert-compute op test module and moved its script entry point to the end of the equivalence test module.

#### Regression evidence

| Gate | Result |
|---|---|
| Focused loader/op/API matrix | 138 passed |
| Repository unit gate | 3,479 passed, 10 skipped (`test_moe_dispatch.py` excluded per native-extension guidance) |
| L0 adapter equivalence | Lossless on all six legacy families (11,820 adapted rows) |
| L1 query equivalence | rel <= 1e-9 with zero carve-outs; all 430 all-to-all shipped-data comparisons pass on current `main` interpolation |
| Lint / formatting | Ruff clean |
| DCO | Both closeout commits signed off; DCO check green |

- `aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py` — registry, unified loaders, adapters, query behavior, and both ops.
- `aic-core/src/aiconfigurator_core/sdk/perf_database.py` — database forwards and coverage probes.
- `tests/unit/sdk/database/test_moe_a2a_query_equivalence.py` and `test_moe_ep_query_equivalence.py` — old-vs-new numerical equivalence.
- `tests/unit/sdk/database/test_moe_a2a_loader.py` and `test_moe_ep_loader.py` — schema and invalid-data contracts.
- `aic-core/API.md` — public API contract.

#### Related issues

- Relates to Linear AIC-1438 (foundation)
- Relates to Linear AIC-1439 (wiring/search/flag deprecation follow-up)
- Relates to Linear AIC-1601 (Rust port follow-up)
- Relates to Linear AIC-1440 (collector follow-up)

## Summary by CodeRabbit

* **New Features**
  * Added unified large-EP MoE all-to-all and expert-compute performance queries.
  * Exposed MoE operations and backend capability information through the SDK.
  * Added APIs to identify available MoE performance coverage.
* **Documentation**
  * Clarified supported access paths and silicon-measured data availability for MoE queries.
* **Tests**
  * Added coverage for data loading, interpolation, fallbacks, coverage discovery, and legacy-data equivalence.


Arsene12358:
Rebased onto main @ dc4caca (post-#1460), per the coordination plan in the merge-coordination note: #1460 landed first, this stack carries its conventions.

**Merge-base audit.** Baselines were re-captured at the new merge-base with a two-point capture isolating #1460 (main @ 173c9cb vs @ dc4caca): all comparable gate artifacts are byte-identical, independently confirming #1460's zero-numeric-drift claim. One gate case left the matrix: Kimi-K2 8×H200 agg has no feasible config on main since #1423 (bisected; the per-projection BF16 attention weights push the 140.4GB winners over the memory budget) — an upstream behavior change, so the local dual-checkout gate is now 73 artifacts (was 81).

**One behavioral fix rode the rebase** (`e2b884f`): the moe_ep legacy adapters now follow the first-wins skip-on-key-conflict contract that #1423 gave their oracle loaders (`load_wideep_{context,generation}_moe_data`, `load_wideep_moe_compute_data`). This was caught by the harness as a 4.3% `context_moe` divergence on sglang 0.5.10 (multi-source data): with the fix, `EPMoE` matches the post-#1423 legacy query bit-exact (395.02816ms on the probe point). A cross-source regression test now pins the mechanism. The a2a adapters already kept first.

Fresh gates at `e2b884f`: strict-provenance unit 2891, cross-package 68, dual-checkout harness **73 PASS / 0 LEDGERED / 0 FAIL** vs the dc4caca baselines.

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
```diff
@@ -0,0 +1,1294 @@
...
+            int(row["ep_size"]),
+            int(row["node_num"]),
+            int(row["hidden_size"]),
+            int(row["topk"]),
+            int(row["num_experts"]),
+            _normalize_sms(row.get("sms")),
+            int(row["num_tokens"]),
+        )
+        latency = float(row["latency"]) / 1000.0  # collector records us; leaves are ms
+        power = float(row.get("power", 0.0))
```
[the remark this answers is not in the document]
Arsene12358: Fixed, in the two parts you outlined:

- **power** — the `_row_power` normalizer (null/NaN/absent → 0.0, "not measured") now lands on this PR where the loaders are born (f74b652, hoisted from later in the stack where it had been staged as the collectors PR's declared loader rider), covering every power read in the module — both new-schema loaders and all four legacy adapters — with regression tests in both loader test files.
- **latency** — we took the hard-error branch of your suggestion: it's schema-required, so `_require_latency` (13b6c04) refuses the load with a named error ("latency is schema-required; refusing to load corrupt perf data") instead of the bare `float("")` ValueError; coercing to 0.0 ms would poison consumers silently. Scope note: the legacy adapters keep their oracle loaders' bare-float behavior deliberately (parity with `operations/moe.py`).

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
```diff
@@ -0,0 +1,1294 @@
...
+
+_A2A_PHASES = ("prepare", "dispatch", "combine")
+
+
+def _validate_a2a_request(comm_backend: str, phase: str) -> None:
+    """Shared ctor/query validation: unknown backend or phase is a ValueError."""
+    if comm_backend not in MOE_A2A_BACKENDS:
+        raise ValueError(f"Invalid comm_backend '{comm_backend}'. Must be one of {sorted(MOE_A2A_BACKENDS)}")
+    if phase not in _A2A_PHASES:
+        raise ValueError(f"Invalid phase '{phase}'. Must be one of {list(_A2A_PHASES)}")
```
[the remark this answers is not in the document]
Arsene12358: Adopted verbatim (13b6c04), including your caveat check: the block builder iterates `spec.comm_phases` itself and the coverage probe walks table keys without calling the validator, so no live path relied on the looser behavior — the check is a pure guard that fails where the intent is expressed. One consequence worth noting: the Rust twin had deliberately test-pinned the OLD behavior (`undeclared_but_valid_phase_is_a_data_miss_not_a_config_error`), so #1464 mirrors the tightened contract (36ba3cf) — `('deepep_ht', 'prepare')` is now an `InvalidEngineConfig` there too, and the pinning test flipped with it.

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
```diff
@@ -0,0 +1,1294 @@
...
+        self._quant_mode = quant_mode
+        self._workload_distribution = workload_distribution
+        self._attention_dp_size = attention_dp_size
+        self._inference_phase = inference_phase
+        self._kernel_source = kernel_source
+        self._is_gated = is_gated
+        # 3 GEMMs for gated (gate, up, down), 2 GEMMs for non-gated (up, down).
+        # EP-only family: no moe_tp division (moe_tp == 1 by construction).
+        num_gemms = 3 if is_gated else 2
+        self._weights = hidden_size * inter_size * num_experts * quant_mode.value.memory * num_gemms // moe_ep_size
```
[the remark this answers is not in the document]
Arsene12358: Confirmed against the merge-base: the legacy `TrtLLMWideEPMoE._weights` sizes by `num_experts` (`operations/moe.py:1435` at dc4caca), so this op is deliberately parity-pinned — `get_weights()` must byte-match the retired classes on the shipped gb200 EPLB artifacts per the series' zero-regression contract (the case05 EPLB gate artifacts match exactly because of this). Took your fallback: 13b6c04 adds the pinning comment at the site, and the physically-correct `num_slots` sizing is filed as **AIC-1674** with the process it needs — it moves the memory column on every EPLB config, and both twins (this op and the Rust `EpMoeOp`) must change together, so it's an intentional-delta follow-up rather than a ride-along here.

jasonqinzhou:
### Naming suggestion: rename `EPMoE` to `MoEExpertCompute`

Could we rename this surface before the new public contract lands? `EPMoE` reads like the entire expert-parallel MoE block, but the operation represents only the routed-expert compute between dispatch and combine:

```text
MoEAllToAll(dispatch)
→ MoEExpertCompute
→ MoEAllToAll(combine)
```

It excludes the router, dispatch/combine communication, and the separately modeled shared-expert branch. `MoEExpertCompute` therefore describes the ownership much more directly; the documentation can call it “routed-expert compute” where that distinction matters.

I suggest carrying the rename consistently through the full stacked contract:

| Current | Suggested |
|---|---|
| `EPMoE` | `MoEExpertCompute` |
| `EpMoeOp` | `MoeExpertComputeOp` |
| `moe_ep_perf.parquet` | `moe_expert_compute_perf.parquet` |
| `query_moe_ep()` | `query_moe_expert_compute()` |
| `moe_ep_compute_coverage()` | `moe_expert_compute_coverage()` |

Because these APIs and the producer schema are new and unmerged, this is the cheapest point to make the coordinated change across #1442, #1448, #1464, and #1470, before the names become compatibility obligations.

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
```diff
@@ -0,0 +1,1354 @@
...
+    if comm_dtype in phase_slice:
+        return phase_slice[comm_dtype]
+    if comm_dtype == "fp8_block" and "fp8" in phase_slice:
+        logger.debug(
+            "moe_a2a: comm_dtype 'fp8_block' not collected; normalizing to 'fp8' "
+            "(behavioral mode reusing the fp8 comm tables; %s)",
+            query_context,
+        )
+        return phase_slice["fp8"]
+    if len(phase_slice) == 1:
```
jasonqinzhou: **[P1] Restrict the sole-dtype fallback to untyped DeepEP data.** `phase_slice` is typed for the TRT-LLM adapters, so a sole key is not an interchangeable stand-in for a different requested precision. Shipped GB200 `NVLinkOneSided` contains only `nvfp4`; this branch makes `bfloat16`, `fp8`, and `fp8_block` requests all return the same `0.0128959998 ms` nvfp4 leaf, while the legacy query raises `PerfDataNotAvailableError` for those uncollected modes. Matched two-sided rows show bf16/nvfp4 dispatch ratios from 0.56x to 3.48x, so the substitution is material, and the dtype-agnostic coverage probe could let PR 2 admit such an unsupported config. Please preserve the explicit `fp8_block -> fp8` alias but limit sole-value fallback to genuinely untyped legacy DeepEP data (the `default` slice, or an equivalently backend-gated rule); typed slices should raise the existing named miss. Add a sole-nvfp4 one-sided regression while retaining the DeepEP-default fallback case.

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
```diff
@@ -0,0 +1,1354 @@
...
+
+    # ------------------------------------------------------------------
+    # Op contract
+    # ------------------------------------------------------------------
+
+    def query(self, database: PerfDatabase, **kwargs) -> PerformanceResult:
+        # Attention DP globalizes tokens: the A2A dispatch delivers every DP
+        # rank's tokens to the experts (same scaling as legacy MoE /
+        # TrtLLMWideEPMoE query()).
+        num_tokens = kwargs.get("x") * self._attention_dp_size
```
jasonqinzhou, line 1411: **[P1] Preserve the SGLang context-EPLB token correction.** `SGLangEPMOEModel` currently passes `enable_eplb=self.config.enable_eplb`, and legacy `MoE._query_moe_table` uses `int(num_tokens * 0.8)` for context EPLB before lookup. `EPMoE` has no corresponding flag and this line always queries raw ADP-scaled tokens. On shipped `h200_sxm`/SGLang `0.5.6.post2` data (`power_law_0.6`, topk=8, experts=256, hidden=7168, inter=2048, TP=1, EP=2), 32 tokens returns `1.2833225 ms` here versus `1.0025957 ms` through the live EPLB path, 28% high. The L1 equivalence test explicitly omits `enable_eplb`, so it does not cover this mode. Please thread the EPLB token-adjustment contract through `EPMoE` (or explicitly record and validate its removal as an intentional modeling change) and add an `enable_eplb=True` equivalence case.

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
```diff
@@ -0,0 +1,1354 @@
...
+        # Verbatim wideep-MoE roofline (TrtLLMWideEPMoE._query_compute_table's
+        # get_sol): num_slots (not num_experts) sizes the weight-read term —
+        # EPLB redundant mode replicates experts across slots. sglang-adapted
+        # slices pin num_slots == num_experts, where this reduces exactly to
+        # the sglang oracle's SOL (MoE._query_moe_table's get_sol). num_gemms
+        # is pinned to 3 (gated): the forward has no is_gated axis and both
+        # legacy queries default is_gated=True (neither legacy wideep op
+        # forwards it from query()). The SOL only shapes the beyond-range
+        # boundary util-hold; in-range lookups are raw lerp on measured points.
+        num_gemms = 3
```
jasonqinzhou: **[P2] Honor `is_gated` in the query roofline.** The constructor exposes and stores `is_gated` and uses it for resident weights, but this query always assumes three GEMMs. Legacy SGLang `MoE.query` does forward the flag; on the shipped generation `power_law_0.6` EP32 slice at 8192 tokens (2x the collected maximum), the unified result is `0.2308887 ms` versus `0.2264363 ms` for the legacy non-gated path. Please thread `is_gated` into the forward/SOL calculation and add a non-gated overflow-equivalence case, or remove/document the unsupported parameter before making this a public operation.

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
```diff
@@ -0,0 +1,1354 @@
...
+        # Attention DP globalizes tokens: the A2A dispatch delivers every DP
+        # rank's tokens to the experts (same scaling as legacy MoE /
+        # TrtLLMWideEPMoE query()).
+        num_tokens = kwargs.get("x") * self._attention_dp_size
+        kernel_source = self._kernel_source
+        if kernel_source is None:
+            kernel_source = self._resolve_kernel_source(database, self._quant_mode)
+        result = database.query_moe_ep(
+            kernel_source,
+            self._quant_mode,
```
jasonqinzhou: **[P2] Preserve the per-call `quant_mode` override.** Both expert-compute operations this class replaces use `kwargs.get("quant_mode")`, but `EPMoE` always resolves the kernel and queries with `self._quant_mode`. A direct parity probe constructed all three with `fp8_block` and called `query(..., quant_mode=nvfp4)`: both legacy ops forwarded `nvfp4`, while this op still forwarded `fp8_block`. With `kernel_source=None`, that can also choose the wrong Blackwell kernel (`deepgemm` versus `moe_torch_flow`). Please derive one effective quant mode from `kwargs`, use it for both kernel resolution and `query_moe_ep`, and add an op-level parity test.

jasonqinzhou (changes requested):

## What this PR does

This PR lays the aic-core foundation for model-agnostic large-scale expert parallelism: unified MoE communication and expert-compute schemas, lossless adapters for shipped legacy tables, new operation/query APIs, backend metadata, and coverage probes. Model construction and candidate-search wiring are intentionally deferred to the next PR.

The three follow-up commits are well scoped: they restore the #1423 shared-layer first-wins contract with a cross-source regression, distinguish optional power from required latency, and align phase validation with the backend registry while closing every CodeRabbit thread.

## Why this score

6.0/10, REQUEST_CHANGES. The force-rebase preserves all 12 previously reviewed patches and adds three useful follow-ups; all 199 focused tests and 19 live checks pass. Two blocking behavioral regressions still make the new public operation unsafe for affected configurations: typed TRT communication slices silently substitute the wrong precision through the sole-dtype fallback, and SGLang context EPLB still loses its active token correction. The non-gated overflow roofline and per-call quant-mode contracts also remain inconsistent with the legacy operations.

Harrilee (approved):

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
Arsene12358: Fixed. The sole-value fallback is now gated on the slice being genuinely untyped: step 3 of `_resolve_comm_dtype_slice` only fires when the sole key IS `"default"` (the legacy DeepEP shape); a typed sole key (e.g. nvfp4-only NVLinkOneSided) raises the named `PerfDataNotAvailableError` listing the available dtypes. The explicit `fp8_block -> fp8` alias is untouched. Regressions: `test_dtype_fallback_to_sole_untyped_default` (DeepEP default slice still falls back) and `test_sole_typed_dtype_raises_named_miss` (sole-nvfp4 one-sided combine raises for bf16/fp8/fp8_block) in tests/unit/sdk/database/test_moe_alltoall_op.py.

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
Arsene12358: Fixed. `EPMoE` gained `enable_eplb` (ctor + per-call kwarg), and `_query_ep_table` applies the legacy context correction — `num_tokens = int(num_tokens * 0.8)` — exactly on the legacy gate: `enable_eplb and inference_phase == "context" and kernel_source in _SGLANG_ADAPTED_KERNEL_SOURCES` (moe.py:649 semantics; the trtllm path never had the correction and still doesn't). The L1 equivalence suite gained an `enable_eplb=True` sweep (`test_eplb_context_equivalence_sweep`) that matches the live sglang EPLB path leaf-for-leaf, including your h200 0.5.6.post2 repro slice. The Rust twin already carried the 0.8 correction and now has the same gate.

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
Arsene12358: Fixed. `is_gated` now flows into the query roofline: `_query_ep_table` takes it and the beyond-range SOL uses `num_gemms = 3 if is_gated else 2` (matching legacy `MoE.query`); the Rust `ep_moe` SOL got the same threading (`MoeEpTable::query(..., is_gated)` -> `ep_sol_latency_ms`). Non-gated 2x-max overflow equivalence case added (`test_non_gated_overflow_equivalence`).

aic-core/src/aiconfigurator_core/sdk/operations/moe_comm.py
Arsene12358: Fixed. `EPMoE.query` derives one effective quant mode — `quant_mode = kwargs.get("quant_mode") or self._quant_mode` — and uses it for BOTH kernel resolution and `query_moe_ep`, mirroring the two ops it replaces. Op-level parity probe added (`test_per_call_quant_mode_override`): ctor nvfp4 (misses), `query(..., quant_mode=fp8_block)` hits the fp8_block leaf.

Arsene12358:
**Re: naming suggestion (`EPMoE` -> `MoEExpertCompute`)** — Adopted across the stacked contract as proposed: `EPMoE` -> `MoEExpertCompute`, `EpMoeOp` -> `MoeExpertComputeOp`, `moe_ep_perf.parquet` -> `moe_expert_compute_perf.parquet`, `query_moe_ep()` -> `query_moe_expert_compute()`, `moe_ep_compute_coverage()` -> `moe_expert_compute_coverage()` — one rename commit per PR at its layer (SDK op + loader + coverage API on this PR; call sites on PR 2; the Rust op/table/wire vocabulary on PR 3, still inside the unreleased v7 bump; the collector table filename on PR 4). The design docs now describe the op as routed-expert compute between dispatch and combine.

Arsene12358:
Rebased onto main `ed13cfc` (post #1384/#1474/#1496/#1512): `9a31043`. Reconciliation on this PR is textual only — the opspec-coverage EXEMPT dict unions #1384's FPMForwardOp entry with the large-EP entries. Local gates re-verified at the stack tip: unit 3122 + cross_package 72, parity suite 430 passed / 0 skipped, and the workspace ground-truth harness at 73 PASS / 0 FAIL against baselines re-captured at `ed13cfc` (main's only drift since `ac9897c` is the #1384 `forward_model` field; zero numeric movement).

jasonqinzhou:
Exact-head follow-up on `9a310436c87c77f0140ece6f9a9be897ee51a661`:

The four behavioral blockers from my earlier review are fixed on the current head: typed TRT communication slices now miss instead of cross-substituting precision, SGLang context EPLB preserves its token correction, `is_gated` reaches the overflow roofline, and the per-call quant override drives both kernel resolution and lookup.

I see three remaining closeout items for PR 1:

1. **Harden the new-schema numeric boundary.** `_require_latency()` rejects null/empty cells but still accepts `NaN` and `±Inf`; `_row_power()` treats `NaN` as unmeasured but still accepts `±Inf`. Required latency should reject every non-finite value with the named corrupt-data error. Optional power should preserve the documented NaN-as-unmeasured behavior while refusing infinite values. Please cover both `moe_a2a` and `moe_expert_compute` loaders.
2. **Finish the public rename cleanup.** Rename `test_ep_moe_op.py` to `test_moe_expert_compute_op.py`, move the mid-file `__main__` pytest block to the end, and refresh the PR title/body so the public surface says `MoEExpertCompute` / `moe_expert_compute_perf` rather than the pre-rename `EPMoE` / `moe_ep_perf` wording.
3. **Rebase onto current `main`.** GitHub currently reports `DIRTY/CONFLICTING`; after resolving the conflict, the focused loader/equivalence tests, core API contract, parity, and regression gates need to run on the new exact head.

I can prepare and push this small follow-up plus the rebase directly to `feat/large-ep-a2a-sdk`. Once the new head is conflict-free and required CI is green, I’ll re-review the exact head, approve, and merge this PR before moving to PR 2.

jasonqinzhou:
Closeout update for exact head `d7132114c1afb740e40a2942984da76ff072a0cd`:

- Updated the branch to current `main@90f820f6ebdbfd810cd2369938ab951841635833` without rewriting contributor history (`96f37dc2`). The only textual conflict was the OpSpec exemption list; native FPM remains covered by current main, while this PR retains the two new large-EP exemptions.
- Hardened both unified loaders: required latency rejects null/NaN/+Inf/-Inf; optional power treats null/NaN as unmeasured and rejects measured infinities.
- Renamed `test_ep_moe_op.py` to `test_moe_expert_compute_op.py` and moved the equivalence module's `__main__` block to the end.
- Reconciled DeepEP-HT phase queries with current main's nonlinear joint-log Grid frontier hold: the summed dispatch+combine curve is resolved first, then the result is apportioned by phase share. This restored exact old-vs-new parity at the overflow probes.
- Refreshed the PR title/body to use the public `MoEExpertCompute` / `moe_expert_compute` terminology and current validation evidence.

Local validation:

- Ruff check + format: clean
- Focused loader/op/API/equivalence matrix: 138 passed
- Repository unit gate: 3,479 passed, 10 skipped, 749 deselected (`test_moe_dispatch.py` excluded per native-extension guidance)
- DCO: both closeout commits signed off; remote DCO check is green

Fresh exact-head CI is running. Once required checks are green and the head remains unchanged, I will replace my stale change request with an approval and merge.

jasonqinzhou (approved):
Approved at exact head d7132114c1afb740e40a2942984da76ff072a0cd after current-main integration, 0 unresolved review threads, all exact-head CI checks green, 3,479 local unit tests passing, and shipped-data all-to-all/expert-compute equivalence restored.

PR #1444, opened 2026-07-30
fix(sdk): detect block FP8 compressed-tensors weights

jasonqinzhou:
## Summary

- classify compressed-tensors FP8 weights with strategy: block or a non-empty block_structure as fp8_block
- preserve architecture-based support for the RedHatAI DeepSeek-V4 variant without registering it as a default model
- add parser and mixed-quantization regression coverage

## Root cause

The compressed-tensors parser only inspected weight bit width and type. The RedHatAI checkpoint attention weights are 8-bit floating point with strategy: block and block_structure: [128, 128], but AIConfigurator resolved them as plain fp8.

That made the DeepSeek-V4 attention lookup request (FMHA=bfloat16, KV=fp8, GEMM=fp8). The shipped B200 and H200 SGLang 0.5.14 data contains the corresponding fp8_block GEMM rows, so the plain-FP8 lookup produced no configurations.

## Impact

The checkpoint now resolves attention GEMM to fp8_block while preserving its FP4 routed-expert mode. The exact B200 support-matrix workload produces ranked aggregate and disaggregated recommendations instead of failing every configuration.

This variant is intentionally not added to DefaultHFModels or PR_CASES, and no cached model config is added, so the daily generator will not create a new exact support-matrix row for it. It continues to use the existing architecture rows and loads its Hugging Face config on demand.

H200 SILICON now reaches a separate missing w4a16_mxfp4_cutlass MoE workload-distribution slice; H200 HYBRID succeeds. This PR intentionally does not relabel that separate silicon-data boundary.

## Validation

- Ruff check and format on changed Python files
- 89 parser and quant-inference unit tests passed
- wider affected-unit run: 378 passed; 2 pre-existing network-dependent Task tests were blocked by Hugging Face HTTP 429
- exact B200/SGLang 0.5.14 SILICON support-matrix command on final head: 9,410 aggregate and 73 disaggregated results, exit code 0
- H200/SGLang 0.5.14 HYBRID workload: 5,817 aggregate and 135 disaggregated results, exit code 0
- verified RedHatAI/DeepSeek-V4-Flash-NVFP4-FP8 is absent from the generated default-model population

## Tracking

- [NVBugs 6531573](https://nvbugspro.nvidia.com/bug/6531573)


tests/unit/sdk/test_utils.py
```diff
@@ -1139,6 +1139,20 @@ def test_fp8_base_algo(self):
...
 
+    def test_block_fp8_base_algo(self):
+        """RedHatAI DSV4 compressed-tensors weights use block-scaled FP8."""
+        from aiconfigurator.sdk.utils import parse_compressed_tensors_quant
+
+        cfg = self._make_quant_config(8, "float")
+        cfg["config_groups"]["group_0"]["weights"].update(
+            {
+                "strategy": "block",
+                "block_structure": [128, 128],
```
jasonqinzhou: **[P2] Cover the two independent block-FP8 signals separately.** The production branch accepts either `strategy == "block"` or a non-empty `block_structure`, but this test supplies both. It would remain green if the `or` regressed to `and`, leaving checkpoints that expose only one valid signal unprotected. Please parameterize this case, or add a second test, so strategy-only and `block_structure`-only inputs each resolve to `fp8_block`; the existing plain-FP8 test can remain the negative control.

tests/e2e/support_matrix/test_pr_support_matrix.py
```diff
@@ -33,6 +33,7 @@
 # Add cases only when the whole e2e job still fits that budget.
 PR_CASES: list[tuple[str, str, str]] = [
     ("nvidia/DeepSeek-V3.1-NVFP4", "b200_sxm", "trtllm"),
+    ("RedHatAI/DeepSeek-V4-Flash-NVFP4-FP8", "b200_sxm", "sglang"),
```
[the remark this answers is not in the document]
jasonqinzhou: The dedicated RedHat PR_CASES entry was intentionally removed in eb6b4b57 so this variant reuses existing architecture support instead of adding a new generated support-matrix row. Since the commented smoke case no longer exists, this finding is no longer applicable.

tests/unit/sdk/test_utils.py
jasonqinzhou: Fixed in eeea6efc. The parser regression is now parameterized so strategy-only and block_structure-only inputs are each required to resolve to fp8_block; the plain-FP8 negative control remains unchanged.

Arsene12358 (approved):

simone-chen (approved):

PR #1446, opened 2026-07-31
fix(cli): resolve SGLang DeepEP gate through the perf database

Harrilee:
## Symptom

DeepEP sweeps were silently skipped at the default backend version on every system that actually ships DeepEP data. On `main`, `h100_sxm` and `h200_sxm` — the only two systems carrying `wideep_deepep_normal_perf.parquet` / `wideep_deepep_ll_perf.parquet` — both report:

```
missing required DeepEP perf data:
  .../systems/data/h200_sxm/sglang/0.5.14/wideep_deepep_normal_perf.parquet,
  .../systems/data/h200_sxm/sglang/0.5.14/wideep_deepep_ll_perf.parquet
```

The sweeps were reachable only by passing `--backend-version 0.5.6.post2` explicitly. The path named in that message is itself the tell: it exists in neither tree layout.

## Mechanism

`_sglang_deepep_perf_data_skip_reason` decided availability with `os.path.isfile(resolve_op_data_path(...))` — is there a raw file for this table at the resolved version? The latest declared sglang bucket is `0.5.14`, while the DeepEP tables live at `0.5.6.post2` and are reached through the perf database's nearest-earlier same-backend fallback (design §6.2). The loader finds those rows; a raw-file probe at `0.5.14` cannot see them. The gate and the loader were answering the same question through two different implementations, and they disagreed.

## Fix

Decide availability with the perf database's own source resolution instead of a second, weaker rule.

- The reuse channels of `PerfDatabase._build_op_sources` (declared donors, nearest-earlier same-backend siblings, kernel-identity-gated cross-backend fill) move verbatim to a module-level `_reuse_op_source_records`. `_build_op_sources` now calls it and keeps its existing behaviour, including the low-fidelity warning and the `data_provenance` record shape.
- New `resolve_op_data_sources` answers "which files would a query for this table at this version actually read?" without instantiating — and therefore fully loading — a `PerfDatabase`. It returns existing paths in load-priority order.
- The CLI gate calls that. Gate and loader now share one implementation and cannot drift apart.

The comm-family exclusion asymmetry documented on `_op_file_family_from_path` is untouched: it is still decided from the primary path only, and `test_legacy_layout_comm_op_keeps_pre_v3_siblings` is unmodified. That asymmetry is load-bearing here — when the requested version carries no DeepEP table the primary resolves to a legacy-shaped path, family reads as `None`, and sibling fallback proceeds. The new `test_resolve_op_data_sources_honors_family_layout_comm_exclusion` pins the other side, so a version that *does* carry the table still gets primary-only.

## Effect

Measured on the committed tree, `_sglang_deepep_perf_data_skip_reason(system, None, None)`:

| system | latest sglang | before | after |
| --- | --- | --- | --- |
| `h100_sxm` | 0.5.14 | skipped | **runs** (inherits from 0.5.6.post2) |
| `h200_sxm` | 0.5.14 | skipped | **runs** (inherits from 0.5.6.post2) |
| `gb200`, `gb300`, `b200_sxm`, `b300_sxm`, `l40s`, `rtx_pro_6000_server`, `a100_sxm` | — | skipped | skipped (no DeepEP table at any version) |

Systems with no DeepEP data anywhere still skip, and the reason now names the missing table logically (`<system>/sglang/<version>/<table>`) rather than printing an absolute path that does not exist. When only one of the two tables is missing, only that one is named.

## Why now

We are landing new DeepEP LL data at a new `0.5.12` bucket whose `normal` table stays at `0.5.10` and is inherited rather than duplicated. Under the old raw-file gate, `--backend-version 0.5.12` would have skipped the DeepEP sweep entirely even with a perfect LL table, so this is on the critical path for that data rather than a cleanup. Verified against a synthetic tree of exactly that shape: LL at `comm/sglang/0.5.12/`, `normal` only at `0.5.10` — skipped before, runs after.

## Verification

- **Behaviour preservation of the extraction:** 9,000 randomized synthetic-tree cases comparing `_build_op_sources` on this branch against `origin/main`'s, asserting identical `(path, kernel_source_filter)` lists *and* identical `data_provenance` records. Zero divergence, with coverage across all four channels (8,684 primary / 137 declared_reuse / 1,932 fallback / 6,631 cross_backend) and all 17,384 provenance rows keeping exactly the original key set.
- **Gate/loader agreement:** 12,000 cases asserting `resolve_op_data_sources` returns exactly the sources `_build_op_sources` admits and finds on disk, in the same order, across both tree layouts, partial markers, unparseable versions and `shared_layer` on/off.
- `ruff check .` and `ruff format --check .` clean; `pytest -m unit` 2786 passed, 10 skipped, 0 failures.
- `CODEOWNERS` needs no regeneration: strict coverage 2722/2722 with 0 catch-all-only, and regeneration is byte-identical to the committed artifact.

One scoping note for reviewers: with multiple `--systems-paths`, `resolve_op_data_sources` answers from the first root holding any source for the op, whereas `get_database` selects the first root holding the requested version dir. These coincide for the single-root default; the behaviour is documented on the function.

## Summary by CodeRabbit

* **New Features**
  * Added a public SDK utility for discovering all performance-data sources available to an operation, including shared, inherited, fallback, and cross-backend data.
  * Added an option to limit discovery to the primary data source.

* **Improvements**
  * Performance data can now be reused across compatible versions and backends more consistently.
  * DeepEP availability checks now recognize inherited performance data and provide clearer messages when required tables are unavailable.
  * Improved handling of communication performance data and provenance information.

### DIFF
--- a/aic-core/src/aiconfigurator_core/sdk/perf_database.py
+++ b/aic-core/src/aiconfigurator_core/sdk/perf_database.py
@@ -1888,6 +1888,231 @@ def _op_file_family_from_path(primary_path: str, system_data_root: str) -> str |
     return None
 
 
+def _reuse_op_source_records(
+    op_filename_enum: PerfDataFilename,
+    primary_path: str,
+    system_data_root: str,
+    *,
+    backend: str,
+    version: str,
+    enable_shared_layer: bool,
+    strict_provenance: bool,
+    kernel_source_manifest_entries: dict[str, tuple[dict, ...]],
+) -> list[dict[str, object]]:
+    """Reuse-channel sources (design §6.2-§6.4) for one op file, in priority order.
+
+    This is everything ``PerfDatabase._build_op_sources`` admits AFTER the
+    active backend/version's own primary file: declared donors, nearest-earlier
+    same-backend siblings, then kernel-identity-gated cross-backend fill. Only
+    sources whose file exists are returned; each record carries ``version``,
+    ``path``, ``channel`` and ``ks_filter``, plus ``low_fidelity`` on
+    cross-backend records so the caller can warn about them.
+
+    Module-level (rather than a method) so the CLI can ask "would the database
+    find rows for this op at this version?" via ``resolve_op_data_sources``
+    without instantiating -- and therefore fully loading -- a ``PerfDatabase``.
+    Both callers share this one implementation so the ordering and the
+    exclusions below cannot drift apart.
+    """
+    if not enable_shared_layer:
+        return []
+    if op_filename_enum in (PerfDataFilename.nccl, PerfDataFilename.oneccl):
+        return []
+    if _op_file_family_from_path(primary_path, system_data_root) == _COMM_FAMILY_DIR:
+        return []
+
+    op_file_basename = op_filename_enum.value
+    backend_lower = backend.lower()
+    records: list[dict[str, object]] = []
+
+    # Channel 2 (design §6.3): declared donors from the REQUESTED version
+    # dir's reuse.yaml, in file order. Duplicate (table, from_version)
+    # entries in one reuse.yaml would otherwise admit the same source
+    # twice -- table is fixed to op_file_basename for this call, so
+    # `declared_donor_versions` membership alone is the (table,
+    # from_version) dedupe key; first occurrence wins (AIC-1503 PR4
+    # task 5, FIX 2).
+    declared_donor_versions: set[str] = set()
+    for reuse_entry in _requested_version_reuse_entries(
+        system_data_root, backend_lower, version, op_file_basename, strict=strict_provenance
+    ):
+        from_version = reuse_entry["from_version"]
+        if from_version in declared_donor_versions:
+            logger.debug(
+                "Duplicate declared-reuse entry for table %s from_version %s under %s; first occurrence wins.",
+                reuse_entry["table"],
+                from_version,
+                system_data_root,
+            )
+            continue
+        donor_path = resolve_op_data_path(system_data_root, backend_lower, from_version, op_file_basename)
+        if not os.path.isfile(donor_path):
+            continue
+        records.append(
+            {
+                "version": from_version,
+                "path": donor_path,
+                "channel": "declared_reuse",
+                "ks_filter": None,
+            }
+        )
+        declared_donor_versions.add(from_version)
+
+    # Channel 1 aka §6.2 (nearest-earlier same-backend fallback). Unparseable
+    # sibling versions can't be ordered against the requested version, so
+    # they're excluded here (logged once) — an explicit declaration still
+    # works for them. Versions already admitted as declared donors above
+    # are excluded too — the dominant real reuse.yaml pattern points
+    # BACKWARD at an earlier sibling, and without this exclusion that same
+    # physical source would be listed twice (channels declared_reuse AND
+    # fallback), doubling I/O and duplicating data_provenance rows.
+    requested_parsed = parse_support_matrix_version(version)
+    if requested_parsed is not None:
+        sibling_versions = {v for v, _ in _iter_backend_version_dirs(system_data_root, backend_lower)}
+        sibling_versions.discard(version)
+        sibling_versions -= declared_donor_versions
+        earlier_versions = []
+        for sibling_version in sibling_versions:
+            parsed = parse_support_matrix_version(sibling_version)
+            if parsed is None:
+                _warn_unparseable_sibling_version_once(system_data_root, backend_lower, sibling_version)
+                continue
+            if parsed >= requested_parsed:
+                continue  # Never admit newer-than-requested implicitly.
+            earlier_versions.append((parsed, sibling_version))
+        earlier_versions.sort(key=lambda item: item[0], reverse=True)  # nearest-earlier first
+        for _, sibling_version in earlier_versions:
+            sibling_path = resolve_op_data_path(system_data_root, backend_lower, sibling_version, op_file_basename)
+            if not os.path.isfile(sibling_path):
+                continue
+            records.append({"version": sibling_version, "path": sibling_path, "channel": "fallback", "ks_filter": None})
+
+    # Channel 4 aka §6.4 (cross-backend fill, kernel-identity gated). Same
+    # mechanism as before; the active backend is excluded from
+    # `ordered_frameworks` because channels 2-3 above already cover it.
+    per_framework_filter: dict[str, set[str]] = defaultdict(set)
+    per_framework_fallback: dict[str, set[str]] = defaultdict(set)
+    for entry in kernel_source_manifest_entries.get(op_file_basename, ()):
+        frameworks_lower = {fw.lower() for fw in entry.get("frameworks") or []}
+        if backend_lower not in frameworks_lower:
+            continue  # Active backend isn't listed as a consumer of this kernel_source.
+        ks = entry.get("kernel_source")
+        if not ks:
+            continue
+        tier = entry.get("tier")
+        if tier in ("shared", "shared_fallback"):
+            for fw in frameworks_lower:
+                per_framework_filter[fw].add(ks)
+            if tier == "shared_fallback":
+                for fw in frameworks_lower:
+                    per_framework_fallback[fw].add(ks)
+
+    ordered_frameworks = sorted(set(per_framework_filter) - {backend_lower})
+
+    # Sort key for newest-first ordering. Parseable PEP 440 versions form one
+    # group and always rank above unparseable strings — guarantees `1.10.0`
+    # beats `1.2.0` regardless of the lexicographic accident.
+    def _newest_first(candidate_version: str) -> tuple:
+        parsed = parse_support_matrix_version(candidate_version)
+        return (1, parsed) if parsed is not None else (0, candidate_version)
+
+    for framework in ordered_frameworks:
+        ks_filter = per_framework_filter[framework]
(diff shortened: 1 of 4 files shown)


aic-core/src/aiconfigurator_core/sdk/perf_database.py
```diff
@@ -1888,6 +1888,231 @@ def _op_file_family_from_path(primary_path: str, system_data_root: str) -> str |
...
+    find rows for this op at this version?" via ``resolve_op_data_sources``
+    without instantiating -- and therefore fully loading -- a ``PerfDatabase``.
+    Both callers share this one implementation so the ordering and the
+    exclusions below cannot drift apart.
+    """
+    if not enable_shared_layer:
+        return []
+    if op_filename_enum in (PerfDataFilename.nccl, PerfDataFilename.oneccl):
+        return []
+    if _op_file_family_from_path(primary_path, system_data_root) == _COMM_FAMILY_DIR:
```
jasonqinzhou, line 1921: **[P1] Make DeepEP reuse an explicit comm-family policy.** This guard is driven by `primary_path`, but when the requested comm table is absent, `resolve_op_data_path()` returns a legacy-shaped *missing* path. `_op_file_family_from_path()` then returns `None`, so current `h100_sxm/sglang/0.5.14` admits `comm/sglang/0.5.6.post2` as a fallback. That contradicts Collector V3 design §6.5 and this method's own contract that `comm` is excluded from sibling-version reuse entirely. The earlier comment's "zero live exposure" assumption is disproved here: the committed tree is family-first, yet absence of one requested file triggers the exception. Because this PR turns that loophole into the default CLI gate, it can silently optimize with comm curves from a different SGLang version. Please either identify the family independently of file existence and keep comm primary-only, or land a narrow, explicit DeepEP reuse policy with compatibility evidence, provenance/audit coverage, and tests; do not pin the behavior to a missing path looking legacy.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
```diff
@@ -1888,6 +1888,231 @@ def _op_file_family_from_path(primary_path: str, system_data_root: str) -> str |
...
+                primary_path,
+                system_data_root,
+                backend=backend,
+                version=version,
+                enable_shared_layer=shared_layer,
+                strict_provenance=effective_strict,
+                kernel_source_manifest_entries=_load_op_kernel_source_manifest_entries(systems_root),
+            )
+        )
+        if paths:
```
jasonqinzhou, line 2111: **[P2] Select the same declared root/version as `get_database()`.** This returns from the first systems root that yields this individual op, not the first root that owns the requested `(backend, version)`. I reproduced root A declaring `sglang/0.5.14` without DeepEP and root B carrying an older donor: `get_database()` selected A while this helper returned B; separate normal/LL calls can even mix roots. The same gap lets `h100_sxm/sglang/999.0` return four GEMM source files although `get_database()` rejects that undeclared version. This recreates the gate/loader drift and violates the documented `--systems-paths` first-match contract. Please apply the same declared, non-partial requested-version/root admission as `get_database()`, then resolve all required tables only within that selected root, with undeclared-version and two-root regressions.

src/aiconfigurator/cli/main.py
```diff
@@ -1324,19 +1321,18 @@ def _sglang_deepep_perf_data_skip_reason(
             continue
 
         for filename in _SGLANG_DEEPEP_REQUIRED_FILES:
-            resolved_path = _get_backend_data_path(
+            sources = perf_database.resolve_op_data_sources(
```
jasonqinzhou, line 1324: **[P2] Keep formula-only modes from scheduling inherited DeepEP sweeps.** `resolve_op_data_sources()` defaults `shared_layer=True`, but this gate has no `database_mode` and both call sites invoke it for every mode. On the committed h100 data, `build_default_tasks(..., database_mode="EMPIRICAL")` now adds `agg_deepep` and `disagg_deepep` at the default 0.5.14 version, while the actual database has `enable_shared_layer=False` and a direct DeepEP EMPIRICAL query raises `NotImplementedError` (the SOL path is also unimplemented). These optional experiments are therefore known failures. Thread the database/shared-layer policy into the gate and skip modes without a DeepEP formula implementation; add SOL and EMPIRICAL task-set regressions.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
```diff
@@ -1888,6 +1888,231 @@ def _op_file_family_from_path(primary_path: str, system_data_root: str) -> str |
...
+    list when no root has it. Raises ``ValueError`` for an op filename that is
+    not a known ``PerfDataFilename``.
+    """
+    op_filename_enum = op_filename if isinstance(op_filename, PerfDataFilename) else PerfDataFilename(str(op_filename))
+    if systems_paths is None:
+        systems_paths = get_systems_paths()
+    elif isinstance(systems_paths, str):
+        systems_paths = [systems_paths]
+
+    effective_strict = _strict_provenance_enabled(strict_provenance)
```
jasonqinzhou, line 2078: **[P2] Enforce strict provenance before reporting sources.** `effective_strict` currently reaches partial-directory and malformed-`reuse.yaml` handling, but this helper never runs `_check_strict_provenance_for_request()`. In a synthetic root with a primary parquet and no `collection_meta.yaml`, `resolve_op_data_sources(..., strict_provenance=True)` returned the table path while strict `get_database()` rejected the exact request with the expected missing-sidecar `ValueError`. That breaks the new public helper's advertised loader parity and its fail-closed option. After selecting the requested root/version, run the same request-scoped provenance validation, including declared donors, and add missing/uncovered-sidecar parity tests.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.2/10 · high confidence**

## What this PR does

This PR fixes the CLI's SGLang DeepEP availability gate so a latest backend bucket can discover DeepEP tables inherited from earlier performance-data versions. It extracts the PerfDatabase reuse channels into a shared resolver, calls that resolver from the CLI gate, and adds source-ordering and inherited-table tests.

The author did unusually careful work to consolidate source ordering into one implementation, preserve provenance shape, and validate the extraction with both focused tests and extensive randomized parity checks; the updated head also cleanly incorporates current main without altering the reviewed patch.

## Why this score

Score 6.2/10 with REQUEST_CHANGES. The head update is only a clean merge of current main and leaves the effective four-file PR patch unchanged, so all four previously reported issues remain. Most importantly, the default DeepEP gate still relies on an accidental comm-family reuse escape hatch; the resolver also still diverges from get_database for declared-version/root selection and strict provenance, while formula-only modes still schedule known-failing inherited DeepEP experiments.

tianhaox (commented):
Reviewed with the diff side-by-side against `main`'s `_build_op_sources`, plus `resolve_op_data_path`'s family/legacy fallback, the manifest cache, the compat shim re-export, and the op-loader call graph. The fix is correct and the extraction is faithful — LGTM overall; everything below is non-blocking.

**What I verified**

- Channels 2/1(§6.2)/4 in `_reuse_op_source_records` are a line-for-line move: dedupe keys, unparseable-sibling warn-once, `_newest_first` ordering all preserved. The only semantic delta is the low-fidelity warning moving from inline to a `low_fidelity` marker popped by `_build_op_sources` — behavior-equivalent.
- Primary semantics are self-consistent on both sides: `_build_op_sources` still admits a nonexistent primary into provenance (`exists: False`), while `resolve_op_data_sources` returns existing files only — which is exactly its documented contract, and the ordering test compares against `exists`-filtered provenance. Partial-dir handling (refuse primary, reuse channels still fill) matches.
- The load-bearing comm asymmetry holds: `resolve_op_data_path` only returns a family-shaped path when the file exists, so "no table at requested version → legacy-shaped primary → family `None` → fallback proceeds" is real, and both directions are now pinned by tests.
- Gate/loader agreement covers every query path: DeepEP loading (`moe.py`) and all other ops go through `database._build_op_sources`, including the Rust query engine (data loading stays Python-side).
- No perf concern: `_load_op_kernel_source_manifest_entries` is `functools.cache`d; the gate makes at most 2 systems × 2 files calls.

**Non-blocking observations**

1. *"Cannot drift apart" is slightly overstated.* Only channels 2–4 are shared; the primary admission logic (isfile + partial gate) is still two hand-written copies — one in `_build_op_sources`, one inline in `resolve_op_data_sources`. The 12k-case equivalence run in the PR description was a one-off harness, and the committed tests don't exercise "primary exists but partial → resolve excludes the primary yet still returns reuse sources". Suggest one small test pinning that branch, since it's the only duplicated (rather than shared) piece.
2. *`shared_layer`/`database_mode` isn't threaded into the gate.* `resolve_op_data_sources` defaults `shared_layer=True`, while formula-only database modes load with the shared layer off (`_shared_layer_enabled`). `build_default_tasks` has `database_mode` in scope at both call sites. Harmless for the SILICON default, and formula modes likely never consume these tables — but worth either threading it through or noting the assumption in the gate's docstring.
3. The multi-root scoping divergence vs `get_database` (first root with *any* source vs first root with the requested version dir) is honestly disclosed in the docstring and PR description; fine for the single-root default.

Also flagging for the approver: this intentionally spans aic-core SDK (new public API) and its CLI consumer — that coupling is the fix itself (the gate must share the loader's implementation), not a drive-by.

The skip-message change from a nonexistent absolute path to the logical `<system>/sglang/<version>/<table>` form is a real usability win, and the PR description (symptom/mechanism/effect table/why-now) sets a high bar.

PR #1447, opened 2026-07-31
feat(models): add StepFun Step-3.7-Flash (hybrid SWA/global + shared-expert MoE + MTP)

chanh:
Adds the StepFun **Step-3.7-Flash** decoder to AIConfigurator: a 1:3 global:sliding-window attention pattern, dense-first-K then MoE FFN with one always-on shared expert per MoE layer, and MTP (`num_nextn_predict_layers`).

It reuses the existing `HybridMoEModel` operation pipeline and adds the Step-specific behavior that pipeline did not previously model:

- **Authoritative checkpoint schema** — recognizes `Step3p7ForConditionalGeneration` and nested `Step3p5ForCausalLM`, unwraps `text_config`, normalizes the vendor MoE/attention fields, and trims MTP-only `layer_types`.
- **Hybrid attention geometry** — models 96 query heads for sliding attention versus 64 for global attention, per-head Q/K normalization, and the head-wise attention gate.
- **Window-capped KV growth** — sliding-attention layers cap at their configured window while global layers keep growing.
- **Shared expert and context parallelism** — appends the always-on dense SwiGLU shared expert and applies the same CP audit/splitting as the inherited context pipeline.
- **Faithful quantization** — keeps the registered `stepfun-ai/Step-3.7-Flash` base checkpoint in BF16; the separate FP8 artifact is not conflated with it.
- **Backend memory accounting** — registers `STEP3P7` for MoE activation coefficients and dispatch workspace in TRT-LLM, vLLM, and SGLang, using the residual hidden width for the hybrid model's workspace.

**No new performance database** — MoE and attention costs use the existing perf DB through HYBRID-mode interpolation because the exact 288-expert MoE shape has no measured silicon slice.

### What's included

- `models/step3p7.py` — `@register_model("STEP3P7") Step3p7Model(HybridMoEModel)`.
- Authoritative and compatibility architecture mappings plus nested-text-config handling.
- Cached BF16 Hugging Face config, default-model registration, frontend compatibility alias, and import-contract leaf entry.
- Step-specific hybrid attention, shared-expert, CP, KV-cache, parser, public-factory, and backend-memory regression coverage.

### Verification

- `uv run ruff check .`
- `uv run ruff format --check .` — 540 files already formatted
- `uv run pytest tests/unit/sdk/backends tests/unit/sdk/models/test_step3p7.py -q -p no:timeout` — 122 passed
- `uv run pytest -m unit -q -p no:timeout --ignore=tests/unit/sdk/database/test_moe_dispatch.py` — 3,249 passed, 10 skipped (the ignored test requires the optional Torch dependency)

The published checkpoint parses to 45 decoder layers (33 sliding / 12 global), 42 MoE / 3 dense layers, 288 experts with top-8 routing, and three MTP layers.

## Summary by CodeRabbit

* **New Features**
  * Added support for the Step-3.7-Flash model.
  * Added model configuration, architecture recognition, and registration.
  * Added hybrid attention, mixture-of-experts, shared experts, and sliding-window KV-cache support.
  * Added support for nested model settings and separate sliding-attention head counts.
  * Added Step-3.7-Flash to the default model catalog.
  * Added context-parallel support for shared-expert operations.
  * Added compatibility across supported SDK packages.

* **Tests**
  * Added coverage for model mapping, configuration parsing, validation, attention settings, shared experts, context parallelism, KV-cache calculations, and backend memory accounting.


tianhaox:
Reviewed the model definition against the `HybridMoEModel`/`Gemma4MixModel` base code. The overall structure is sound and the shipped unit tests check the right invariants, but there are three confirmed correctness issues to fix before merge, plus a few structural suggestions.

## Blocking

**1. Shared-expert ops bypass the CP audit and are mis-scaled under `cp_size > 1`** — `step3p7.py::_append_shared_expert_ops`

`_append_shared_expert_ops()` extends `context_ops` **after** `super().set_hybrid_config()` has already run `_build_context_ops()`, which is where the CP pass assigns `op._seq_split = cp` to every `_CP_AWARE` op and raises `NotImplementedError` for un-audited ones (`hybrid_moe.py:361-379`). `Step3p7Model` inherits `supports_cp("sglang") == True` (`hybrid_moe.py:29-31`) and the model factory gate (`models/__init__.py:96-104`) has no STEP3P7 exception, so sglang + `cp_size=2` constructs fine — and the 42 layers of shared-expert GEMMs are then modeled at the **full** prefill token count (GEMMs divide tokens by `_seq_split`, see `gemm.py:750`), ~2x over-counted at cp=2, silently. The docstring's "cp_size==1 for the Step-3.7 sweeps here" is an assumption, not an enforced constraint. Minimal fix: override `supports_cp` to return `False` until the shared-expert ops are CP-audited; better fix: emit them inside the audited pipeline.

**2. The cached config for the plain (BF16) model name carries a fabricated `quantization_config`** — `model_configs/stepfun-ai--Step-3.7-Flash_config.json`

Checked against upstream: `huggingface.co/stepfun-ai/Step-3.7-Flash/raw/main/config.json` has **no** `quantization_config` (plain bf16 checkpoint); the fp8 block lives only in the separate `stepfun-ai/Step-3.7-Flash-FP8` repo. The cached config here embeds the FP8 repo's quant block under the plain name, and that name is added to `DefaultHFModels`. Consequences: `utils.py:1070-1072` infers `quant_algo=fp8_block`, so `_infer_quant_modes_from_raw_config` (`helpers.py:239-241`) defaults the BF16 model to fp8_block GEMM/MoE modes (≈halved weight-memory estimate), and `HybridMoEModel._validate_fp8_block_quantized_moe_config` then rejects `moe_tp >= 4` (1280/4 % 128 != 0) for topologies that are perfectly valid in BF16. Fix: strip `quantization_config` from the plain-named config (repo convention — compare `Qwen--Qwen3-235B-A22B` vs `-FP8`, `zai-org--GLM-5` vs `-FP8`).

**3. STEP3P7 is missing from the backend memory tables** — `base_backend.py:55`, `trtllm_backend.py:46`, `sglang_backend.py:26`

`MOE_WORKSPACE_FAMILIES` and `ACTIVATION_COEFFICIENTS` both list `GEMMA4MIX` (the model this class explicitly mirrors) but not `STEP3P7`. An unlisted family falls through to the dense default coefficients (`base_backend.py:1616`; {1:10} vs the MoE-family {1:22} — 2.2x at TP1) and skips the MoE dispatch workspace entirely (`base_backend.py:1621`). For a 288-expert fp8-block MoE that's roughly 1–3 GiB/GPU of overstated free memory / KV budget → recommended `max_batch_size`/`max_num_tokens` risk OOM on real GPUs. Direct precedent: GEMMA4MOE needed exactly this fix inside its own introduction PR (#1126, b93de8cec "GEMMA4MOE was falling through to the default").

## Suggestions (non-blocking)

4. **Plumb `share_expert_dim` through `HybridMoEConfig` instead of a private attribute poked by `create()`.** The established pattern is a parsed config field (`Qwen35Config.shared_expert_inter_size`, `common.py:229`, consumed in `qwen35.py`). As shipped, any construction path that bypasses `create()` silently drops the shared expert (`_share_expert_dim` defaults to 0 and `_append_shared_expert_ops` early-returns) — the PR's own test has to hand-set `model._share_expert_dim = 1280` before `set_hybrid_config`, which is the tell. Also `n_shared_experts` is never read: a checkpoint with `n_shared_experts=2` would silently undercount the shared-expert FFN by 2x. A `shared_expert_inter_size` field on `HybridMoEConfig` (emitted by `HybridMoEModel` when > 0) would also make this available to future hybrid models declaratively.

5. **The window-capped KV trio belongs in `HybridMoEModel`, not a third leaf-class copy.** `get_kvcache_bytes_per_sequence`/`get_kvcache_max_tokens`/`_swa_global_counts` are near-verbatim `Gemma4MixModel` (`gemma4.py:453-499`). `HybridMoEModel` already carries `sliding_window_size` + `attn_layer_pattern` yet has no KV overrides — which means its existing users (MiMo-V2-Flash, Llama4 via `attention_chunk_size`) currently model window-aware attention *compute* but linear KV *memory*, over-counting SWA KV today. Implementing this once in the base class (geometry-aware like Gemma4's version, honoring `swa_num_kv_heads`/`swa_head_dim`) both removes the duplication and heals those models.

6. **`get_kvcache_elements_per_token` override is dead weight** — byte-equivalent to `BaseModel`'s GQA branch (`base.py:231-232`), including the ceil-division. Safe to delete.

7. Minor: `Step3p5FlashForCausalLM` is mapped in `ARCHITECTURE_TO_MODEL_FAMILY` with no cached config, test, or schema verification — the parse branch hard-requires Step-3.7-style `layer_types`, so a real Step-3.5 config either hard-fails or silently mis-parses; consider dropping the mapping until a real config is validated (the PR body's caution about `DefaultHFModels` applies here too). Also worth a shared `_parse_layer_types(...)` helper in `utils.py` (this branch is the third copy of the same validation, after Gemma4 and Qwen3.5), a `sliding_window > 0` sanity check when `sliding_attention` layers are present (a `null` window currently degrades silently to all-global, ~3.4x KV over-estimate at 16K), and syncing the Rust `ModelFamily` reference mirror (`enums.rs:322-338` — no runtime consumer, so purely hygiene).

Verification notes: findings 1–3 were each traced through the base code (quotes above); the 16K KV-ratio test math (33·512 + 12·16384 token-layers → 0.2896) checks out.

aic-core/src/aiconfigurator_core/sdk/common.py
```diff
@@ -614,6 +617,8 @@ def get_default_models() -> set[str]:
     "MiniMaxM2ForCausalLM": "MOE",
     "MiniMaxM3ForCausalLM": "MINIMAXM3",
     "MiniMaxM3SparseForConditionalGeneration": "MINIMAXM3",
+    "Step3p7FlashForCausalLM": "STEP3P7",
```
jasonqinzhou, line 706: **[P1] Parse the authoritative Step-3.7 checkpoint schema.** These architecture strings do not exist in the target checkpoint. The published `stepfun-ai/Step-3.7-Flash` config uses top-level `Step3p7ForConditionalGeneration` with a nested `text_config` architecture of `Step3p5ForCausalLM`; the current parser rejects that config before reaching this family. The exact default ID only appears to work because its bundled file rewrites the schema, so a local copy of the same checkpoint and the real Step-3.5/FP8 variants fail. The authoritative text config also uses `num_attention_groups`, `moe_num_experts`, `moe_top_k`, and `moe_layers_enum`, and carries 48 `layer_types` entries for 45 decoder plus 3 MTP layers. Please map and unwrap the real schema, normalize those fields (including trimming MTP-only entries where appropriate), cache the authoritative config, and add a parser/public-factory regression based on [the immutable upstream config](https://huggingface.co/stepfun-ai/Step-3.7-Flash/blob/5f6244077ac62e04eec3f320501ff8c2b293373a/config.json).

aic-core/src/aiconfigurator_core/model_configs/stepfun-ai--Step-3.7-Flash_config.json
```diff
@@ -0,0 +1,79 @@
...
+  ],
+  "first_k_dense_replace": 3,
+  "num_experts": 288,
+  "num_experts_per_tok": 8,
+  "moe_intermediate_size": 1280,
+  "share_expert_dim": 1280,
+  "n_shared_experts": 1,
+  "scoring_func": "sigmoid",
+  "num_nextn_predict_layers": 3,
+  "quantization_config": {
```
jasonqinzhou: **[P1] Do not model the BF16 checkpoint as FP8-block.** The registered base ID's [authoritative config](https://huggingface.co/stepfun-ai/Step-3.7-Flash/blob/5f6244077ac62e04eec3f320501ff8c2b293373a/config.json) declares BF16 and has no `quantization_config`; FP8 is published as the separate [`Step-3.7-Flash-FP8`](https://huggingface.co/stepfun-ai/Step-3.7-Flash-FP8/blob/b3d7916fccac844cca050d7520f2aaa513f9a84f/config.json) artifact. This injected block makes public `get_model()` resolve the base ID to FP8-block GEMM/MoE and FP8 KV. At TP8, the current context operation list accounts for 23.868 GiB of weights versus 46.537 GiB with explicit BF16, materially changing fit, capacity, and latency predictions. Please keep this cache faithful to the BF16 artifact, or register/cache the actual FP8 ID with its authoritative exclusion metadata, and add a factory-path quant-mode assertion.

aic-core/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -756,6 +756,34 @@ def _parse_hf_config_json(config: dict) -> dict:
...
+            raise ValueError(f"Step3p7 layer_types length {len(layer_types_raw)} != num_hidden_layers {layers}")
+        if any(lt not in ("sliding_attention", "full_attention") for lt in layer_types_raw):
+            raise ValueError("Step3p7 layer_types must contain only 'sliding_attention' or 'full_attention'")
+        attn_pattern = tuple(1 if lt == "full_attention" else 0 for lt in layer_types_raw)
+        first_k_dense = int(config.get("first_k_dense_replace", 0) or 0)
+        moe_freq = tuple(0 if i < first_k_dense else 1 for i in range(layers))
+        extra_params = HybridMoEConfig(
+            attn_layer_pattern=attn_pattern,
+            moe_layer_freq=moe_freq,
+            # SWA and global layers share head geometry -> 0 = fall back to defaults.
```
jasonqinzhou: **[P1] Preserve Step's per-attention-type query-head geometry.** The authoritative text config uses 64 query heads for full attention but switches sliding layers to 96 through `attention_other_setting` (KV groups and head size remain 8 and 128), and [the model implementation applies that switch](https://huggingface.co/stepfun-ai/Step-3.7-Flash/blob/5f6244077ac62e04eec3f320501ff8c2b293373a/modeling_step3p7.py). `HybridMoEModel` instead sizes both groups from the global 64-head value. At TP8 the current SWA QKV width, projection input, and FMHA head count are 1280, 1024, and 8; they should be 1792, 1536, and 12. Because this affects 33 of 45 layers, weights and latency are materially undercounted. Please carry an SWA query-head count through parsing/model construction and assert the exact per-type operation dimensions from a faithful checkpoint fixture.

aic-core/src/aiconfigurator_core/sdk/models/step3p7.py
```diff
@@ -0,0 +1,141 @@
...
+        super().__init__(*args, **kwargs)
+        # Set by create() before set_hybrid_config(); default 0 = no shared expert.
+        self._share_expert_dim: int = 0
+
+    # ------------------------------------------------------------------ #
+    # Shared expert: one always-on dense SwiGLU per MoE layer.
+    # ------------------------------------------------------------------ #
+    def set_hybrid_config(self, cfg) -> None:
+        super().set_hybrid_config(cfg)
+        self._append_shared_expert_ops()
```
jasonqinzhou, line 70: **[P2] Apply CP wiring after adding the shared expert.** `HybridMoEModel.set_hybrid_config()` completes the SGLang CP audit and applies `seq_split=cp` before `_append_shared_expert_ops()` runs. A public `get_model(..., cp_size=2, backend_name='sglang')` probe succeeds, but inherited dense FFN operations have `_seq_split=2` while all `context_shared_expert_*` operations remain at 1, charging the full shared-expert token workload on every CP rank. The comment's `cp_size==1` assumption does not constrain this class because it inherits SGLang CP support. Please add these operations before CP wiring, apply the same audit/split explicitly, or reject CP for this subclass, and add a factory-path CP regression.

aic-core/src/aiconfigurator_core/sdk/models/step3p7.py
```diff
@@ -0,0 +1,141 @@
...
+"""
+
+from __future__ import annotations
+
+from aiconfigurator_core.sdk.models.base import register_model
+from aiconfigurator_core.sdk.models.hybrid_moe import HybridMoEModel
+
+
+@register_model("STEP3P7")
+class Step3p7Model(HybridMoEModel):
```
jasonqinzhou, line 33: **[P2] Include the Step-specific Q/K normalization and attention gate.** [The authoritative attention block](https://huggingface.co/stepfun-ai/Step-3.7-Flash/blob/5f6244077ac62e04eec3f320501ff8c2b293373a/modeling_step3p7.py) applies per-head Q and K RMSNorm on every layer, and `use_head_wise_attn_gate=true` adds a `hidden_size -> num_heads` `g_proj` plus sigmoid/multiply before the output projection. `ContextAttention` and `GenerationAttention` already expose `use_qk_norm`, but the reused HybridMoE path does not enable it, and no gate operations are emitted. Please carry these Step flags into the operation graph and add focused weight/latency-shape assertions.

jasonqinzhou (changes requested):

## What this PR does

This PR adds StepFun Step-3.7-Flash decoder support by registering a STEP3P7 model family, parsing its hybrid full/sliding attention and dense/MoE layer pattern, adding shared-expert operations, capping sliding-window KV growth, and preserving the legacy SDK import surface.

The implementation keeps the new family localized, mirrors the proven Gemma4 KV-cache inverse carefully, and adds clean coverage for registration, layer splitting, KV sizing, and cross-package import identity.

## Why this score

Score 4.2/10 with REQUEST_CHANGES. The structure and tests are thoughtful, but the registered public checkpoint is not modeled from its authoritative schema: real local/HF configs are rejected, the BF16 base ID is silently forced to FP8-block, and 33 sliding layers are built with 64 rather than 96 query heads. The shared-expert CP path and two Step-specific attention operations also remain unmodeled, so the current estimates are not reliable enough to merge.

chanh:
All five review points are addressed. Pushed through `df7cea2f`.

Everything below was validated against the **published `stepfun-ai/Step-3.7-Flash` config and `modeling_step3p7.py`**, not against the fixture this PR originally shipped. That distinction turned out to matter — the curated fixture agreed with the code, so it caught none of these.

**1. Architecture names.** The originals were invented. The real checkpoint is `Step3p7ForConditionalGeneration` wrapping a nested `text_config` whose architecture is `Step3p5ForCausalLM`. All four spellings now map to `STEP3P7`, and the decoder is unwrapped via `MULTIMODAL_TEXT_CONFIG_KEY` — the parser previously read `num_hidden_layers` from the top level, where it does not exist.

**2. Layer topology.** `layer_types` is sized over the decoder *plus* the `num_nextn_predict_layers` MTP layers (48 entries for 45 layers), so it is now trimmed. MoE placement comes from `moe_layers_enum`, not `first_k_dense_replace` — without it all 45 layers were modelled as MoE instead of 42. Now resolves to 33 sliding / 12 global, 42 MoE / 3 dense.

**3. SWA head count.** Sliding layers declare 96 query heads via `attention_other_setting` against 64 global. The first fix here was incomplete: it resized the QKV and out-projection GEMMs but left `ContextAttention`/`GenerationAttention` on `self._num_heads`, so FMHA — the dominant term for sliding layers — was still costed at 64 heads. Fixed at all four SWA sites.

**4. Shared-expert CP.** `_append_shared_expert_ops` runs after `_build_context_ops` has wired context parallelism, so those ops kept `_seq_split=1` while the inherited dense FFN carried `cp`, and they bypassed the un-audited-op guard — a wrong number rather than an error. The audit loop is now `HybridMoEModel.apply_cp_to_context_ops`, called from both places, with the guard semantics unchanged for the other families.

**5. Q/K norm and attention gate.** `Step3p7Attention` builds `q_norm`/`k_norm` unconditionally — there is no config flag, which is why none appears in the config. `use_qk_norm` was never reaching the ops because `BaseModel` only reads it when `extra_params` is a dict and the hybrid path passes a `HybridMoEConfig`. `use_head_wise_attn_gate` is set in the real config; `g_proj` is `hidden_size -> num_attention_heads`, sized off *that layer's* head count, so the gate is 96-wide on sliding layers and 64-wide on global ones. Emitted as a GEMM plus sigmoid-multiply `ElementWise` before the out-projection at all eight sites.

Both new flags default to off in `HybridMoEConfig`, since the base class is shared with MiMo-V2-Flash, Llama 4 and Gemma 4. There is a test asserting no gate ops appear for a config that doesn't request one.

Also fixed along the way: the published config spells the MoE width `moe_num_experts` / `moe_top_k`, neither of which was read, so `num_experts` resolved to 0 and constructing the model asserted out.

**Tests.** 3376 pass, lint clean. New coverage asserts on the op graph rather than the parse result — the CP and SWA tests both fail on the pre-fix code (`{1} == {2}` and `{64} == {96}`), which the parse-level assertions did not. Two `EngineSpec schema 6 vs 4` failures in `test_rust_engine_step.py` and `test_cli_api.py` are pre-existing and reproduce identically on an unmodified `main`.

chanh:
@jasonqinzhou this is ready for another look when you have a moment — all five points from your review are addressed in the comment above. (I don't have write access here, so I can't re-request review through the UI.)

aic-core/src/aiconfigurator_core/sdk/common.py
chanh: Fixed, and you were right on every specific. The architecture strings in the original diff did not exist — I had registered names the curated fixture invented, so real checkpoints were still rejected while the bundled file made the default ID appear to work.

Now: `Step3p7ForConditionalGeneration` and `Step3p5ForCausalLM` both map to `STEP3P7`, and the nested decoder is unwrapped via the existing `MULTIMODAL_TEXT_CONFIG_KEY` (the parser previously read `num_hidden_layers` from the top level, where it does not exist). `layer_types` is trimmed by `num_nextn_predict_layers` — 48 entries for 45 decoder + 3 MTP. MoE placement comes from `moe_layers_enum`; without it all 45 layers were modelled as MoE instead of 42. `moe_num_experts` / `moe_top_k` are read too — that one was not in your list, and without it `num_experts` resolved to 0 and constructing the model asserted out. `num_attention_groups` feeds the SWA KV head count.

Verified end to end against the published config: 33 sliding / 12 global, 42 MoE / 3 dense, 288 experts, top-8. Regression test `test_step3p7_parses_the_published_checkpoint_schema` uses the real nested schema rather than the curated one.

aic-core/src/aiconfigurator_core/model_configs/stepfun-ai--Step-3.7-Flash_config.json
chanh: Fixed — `quantization_config` is gone from the BF16 base config. Only `Step-3.7-Flash-FP8` declares the dynamic 128x128 block FP8, so the base artifact is no longer misclassified.

aic-core/src/aiconfigurator_core/sdk/utils.py
chanh: Fixed as part of the schema work — the parser normalizes the vendor field spellings (`moe_num_experts`, `moe_top_k`, `moe_layers_enum`, `num_attention_groups`) and trims the MTP tail off `layer_types`. Detail in the reply on the `common.py` thread.

aic-core/src/aiconfigurator_core/sdk/models/step3p7.py
chanh: Fixed in `167a2168`. The CP audit loop at the end of `_build_context_ops` is now `HybridMoEModel.apply_cp_to_context_ops`, called both there and again after `_append_shared_expert_ops` runs. Guard semantics are unchanged, since the method is shared with MiMo-V2-Flash, Llama 4 and Gemma 4.

Your diagnosis was exact, including the detail that those ops also bypassed the un-audited-op guard — so the miss surfaced as a wrong number rather than an error. Regression test builds through the factory path at `cp_size=2` on sglang and asserts `_seq_split == cp` on the `context_shared_expert_*` ops; it fails `{1} == {2}` without the fix.

aic-core/src/aiconfigurator_core/sdk/models/step3p7.py
chanh: Fixed in `df7cea2f`. Both halves:

`use_qk_norm` — `Step3p7Attention` builds `q_norm`/`k_norm` unconditionally, so there is no config flag to read, which is why none appears in the published config. The reason it was not reaching the ops is that `BaseModel` only reads `use_qk_norm` when `extra_params` is a dict, and the hybrid path passes a `HybridMoEConfig` dataclass. It now travels on the config object to all 8 attention ops.

`use_head_wise_attn_gate` — emits `g_proj` as a GEMM plus a sigmoid-multiply `ElementWise` before the out projection, at all 8 sites. Worth noting it is sized off *that layer’s* head count, not a single shared width: 96 on sliding layers, 64 on global. Shape assertions pin `_n == heads` and `_k == hidden_size` per layer type, plus the gate width against the projection it scales.

Both default off in `HybridMoEConfig`, with a test asserting no gate ops are emitted for a config that does not request one.

aic-core/src/aiconfigurator_core/sdk/common.py
```diff
@@ -129,6 +132,7 @@ class HybridMoEConfig:
 
     attn_layer_pattern: tuple[int, ...]  # per-layer: 0=SWA/local, 1=global
     moe_layer_freq: tuple[int, ...]  # per-layer: 0=dense, 1=MoE
+    swa_num_heads: int = 0
```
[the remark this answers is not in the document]
chanh: Good catch — fixed. `swa_num_heads` is now appended after `dense_inter_size` rather than inserted ahead of `swa_num_kv_heads`, so a legacy positional call keeps its original meaning instead of silently feeding a query-head count into the KV-head slot.

Added `test_hybrid_moe_config_positional_layout_is_stable`, which pins the pre-existing 8-field prefix and constructs the dataclass positionally to check the fields land where they used to. The two flags added for the Q/K-norm and attention-gate work are appended as well.

jasonqinzhou:
Fixed the remaining blocking item from @tianhaox’s review ([original comment](https://github.com/ai-dynamo/aiconfigurator/pull/1447#issuecomment-5175158620)) in `b12fd77a`: `STEP3P7` now uses MoE activation coefficients and dispatch-workspace accounting in TRT-LLM, vLLM, and SGLang, with the residual hidden width used for this hybrid model’s workspace. Added a three-backend regression test. Local validation: 122 focused tests passed, 3,249 unit tests passed (10 skipped), and repository-wide Ruff checks passed.

The original note is a top-level conversation comment, so GitHub does not expose a Resolve control for it; this follow-up records that its remaining blocker is addressed.

jasonqinzhou (approved):
All requested changes are addressed, all seven review threads are resolved, and the complete CI suite is green on `b12fd77a`. Approving.

PR #1452, opened 2026-07-31
feat(collector): add Step-3.7-Flash case planning and routing parity

chanh:
## Summary

Adds the StepFun Step-3.7-Flash collector case plan and the bundled model metadata needed to preserve BF16 and FP8 artifact identity.

This is a coordinated collector/model-config update. Maintainer approval was given for the cross-boundary scope.

## What changed

- Adds separate bundled configs for `stepfun-ai/Step-3.7-Flash` and `stepfun-ai/Step-3.7-Flash-FP8`.
- Registers the FP8 artifact for offline SDK config loading.
- Plans global and sliding-window attention as correlated physical topologies:
  - global: 64 query heads / 8 KV heads / window 0
  - sliding: 96 query heads / 8 KV heads / window 512
- Plans BF16 and FP8 MoE artifacts as separate physical model paths for the 4096 / 1280 / top-8 / 288-expert shape.
- Preserves Step-3.7's routed-expert contract, including router bias and a routed scaling factor of 3.0.
- Preserves an explicitly declared scaling factor of 0.0.
- Adds regression coverage for artifact identity, MoE geometry, attention topology, routing, deduplication, and offline FP8 config loading.

## Scope clarification

This PR does **not** add or modify measured performance rows, Parquet files, or collection provenance. The earlier H100 measurement payload was withdrawn; measured rows can be published separately through the supported provenance workflow.

## Validation

- `ruff check`: passed
- `ruff format --check`: passed
- focused collector/config tests: 112 passed
- full unit gate (excluding the repository's standard `test_moe_dispatch.py` exclusion): 3,262 passed, 10 skipped

## Summary by CodeRabbit

* **New Features**
  * Added support for the Step-3.7-Flash and Step-3.7-Flash FP8 model configurations.
  * Added coverage for GEMM, mixture-of-experts, attention, and multi-framework evaluation scenarios.
  * Added FP8 quantization support with block-based weight scaling.

* **Bug Fixes**
  * Improved compatibility with alternate model routing and scaling configuration names.

* **Tests**
  * Added regression coverage for Step-3.7 routing behavior and expanded model-case validation.


tianhaox:
Did a full data-quality pass on the parquet (base vs PR head) plus a before/after HYBRID-vs-measured comparison, and reviewed the collector/provenance files against the repo's collector rules.

## Data quality

Structurally clean:

- All 46,703 base rows preserved byte-for-byte; exactly 2,349 appended rows; same schema; no duplicate keys (within the appended set, and zero overlap with existing keys); no NaN or non-positive latencies.
- Coverage matches the declared plan exactly: 29 (dtype, tp, ep) cells x 27 token points x 3 distributions; the only absent cells are the 12 documented fp8_block TP>=4 divisibility gaps.
- `kernel_source` values are internally consistent and match the existing base table's dispatch split (fp8 → modelopt/flashinfer-cutlass; fp8_block → triton at EP=1, flashinfer-cutlass at EP>=2; bf16 → triton).
- The flip evidence in the PR body reproduces (e.g. fp8 tp1 tok64 = 7.216 ms silicon).

One real issue from the unlocked clocks: **8 of 87 latency curves contain clearly inflated points** — the next-larger token count is >30% *faster*, which is physically implausible. Worst offenders:

| cell | token point | latency | next point |
|---|---|---|---|
| fp8 tp16 ep1 power_law_1.01 | 2 | 1.809 ms | 4 → 0.153 ms (12x) |
| bf16 tp4 ep1 power_law_1.2 | 4 | 1.920 ms | 8 → 0.194 ms (10x) |
| fp8_block tp2 ep1 balanced | 64 | 2.236 ms | 80 → 0.817 ms (2.7x) |
| bf16 tp32 ep1 (all 3 dists) | 64 | 0.60–1.08 ms | 80 → 0.36–0.51 ms |
| bf16 tp1 ep16 / fp8_block tp1 ep32 | 1 | ~1.6–1.8x the tok-2 point | |

These sit at low token counts — exactly the decode/TPOT interpolation region — and will be served as exact `silicon` hits. Recommend re-measuring (or dropping) these 8 points before merge.

## HYBRID vs measured — how far off was the interpolation this heals?

Queried the same 330-point grid (11 dtype/tp/ep cells x 15 token counts x 2 distributions) under `DatabaseMode.HYBRID`, before and after this PR's rows. Before: all 330 points served as `source='empirical'`; after: all 330 `source='silicon'`. Ratio (HYBRID estimate / measured):

| dtype | median | 5%–95% | worst |
|---|---|---|---|
| bfloat16 | 0.90 | 0.68–1.06 | 0.62 |
| fp8 | 0.87 | 0.65–1.03 | 0.50 |
| fp8_block | 0.91 | 0.63–1.05 | 0.36* |

HYBRID was systematically optimistic for this shape: ~9–13% low at median, near-unbiased at tp1/ep1 but degrading to 0.75–0.86 median at tp8/ep>=8 with individual points ~2x low — and the 1-token decode points at tp8/ep8 were ~40–50% underestimated, which directly inflates predicted decode throughput. So this collection is a real accuracy win, especially for parallel topologies. (*The 0.36 worst point is the fp8_block tp2 tok64 outlier flagged above — likely the measurement, not HYBRID, is wrong there; another reason to re-measure it.)

## Findings

**1. Cached configs: the plain-named (BF16) config embeds the FP8 repo's `quantization_config`, and the `-FP8` config is unreachable.** Upstream `stepfun-ai/Step-3.7-Flash` has no `quantization_config` (verified against the HF repo; the quant block belongs to `stepfun-ai/Step-3.7-Flash-FP8`). Both JSONs here are byte-identical, so the BF16 name gets inferred as fp8_block (`utils.py:1070` → wrong default quant modes, ~halved weight-memory, and fp8_block's moe_tp>=4 divisibility ValueError applied to a BF16 model). Separately, `stepfun-ai/Step-3.7-Flash-FP8` is not added to `DefaultHFModels`, and `_load_model_config_from_model_path` only consults the bundled cache for paths in that set — so the new `-FP8` config can't be loaded by the SDK at all (offline runs of the exact model name this data was collected under will attempt an HF download). Fix: strip `quantization_config` from the plain config, keep it in `-FP8`, and register the `-FP8` path. Also note both this PR and #1447 add the identical plain config blob — worth rebasing this PR on #1447 to de-duplicate, as the PR body anticipates.

**2. `collection_meta.yaml`: the hand-authored `appended:` block is outside the sidecar contract.** No other `collection_meta.yaml` in the tree has this key; the deterministic writer (`collector/provenance.py:59` `_TABLE_FIELD_ORDER`) cannot round-trip it, and `collect.py:1487-1494` raises on an existing `provenance: legacy` sidecar anyway — so the block dies on the next real collection either way. It's also cosmetic under strict provenance: with top-level `provenance: legacy` retained, the 2,349 fresh rows still load as legacy-tier with no per-table hash attestation. The sanctioned recording is the standard table-level fields (`collector_hash`, `case_plan_hash`, `collected_at`, `rows`, `status`) written by the collector, replacing the legacy sidecar rather than annotating it. The `runtime.image`/`image_digest` additions are fine (both are in `_RUNTIME_FIELD_ORDER`).

**3. Cases YAML: `all_frameworks_op_cases: moe: cases: all` activates sglang/trtllm collection without the routing-identity fields.** Step-3.7 is sigmoid-scored, but the sglang collector reads scoring from the case (`collect_moe.py:197/227` ← `sglang_moe_scoring_func`, default `"softmax"` in `case_generator.py:1044`), not from the model config. Sigmoid-MoE precedent rows (`DeepseekV3ForCausalLM_cases.yaml:29-35`) explicitly declare `sglang_moe_scoring_func: sigmoid` + routing fields. As written, the first sglang campaign for this architecture would benchmark softmax routing under Step-3.7's name — mislabeled rows with no failure record. The vLLM data in this PR is unaffected (vllm's `collect_moe.py:92` resolves `scoring_func` from the packaged config). Fix: add the `sglang_moe_*` fields to the moe row, or scope moe activation to vllm until then.

## Explicitly checked and fine

- Leaving the 12 fp8_block TP>=4 cases to fail classified (rather than encoding a TP-conditional exclusion in YAML) is exactly what `failure_handling.md` mandates — the constraint is TP-conditional and no sanctioned declarative field can express it. Good call documenting it.
- `model_aliases` for the BF16 artifact on shape-only moe/attention rows matches the Qwen3-30B-A3B precedent (conditional on fixing the config split in finding 1, which is what keeps the two artifacts provably kernel-equivalent for these ops).
- The updated test counts (1986/53622 and 5010) were independently reproduced against the base checkout with this PR's YAML materialized — they're correct, not just re-pinned. Note they depend on the plain Step-3.7 config JSON existing (satisfied by either PR's copy).

aic-core/src/aiconfigurator_core/model_configs/stepfun-ai--Step-3.7-Flash-FP8_config.json
```diff
@@ -0,0 +1,79 @@
...
+    "sliding_attention",
+    "full_attention"
+  ],
+  "first_k_dense_replace": 3,
+  "num_experts": 288,
+  "num_experts_per_tok": 8,
+  "moe_intermediate_size": 1280,
+  "share_expert_dim": 1280,
+  "n_shared_experts": 1,
+  "scoring_func": "sigmoid",
```
jasonqinzhou, line 69: **[P1] Recollect through the same Step3p5 routing path that vLLM serves.** The authoritative config declares `use_moe_router_bias=true` and `moe_router_scaling_factor=3.0`, and vLLM 0.24.0's Step3p5 expert block passes the correction bias and scale into `FusedMoE`. This curated config stops at `scoring_func`, so `_resolve_moe_runtime_config()` currently produces `use_routing_bias=False` and `routed_scaling_factor=1.0`. Consequently, all 2,349 appended rows were benchmarked through a different MoE invocation from serving. Please preserve/normalize the full routing contract in the config and resolver, then recollect the parquet; metadata changes alone cannot repair these timings.

aic-core/src/aiconfigurator_core/systems/data/h100_sxm/moe/vllm/0.24.0/collection_meta.yaml
```diff
@@ -6,6 +6,25 @@ provenance: legacy
...
+  image: vllm/vllm-openai:v0.24.0
+  image_digest: sha256:251eba5cc7c12fed0b75da22a9240e582b1c9e39f6fbc064f86781b963bd814f
 tables:
   moe_perf:
     status: complete
+    # Base rows are legacy (provenance: legacy). This release APPENDS 2349
+    # freshly-measured Step-3.7-Flash MoE rows (288 experts, inter 1280, topk 8,
+    # hidden 4096) on NVIDIA H100 80GB HBM3 under vllm 0.24.0, healing the
+    # previously HYBRID-interpolated 288-expert / inter-1280 grouped-GEMM shape.
+    appended:
```
jasonqinzhou: **[P1] Publish these fresh rows through the supported provenance contract.** `appended:` is not part of the `collection_meta.yaml` schema, so the strict loader and CI ignore the nested hashes; with `provenance: legacy` still present and no `collector_ref`, this entire table remains legacy rather than attesting the 2,349-row append. The canonical finalizer explicitly rejects mixing fresh provenance into a legacy directory. Please use a fresh runtime directory or recollect/migrate the full table through the official finalizer, retaining checkpoint evidence. If the 12 block-FP8 cases are unresolved checkpoint failures, canonical status must be `partial`; if they are intentionally unrepresentable, exclude them from the attested plan rather than asserting an ad hoc exception.

aic-core/src/aiconfigurator_core/systems/data/h100_sxm/moe/vllm/0.24.0/collection_meta.yaml
```diff
@@ -6,6 +6,25 @@ provenance: legacy
...
+      model: stepfun-ai/Step-3.7-Flash-FP8
+      collected_at: '2026-07-31'
+      rows: 2349
+      collector_hash: sha256:520930ea82b1d49b16a2eda39b08bc0b35378f1f001a281b0d91bc993af8b710
+      case_plan_hash: sha256:ba180b03dd3f08260c00379bac2db553399c3b53b76505faf0e19f80619c34c2
+      dtypes: [bfloat16, fp8, fp8_block]
+      # Honest caveats:
+      #  - clocks were UNLOCKED (shared SLURM, no sudo) -> slightly noisier than
```
jasonqinzhou: **[P1] Recollect and audit the isolated silicon timing spikes.** The unlocked-clock caveat does not account for three discontinuities that exact Parquet queries still return unchanged: fp8/power_law_1.01/TP16 at 2 tokens is 1.80919 ms (16.4x its log-neighbor interpolation), bfloat16/power_law_1.2/TP4 at 4 tokens is 1.91972 ms (14.4x), and fp8_block/balanced/TP2 at 64 tokens is 2.23554 ms (2.8x). These points can directly distort configuration ranking. Please recollect the affected cases under controlled conditions and audit every new curve for similar spikes; do not hand-smooth or synthesize replacements.

collector/cases/models/Step3p7FlashForCausalLM_cases.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+model_case_values:
+  attention:
+    - model_path: stepfun-ai/Step-3.7-Flash-FP8
+      model_aliases:
+        - stepfun-ai/Step-3.7-Flash
+      num_attention_heads: 64
+      num_key_value_heads: 8
+      head_dim: 128
+      window_sizes: [0, 512]
```
jasonqinzhou: **[P2] Split the global and sliding-attention topology rows.** The pinned vLLM Step3p5 implementation substitutes 96 query heads for `sliding_attention` layers while global layers retain 64. This row cross-products `num_attention_heads: 64` with both `[0, 512]`, so the generated plan contains nonexistent 64-head SWA cases and omits the real 96-head SWA cases. Please declare separate correlated rows for `(64 Q, 8 KV, window 0)` and `(96 Q, 8 KV, window 512)`.

aic-core/src/aiconfigurator_core/model_configs/stepfun-ai--Step-3.7-Flash_config.json
```diff
@@ -0,0 +1,79 @@
...
+  ],
+  "first_k_dense_replace": 3,
+  "num_experts": 288,
+  "num_experts_per_tok": 8,
+  "moe_intermediate_size": 1280,
+  "share_expert_dim": 1280,
+  "n_shared_experts": 1,
+  "scoring_func": "sigmoid",
+  "num_nextn_predict_layers": 3,
+  "quantization_config": {
```
jasonqinzhou: **[P2] Do not mark the base artifact as block-FP8.** The authoritative `stepfun-ai/Step-3.7-Flash` config has no `quantization_config`; only `Step-3.7-Flash-FP8` declares dynamic 128x128 block FP8. These two bundled files are currently byte-identical, so the base artifact is misclassified and cannot reproduce its actual runtime policy. Please remove this block from the base config and preserve artifact-specific quantization when splitting the physical MoE cases.

jasonqinzhou (changes requested):

## What this PR does

This PR adds 2,349 measured H100/vLLM 0.24.0 MoE timing rows for StepFun Step-3.7-Flash, together with curated model configs, a Collector case plan, provenance notes, and updated case-count snapshots intended to replace HYBRID interpolation for the 288-expert/intermediate-1280 shape with silicon data.

The latest commit promptly brings both affected Collector count snapshots back in sync, and the broader append still preserves all 46,703 prior logical rows while documenting its image, hashes, topology, and collection caveats in enough detail to audit the intended coverage.

## Why this score

4.0/10, REQUEST_CHANGES. The head update fixes the prior red Collector count assertion: both changed tests pass locally and the required unit job is now green. The central data outcome is still unsafe, however, because the benchmark used different Step3p5 routing semantics from vLLM serving, the fresh hashes remain hidden under an unsupported legacy-provenance extension, and three exact silicon rows retain severe isolated timing spikes. The attention plan and base artifact config also retain two correctness defects.

chanh:
Pushed `53567cfa`, covering the two P2s that don't need a GPU.

**[P2] Split the global and sliding-attention topology rows.** Done — two correlated rows now, `(64 Q, 8 KV, window 0)` and `(96 Q, 8 KV, window 512)`, following the MiMo-V2-Flash precedent of one row per attention type with an explicit `id`. The plan now generates `{(64,8,128,0), (32,4,128,0), (16,2,128,0), (8,1,128,0)}` and `{(96,8,128,512), (48,4,128,512), (24,2,128,512), (12,1,128,512)}`, with no 64-head SWA cases.

Worth noting for the test strategy: **the case count is 8 either way**, before and after. The aggregate assertion could never have distinguished 8 correct cases from 8 incorrect ones — which is the same point raised on the count test. The new topology test asserts the exact set and fails on the old plan.

**Assert the new cases, not only the total.** Done — the Step-3.7-Flash rows are now checked for `hidden_size == 4096`, `inter_size == 1280`, `topk == 8`, `num_experts == 288`. One correction to the request: the BF16 path is a `model_aliases` entry on the FP8 row, so it *selects* that row rather than materializing a second set of cases (`_matches_model_path`). Asserting on both `model_name` values would fail. The alias is covered instead in the attention test, which runs both paths and asserts they resolve to identical topology.

**One thing found beyond the review.** The same routing-contract break flagged for vLLM also existed on the sglang side of this case plan. Those `sglang_moe_*` fields are read from the case YAML, not from the HF config, and this MoE row declared none of them — so they sat at the defaults of softmax, no correction bias, no scaling, while `Step3p7MoE.router_bias_func` in the published `modeling_step3p7.py` applies sigmoid over the gate, adds `router_bias` before top-k, renormalizes (`norm_expert_weight`), and scales by `moe_router_scaling_factor: 3.0`. Now declared explicitly on the row. The model declares no expert grouping, so `num_expert_group`/`topk_group` stay unset.

3362 tests pass, lint clean. The two `EngineSpec schema 6 vs 4` failures are pre-existing and reproduce identically on an unmodified `main`.

Still open on this PR are the two provenance/timing P1s. Both are resolved by the same choice — whether to recollect the full 47,675-row `0.24.0` table or to split this PR and land the collector fix on its own. I've flagged that decision to the repo owner; if you have a preference between them, that would settle it. Note that the coderabbit thread above independently asks to split the non-collector changes out of this cohort, which points the same way.

chanh:
Status on the two data P1s — **both are moot for this PR as it now stands.**

When I resolved the binary merge conflict I reverted the legacy `h100_sxm/moe/vllm/0.24.0/` directory to `main`, which withdrew the appended rows and the `collection_meta.yaml` change along with it. The PR's changed-file list is now entirely collector code, model configs, case plan and tests — no `moe_perf.parquet`, no provenance file. GitHub has both review threads marked outdated for the same reason.

That means:

- **[P1] Publish these fresh rows through the supported provenance contract** — no fresh rows are being published here anymore. Nothing to attest.
- **[P1] Recollect and audit the isolated silicon timing spikes** — those three discontinuities are in the existing published table, untouched by this PR.

This lands where the coderabbit thread above asked us to end up: the non-collector data change is split out of the collector task.

On re-landing the data separately, one constraint is worth stating explicitly, because it rules out one of the options offered in the provenance review. Recollecting the full 47,675-row table would re-measure every other model's already-published timings. Data changes here should be **additive only** — new rows may be added, existing measurements should not be silently replaced by a fresh run on different silicon conditions. Since a version directory holds exactly one `moe_perf.parquet` and exactly one provenance record, and the finalizer rejects mixing fresh provenance into a legacy directory, there is currently no supported way to attest an append without rewriting the neighbours.

So the fresh Step-3.7 rows will come back as their own change once there is an additive path — either provenance that can attest an append, or a runtime directory that does not require restating the rest of the table. Happy to take a steer on which of those you'd prefer; it's a schema question rather than something to settle inside this PR.

The collector-side fix that made the rows correct in the first place — the routing contract, `use_routing_bias` / `routed_scaling_factor` reading the vendor key spellings — is in this PR and is independently reviewable.

aic-core/src/aiconfigurator_core/model_configs/stepfun-ai--Step-3.7-Flash-FP8_config.json
chanh: Addressed. The curated configs now carry `use_moe_router_bias: true` and `moe_router_scaling_factor: 3.0`, and `_resolve_moe_runtime_config()` reads both the canonical and vendor spellings, so Step-3.7 resolves to `use_routing_bias=True` / `routed_scaling_factor=3.0`. Regression test in `test_step3p7_preserves_vendor_routing_contract`.

On the recollect: the 2,349 appended rows are no longer part of this PR — the parquet and provenance changes went away when I resolved the binary merge conflict, so nothing benchmarked through the wrong invocation is being published here. The resolver fix that makes a future collection correct is what remains. See the PR comment for why a full-table recollect is not the path we want to take.

collector/cases/models/Step3p7FlashForCausalLM_cases.yaml
```diff
@@ -0,0 +1,55 @@
...
+      tensor_parallel_sizes: [1, 2, 4, 8]
+  moe:
+    - model_path: stepfun-ai/Step-3.7-Flash-FP8
+      model_aliases:
+        - stepfun-ai/Step-3.7-Flash
+      hidden_size: 4096
+      inter_size: 1280
+      topk: 8
+      num_experts: 288
```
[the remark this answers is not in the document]
chanh: Addressed in `53567cfa`. The attention row was cross-producting `num_attention_heads: 64` with `window_sizes: [0, 512]`; it is now two correlated rows, `(64 Q, 8 KV, window 0)` and `(96 Q, 8 KV, window 512)`, following the MiMo-V2-Flash precedent of one row per attention type with an explicit `id`.

aic-core/src/aiconfigurator_core/systems/data/h100_sxm/moe/vllm/0.24.0/collection_meta.yaml
chanh: No longer applicable to this PR. The `collection_meta.yaml` change and the appended parquet rows were both withdrawn when I resolved the binary merge conflict — the `0.24.0` directory is back to `main`, so this PR publishes no rows and creates no provenance record. The `appended:` key you flagged is gone with it.

When the rows are re-landed separately, the constraint you identified still stands and we would like your steer on it: a version directory holds exactly one `moe_perf.parquet` and one provenance record, and the finalizer rejects fresh provenance in a legacy directory — so there is no way to attest an append without restating the neighbours. Recollecting the full 47,675-row table would re-measure every other model’s published timings under different (unlockable) clock conditions, which we do not want to do. Would you prefer provenance that can attest an append, or a runtime-directory convention that does not require restating the rest of the table?

aic-core/src/aiconfigurator_core/systems/data/h100_sxm/moe/vllm/0.24.0/collection_meta.yaml
chanh: No longer applicable to this PR — the three discontinuities you found (fp8/power_law_1.01/TP16 at 2 tokens, bfloat16/power_law_1.2/TP4 at 4 tokens, fp8_block/balanced/TP2 at 64 tokens) are in the existing published table, which this PR no longer modifies.

Worth recording for whoever picks up the audit: clock locking is not available to us on either Athena or SLURM (both unprivileged), so the unlocked-clock caveat cannot be removed by re-running. That makes a re-measurement a substitution rather than a reproduction, which is why we would rather not silently replace those points.

collector/cases/models/Step3p7FlashForCausalLM_cases.yaml
chanh: Addressed in `53567cfa` — the case plan now declares the two attention topologies as separate correlated rows rather than a cross-product. Detail in the reply on the other thread for this file.

aic-core/src/aiconfigurator_core/model_configs/stepfun-ai--Step-3.7-Flash_config.json
chanh: Addressed. `quantization_config` is removed from the BF16 base config; only `Step-3.7-Flash-FP8` declares the dynamic 128x128 block FP8. The two files are no longer byte-identical.

aic-core/src/aiconfigurator_core/model_configs/stepfun-ai--Step-3.7-Flash_config.json
```diff
@@ -0,0 +1,79 @@
...
+    "quant_method": "fp8",
+    "weight_block_size": [
+      128,
+      128
+    ],
+    "activation_scheme": "dynamic"
+  }
+}
```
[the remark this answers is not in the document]
chanh: Addressed — this PR no longer changes any file under `aic-core/src/aiconfigurator_core/systems/data/`. The parquet and `collection_meta.yaml` changes were withdrawn, so what remains is the collector code, the case plan, the model configs and tests. That is the split you asked for.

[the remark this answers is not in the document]
jasonqinzhou: Resolved on the current head (`e1fce476`). This maintenance pass was explicitly approved as a coordinated collector/model-config change. The bundled BF16 and FP8 configs are required to preserve physical artifact identity, the FP8 artifact is registered for offline loading, and this PR contains no systems-data or performance-row changes. The updated title/body now declares that scope; the full unit gate passed (3,262 passed).

aic-core/src/aiconfigurator_core/model_configs/stepfun-ai--Step-3.7-Flash-FP8_config.json
jasonqinzhou: Verified resolved on the current head (`e1fce476`). Both bundled configs carry the Step-3.7 routing fields, the resolver accepts the canonical/vendor spellings, and the regression pins router bias plus the 3.0 routed scale (including preservation of explicit 0.0). No recollection is required in this PR because all measured rows and provenance changes were withdrawn.

aic-core/src/aiconfigurator_core/systems/data/h100_sxm/moe/vllm/0.24.0/collection_meta.yaml
jasonqinzhou: Verified obsolete on the current head (`e1fce476`). The PR no longer changes the H100 performance table, so none of the cited timing points are introduced or modified here. The stale measured-data claims have also been removed from the title and body.

collector/cases/models/Step3p7FlashForCausalLM_cases.yaml
jasonqinzhou: Verified resolved on the current head (`e1fce476`). The plan uses correlated rows for global `(64 Q, 8 KV, window 0)` and sliding `(96 Q, 8 KV, window 512)` attention, with direct regression coverage in the collector tests.

aic-core/src/aiconfigurator_core/systems/data/h100_sxm/moe/vllm/0.24.0/collection_meta.yaml
jasonqinzhou: Verified obsolete on the current head (`e1fce476`). The PR has no Parquet, `collection_meta.yaml`, or other systems-data changes; the updated PR body explicitly says measured rows must be published separately through the supported provenance workflow.

aic-core/src/aiconfigurator_core/model_configs/stepfun-ai--Step-3.7-Flash_config.json
jasonqinzhou: Verified resolved on the current head (`e1fce476`). The BF16 base config has no `quantization_config`; only the FP8 artifact declares dynamic 128x128 block FP8. The MoE plan now uses separate physical model paths, and the artifact-filter regression verifies both identities and the 4096/1280/top-8/288-expert shape.

jasonqinzhou (approved):
Verified the current head after resolving the review findings. BF16/FP8 artifact identity, Step-3.7 routing, attention topology, and offline config loading are covered by regressions; all review threads are resolved; local full unit validation and all exact-head GitHub checks pass.

PR #1455, opened 2026-08-01
feat(sdk): GEMM quant-transfer ladder via shared primitive + validate-gate mirror

tianhaox:
## What

Prerequisite mechanism for #1392 (`nvfp4_wo`) — adds **no new quant**. Gives GEMM the empirical quant-transfer ladder MoE already has, unified behind one shared primitive, and cleans the `task_v2` validate gate so it mirrors (never exceeds) the query ladder's admission policy.

After this lands, adding a data-less quant is **one enum line + one util-level line** (both languages): HYBRID/EMPIRICAL estimates end-to-end via util borrowing, SILICON still rejects early. Full recipe for the #1392 author: `docs/design/how_to_add_a_gemm_quant_mode.md`. Design + as-built decisions: `docs/design/gemm_quant_transfer_design.md`.

## How

**One borrow primitive, labels derived from the reference's relation** (`util_empirical.quant_transfer_grid`): correction is `e(profile_query)/e(profile_ref)` from a per-op util-LEVEL table — identically 1 within a profile, so `xshape` (same quant) and `xquant` (same profile) are the ratio-1 degenerations of the cross-profile borrow; three confidence labels over one mechanism, not three mechanisms.

- **MoE**: refactored onto the primitive with zero behavior change (pinned by the existing unit suite and the Rust parity oracles).
- **GEMM** (new): own depth-3 `(m,n,k)` grid (its `xshape` relation class is structurally empty — the own grid already spans every collected shape) → `xquant` → `xprofile`. `_GEMM_QUANT_UTIL_LEVEL` seeded from compute-bound medians (`m≥1024, n,k≥2048`) across b200/h200/h100 × trtllm/vllm/sglang; method + LOO evidence on the table (LOO `nvfp4 ← fp8`: 22–33% MAPE, comparable to MoE's ~24% xprofile tier).
- **Compute-first XPROFILE ordering for GEMM** (`(|Δcompute|, |Δmemory|)` lexicographic): the compute factor (activation precision) determines the dense kernel's compute family, so a weight-only quant deterministically borrows **bfloat16** data — under plain L1 it would tie between bf16 and fp8 (e.g. `(0.5625, 1)`: both 1.4375) and fall to a file-order lottery. MoE keeps its validated L1 metric unchanged.
- **Validate gate** (`task_v2._check_role_against_db`): `gemm` now gets `profile_transfer`; new XPROFILE reachability = non-SILICON + `XPROFILE` in the resolved policy + query profile **listed** in the op's level table (deliberately stricter than the runtime default fallback — enforces the recipe) + ≥1 other-profile collected quant. Dead `"nvfp4_wo": "bfloat16"` alias removed (no such enum member; the normalize-to-bf16 design was reviewed out of #1392); the real `w4a16_mxfp4_cutlass → w4a16_mxfp4` alias kept.
- **Rust point-for-point**: `GemmGrids.quant_order` records first-seen (file row) quant order (the `BTreeMap` iterates alphabetically, but tie-breaks are pinned to Python's dict-insertion order) + `GemmTable::available_quants()`; ladder in `gemm_empirical`; `GEMM_QUANT_UTIL_LEVEL` const; oracle tests pinned to Python at 1e-9.

## Behavior change (intentional — the mechanism working)

HYBRID/EMPIRICAL queries for existing quants with no data on a system change from raising `EmpiricalNotImplementedError` to a policy-gated, provenance-tagged estimate under the default (aggressive) policy — e.g. nvfp4 GEMM on Hopper. All previously-computable paths (silicon, own-grid empirical, covered hybrid) are bit-unchanged; the ladder only extends behind the existing miss, and `transfer_policy="balanced"`/`"conservative"`/`"off"` still reject.

Parity-case archaeology: the `minimax-m25-nvfp4-h200` HYBRID case now computes on both sides — the old raise came from the **GEMM** op (the old test comment misattributed it to MoE, whose ladder always had a reachable reference there). The error-symmetry contract is preserved in a new `...-hybrid-balanced-miss` case and the FFI typed-miss test, both pinned with `transfer_policy="balanced"` (no same-profile sibling exists for `(0.5625, 4)`, so the miss is stable by construction).

## Verification

- Python unit: 2943 passed (3 pre-existing collector/GLM failures verified present at base via stash).
- `cargo test --workspace --all-features`: green.
- Engine-step parity: 307 passed; compile-engine parity + goldens: 68 passed.
- New mechanism tests: `tests/unit/sdk/database/test_gemm_quant_transfer.py` (xquant exactness, weight-only-borrows-bf16 guarantee, policy gating, SILICON rejection, covered-quant invariance, level-table completeness) + gate admission tests in `test_task_config.py` + Rust ladder oracles in `operators/gemm.rs`.

## Summary by CodeRabbit

- **New Features**
  - Added GEMM quantization transfer for empirical estimates, including same-profile and cross-profile fallback.
  - Added utilization-based scaling and provenance tracking.
  - Extended hybrid-mode validation for reachable quantization profiles under configured policies.
  - Aligned MoE and GEMM transfer behavior.

- **Bug Fixes**
  - Improved policy-aware handling of missing quantization data.
  - Removed the obsolete `nvfp4_wo` alias.

- **Documentation**
  - Added guidance for quantization transfer and new GEMM quantization modes.

### DIFF
--- a/src/aiconfigurator/sdk/task_v2.py
+++ b/src/aiconfigurator/sdk/task_v2.py
@@ -1408,9 +1408,15 @@ class Task:
             alias = validation_aliases.get(name)
             if alias and alias in modes:
                 return
-            if profile_transfer and xquant_enabled and _profile_reachable(mode, modes):
+            # fp8_static is a COMPOSITE mode: its base GEMM is transferable, but
+            # it also requires compute_scale/scale_matrix overhead tables that
+            # have no transfer ladder (by design). Its admission stays purely
+            # data-driven (fp8_static in modes iff all three tables exist, see
+            # _gemm_key_names).
+            transfer_ok = profile_transfer and mode is not common.GEMMQuantMode.fp8_static
+            if transfer_ok and xquant_enabled and _profile_reachable(mode, modes):
                 return  # XQUANT-reachable in HYBRID/EMPIRICAL (same-profile data)
-            if profile_transfer and xprofile_enabled and _xprofile_reachable(op, mode, modes):
+            if transfer_ok and xprofile_enabled and _xprofile_reachable(op, mode, modes):
                 return  # XPROFILE-reachable (calibrated level + any other-profile data)
             exc_type = UnsupportedWideepConfigError if op.startswith("wideep_") else ValueError
             raise exc_type(
--- a/src/aiconfigurator/sdk/task_v2.py
+++ b/src/aiconfigurator/sdk/task_v2.py
@@ -1349,18 +1349,32 @@ class Task:
         xquant_enabled = _non_silicon and common.TransferKind.XQUANT in _policy
         xprofile_enabled = _non_silicon and common.TransferKind.XPROFILE in _policy
 
-        def _profile_reachable(mode: Any, supported_names: list) -> bool:
-            enum_cls = type(mode)
+        from aiconfigurator.sdk.operations import gemm as gemm_ops
+        from aiconfigurator.sdk.operations import moe as moe_ops
+
+        _xprofile_level_known = {
+            "gemm": gemm_ops.xprofile_util_level_known,
+            "moe": moe_ops.xprofile_util_level_known,
+            "wideep_context_moe": moe_ops.xprofile_util_level_known,
+            "wideep_generation_moe": moe_ops.xprofile_util_level_known,
+        }
+
+        def _mode_profile(mode: Any) -> tuple:
             val = getattr(mode, "value", None)
-            qp = (getattr(val, "memory", None), getattr(val, "compute", None))
+            return (getattr(val, "memory", None), getattr(val, "compute", None))
+
+        def _supported_profiles(mode: Any, supported_names: list) -> list[tuple]:
+            enum_cls = type(mode)
+            out = []
             for nm in supported_names:
                 try:
-                    other = enum_cls[nm].value
+                    out.append(_mode_profile(enum_cls[nm]))
                 except (KeyError, AttributeError):
                     continue
-                if (getattr(other, "memory", None), getattr(other, "compute", None)) == qp:
-                    return True
-            return False
+            return out
+
+        def _profile_reachable(mode: Any, supported_names: list) -> bool:
+            return _mode_profile(mode) in _supported_profiles(mode, supported_names)
 
         def _xprofile_reachable(op: str, mode: Any, supported_names: list) -> bool:
             """XPROFILE admission: the op's util-LEVEL table must list the query
@@ -1369,28 +1383,11 @@ class Task:
             enforces the enum-line + level-line add-a-quant recipe), and the DB
             must carry at least one quant of a DIFFERENT profile to borrow from
             (any collected quant is a viable nearest-profile reference)."""
-            from aiconfigurator.sdk.operations import gemm as gemm_ops
-            from aiconfigurator.sdk.operations import moe as moe_ops
-
-            level_known = {
-                "gemm": gemm_ops.xprofile_util_level_known,
-                "moe": moe_ops.xprofile_util_level_known,
-                "wideep_context_moe": moe_ops.xprofile_util_level_known,
-                "wideep_generation_moe": moe_ops.xprofile_util_level_known,
-            }.get(op)
+            level_known = _xprofile_level_known.get(op)
             if level_known is None or not level_known(mode):
                 return False
-            enum_cls = type(mode)
-            val = getattr(mode, "value", None)
-            qp = (getattr(val, "memory", None), getattr(val, "compute", None))
-            for nm in supported_names:
-                try:
-                    other = enum_cls[nm].value
-                except (KeyError, AttributeError):
-                    continue
-                if (getattr(other, "memory", None), getattr(other, "compute", None)) != qp:
-                    return True
-            return False
+            qp = _mode_profile(mode)
+            return any(p != qp for p in _supported_profiles(mode, supported_names))
 
         def _check(op: str, mode: Any, *, profile_transfer: bool = False) -> None:
             if mode is None:
--- a/src/aiconfigurator/sdk/task_v2.py
+++ b/src/aiconfigurator/sdk/task_v2.py
@@ -1330,19 +1330,24 @@ class Task:
 
         # supported_quant_mode is a DATA-PRESENCE list (which quants the DB carries
         # tables for), not a backend-capability list. In SILICON that equals what we
-        # can model. In HYBRID/EMPIRICAL the MoE util-empirical path can synthesize a
-        # quant from a collected quant that shares its (memory, compute) profile
-        # (XQUANT cross-quant transfer, see operations/moe.py) -- only MoE implements
-        # this, so only MoE relaxes. Truly-unreachable quants (no same-profile data)
-        # still fail early here rather than crashing late in the sweep.
-        # Admission via the XQUANT cross-quant transfer only holds if (a) we're in a
-        # non-SILICON mode AND (b) the resolved transfer policy actually enables XQUANT.
-        # Otherwise operations/moe.py rejects the quant at query time by policy, so
-        # validate must not pre-admit it (e.g. transfer_policy="off"/"conservative").
-        xquant_enabled = self.database_mode not in (
+        # can model. In HYBRID/EMPIRICAL the util-empirical path (GEMM and MoE; the
+        # shared quant-transfer primitive in operations/util_empirical.py) can
+        # synthesize a quant from a collected sibling: XQUANT borrows within the
+        # same (memory, compute) profile, XPROFILE borrows across profiles rescaled
+        # by the op's util-LEVEL ratio. This gate mirrors exactly what the resolved
+        # policy + DB contents make reachable at query time — truly-unreachable
+        # quants still fail early here rather than crashing late in the sweep.
+        # Admission only holds if (a) we're in a non-SILICON mode AND (b) the
+        # resolved transfer policy actually enables that relation. Otherwise the
+        # query path rejects the quant at run time by policy, so validate must not
+        # pre-admit it (e.g. transfer_policy="off"/"conservative").
+        _non_silicon = self.database_mode not in (
             None,
             common.DatabaseMode.SILICON.name,
-        ) and common.TransferKind.XQUANT in common.resolve_transfer_policy(self.transfer_policy)
+        )
+        _policy = common.resolve_transfer_policy(self.transfer_policy)
+        xquant_enabled = _non_silicon and common.TransferKind.XQUANT in _policy
+        xprofile_enabled = _non_silicon and common.TransferKind.XPROFILE in _policy
 
         def _profile_reachable(mode: Any, supported_names: list) -> bool:
             enum_cls = type(mode)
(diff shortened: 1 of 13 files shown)


src/aiconfigurator/sdk/task_v2.py
```diff
@@ -1367,23 +1399,30 @@ def _check(op: str, mode: Any, *, profile_transfer: bool = False) -> None:
...
+            # query-time ladder.
+            validation_aliases = {"w4a16_mxfp4_cutlass": "w4a16_mxfp4"}
             alias = validation_aliases.get(name)
             if alias and alias in modes:
                 return
             if profile_transfer and xquant_enabled and _profile_reachable(mode, modes):
-                return  # transfer-reachable in HYBRID/EMPIRICAL with XQUANT enabled
+                return  # XQUANT-reachable in HYBRID/EMPIRICAL (same-profile data)
+            if profile_transfer and xprofile_enabled and _xprofile_reachable(op, mode, modes):
+                return  # XPROFILE-reachable (calibrated level + any other-profile data)
```
Arsene12358: **Hard blocker — transfer admission must exclude `fp8_static`.** It is a composite mode: `GemmOp.query` subtracts `compute_scale`/`scale_matrix` overhead tables that this PR deliberately gives no ladder, so profile-reachability admits configs that are still unreachable at runtime and every sweep config then dies in `query_compute_scale` (see the review body for the `b200_sxm/vllm/0.19.0` before/after repro and the 38 affected combos). Suggested guard — validated on this head: the probe returns to fast rejection and `test_task_config.py` + `test_gemm_quant_transfer.py` pass (96 tests), with the new nvfp4 admissions untouched:

```suggestion
            # fp8_static is a COMPOSITE mode: its base GEMM is transferable, but
            # it also requires compute_scale/scale_matrix overhead tables that
            # have no transfer ladder (by design). Its admission stays purely
            # data-driven (fp8_static in modes iff all three tables exist, see
            # _gemm_key_names).
            transfer_ok = profile_transfer and mode is not common.GEMMQuantMode.fp8_static
            if transfer_ok and xquant_enabled and _profile_reachable(mode, modes):
                return  # XQUANT-reachable in HYBRID/EMPIRICAL (same-profile data)
            if transfer_ok and xprofile_enabled and _xprofile_reachable(op, mode, modes):
                return  # XPROFILE-reachable (calibrated level + any other-profile data)
```

Arsene12358 (changes requested):
**REQUEST CHANGES — one hard blocker: the validate gate now admits `fp8_static` configs that cannot run. Everything else verified merge-ready under adversarial re-checking.**

## Verified good first (so the blocker reads in context)

I ran a 1,620-query base↔PR sweep across three real DBs (`h200_sxm/trtllm/1.3.0rc10`, `h200_sxm/vllm/0.19.0`, `b200_sxm/vllm/0.19.0`; every `GEMMQuantMode` × SOL/SILICON/HYBRID/EMPIRICAL × 5 shapes, plus 3 collected MoE slices × 4 token counts × 3 modes per `MoEQuantMode`):

- **0 value changes** on any previously-computable path — the bit-unchanged claim holds on real data.
- All **110** diffs are the intended miss→estimate flips, confined to exactly the five uncovered GEMM quants (`fp8_ootb` ×30; `int4_wo`/`int8_wo`/`nvfp4`/`sq` ×20 each).
- **900 MoE values bit-identical**, including every transfer-resolved (xshape/xquant/xprofile) path — the MoE-refactor zero-behavior-change claim is empirically confirmed, not just suite-pinned.
- All 7 Rust oracle constants independently re-derived from Python (`shared_layer=False` views, fresh process): exact matches, including the b200 `int4_wo` case where compute-first ordering must beat file order. Python/Rust `GEMM_QUANT_UTIL_LEVEL` compared programmatically: identical rows, order, and default.
- Misattribution archaeology confirmed at merge-base: MoE nvfp4 on `h200_sxm/vllm/0.19.0` HYBRID computes (0.11299… ms, identical values post-PR) while GEMM nvfp4 raises — the old parity comment did blame the wrong op.
- `nvfp4_wo` confirmed absent from `GEMMQuantMode`, so the removed validation alias was dead code; removal changes nothing.
- Corner audits clean: borrowed-grid identity keys carry `id(ref.node)` + the policy-bearing `selection_key` + provenance (no cross-policy aliasing), and `LoadedOpData` iteration on unloaded data yields nothing (UserDict over an empty dict), matching Rust's `is_missing_perf_data → empty vec`, so fully data-less ops still raise the typed empirical miss on both sides.

## Hard blocker: `fp8_static` transfer admission — the gate exceeds the ladder

`fp8_static` is a **composite** mode: `GemmOp.query` models it as the dynamic-fp8 GEMM **minus** separately collected `compute_scale`/`scale_matrix` overhead tables, and this PR's own design doc scopes those tables out of the ladder ("the compute_scale/scale_matrix overhead tables get no ladder … it requires data anyway"). The support matrix encodes exactly that: `fp8_static ∈ supported_quant_mode["gemm"]` only when all three fp8 tables exist (`_gemm_key_names`), and `validate()`'s docstring calls this "where fp8_static is gated by overhead-table availability".

The new profile-transfer admission reasons only about `(memory, compute)` — and `fp8_static = QuantMapping(1, 2)`, the same profile as `fp8`. So wherever plain fp8 gemm data exists, XQUANT admits it; wherever *any* gemm data exists, XPROFILE admits it (profile (1,2) is listed in the level table). The ladder can synthesize the base GEMM part, but nothing can synthesize the overhead tables, so the admitted config is still unreachable — the gate now exceeds the ladder, which is precisely what the new gate comment forbids ("truly-unreachable quants still fail early here rather than crashing late in the sweep").

Repro on `b200_sxm/vllm/0.19.0` (fp8 gemm data present, no `quantize/` tables), identical probe on both trees:

```text
# merge-base eed87c1c
validate(HYBRID, fp8_static): REJECTED ValueError: Unsupported gemm quant mode 'fp8_static'
  for system='b200_sxm', backend='vllm', version='0.19.0'. Supported gemm modes: ['bfloat16', 'fp8', ...]

# PR head b647f139
validate(HYBRID, fp8_static): ADMITTED (no raise)          <- the regression
GemmOp.query(fp8_static): EmpiricalNotImplementedError: No empirical compute_scale data is available for m=1024, k=4096.
```

Downstream, `pareto_analysis.py` catches the per-config exception and skips, every config fails identically, and the user burns the whole sweep to get `RuntimeError: No results found for any parallel configuration. Showing last exception: … No empirical compute_scale data …` instead of today's instant, actionable validation error. Op-level behavior is unchanged vs base (`fp8_static` is not among the 110 flips) — the defect is confined to the gate.

Exposure is not exotic: `models/helpers.py` auto-selects `fp8_static` for every per-tensor FP8 checkpoint that doesn't mark dynamic activations (the ModelOpt/NVIDIA default), on every backend, and `cli_recommend` passes `database_mode="HYBRID"` by default (`cli/api.py:381`). 38 shipped combos carry gemm tables but no `quantize/` tables, all affected: `a100_sxm` (all 7 combos — no quantize dir at all), `h100_sxm` (sglang 0.5.6.post2/0.5.9/0.5.10, vllm 0.14.0/0.19.0), `h200_sxm` (sglang ×3, vllm 0.14.0/0.18.0/0.19.0), `b200_sxm` (sglang 0.5.9, vllm 0.19.0), `b300_sxm` (sglang 0.5.9/0.5.10, vllm 0.19.0), `gb200`+`gb300` (sglang 0.5.9/0.5.10, vllm 0.14.0/0.19.0 each), `b60` (vllm 0.12.0/0.20.0), `l40s` (sglang 0.5.9/0.5.10, vllm 0.14.0), `rtx_pro_6000_server` (sglang 0.5.10, vllm 0.19.0).

Why the existing regression test didn't catch it: `test_validate_agg_fp8_static_on_sglang_is_data_driven` leaves `database_mode` unset, and the gate treats `None` as non-relaxing — the admission only opens with an explicit HYBRID/EMPIRICAL, which is exactly what `cli_recommend` sends.

## Suggested fix (validated on this head)

The `transfer_ok` guard is posted as an inline suggestion on `task_v2.py`. I chose the simple exclusion over a data-driven mirror (checking overhead-table presence inside the gate) deliberately: overhead data without base fp8 data doesn't ship, so the extra fidelity buys nothing and would re-implement `_gemm_key_names` in a second place.

Please also pin the contract with a HYBRID-mode sibling of the existing sglang test:

```python
def test_validate_fp8_static_not_transfer_admitted_in_hybrid():
    """fp8_static is a composite mode (fp8 base minus compute_scale/scale_matrix);
    the overhead tables have no transfer ladder, so HYBRID must NOT admit it via
    profile transfer on combos without quantize data — validate keeps failing
    fast instead of the sweep dying late in query_compute_scale."""
    t = Task(
        serving_mode="agg",
        model_path="Qwen/Qwen3-32B",
        system_name="b200_sxm",
        backend_name="vllm",
        backend_version="0.19.0",
        database_mode="HYBRID",
    )
    t.gemm_quant_mode = common.GEMMQuantMode.fp8_static
    with pytest.raises(ValueError, match="Unsupported gemm quant mode 'fp8_static'"):
        t.validate()
```

Validation of the patch on this head: the probe flips back to REJECTED with the original message, and `test_task_config.py` + `test_gemm_quant_transfer.py` stay green (96 passed) — the new nvfp4 admission tests are untouched, so the exclusion is surgical.

## Non-blocking notes

- Both CodeRabbit nitpicks triaged as genuinely trivial: the pooled cache tag is cosmetic (correctness holds — grid identity carries node id, policy, and provenance), and the int4_wo-dataless premise pin is optional hardening.
- Spot-checked the `_GEMM_QUANT_UTIL_LEVEL` `[data]` derivation on the three stacks above (compute-bound-region medians, m≥1024, n,k≥2048): bf16 0.685–0.763 sits inside the quoted 0.55–0.79; the fp8-profile and nvfp4 medians land within ~0.01–0.05 of the quoted endpoints, consistent with my 3-stack subset vs your 9-stack set. Method reproduces; no action needed.

src/aiconfigurator/sdk/task_v2.py
tianhaox: Fixed in 88b6d26ec — applied your `transfer_ok` guard verbatim and added the suggested `test_validate_fp8_static_not_transfer_admitted_in_hybrid` regression test (pins the HYBRID rejection on `b200_sxm/vllm/0.19.0`). Design doc's gate section now records the composite-mode exclusion so the mirror contract stays accurate. Both CodeRabbit nitpicks also taken (pooled cache tag rename + `_assert_dataless` premise pin). Affected suites: 209 passed.

Arsene12358 (approved):
**APPROVE — the blocker fix in 88b6d26e is verified; this is merge-ready.**

Re-verified the fix independently on the new head: the `transfer_ok` guard and the regression test match the validated suggestion; the `b200_sxm/vllm/0.19.0` probe returns to the fast, actionable rejection (`ValueError: Unsupported gemm quant mode 'fp8_static' …`); the four touched suites pass locally (146 tests, including the new HYBRID pin and the `_assert_dataless` premise checks); and since the fix also renamed the pooled cache tag, I re-ran the 1,620-query real-DB value sweep against the previously reviewed head — 0 diffs, so nothing numeric moved. CI is green across all 19 checks.

With the composite-mode exclusion recorded in the design doc, the gate↔ladder mirror contract now holds everywhere I could probe it. Nice turnaround — and the shared-primitive design held up well under adversarial re-verification (900 MoE values bit-identical across the refactor, all 7 Rust oracles re-derived exactly).

PR #1459, opened 2026-08-02
feat(sdk): model PP stage imbalance and pipe fill instead of ideal linear speedup

tianhaox:
#### Overview:

PP was modeled as an **ideal linear speedup**. A step's latency was the *whole*
model's (`_num_layers` is never divided by `pp_size`) and throughput was simply
multiplied by `pp_size`. That is exact only when every stage costs the same and
the pipe is always full — neither holds.

This adds `PipelineSchedule`, which computes the two factors that were missing:
stage imbalance and pipe fill. `pp_size == 1` is bit-identical to before.

#### Details:

**What was wrong**

A pipeline advances at `max_i(t_i)`, not the average. The un-sharded `lm_head`
sits alone on the last stage, so for a **decode** step it sets the cycle.
Qwen3-32B (64 layers, vocab 151936) on H20 at `pp=8`:

```
stage times (ms): [3.153, 3.150, 3.150, 3.150, 3.150, 3.150, 3.150, 3.901]
                                                                    ^ lm_head
ideal cycle 3.244   realized 3.901   ->  balance 83.2%  ->  6.65x, not 8x
```

**Prefill** steps stay at 99.9% — per-layer work dominates there, so this is a
decode-side effect.

Two more gaps: `num_layers % pp_size != 0` only logged *"we're nothing to
correct this"* while the fattest stage actually gates the cycle (64 layers over
6 stages costs `11/(64/6)`), and a starved pipe still scaled linearly.

**The model**

```
cycle_time     = max_i(stage_time_i) + p2p_per_hop
balance_factor = (step_total / pp_size) / cycle_time   # per-microbatch latency effect
fill_factor    = min(1, num_microbatches / pp_size)    # throughput-only effect
```

Op placement reuses the naming contract the model classes already require
(`GPTModel` docstring: *"Same for logits_gemm"*): `*embedding*` → first stage,
`*logits_gemm*` → last, `*p2p*` → link (charged once per hop, not as a
whole-step total), everything else per-layer. `warn_on_unclassified_ops` flags a
new head-like op that would otherwise be silently smeared across every stage.

Layer partition defaults to an even split with the remainder on the leading
stages (vLLM `get_pp_indices` / TRT-LLM); an explicit partition is accepted.

**Where it plugs in** — two touch points in `run_agg`:

```python
mix_step_latency_ms     /= pp_schedule.balance_factor(mix_per_ops, model._num_layers)
genonly_step_latency_ms /= pp_schedule.balance_factor(genonly_per_ops, model._num_layers)
output_throughput = output_throughput * scale_factor * pp_schedule.fill_factor()
```

Inflating the step latency to the real traversal time (`pp * cycle`) is the
load-bearing choice: TTFT, TPOT, `_total_step_latency_ms`, `_step_throughput`
and `_throughput_cap` all derive from it, so they stay consistent without
individual patches.

**Why orchestration and not the op layer** — per `.claude/rules/rust-core/parity.md`,
changing op query math obligates a mirrored Rust implementation plus oracle
anchors. This consumes the per-op breakdown *after* the step is evaluated, so
both engine-step paths benefit with no parity surface touched. Same layering
AFD's pipeline model already uses.

**Measured effect** (Qwen3-32B, 8x H20, isl=4000/osl=500/bs=32 — prefill-dominated,
so the correction is mild here; decode-heavy workloads move much more):

| tp/pp | tokens/s/gpu before → after |
|---|---|
| 8/1 | 207.72 → 207.72 (unchanged) |
| 4/2 | 268.37 → 266.78 |
| 2/4 | 309.94 → 305.85 |
| 1/8 | 344.44 → **333.21** |

- `aic-core/src/aiconfigurator_core/sdk/pipeline.py` — the whole model, ~200 lines
- `aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py` — two touch
  points in `run_agg` plus the `_pipeline_schedule` backend hook
- `docs/PIPELINE_PARALLEL_MODELING.md` — derivation and the known-gaps list

#### Testing

- New `tests/unit/sdk/test_pipeline_schedule.py`: 40 cases (placement, partition
  incl. uneven splits, max-not-average cycle, P2P hop amortization + overlap,
  starvation, the `pp_size == 1` identity, validation).
- Full `tests/unit`: **2974 passed**. The 3 failures in
  `tests/unit/collector/sglang/test_collect_mla_module.py` (GLM-5 checkpoint
  naming) are **pre-existing** — reproduced on a clean baseline.
- `pp_size == 1` verified bit-identical end to end on H20: TTFT 650.722 /
  TPOT 18.195 / 207.72 tokens/s/gpu, unchanged.

#### Known gaps (documented in the design doc, deliberately not in this PR)

1. **`fill_factor` is linear** — the real curve is likely steeper once scheduler
   and sync overhead are counted. Needs silicon calibration.
2. **Chunked prefill x PP is not modeled** — adjacent chunks of one request have
   a RAW dependency on the KV they write, so they cannot occupy the pipe
   simultaneously. `fill_factor` is the intended hook; the coefficient needs
   measurement.
3. **P2P still hardcodes `inter_node_bw`** (~9x pessimistic intra-node).
   `SystemSpec.get_p2p_bandwidth(num_gpus)` already implements the topology
   tiers on both sides, but the correct selector is `tp * pp`, which the op is
   not constructed with — fixing it touches every model's `P2P` construction
   plus the Rust mirror, so it belongs in its own change. Impact at `pp=8`:
   0.67% of a decode step, 2.2% of a prefill step, conservative direction.
4. **PP is still excluded from the automatic search** — `build_disagg_parallel_lists`
   takes `should_enable_pp` (default `False`) and no caller passes `True`.
   Turning it on should wait until 1 and 2 are calibrated.

#### Related Issues:

- Relates to: PP modeling alignment (no tracked issue yet)

## Summary by CodeRabbit

* **New Features**
  * Added pipeline-parallel performance modeling with configurable stage placement, layer partitioning, and microbatch scheduling.
  * Estimates now account for stage balance, communication overhead, pipeline fill, and overlap efficiency.
  * Added diagnostics for incomplete or potentially misclassified pipeline operations.

* **Bug Fixes**
  * Improved latency and throughput estimates for multi-stage pipelines while preserving single-stage behavior.

* **Documentation**
  * Added guidance on pipeline modeling, configuration, integration, and known limitations.


natoscott:
## Multi-backend PP validation data — 5 models, 3 backends, 8xH200

I recently ran an independent PP validation campaign that may be useful context for this PR. The data covers the regime the silicon validation in §6 of `docs/PIPELINE_PARALLEL_MODELING.md` targets (single-node, 8 GPUs held constant across TP/PP splits) but extends it across backends and model sizes.

### Methodology

- **Hardware:** 8x H200 SXM (single node, NVLink domain)
- **Backends:** vLLM 0.9.1, SGLang v0.5.13.post1, TRT-LLM 1.3.0rc18 (pytorch backend)
- **Models:** Qwen3-0.6B, 8B, 14B, 32B-FP8, Llama-3.1-70B-FP8 (TRT-LLM incompatible with this model's FP8 quantization)
- **PP configs:** TP=8/PP=1, TP=4/PP=2, TP=2/PP=4, TP=1/PP=8
- **Workload:** ISL=4000, OSL=1000, max_duration=120s, guidellm concurrent mode
- **Metric:** Peak output tok/s/gpu across concurrency sweep (1, 2, 4, 8, 16, 32), normalized to TP=8/PP=1 baseline

All three backends target the same models on the same hardware with the same workload, so throughput ratios are directly comparable.

### Observed PP efficiency

| Config | vLLM | SGLang | TRT-LLM |
|---|---|---|---|
| **Qwen3-0.6B** | | | |
| TP=4/PP=2 | 0.791 | 0.991 | 0.775 |
| TP=2/PP=4 | 0.563 | 0.559 | 0.135 |
| TP=1/PP=8 | 0.349 | 0.307 | 0.106 |
| **Qwen3-8B** | | | |
| TP=4/PP=2 | 0.798 | 0.815 | 0.759 |
| TP=2/PP=4 | 0.559 | 0.569 | 0.193 |
| TP=1/PP=8 | 0.315 | 0.315 | 0.106 |
| **Qwen3-14B** | | | |
| TP=4/PP=2 | 0.763 | 0.807 | 0.689 |
| TP=2/PP=4 | 0.517 | 0.492 | 0.296 |
| TP=1/PP=8 | 0.308 | 0.264 | 0.156 |
| **Qwen3-32B-FP8** | | | |
| TP=4/PP=2 | 0.783 | 0.761 | 0.679 |
| TP=2/PP=4 | 0.525 | 0.410 | 0.564 |
| TP=1/PP=8 | 0.308 | 0.231 | 0.185 |
| **Llama-70B-FP8** | | | |
| TP=4/PP=2 | 0.789 | 0.662 | -- |
| TP=2/PP=4 | 0.456 | 0.542 | -- |
| TP=1/PP=8 | 0.221 | 0.163 | -- |

### How much of the PP penalty does `balance_factor` explain?

I decomposed the observed efficiency into the `balance_factor` component (stage imbalance from lm_head placement) and a scheduling residual. Using a rough decode-step cost model (per-layer ~ `4h^2 + 3h*intermediate`, lm_head ~ `h*vocab`):

| Model | PP | Total penalty | balance_factor explains | Scheduling residual |
|---|---|---|---|---|
| Qwen3-8B | 2 | 20.2% | 14.7% (73%) | 5.5% |
| Qwen3-8B | 8 | 68.5% | 40.7% (59%) | 27.8% |
| Qwen3-14B | 8 | 69.2% | 33.1% (48%) | 36.1% |
| Qwen3-32B-FP8 | 8 | 69.2% | 23.6% (34%) | 45.6% |
| Llama-70B-FP8 | 8 | 77.9% | 11.5% (15%) | 66.4% |

The `balance_factor` accounts for 15-73% of the total PP penalty, with the fraction *decreasing* for larger models (where per-layer compute dominates lm_head cost and the stages are more balanced). The majority of the penalty — especially for larger models — comes from the scheduling residual flagged in §6.3 of `docs/PIPELINE_PARALLEL_MODELING.md`.

### The scheduling residual has structure

The residual (observed_eff / balance_factor) fits `r = 1 - s * log2(pp)` reasonably well for vLLM and SGLang. The coefficient `s` represents the scheduling/synchronization overhead component:

| Model | vLLM s | SGLang s | TRT-LLM s |
|---|---|---|---|
| Qwen3-8B | 0.158 | 0.150 | 0.283 |
| Qwen3-14B | 0.195 | 0.192 | 0.288 |
| Qwen3-32B-FP8 | 0.212 | 0.250 | 0.256 |
| Llama-70B-FP8 | 0.239 | 0.274 | -- |

Two patterns:
1. **`s` increases with model size** (0.15 -> 0.24 for vLLM). This is counter-intuitive if the overhead were a fixed per-hop ms cost — it suggests the scheduling overhead scales with something beyond simple synchronization (memory bandwidth contention, KV cache management, activation tensor size).
2. **vLLM and SGLang `s` values are close** (within 0.03 for 8B-14B), diverging for 32B+ models. TRT-LLM `s` is consistently higher (~0.27).

### Combined model accuracy

I compared three modeling approaches for vLLM (models 8B-70B, PP=2/4/8):

| Approach | Mean abs error | Max abs error |
|---|---|---|
| Flat empirical coefficient (c=0.228) | 2.7 pp | 9.5 pp |
| balance_factor * (1 - s_per_model * log2(pp)) | 1.8 pp | 4.7 pp |
| balance_factor * (1 - s_mean * log2(pp)) | 4.6 pp | 14.2 pp |

The combined model with per-model `s` is the most accurate, halving max error. But using a single mean `s` per backend is actually *worse* than the flat coefficient — `s` varies enough with model size that collapsing it into one number introduces more error than the balance_factor removes.

For SGLang: flat c gives 5.2/11.7 pp mean/max error; combined per-model gives 3.7/10.7 pp.

For TRT-LLM: all approaches struggle (6-8 pp mean, 16-21 pp max) due to the non-monotonic PP=4 behavior described below.

### TRT-LLM PP=4 cliff

TRT-LLM shows a non-monotonic PP efficiency pattern across all model sizes tested:

| Model | PP=2 c | PP=4 c | PP=8 c |
|---|---|---|---|
| Qwen3-0.6B | 0.225 | **0.432** | 0.298 |
| Qwen3-8B | 0.241 | **0.404** | 0.298 |
| Qwen3-14B | 0.311 | **0.352** | 0.281 |
| Qwen3-32B-FP8 | 0.321 | 0.218 | 0.272 |

(Per-PP-level coefficient `c` from `eff = 1 - c * log2(pp)`)

For small/medium models (0.6B-14B), the PP=4 per-level coefficient spikes to 0.35-0.43 — meaning PP=4 is disproportionately worse than either PP=2 or PP=8 would predict from a smooth curve. The per-concurrency ITL data confirms this is consistent across all concurrency levels (not a measurement artifact):

```
TRT-LLM Qwen3-8B conc=1 ITL:
  TP=8/PP=1:  2.9 ms
  TP=4/PP=2:  5.7 ms  (2.0x)
  TP=2/PP=4: 17.1 ms  (5.9x)  <-- step function
  TP=1/PP=8: 18.7 ms  (6.4x)
```

This pattern is absent from vLLM and SGLang (smooth degradation) and may relate to TRT-LLM's MPI-based pipeline implementation handling the TP=2 per-stage configuration differently. The effect diminishes for 32B-FP8 where per-stage compute is larger relative to the overhead.

This appears related to the non-monotonic saturation behavior identified in §6.3 of `docs/PIPELINE_PARALLEL_MODELING.md` — the scheduling overhead may depend on the interaction between TP and PP configuration within each stage, not just PP alone.

### Implications

The `balance_factor` is structurally correct — it captures the right mechanism (lm_head imbalance). The data shows:

1. **The scheduling residual is the dominant component** for models >= 14B (52-85% of total penalty). Any calibration effort should prioritize this over refining balance further.
2. **The residual is backend-specific.** vLLM and SGLang share similar PP scheduling behavior; TRT-LLM is materially different. This has implications for whether the correction belongs in `BaseBackend` vs per-backend overrides.
3. **The residual scales with model size** in a way a single per-backend constant doesn't capture. The `fill_factor` hook may be the right place to absorb this, but the relationship needs characterization.
4. **TRT-LLM's PP=4 cliff suggests the overhead depends on the TP*PP interaction**, not just PP alone.

The raw benchmark data (224 result files across all backends) and analysis scripts are available if useful for calibration.

PR #1461, opened 2026-08-02
feat: run forward_model="fpm" on the compiled Rust engine (Op::FpmForward)

liyuanzhe1991:
#### Overview:

Run `forward_model="fpm"` on the compiled Rust engine. The Python FPM
implementation (#1384) modeled whole-model forward passes from collected
`fpm_forward_perf` data but was force-routed to the Python engine step
("FPM models compile to no Rust op variant yet"). This PR is the M2/M3
Rust half of the approved plan (`docs/fpm/aic-fpm-modeling-plan.md`): a
native `Op::FpmForward` whole-model variant, so FPM sweeps get the same
compiled-engine speedup as op_level — validated **bit-exact** against the
Python reference on a real 20-cell MiniMax-M2.7 database.

> **Stacked on #1384** (`feature/fpm-modeling`). Until it merges, this
> diff includes its commits; the new work is the 4 commits after
> `badc7920`. Will rebase onto main once #1384 lands.

#### Details:

- **`perf_database/fpm_forward.rs`** (new): strict loader for the collector
  pair `fpm_forward_perf.parquet` + sha256'd `.metadata.json` (schema v5
  gates, per-row validation, duplicate physical-row-key rejection, cell
  grouping, per-phase domain bounding boxes, prebuilt `SiteIndex`).
  Deliberately no shared-layer inheritance — FPM data is valid only for its
  exact backend/version, mirroring Python.
- **`perf_interp.rs`**: adds Python's `own_curve_coverage_fallback` flag
  (default `false`; GEMM behavior byte-identical) and floors site logs at
  `1e-12` like Python `_site_index` — the old `max(1)` shortcut pushed
  legitimate KV=0 prefill sites ~40 log2-units away from their own queries.
- **`operators/fpm_forward.rs` + `operators/fpm_sol.rs`** (new):
  `Op::FpmForward` with the exact Python query semantics (iteration-total
  coordinates, inclusive domain gate BEFORE interpolation — FPM never
  extrapolates — `query_pass_baseline` KV-floor); the SOL roofline mirrors
  Python's `DatabaseMode.SOL` chain per op family with float coordinates
  and Python's floor/ceil placement, evaluated lazily (exact hits never
  invoke it).
- **`engine/runtime.rs`**: FPM branches for the mixed step (marginal-decode
  composition, Python's `s = isl + osl//2 + 1` convention), genonly step,
  and ForwardPassMetrics rank dispatch (queries raw iteration totals);
  `Engine::build` validates spec shape recursively and rejects MTP.
- **Wire**: `ENGINE_SPEC_SCHEMA_VERSION` 2 → 3 in lockstep
  (`config.rs` + `engine.py`); `_to_opspec` gains the recursive
  `FpmForward` converter (identity strings normalized on the producer side;
  `sol_ops` carry the original granular list).
- **`base_backend.py`**: the three `forward_model != "fpm"` force-python
  gates are lifted; FPM routes through `should_use_rust_engine_step` like
  op_level. `rust_engine_step` re-raises Rust perf-DB-miss ValueErrors as
  `PerfDataNotAvailableError` so sweeps keep one error taxonomy across
  routes; sweeps treat FPM out-of-domain points as unanswerable.
- **Review hardening**: an 18-finding adversarial review was run and all
  confirmed items fixed or documented (engine-handle cache key now carries
  raw quant names incl. comm; lazy SOL unblocks DSA-family models on grid
  points; loader null/ordering/JSON-float parity; saturating adds; inner
  RuntimeConfigs inherit `engine_step_backend`). Deliberate divergences are
  documented in `docs/fpm/aic-fpm-rust-port-design-20260802.md`.

**Validation**
- Full-grid acceptance on the real merged M2.7 database (20 cells /
  10 topologies): 36,986 grid points, 238 in-domain off-grid probes,
  30 mixed + 10 genonly steps — Python and Rust agree **bit-exactly**
  (max rel error 0.0); 2 out-of-domain probes error symmetrically.
- `cargo test`: 225 passed (plus the 6 pre-existing baseline failures
  unrelated to this PR); unit suite 2220 passed; parity CI suites 288
  passed including 14 new FPM-mode cases over a synthetic dataset.

**Known scope limits** (documented, follow-ups tracked): DSA (GLM-5.2) /
MSA (MiniMax-M3) / MLA-module SOL formulas are not ported — exact-hit and
in-curve FPM queries work for those models; interpolation-transfer paths
miss with a structured error. MiniMax-M3's MSA ops have no OpSpec
converter yet, so M3+fpm requires `engine_step_backend="python"`.

1. `aic-core/rust/aiconfigurator-core/src/operators/fpm_forward.rs` — the
   whole-model op; compare against the Python reference
   `aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py`.
2. `aic-core/rust/aiconfigurator-core/src/operators/fpm_sol.rs` — the
   SOL-mode formulas (provenance file:line comments point at each Python
   `get_sol`).
3. `aic-core/rust/aiconfigurator-core/src/engine/runtime.rs` — the FPM
   mixed/genonly/rank branches vs `base_backend._get_fpm_mix_step_latency`.
4. `aic-core/rust/aiconfigurator-core/src/perf_database/perf_interp.rs` —
   the shared-engine change (`own_curve_coverage_fallback` + site-log
   floor); confirm GEMM is untouched.
5. `aic-core/src/aiconfigurator_core/sdk/rust_engine_step.py` — the error
   translation and engine-handle cache key.
6. `docs/fpm/aic-fpm-rust-port-design-20260802.md` — design contract,
   review addenda, and documented divergences.

- Relates to #1384 (stacked on it: Python FPM modeling, the semantic
  reference this port is validated against)
- Relates to #1365 (FPM collector — producer of the `fpm_forward_perf`
  data contract this consumes)

## Summary by CodeRabbit

- **New Features**
  - Added selectable whole-model forward-pass modeling with `op_level` and `fpm` modes.
  - Added the `--forward-model` CLI option and configuration support.
  - Added FPM prefill, decode, mixed-step, interpolation, latency estimation, and weight reporting.
  - Added validated performance-data loading with checksum and metadata checks.

- **Bug Fixes**
  - Improved model selection and cache separation.
  - Handled unavailable performance data during configuration sweeps.
  - Prevented unsupported FPM configurations and out-of-domain requests.

- **Documentation**
  - Documented forward-model options, data requirements, interpolation behavior, and compatibility limitations.


simone-chen (commented):
Two integration findings. I think we can merge it after these are fixed, and then I can connect the aic-fpm backend to fpm-arena

aic-core/rust/aiconfigurator-core/src/engine/runtime.rs
```diff
@@ -932,6 +1134,57 @@ impl Engine {
         let has_prefill = sched.num_prefill_requests > 0;
         let has_decode = sched.num_decode_requests > 0;
```
simone-chen: A fully prefix-cached payload can retain prefill request/KV metadata while scheduling no fresh prefill compute. For example:

```text
num_prefill_requests = 1
sum_prefill_tokens = 0
sum_prefill_kv_tokens = 4096
num_decode_requests = 8
```

This iteration is decode-only. The count-based check instead queries prefill at zero (outside the FPM domain) and treats decode as marginal work riding a prefill pass. Align with `IterationFeatures`:

```suggestion
        let has_prefill = sched.sum_prefill_tokens > 0;
        let has_decode =
            sched.num_decode_requests > 0 || sched.sum_decode_kv_tokens > 0;
```

aic-core/src/aiconfigurator_core/sdk/engine.py
```diff
@@ -656,6 +659,27 @@ def recurse(child: Any) -> dict:
             }
         }
 
+    # Whole-model FPM op (forward_model="fpm"): recursive like the composites —
+    # `sol_ops` carries the model's original granular list as the roofline
+    # source. The identity strings were normalized by _norm_identity at op
+    # construction; Rust compares them verbatim.
+    if isinstance(op, FPMForwardOp):
```
simone-chen, line 673: `RustForwardPassPerfModel.from_native()` currently calls compile_engine() without forward_model, so it always creates the existing op_level engine. FPM arena cannot select FPM through AIC’s supported predictor API.

simone-chen (approved):

aic-core/rust/aiconfigurator-core/src/engine/runtime.rs
liyuanzhe1991: Fixed in a7f46a19: `rank_latency_ms` now dispatches on tokens, aligned with `IterationFeatures` (`has_prefill = sum_prefill_tokens > 0`; `has_decode = num_decode_requests > 0 || sum_decode_kv_tokens > 0`). Your exact payload — prefill metadata retained, zero fresh prefill tokens, 8 decode requests — is now a decode-only dispatch test on the FPM engine (full decode latency, no zero-token prefill query, no marginal composition).

aic-core/src/aiconfigurator_core/sdk/engine.py
liyuanzhe1991: Fixed in a7f46a19: `EngineConfig` gains an additive-optional `forward_model` field, threaded through `EngineBuildRequest` / `AicEngineBuilder` into `compile_engine` (new keyword; the public-api signature pin updated with it). An e2e parity test builds `RustForwardPassPerfModel.from_native({..., "forward_model": "fpm"})` and asserts a decode-only estimate hits the `fpm_forward` table's exact row — the FPM arena now selects the whole-model engine through the supported predictor API.

PR #1470, opened 2026-08-03
feat(collector): add unified large-EP moe_a2a and moe_expert_compute producers

Arsene12358:
## Overview

This is PR 3, the final producer-side change in the model-agnostic large-EP refactor series (Linear AIC-1440). It targets `main` directly at exact head `31c2b3fbede5be259dfa767ec16195d58da42a03`, after a normal merge of current `main` (`11354af37474f6e991c24af6692171becce8a5f7`) into the existing branch. No history was rewritten.

It adds the producer half of the unified large-EP data contract already consumed on `main`:

- generalized `moe_ep` expert-compute collectors for SGLang and TensorRT-LLM, with a dormant vLLM path pending an image pin;
- a standalone multi-node `moe_a2a` collector for DeepEP HT/LL;
- TensorRT-LLM all-to-all emission into the unified schema;
- Collector V3 provenance, manifest-pinned runtime identity, Slurm launchers, collection documentation, and retirement of the older manual wideEP pipeline.

The canonical table names are `moe_a2a_perf` and `moe_expert_compute_perf`.

## Contract and scope

This is the declared producer half of the coordinated data-contract change described in `.claude/rules/collector/layer_permissions.md`. The consumer SDK/Rust implementation already on `main` remains canonical.

After integrating `main`, the PR is narrowed to 44 files (7,155 additions, 873 deletions):

- 31 files under `collector/**`;
- 12 files under `tests/unit/collector/**`;
- one explicitly coordinated two-sided schema test: `tests/unit/sdk/database/test_collector_schema_contract.py`;
- no SDK or Rust production-code changes.

No data is collected in this PR. It contains code, cases, tests, launchers, and docs only.

## Review follow-ups included

The exact-head tree retains and tests the original ten resolved review fixes, plus six final CodeRabbit threads:

1. preserve SGLang model identity;
2. classify missing grouped token points as failures;
3. benchmark the declared `moe_a2a` routing workload;
4. attest the actually launched container image and variant;
5. treat `log_perf == false` as a classified persistence failure and synchronize distributed writers;
6. fail closed on stale append-mode staging artifacts;
7. deduplicate TensorRT-LLM synthetic invocations by runtime/consumer identity;
8. prevent diagnostic LL transports from producing publishable data;
9. avoid zero-case default launcher worlds;
10. complete and positively test the dormant vLLM activation checklist;
11. correct Slurm defaults and job-scoped output paths;
12. reject incompatible resumed CSV schemas without modifying the existing file;
13. document the SGLang CSV example correctly and report all unconditional token drops;
14. split the chained `pytest.approx` assertion;
15. pin the actual `MOE_EP_HEADER` collector symbols in the two-sided schema test;
16. propagate every `log_perf` lock/schema/write failure as a typed task failure, including legacy callers that do not inspect its success value.

All 16 review threads are resolved.

## Validation at exact head

| Gate | Result |
|---|---|
| Main integration | merge commit `202176c6`; parents `27e0f8b7` + `11354af3` |
| Final review fixes | signed-off commits `31fd39c7` and `31c2b3fb` |
| Diff scope/hygiene | 44 files; no unexpected outside-scope file; `git diff --check` clean |
| Affected final-fix suite | 74 passed |
| Focused review-regression suite | 250 passed |
| Full collector + schema + collector-data checks | 811 passed, 6 skipped |
| Cross-package contract suite | 73 passed |
| Full unit suite | 4,095 passed, 50 skipped |
| Ruff | repository lint clean; 596 files format-clean |
| Launchers | both Slurm scripts pass `bash -n` |

Fresh GitHub CI is running against exact head `31c2b3fb` and remains the merge gate.

## Reviewer starting points

- `collector/wideep/sglang/collect_moe_a2a.py` — distributed identity, HT/LL benchmarks, persistence, and sidecar lifecycle.
- `collector/wideep/sglang/collect_deepep_moe.py` — SGLang expert-compute producer.
- `collector/wideep/trtllm/collect_moe_compute.py` — TensorRT-LLM expert-compute producer and deduplication.
- `collector/network/slurm/collect_trtllm_alltoall.py` — unified TensorRT-LLM all-to-all emission.
- `tests/unit/sdk/database/test_collector_schema_contract.py` — the two-sided producer/consumer schema contract.
- `collector/network/slurm/README.md` — launch and publish procedure.

## Deferred hardware verification

The first collection run through the auto-collect playbook must still perform runtime smoke tests in the pinned images, exercise real-data consumer queries, and validate the documented HT/LL hardware risks. Passing unit/contract tests here is not a claim that new measurements already exist.

## Related

- Relates to Linear AIC-1440
- Series: #1442, #1448, #1464, #1470

## Summary by CodeRabbit

- **New Features**
  - Added standalone MoE all-to-all collection for SGLang and TensorRT-LLM, including multi-node Slurm workflows.
  - Added unified Parquet outputs with metadata, provenance sidecars, and failure reporting.
  - Added large-EP model cases, workload planning, and dormant vLLM expert-compute support.

- **Documentation**
  - Updated guidance for `moe_ep`, runtime requirements, output formats, migration, and data merging.

- **Compatibility**
  - Standardized MoE operation names, schemas, latency handling, and performance files across frameworks.


Arsene12358:
Rebased onto main @ dc4caca via the rebased #1464. The collector tree absorbed main #1356's trtllm 1.3.0rc20 upgrade of the very collector this PR retargets, so the carries are rc20-flavored (`add51e8`):

- The rc20 `_SlotLoraPaddedRunner` TorchBind pad shim, its run-time installer, and the `trtllm>=1.3.0rc20` compat pin arrived via the merge and are kept verbatim; the rc5-era SM120 nvfp4 generation-time skips stay deleted (main hardware-verified them clean, and they were silent filters `layer_permissions.md` forbids).
- Population pins re-derived at the rc20 reality: the fp8_block floor moved to min_sm 90 (SM87-89 lost it, SM100/SM120 gained it alongside nvfp4 → 84 cases each for DSV3), and the `(True, 288)` EPLB layout now exists only for models with ≤288 experts (replication identity `num_slots >= num_experts`), so the full declared sweep is 770 (was 820; the five 384-expert entries — the Kimi trio and the DeepSeek-V4-Pro pair — collect 28 instead of 38 each). The inline retired-getter transcription in the tests carries the same chain, so the kept/added/removed parity check stays independent.
- The closed drop-accounting log now accumulates the conditional EPLB axis per recipe instead of assuming ×3, keeping `expanded − slot drops == emitted` arithmetically closed.
- `submit_trtllm_alltoall.sh`'s default image resynced to the manifest's rc20 digest — flagged by exactly the manifest-sync drift-guard test this PR introduced.

Fresh gates at `add51e8`: collector suite 705, `tools/perf_database/check_collector_data.py` (renamed from `audit_collector_data.py` by #1420) R1–R6 all pass, strict unit 3254, cross-package 70, harness **73/0/0** vs the dc4caca baselines, ruff clean, launchers `bash -n` clean.

collector/wideep/sglang/collect_deepep_moe.py
```diff
@@ -951,54 +1117,106 @@ def run_moe(
...
+        )
 
+    print(
+        f"moe_ep: {len(cases)} cases from {len(recipes)} moe recipes "
+        f"(dropped: {dropped_not_declared} not declared wideep, "
+        f"{dropped_not_ep} not tp=1/ep>1, {dropped_quant} quant not allowed; "
+        f"{len(recipes) - dropped_not_declared - dropped_not_ep - dropped_quant - len(cases)} "
+        f"deduplicated onto (model, ep))"
+    )
+    return [cases[key] for key in sorted(cases)]
```
jasonqinzhou, line 1271: **[P1] Keep model identity in each SGLang `moe_ep` task.** The dictionary is keyed by `(model_name, ep_size)`, but this return discards the model key and schedules only the eight shape arguments. With no `COLLECTOR_MODEL_PATH`, the current getter emits 62 cases spanning five geometries but only 46 unique argument lists, while `_selected_moe_model_id()` defaults every subprocess to `deepseek-ai/DeepSeek-V3`. Non-DeepSeek cases therefore load the wrong checkpoint and fail the declaration assertions; duplicate cross-model arguments also share task identity. Please thread `recipe.model_name` through `run_moe_ep` and the subprocess into the model resolver, or constrain expansion to an explicitly selected model before scheduling. Add a registry-surface test covering at least two model families and the unfiltered/full path.

collector/wideep/trtllm/collect_moe_compute.py
```diff
@@ -911,6 +1064,11 @@ def run_wideep_moe_compute(
     # =========================================================================
     for num_tokens in num_tokens_lists:
         if num_tokens > max_tokens:
+            # Runtime token-point drop, logged with its reason: the dry run
+            # above established that this token count cannot be allocated/run
+            # on this device. Drops one token point inside the queued case,
+            # never the case.
+            print(f"moe_ep: dropping num_tokens={num_tokens} because the dry run failed above {max_tokens} tokens")
```
jasonqinzhou: **[P1] Classify failed inner token points instead of shrinking a successful task.** `num_tokens_lists` is part of the queued outer case, but this path omits every declared token above the dry-run ceiling and then returns success. The executor checkpoint and V3 case-plan hash still represent the full list, with no failure record, so a partial curve is indistinguishable from complete coverage. The SGLang twin has the same contract break at its context/decode `continue` paths. Please split token points into tasks, attach an exact retained inner manifest at generation time, use the sanctioned live-memory generation filter, or raise a classified case failure; logging the drop is not sufficient.

collector/wideep/sglang/collect_moe_a2a.py
```diff
@@ -0,0 +1,1289 @@
...
+    is the framework's own path and is used directly.
+    """
+    import deep_ep
+    import torch
+    import torch.distributed as dist
+
+    shape = case.shape
+    num_ranks = identity.world_size
+    num_nodes = identity.node_num
+    num_topk_groups = min(num_nodes, HT_MAX_TOPK_GROUPS)
```
jasonqinzhou: **[P1] Preserve serving routing semantics in the `moe_a2a` workload.** `get_moe_a2a_shapes()` collapses model recipes to `(hidden_size, topk, num_experts)`, then this line applies DeepSeek-style `min(node_num, 4)` node-group routing to every retained shape. Current Kimi and GLM declarations carry `sglang_moe_num_expert_group=1` / `sglang_moe_topk_group=1`, so on a multi-node world this changes the destination set and traffic pattern while the resulting rows have no routing identity. Please either correlate declared routing into the case plan and coordinate the necessary consumer key, or explicitly approve and document one canonical synthetic workload and exclude shapes it does not represent. Add a cross-family population/workload test.

collector/wideep/sglang/collect_moe_a2a.py
```diff
@@ -0,0 +1,1289 @@
...
+    try:
+        installed_public = Version(installed_version).public
+    except InvalidVersion as error:
+        raise MoeA2ADeclarationError(f"invalid installed sglang version {installed_version!r}") from error
+    if installed_public != Version(runtime.version).public:
+        raise MoeA2ADeclarationError(
+            f"moe_a2a collection requires sglang {runtime.version} (manifest wideep_sglang pin), "
+            f"found {installed_version}; use {runtime.image()}"
+        )
+    image, image_digest = _split_image_digest(runtime.image())
```
jasonqinzhou: **[P1] Attest the container image that actually launched the collector.** Calling `runtime.image()` with no variant records `deepseek-v4-blackwell`, while the GB200 launcher defaults to the manifest’s `deepseek-v4-grace-blackwell` variant. The two tests independently pin those conflicting values, so every default run would carry incorrect runtime provenance. In addition, both launchers allow a `CONTAINER_IMAGE` override without plumbing the actual ref/digest into the sidecar. Please pass and validate the launched image, including its manifest variant/digest, through the collector/finalizer and add a launcher-to-sidecar identity test.

collector/wideep/sglang/collect_moe_a2a.py
```diff
@@ -0,0 +1,1289 @@
...
+) -> None:
+    """Emit the case's two rows, ``combine`` before ``dispatch`` (D5).
+
+    The consumer's store nests ``phase`` directly under ``comm_backend``, so
+    alphabetical phase emission keeps that level's insertion order ascending
+    too.
+    """
+    for phase in (PHASE_COMBINE, PHASE_DISPATCH):
+        timing = timings[phase]
+        log_perf(
```
jasonqinzhou: **[P1] Treat `log_perf` failure as a classified case failure.** `collector.helper.log_perf` catches lock/open/write/fsync errors and returns `False`; it does not raise. This return is discarded here, by the TensorRT-LLM standalone writer, and by the new registered `moe_ep` writers. Once any earlier row exists, a later failed write can therefore leave `failure_count == 0`, finalize a partial parquet, and emit `status: complete`. Please check every new writer’s result and convert `False` into a classified task failure. For distributed writers, synchronize the rank-0 persistence result before peers enter the next collective, and add a regression that forces `log_perf` to return false.

collector/network/slurm/submit_moe_a2a.sh
```diff
@@ -0,0 +1,143 @@
...
+    if [ $((NUM_GPUS % GPUS_PER_NODE)) -ne 0 ]; then
+        echo "Error: ${NUM_GPUS} GPUs is not a multiple of ${GPUS_PER_NODE} GPUs/node"
+        exit 1
+    fi
+    NUM_NODES=$((NUM_GPUS / GPUS_PER_NODE))
+    TASKS_PER_NODE=${GPUS_PER_NODE}
+
+    JOB_NAME="${ACCOUNT}-moe_a2a.${NUM_GPUS}gpu"
+    OUTPUT_DIR="${SCRIPT_DIR}/results/moe_a2a_${NUM_GPUS}gpu"
+    mkdir -p "${OUTPUT_DIR}"
```
jasonqinzhou: **[P1] Fail closed on stale append-mode staging artifacts.** This deterministic output directory is reused, while `log_perf` opens the staging CSV in append mode and the launcher performs no attempt/resume validation. After a preemption, resubmitting appends a full run after stale partial rows; finalization can then produce duplicates while the new sidecar attests only the current plan and reports complete. Mode-specific SGLang runs also share this directory. Please use unique attempt/job-id directories or refuse any owned artifacts unless an explicit, validated resume protocol is selected. Apply the same protection to the TensorRT-LLM launcher and add an interrupted-rerun regression.

collector/wideep/trtllm/collect_moe_compute.py
```diff
@@ -678,13 +812,26 @@ def get_wideep_moe_compute_all_test_cases():
                     ]
                 )
 
+    # D5 sorted emission: non-token key axes first, deterministic regardless
+    # of the recipe product order.
+    test_cases.sort(key=lambda case: (case[10], case[0], case[8], case[12], case[13], case[14]))
```
jasonqinzhou, line 817: **[P2] Deduplicate TensorRT-LLM synthetic invocations on runtime and consumer identity.** The corrected SM90 getter now returns 770 tasks, but only 224 real run-argument tuples remain after removing `model_name`, which `run_moe_ep` does not use. The other 546 tasks rerun identical synthetic kernels and emit rows with identical consumer keys, wasting substantial GPU time and making the loaded value depend on duplicate row order/noise. Please schedule one case per real invocation plus persisted-key identity, retain the contributing models in provenance/logging separately, and test that the full plan has no duplicate runtime/row identities.

collector/wideep/sglang/collect_moe_a2a.py
```diff
@@ -0,0 +1,1289 @@
...
+        help="GPUs per node in this allocation; node_num = WORLD_SIZE // gpus-per-node (a persisted key column)",
+    )
+    parser.add_argument(
+        "--modes",
+        default=f"{COMM_BACKEND_HT},{COMM_BACKEND_LL}",
+        help="comma-separated DeepEP kernel families to collect",
+    )
+    parser.add_argument("--output-path", default=os.getcwd())
+    parser.add_argument("--allow-mnnvl", action="store_true", help="allow MNNVL for the low-latency buffer")
+    parser.add_argument("--disable-nvlink", action="store_true", help="disable NVLink for the low-latency buffer")
```
jasonqinzhou: **[P2] Prevent alternate low-latency transports from colliding with default data.** `--disable-nvlink` and `--allow-mnnvl` materially change the LL `deep_ep.Buffer` construction, but neither the persisted row identity nor `case_plan_ids()` nor the runtime sidecar records those settings. A diagnostic transport run can therefore overwrite or mix with default `deepep_ll` data under an identical provenance hash. If these switches are diagnostic-only, refuse to finalize publishable data when they are set; otherwise coordinate an approved transport identity across rows, case IDs, sidecars, and consumers.

collector/network/slurm/submit_moe_a2a.sh
```diff
@@ -0,0 +1,143 @@
...
+# The value is duplicated here because login nodes cannot be assumed to have
+# the repo's Python deps; it is kept in sync with the manifest by
+# tests/unit/collector/test_network_layout.py::test_submit_moe_a2a_default_image_matches_manifest.
+CONTAINER_IMAGE="${CONTAINER_IMAGE:-deepseek-v4-grace-blackwell}"
+CONTAINER_MOUNTS="${CONTAINER_MOUNTS:-${REPO_DIR}:${REPO_DIR}}"
+ACCOUNT="${ACCOUNT:-coreai_tritoninference_triton3}"
+PARTITION="${PARTITION:-gb200}"
+
+# Defaults
+GPU_LIST="8,16,32,48,64,72"
```
jasonqinzhou: **[P2] Do not submit zero-case worlds from the default launchers.** Current GPU-free plan expansion shows that the advertised DeepEP world of 72 drops all six shapes because none has an expert count divisible by 72, so that full-rack job reserves resources only to fail. The TensorRT-LLM default list has the same problem at 48 and 72 for its sole 256-expert shape. Please remove unsupported defaults or preflight every requested world before `sbatch`, with a test asserting that each advertised default expands to at least one case.

collector/wideep/README.md
```diff
@@ -13,8 +13,68 @@ runtime independently from the non-WideEP framework entry.
...
+runtime is pinned. Its dormancy is itself under test
+(`tests/unit/collector/test_vllm_collect_moe_ep.py`): no registry module, no
+manifest family, no hash-closures entry may exist before enrollment.
+Enrollment is one coordinated change:
+
+1. Pin a vLLM-DeepEP image: add a `wideep_vllm` entry to
+   `collector/framework_manifest.yaml` (`base_framework: vllm`,
+   `collector_dir: "collector/wideep/vllm"`, `data_backend: vllm`, a
+   `default.version` and digest-pinned `images`).
+2. Create `collector/wideep/vllm/registry.py` with
```
jasonqinzhou, line 59: **[P2] Complete the dormant vLLM activation checklist.** Creating this registry file is not enough to enroll the framework: `collector.framework_manifest._REGISTRY_MODULES` also needs a `wideep_vllm` entry or registry/runtime resolution raises, and `collector/kernel_source_backends.yaml` currently maps `deepep_moe` only for `framework: sglang`, leaving activated vLLM rows without a verified backend mapping. Please add both enrollment surfaces to this coordinated checklist and to the positive activation tests so following the documentation produces a valid provenance/catalog configuration.

jasonqinzhou (changes requested):

## What this PR does

This final stacked PR adds the producer half of the unified large-EP data contract: generalized SGLang and TensorRT-LLM moe_ep collectors, standalone DeepEP and TensorRT-LLM moe_a2a collection, Collector V3 provenance and launchers, SDK null-power handling, documentation, and retirement of the older wideEP pipeline surfaces. The current head rebases the unchanged collector patch stack onto the updated feat/large-ep-rust base, where the prior null-power loader follow-up now lives.

The update cleanly preserves all twelve remaining PR patches across the newer SDK/Rust base and removes the now-redundant null-power commit from the stack, while the expanded loader, phase-validation, collector-contract, and parity tests all stay green.

## Why this score

The recommendation remains REQUEST_CHANGES at 5.0/10. The head update changes only the merge-base integration and does not touch any prior finding anchor, so six P1 producer-safety issues remain: SGLang can still lose model identity, grouped token points can disappear without failures, moe_a2a can benchmark the wrong routing workload, sidecars can attest the wrong image, write failures can finalize partial data as complete, and interrupted reruns can mix stale rows. Four P2 identity, efficiency, launcher, and activation-contract issues also remain; no new PR-owned finding was introduced by the update.

collector/wideep/sglang/collect_deepep_moe.py
Arsene12358: Fixed. The SGLang `moe_ep` case list now carries `model_name` as its 9th argument, threaded `run_moe_ep` -> subprocess -> `run_moe_benchmark`, and the subprocess resolves the CASE's model (`_case_model_path`): `COLLECTOR_MODEL_PATH` (single-model plans) wins, otherwise the declared model downloads/loads — never a defaulted DeepSeek-V3 checkpoint for another family's case. Cross-model duplicate argument lists are now distinct task identities. Registry-surface regression: `test_full_population_covers_every_declared_wideep_model` asserts the full path spans every declared family (>= 2), that shape-colliding cases stay distinct, plus `test_cases_resolve_their_own_declared_model` for the resolver precedence.

collector/wideep/trtllm/collect_moe_compute.py
Arsene12358: Fixed, via your fourth option (classified case failure) plus a generation-time manifest for the deterministic twin:
- trtllm `collect_moe_compute`: the dry-run walk-down is gone. The largest declared token point either dry-runs or the case raises `MoeEpBenchmarkError` (classified, recorded); the `num_tokens > max_tokens` runtime drop is deleted. The autotuner keeps its shrink (tuning-quality only — every declared point is still benchmarked).
- sglang context: the empty-uniform predicate (`num_token * topk * ep // num_experts == 0`) is pure declared arithmetic, so it moved into `get_moe_prefill_test_cases` (exact retained manifest, logged drops); reaching the runtime branch now raises as an invariant breach.
- sglang decode: the stochastic masked_m buffer exceedance raises classified instead of thinning the curve.
Pins: `test_no_silent_case_skipping_and_the_dry_run_raises` (trtllm) / `test_no_silent_case_skipping_in_the_benchmark_loops` (sglang) assert the drops are gone, `test_prefill_manifest_excludes_empty_uniform_workloads_at_generation` covers the manifest.

collector/wideep/sglang/collect_moe_a2a.py
Arsene12358: Fixed with declared routing. `MoeA2AShape` now carries the model rows' `sglang_moe_num_expert_group`/`sglang_moe_topk_group` as non-identity metadata; `get_moe_a2a_shapes` raises loudly when two models sharing a (hidden, topk, experts) key declare different routing (none do today). `run_ht_case` derives its node-group budget from the declaration (`ht_num_topk_groups`): DeepSeek-style rows keep `min(nodes, topk_group=4)` — bit-identical to the old behavior — while Kimi/GLM rows (1/1) get global routing (the masking degenerates to plain top-k), which is their serving traffic. The routing identity is in the case plan (`case_plan_ids` payload carries both fields); the persisted key is unchanged, so no consumer coordination is needed. Cross-family tests: `test_shapes_carry_the_declared_routing`, `test_conflicting_routing_declarations_raise`, `test_ht_topk_group_budget_derives_from_the_declaration`, `test_case_plan_ids_carry_the_routing_identity`.

collector/wideep/sglang/collect_moe_a2a.py
Arsene12358: Fixed. Both launchers pass `--image-ref "${CONTAINER_IMAGE}"` (exactly what srun --container-image runs, overrides included), and both collectors' `resolve_runtime_meta` now require it, validate it against the manifest's pinned image variants (unknown ref -> declaration error, no fallback to the manifest default), and attest `image` + `image_variant` (+ digest when pinned) in the sidecar. The GB200 default run now records `grace_blackwell`, not the wrong `default` variant. Launcher-to-sidecar identity tests: `test_submit_launchers_pass_the_launched_image_to_the_collector`, `test_runtime_meta_records_the_launched_image_variant`, `test_runtime_meta_requires_the_launched_image`, `test_runtime_meta_rejects_an_image_the_manifest_does_not_pin` (+ trtllm twins).

collector/wideep/sglang/collect_moe_a2a.py
Arsene12358: Fixed. Every new writer checks `log_perf`'s result and converts `False` into a classified failure, in the repo's established `if not log_perf(...): raise` shape: the sglang moe_ep context/generation writers, the trtllm moe_ep writer, the (dormant) vllm writer, `_emit_case_rows` in the a2a collector, and the trtllm alltoall writer. For the distributed writers, rank 0's persistence result is agreed before peers enter the next collective: the a2a collector emits inside the per-case try so the existing all_reduce(MAX) covers it; the trtllm alltoall loop all_reduces a dedicated `persist_failed` flag (recorded as `alltoall_persistence`). Regression: the six writers joined `test_log_perf_false_directly_raises`'s AST parametrization, which forces the guard-with-raise shape on every `log_perf` call site.

collector/network/slurm/submit_moe_a2a.sh
Arsene12358: Fixed, fail-closed on both sides. Collector-side: both standalone collectors refuse to start when the output dir already holds owned artifacts (`<stem>.*`, `collection_meta.yaml`, `errors_*.json` — `helper.stale_output_artifacts`), with the message pointing at a fresh `--output-path` and stating no validated resume protocol exists. Launcher-side: both launchers write into a job-scoped subdirectory (`.../job_${SLURM_JOB_ID}`, resolved inside the job), so a resubmission gets a fresh dir and a same-jobid requeue hits the refusal instead of appending after stale rows. Interrupted-rerun regression: `test_stale_output_artifacts_fail_closed` (seeded stale CSV + errors file -> refusal; clean dir -> passes) plus the launcher-side `test_submit_launchers_use_per_job_output_dirs`.

collector/wideep/trtllm/collect_moe_compute.py
Arsene12358: Fixed. The trtllm getter dedups onto the runtime identity (the full argument tuple minus `model_name`, which `run_moe_ep` uses only as provenance) after the D5 sort, so the kept representative is deterministic; the population log now reconciles the dedup count and retains every contributing model (`... - N deduplicated onto the runtime identity (contributing models: ...)`). Duplicate synthetic reruns and order-dependent duplicate rows are gone. Tests: `test_full_population_covers_every_declared_wideep_model` asserts runtime-identity uniqueness on the full plan and family coverage via the logged contributors; the single-model SM90 log pin gains the `- 0 deduplicated` arithmetic.

collector/wideep/sglang/collect_moe_a2a.py
Arsene12358: Fixed as diagnostic-only. `--allow-mnnvl`/`--disable-nvlink` are now marked DIAGNOSTIC ONLY in help, and a run with either set refuses to finalize: rows stay in the staging CSV, no parquet, no sidecar (`transport_is_default` gates the rank-0 finalize block, printed loudly). Combined with the F19 stale-artifact refusal, a diagnostic run's staging CSV also cannot mix into a later default run in the same directory. Test: `test_alternate_ll_transports_refuse_to_finalize`.

collector/network/slurm/submit_moe_a2a.sh
Arsene12358: Fixed. The advertised defaults now contain only worlds that expand: submit_moe_a2a.sh drops 72 (no declared expert count divides 72; 48 keeps the two 384-expert shapes), submit_trtllm_alltoall.sh drops 48 and 72 (sole 256-expert shape). Both scripts document why, and `test_advertised_default_worlds_expand_to_at_least_one_case` expands every advertised default against the real declarations (GPU-free) so future declaration changes re-derive the lists.

collector/wideep/README.md
Arsene12358: Fixed. The activation checklist now covers both missing enrollment surfaces: step 2 adds the `_REGISTRY_MODULES["wideep_vllm"]` entry alongside registry.py, and step 4 adds the `collector/kernel_source_backends.yaml` `{framework: vllm, kernel_source: deepep_moe}` mapping (with the same citation requirement as the kernel verification item). Both are pinned dormant until then: `test_registry_modules_have_no_wideep_vllm_entry` and `test_kernel_source_backends_have_no_vllm_deepep_moe_mapping` joined the dormancy tests and the README's flip list.

Arsene12358:
Rebased onto the updated PR 2.5: `73298c4` (mechanical re-stack, no collector-side reconciliation needed). Collector suites green at the tip (715+47+31) and the workspace harness holds 73 PASS / 0 FAIL against the `ed13cfc` baselines.

Harrilee (approved):

jasonqinzhou:
I’m taking this through the same closeout process as #1464. Before editing, the live state is:

- head: `27e0f8b787a08af693edc469aadaf9106978218c`
- base: current `main` at `11354af37474f6e991c24af6692171becce8a5f7`
- merge state: CONFLICTING / DIRTY
- review decision: CHANGES_REQUESTED (the prior blocking review is stale)
- review threads: 10 total, 0 unresolved

Planned update:

1. Integrate current `main` normally, without force-pushing. Treat the merged SDK/Rust implementation as canonical and replay only the collector producer series plus its explicitly coordinated frozen-contract test.
2. Audit the resulting diff against the collector layer rules: account for every non-`collector/**` file, remove already-landed stack artifacts, and preserve the reviewed producer/consumer contract rather than reintroducing schema-5 or `moe_ep` naming.
3. Re-verify the ten prior fixes on the converged tree: model identity, classified token failures, declared routing, launched-image attestation, persistence failure handling, stale-output refusal, dedup identity, diagnostic transport isolation, non-empty advertised worlds, and vLLM activation documentation.
4. Refresh the PR body so it describes a direct-to-main collector PR, current `MoeExpertCompute` filenames/contracts, current counts, and the actual post-integration scope.
5. Run the focused collector/provenance/schema suites, launcher syntax, Ruff/format, DCO/diff checks, then monitor the complete exact-head CI and any fresh review feedback.
6. Only after the exact head is mergeable, all required checks are green, review threads are clear, and the stale changes-requested state is superseded: approve and squash-merge with an expected-head guard.

collector/network/slurm/README.md
```diff
@@ -47,7 +47,7 @@ sbatch -N 1 ./slurm_custom_ar_4gpu.sh
 
 # 3. TensorRT-LLM MoE AlltoAll collection (NVLink)
 
-Benchmarks MoE alltoall **dispatch** and **combine** over NVLink. Supports two kernel sources: **NVLinkTwoSided** (WideEP/MNNVL) and **NVLinkOneSided** (CutlassFusedMoE). Results are written to `results/trtllm_alltoall_perf.txt`. Run from `collector/network/slurm/` and configure `CONTAINER_IMAGE`, `CONTAINER_MOUNTS`, `ACCOUNT`, `PARTITION`, `GPU_LIST`, and `GPUS_PER_NODE` in `submit_trtllm_alltoall.sh` before running.
+Benchmarks MoE alltoall **prepare**, **dispatch** and **combine** over NVLink. Supports two kernel sources: **NVLinkTwoSided** (WideEP/MNNVL) and **NVLinkOneSided** (CutlassFusedMoE). Results are written per job in the unified `moe_a2a` schema to `results/moe_a2a_<kernel-source>.<N>gpu/moe_a2a_perf.parquet` with a `collection_meta.yaml` provenance sidecar. The default container image is the manifest `trtllm` pin. Run from `collector/network/slurm/` and configure `CONTAINER_IMAGE`, `CONTAINER_MOUNTS`, `ACCOUNT`, `PARTITION`, `GPU_LIST`, and `GPUS_PER_NODE` in `submit_trtllm_alltoall.sh` before running.
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in `31fd39c7`. The Slurm README now matches both launchers exactly: TensorRT-LLM defaults are `2,4,8,16,32,64`, DeepEP defaults are `8,16,32,48,64`, the zero-case worlds are documented as deliberately absent, and every result path includes the `job_<jobid>/` attempt directory.

tests/unit/collector/test_collect_moe_a2a.py
```diff
@@ -0,0 +1,738 @@
...
+        phase="combine",
+        ep_size=16,
+        node_num=2,
+        shape=SHAPE,
+        num_tokens=512,
+        sms=16,
+        transmit_us=101.25,
+        notify_us=9.75,
+    )
+    assert row["latency"] == pytest.approx(row["transmit_us"] + row["notify_us"]) == pytest.approx(111.0)
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in `31fd39c7`. The chained `pytest.approx` expression is now two direct assertions, so both the transmit+notify relationship and the expected `111.0` latency independently check `row["latency"]`.

tests/unit/sdk/database/test_collector_schema_contract.py
```diff
@@ -0,0 +1,184 @@
...
+The ``moe_a2a_perf`` and ``moe_expert_compute_perf`` tables are produced by collectors and
+consumed by ``aiconfigurator_core.sdk.operations.moe_comm``. Each side pins the
+CSV header independently; this module is the SDK-side half. The collector-side
+twins (which pin the same literals against the actual writers) are:
+
+- ``tests/unit/collector/test_collect_moe_a2a.py::MOE_A2A_HEADER``
+- ``tests/unit/collector/test_collect_trtllm_alltoall.py::MOE_A2A_HEADER``
+- ``tests/unit/collector/sglang/test_collect_moe_ep.py::MOE_EXPERT_COMPUTE_HEADER``
+- ``tests/unit/collector/trtllm/test_collect_moe_ep.py::MOE_EXPERT_COMPUTE_HEADER``
+- ``tests/unit/collector/test_vllm_collect_moe_ep.py::MOE_EXPERT_COMPUTE_HEADER``
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in `31fd39c7`. The docstring/comment now point to the actual `MOE_EP_HEADER` symbols. `_TWIN_PINS` also carries each expected symbol name, and the AST check validates that named assignment’s folded string value rather than accepting the literal under any stale name.

collector/README.md
```diff
@@ -460,17 +460,110 @@ This collects the stock SGLang ops, including:
...
+  column; the adapter derives `node_num = max(1, moe_ep_size // 4)`. New
+  rows overwrite only at 4 GPUs/node; a row with an explicit `num_nodes`
+  column is honored as written.
+
+**`log_perf` freezes the CSV header from the first row**, so optional
+columns are all-or-nothing per file. In particular, power columns exist in a
+staging CSV only if the very first logged row had them: flipping
+`--measure_power` across a `--resume` of the same staging file appends rows
+whose fields no longer match the frozen header — misaligned columns or empty
+cells that fail at load time. Collect any one file with a single power
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in `31fd39c7`. `log_perf` now opens the staging file read/write under its existing writer lock, compares the persisted CSV header with the requested field list, and returns `False` before appending on any schema mismatch (including power off→on and on→off). The caller therefore classifies persistence failure without mutating the file. `test_log_perf_rejects_optional_column_schema_changes_before_append` covers both directions, and the README now documents the fail-closed behavior.

collector/sglang/README.md
```diff
@@ -216,10 +216,10 @@ The simulated EP size is calculated as: `simulated_ep_size = 256 / num_experts *
...
-- `wideep_context_moe_perf.txt`: Prefill phase performance data
-- `wideep_generation_moe_perf.txt`: Decode phase performance data
+- `moe_expert_compute_perf.txt`: both phases in one table, split by the `inference_phase`
+  column (`context` / `generation`)
 
 Output format:
 ```
-framework,version,device,op_name,kernel_source,moe_dtype,num_tokens,hidden_size,inter_size,topk,num_experts,moe_tp_size,moe_ep_size,distribution,latency
+framework,version,device,op_name,kernel_source,moe_dtype,distribution,inference_phase,num_tokens,hidden_size,inter_size,topk,num_experts,num_slots,moe_tp_size,moe_ep_size,latency
 ```
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed both findings bundled in this thread in `31fd39c7`: the output fence is now tagged `csv`, and `get_moe_prefill_test_cases` counts/logs both unconditional filters (below the measurable floor and above the per-rank token budget). The regression asserts the exact dropped populations for EP 64.

collector/helper.py
```diff
@@ -751,6 +751,24 @@ def log_perf(
...
+            if not is_empty:
+                f.seek(0)
+                existing_header = next(csv.reader(f), [])
+                if existing_header != fieldnames:
+                    print(
+                        f"Error writing log: schema mismatch for {perf_filename}: "
+                        f"existing header {existing_header}, requested header {fieldnames}. "
+                        "Use the same measurement settings when resuming or start a fresh staging file."
+                    )
+                    return False
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in `31c2b3fb`. `log_perf` now raises a typed `PerfLogError` for lock acquisition, resumed-schema, and write/fsync failures instead of returning `False`. That central fail-closed contract covers every production caller, including legacy call sites that did not inspect the successful boolean, so the executor classifies the task as failed before adding it to `done_tasks`; checkpoint/provenance status therefore cannot attest the missing row as complete. Regression tests cover schema mismatch (and byte-identical staging), lock timeout, the stale-lock race, and fsync failure. Validation: 131 affected collector tests passed; full unit suite 4,095 passed / 50 skipped; repository Ruff lint and format clean.

jasonqinzhou (approved):
Approved at exact head `31c2b3fbede5be259dfa767ec16195d58da42a03`.

Verified before approval:
- target is `main`, with current `main` integrated by a normal merge commit (no history rewrite)
- remote/local head match and the worktree is clean
- PR is mergeable
- 18/18 exact-head checks are successful, including unit, e2e, Rust/Python parity, collector data/perf sanity, regression snapshots/comparison, lint, DCO, CODEOWNERS, copyright, and CodeRabbit
- 16/16 review threads are resolved
- scope is 44 files: 31 collector files, 12 collector unit-test files, and the explicitly declared two-sided schema-contract test; no SDK or Rust production changes
- local full unit suite: 4,095 passed, 50 skipped; full Ruff and bash syntax checks are clean

The final persistence fix makes lock, schema, and write/fsync failures fail closed for all collector callers instead of allowing incomplete task completion metadata.

PR #1471, opened 2026-08-04
fix(perf-database): reuse framework communication data implicitly

jasonqinzhou:
## Summary

- derive communication reuse from the physical storage namespace instead of an operation allowlist
- allow implicit strictly-earlier reuse only for the validated framework-versioned folders `comm/sglang`, `comm/trtllm`, and `comm/vllm`
- keep `comm/nccl`, `comm/oneccl`, unknown backends, backend mismatches, and ambiguous legacy namespaces primary-only
- disable declared (`reuse.yaml`) and cross-backend fill for every communication table
- enforce the policy in the engine-owned Rust source resolver used by both table loading and Python provenance reporting
- align documentation, marker migration messages, and regression coverage

## Policy

| Storage namespace | Sources |
|---|---|
| `comm/{sglang,trtllm,vllm}/<version>` | primary, then strictly earlier versions from the same storage backend nearest-first |
| `comm/{nccl,oneccl}/<version>` | primary only |
| `comm/<unknown>/<version>` | primary only |
| ambiguous or backend-mismatched legacy paths | primary only |

Primary rows always win; older sources only fill missing shapes. A new table under an existing validated framework folder inherits the policy automatically. A new communication backend fails closed.

Communication tables never consume `reuse.yaml` donors or cross-framework manifest donors.

## Current-engine implementation

Current `main` moved shared-layer resolution into `aic-core/rust/aiconfigurator-core/src/perf_database/source_resolution.rs`. This rebased PR implements the policy there rather than restoring the retired Python resolver.

The resolver now:

- extracts both family and storage backend from a family-first primary path
- infers the family for missing or legacy-shaped primaries from canonical family-first copies of the same table
- restricts communication fallback enumeration to `comm/<same-storage-backend>`
- fails closed for unknown storage backends, path/database backend mismatches, and ambiguous legacy basenames
- keeps NCCL and oneCCL primary-only by table identity

Regression coverage includes the TRT-LLM rc20 → rc10 missing-primary case, an existing legacy primary with forbidden declared and cross-framework decoys, a `comm/futurelib` path under a TRT-LLM database, an ambiguous legacy basename, and a requested/storage-backend mismatch.

## Scope

This PR changes source-selection policy only. It does not add, move, or modify performance data.

## Validation

Rebased onto `main` at `163f66259`.

- Python unit/build suite: **3,544 passed, 10 skipped**
- Rust library suite: **503 passed, 1 ignored**
- Rust/Python engine-step and compile-engine parity: **447 passed**
- Focused reuse, collector validation, and marker migration: **127 passed**
- Focused Rust source-resolution suite: **18 passed**
- Ruff check and format check on every changed Python file
- Rust format check
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved communication performance-data loading for supported framework backends.
  * Earlier versions of the same framework backend can now be used when the primary data is unavailable.
  * NCCL, oneCCL, unknown, and unsupported communication backends remain restricted to primary data, preventing incorrect cross-backend reuse.
  * Added safer handling for legacy, missing, or ambiguous data layouts.

* **Documentation**
  * Updated guidance and validation messages to clarify communication-data reuse rules.


tianhaox (commented):
Direction agreed: the framework-version vs communication-library-version namespace distinction is the right boundary, and the rc20→rc10 backfill regression (introduced by the #1374 tightening) is a real case worth fixing. Loader-side coverage is complete — all 6 comm-family tables in the catalog are migrated to the wrapper, and the test matrix (fail-closed unknown backend, declared+cross-backend decoys, rc20 pin, legacy layout) is thorough. One blocking issue below.

## Blocking: `_compute_perf_db_sources` (engine.py) was not migrated — and since #1454 it is the default query path's source list

`aic-core/src/aiconfigurator_core/sdk/engine.py` (`_compute_perf_db_sources`, the loop at L748-752) calls the bare `_build_op_sources(filename_enum, primary_path, system_data_root)` for **every** `PerfDataFilename`, including all 6 comm tables. Its output is exactly what the compiled Rust engine loads (`"perf_db_sources"`, L823), and #1454 (merged 08-05, after this branch's base c8f735a38) made the Rust engine the **default** engine step.

Before this PR, comm policy was enforced inside `_build_op_sources` itself (op-name early-exit for nccl/oneccl + structural `_op_file_family_from_path` detection), so the engine path was covered. This PR removes both and moves enforcement to a caller-passed `storage_family` — which engine.py never passes. On that path, `storage_family=None` → declared / fallback / cross-backend channels all run for comm tables.

It happens to be behaviorally invisible with today's committed tree, but only by accident:

1. **reuse.yaml**: R3 bans comm reuse.yaml in the committed tree, but user-provided systems dirs (`--systems-path` / local trees) never run the checker. A `comm/trtllm/<v>/reuse.yaml` there is ignored by the Python loaders (wrapper path) yet admitted by the engine path — the Rust engine loads declared donor rows the Python side excludes. Silent Python/Rust drift, and a violation of this PR's own "no communication table uses declared reuse" contract on the default path.
2. **cross-backend**: channel 4 is empty today only because every comm entry in `op_kernel_source_manifest.yaml` lists a single framework. The moment a comm kernel_source is shared across frameworks (your own regression test constructs exactly this shape: `frameworks: [backend, "sglang"]`), the engine path admits cross-framework comm rows the wrapper excludes.
3. The deleted test `test_comm_family_custom_allreduce_gets_primary_only` was the only pin on the bare-call-with-comm-path behavior — it exercised `_sources_for` (plain `_build_op_sources`) with a reuse.yaml + manifest decoy and asserted primary-only. Under this PR it would fail: the plain call now returns primary + declared donor. The replacement tests all go through the wrapper, so the regression is unpinned.

This is not hypothetical about future callers either: **#1442** (large-EP PR 1) adds `moe_comm.py` with four more bare `_build_op_sources` calls on comm tables (`moe_a2a` + the three legacy comm tables as adapter sources). Caller-enforced policy is already being missed by in-flight work.

**Suggested fix** — keep the policy at the choke point: inside `_build_op_sources`, when `storage_family` is not passed, fall back to structural detection from the primary path (the removed `_op_file_family_from_path`) plus the nccl/oneccl op-name check. The explicit kwargs then only *add* information for the missing-primary/legacy cases this PR correctly fixes — they can never *remove* protection. That makes engine.py and #1442 safe with zero changes on their side. Alternatively migrate `_compute_perf_db_sources` to the wrapper (needs an op-file→storage-backend mapping for nccl/oneccl). Either way, please restore a plain-call comm test pinning the engine-path behavior.

Related: please rebase onto current main — the branch base predates #1454, so the severity of this interaction isn't visible from the branch itself.

## Non-blocking

- `_COMM_FRAMEWORK_VERSIONED_BACKENDS` duplicates `common.BackendName`'s values. If the fail-closed decoupling is intentional (a new serving framework must be explicitly validated before comm reuse — reasonable), add cross-reference comments in both places so the next `BackendName` addition doesn't silently ship with primary-only comm.
- Stale comments: `communication.py` module docstring (L21-24) and the NCCL loader comment (~L327-331) still describe the removed "`_build_op_sources` early-exits for nccl/oneccl" behavior.
- Operational note worth a line in the design doc: framework comm tables going from primary-only to fallback-chained means old comm version files can now be **live donors** (rc20→rc10 is exactly that shape). Per-op prune tooling (#1457) must treat comm as reuse-chained from now on.

jasonqinzhou:
Addressed the default-engine blocker in `a730dfcf`: communication reuse is now enforced at the shared `_build_op_sources` boundary, with direct `_compute_perf_db_sources` regression coverage for missing-primary, declared-donor, and cross-framework decoy cases. The branch is rebased onto current `main`, the PR description and stale comments are updated, all 21 checks pass, and no review threads remain unresolved. @tianhaox, could you please re-review when you have a chance?

aic-core/src/aiconfigurator_core/sdk/perf_database.py
```diff
@@ -1861,51 +1859,81 @@ def _gemm_key_names(database) -> list[str]:
...
         return None
     parts = rel.split(os.sep)
     if len(parts) == 4 and parts[0] not in KNOWN_BACKEND_DIRS:
-        return parts[0]
-    # Legacy-layout (3-component) or otherwise-unresolved path: no family
-    # component exists structurally, so we can't detect "comm" here. This is
-    # the transition-window exception described above, not a bug — legacy
-    # comm op files simply keep pre-V3 sibling-reuse behavior.
+        return parts[0].lower()
+    if len(parts) != 3 or os.path.exists(primary_path):
```
Yiming992, line 1891: 🔴 **Blocking: preserve communication policy for an existing legacy primary on the default-engine path.**

This early return makes an existing `<backend>/<version>/<table>` path structurally unclassified. The Python communication loaders compensate by calling `_build_communication_op_sources(...)`, but `_compute_perf_db_sources()` still calls the bare builder, so the compiled-engine source list runs declared and cross-backend channels. A synthetic user systems tree produced:

```text
Python wrapper versions: [2.0.0, 1.0.0]
Engine source versions:  [2.0.0, 3.0.0, 1.0.0, 0.5.14]
Python loaded shapes:    [1024, 2048]
Engine loaded shapes:    [1024, 2048, 4096, 8192]
```

The `3.0.0` rows were declared-reuse decoys and `0.5.14` was a cross-framework decoy, both forbidden by this PR's contract.

**Suggested fix:** give default-engine enumeration authoritative `(storage_family, storage_backend)` metadata for each table and pass it through the shared builder. This should be namespace metadata, not another reuse-policy allowlist; pure legacy trees cannot infer the family from their path. Add a `_compute_perf_db_sources` regression with an existing legacy comm primary, an earlier donor, a declared newer donor, and a cross-framework donor, then assert exact parity with `_build_communication_op_sources`. This spans engine source construction and tests, so a one-click suggestion would be incomplete.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
```diff
@@ -2317,12 +2375,23 @@ def _finish() -> list[tuple[str, Optional[set[str]]]]:
 
         if not self.enable_shared_layer:
             return _finish()
-        if op_filename_enum in (PerfDataFilename.nccl, PerfDataFilename.oneccl):
-            return _finish()
-        if _op_file_family_from_path(primary_path, system_data_root) == _COMM_FAMILY_DIR:
+
+        detected_family = _op_file_family_from_path(primary_path, system_data_root)
+        storage_family_lower = storage_family.lower() if storage_family else None
+        backend_lower = (storage_backend or self.backend).lower()
```
Yiming992, line 2381: 🔴 **Blocking: derive communication reuse from the storage backend, not `self.backend`.**

For a family-first path such as `comm/futurelib/2.0.0/custom_allreduce_perf.parquet`, `_op_file_family_from_path()` returns only `comm`. With no explicit kwarg, this line substitutes a TRT-LLM database's backend, marks the unknown storage namespace as validated, and admits `comm/trtllm/1.0.0/...` as an implicit fallback. The base revision returned primary-only for the same bare call.

**Suggested fix:** make structural detection return the complete namespace, e.g. `(family, backend)`, and use the detected backend for an existing family-first path. Explicit metadata should supplement missing/legacy paths; a mismatch between explicit and structurally detected backends should fail closed rather than weaken the detected boundary. Add a bare-builder regression where `database.backend == "trtllm"` but the primary path is under `comm/futurelib`, asserting primary-only. This requires coordinated helper, caller, and test changes, so a one-click suggestion would be unsafe.

Yiming992 (changes requested):
Requesting changes for two merge-blocking source-policy failures at `a730dfcf`.

## Standards

No standards-based merge blocker found. The branch is focused, DCO-signed, formatted, documented, and well tested.

## Spec / correctness

1. Existing legacy-layout communication primaries are classified correctly by the Python operation wrappers but not by the bare `_build_op_sources` call used to construct the default compiled-engine source map. The engine consequently admits declared and cross-framework rows that the Python loader excludes.
2. Family-first structural detection preserves only the family, not the storage backend. A bare call for an unknown `comm/<storage-backend>` can therefore inherit from the database backend instead of failing closed.

Both failures were independently reproduced. In the legacy parity repro, Python loaded message sizes `[1024, 2048]`, while the engine source map loaded `[1024, 2048, 4096, 8192]`; the extra shapes came from a forbidden declared donor and a cross-framework decoy. In the unknown-backend repro, `comm/futurelib/2.0.0/...` incorrectly admitted `comm/trtllm/1.0.0/...` as fallback; the base revision kept it primary-only.

All 21 CI checks are green, and I reran 231 relevant tests successfully. The current tests miss these intersections: the engine regression covers a missing family-first primary, while the legacy test uses only the explicit Python wrapper, and the unknown-backend test also bypasses the bare structural path. Please add the two regression shapes described inline and rerun the focused reuse and Rust/Python parity suites.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
jasonqinzhou: Addressed in `b76c1012f`. Family-first detection now extracts both `(family, storage_backend)` from the physical path. A `comm/futurelib` primary under a TRT-LLM database is therefore treated as an unknown/mismatched namespace and remains primary-only; it cannot inherit `comm/trtllm`. Matching Python and Rust regressions cover this fail-closed behavior.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
jasonqinzhou: Addressed in `b76c1012f`. Current `main` now owns source resolution in Rust, so the fix is in the shared engine resolver. A legacy or missing primary infers its namespace from canonical family-first copies; communication fallback is then restricted to `comm/<same-storage-backend>`, with declared and cross-backend channels excluded. I added the requested legacy-primary regression with an earlier donor, a declared newer decoy, and a cross-framework decoy, plus Rust coverage. Validation passed: 3,502 Python unit/build tests, 496 Rust tests (1 ignored), and 447 engine/parity tests.

jasonqinzhou:
Rebased onto current `main` (`163f66259`) and pushed the conflict-free head `1065cd025`. The resolution preserves the new `perf_data_reuse_manifest.yaml` ownership/provenance changes while keeping this PR's conservative communication policy.

Local validation is green:
- Python unit/build: 3,544 passed, 10 skipped
- Rust library: 503 passed, 1 ignored
- Rust/Python parity: 447 passed
- Focused reuse/checker/migration: 127 passed
- Ruff, Rust formatting, and `git diff --check`: clean

All prior inline threads remain resolved. @Yiming992 @tianhaox, could you please re-review this exact head? @coderabbitai review

jasonqinzhou:
Addressed the two CodeRabbit nitpicks at `da0483fa4`: clarified requested-backend namespace inference and added direct Rust coverage for ambiguous legacy basenames plus requested/storage-backend mismatches. The focused resolver suite is now 18/18 green; the previously reported full Python, Rust, and parity gates remain green. @coderabbitai review

Yiming992 (changes requested):
Requesting changes at `da0483fa` for two merge-blocking correctness failures and one required parity gap.

## 1. Preserve source priority across all MoE A2A basenames

The resolver orders sources independently for `moe_a2a_perf.parquet`, `wideep_deepep_normal_perf.parquet`, `wideep_deepep_ll_perf.parquet`, and `trtllm_alltoall_perf.parquet`, but `load_moe_a2a_grids` merges the three legacy formats before the new schema. The first new-schema row then unconditionally overwrites an existing legacy coordinate. This loses the PR's promised global precedence: requested-version rows must win and older versions may only fill missing shapes.

I verified this with the identical Rust probe at the fixed point and PR head. The fixture contains a requested TRT-LLM legacy row at 1.5 ms, an existing but partial requested-version `moe_a2a` table, and an older `moe_a2a` row for the queried coordinate at 9,000 μs. At `163f662`, the query returns 1.5 ms and the test passes. At `da0483f`, it fails with `got 9, want 1.5`.

Suggested fix:

- Keep the public/wire `PerfSource` shape unchanged, but expose internal source priority metadata (`channel`, `version`, and source) from `SourceResolver` to the MoE A2A loader.
- Merge sources in global priority tiers: all requested-version sources first, followed by each fallback version nearest-first.
- Within one tier, preserve the existing format rule by loading legacy adapters first and then letting the new schema override legacy rows from the same tier.
- Merge each completed lower-priority tier into the final grid with fill-only semantics (`or_insert`) so it cannot replace a higher-priority coordinate.
- Add the fixed-point/head probe shape as a permanent Rust engine/table-view regression test.

Simply reversing the format load order is not sufficient because it would break the intentional same-priority new-schema-over-legacy behavior.

## 2. Retain the physical backend for legacy-shaped primary paths

`op_file_namespace_from_path` parses only family-first paths. For a legacy-shaped primary, the new `inferred_op_namespaces` helper infers the family but replaces the physical backend with the requested backend. The later mismatch check therefore compares the requested backend with itself.

With `backend=trtllm, version=2.0.0`, an explicit primary at `data/vllm/2.0.0/custom_allreduce_perf.parquet` plus an older table under `comm/trtllm/1.0.0` resolves as `[primary, fallback]`; an unknown `futurelib/2.0.0` primary does the same. Both should be primary-only under the PR's fail-closed contract. I also reproduced the same source list through `PerfDatabase._build_op_sources`.

Suggested fix:

- Parse path location separately from family inference: a family-first path supplies both family and physical backend, while a legacy path supplies its first component as the physical backend and requires only the family to be inferred.
- Never replace that physical backend with `ResolveCtx.backend`.
- Return primary-only when the physical backend is unknown, is not framework-versioned, or differs from the requested backend.
- Add legacy-override tests for `vllm -> trtllm` and an unknown backend such as `futurelib`; the existing family-first mismatch test does not exercise this branch.

## 3. Pin the newly reachable engine behavior

This selection change makes `gb200/trtllm/1.3.0rc20` able to consume the rc10 `trtllm_alltoall` table, but the PR adds no rc20 engine/compile parity case or frozen golden. Resolver-record tests do not pin the resulting engine value. `.claude/rules/rust-core/parity.md` Rule 1 requires an oracle and/or an append-only pinned parity case when a new configuration class becomes reachable.

Suggested fix:

- Add an rc20 engine-step or compile parity case that exercises the rc10 communication fallback.
- Append its frozen golden with `parity_tests/pin_goldens.py` so the selected engine result is visible in the golden diff.

All current GitHub checks are green, and the focused resolver, Python ordering, marker, collector, formatting, and lint checks also pass. The requested tests cover source-layout intersections that the current suites do not exercise.

jasonqinzhou:
Addressed all three blockers from the latest review at current head `65c654283`:

1. **Global MoE A2A source priority** (`492053dfe`): retained internal `channel`/`version`/source metadata and now merge all four A2A basenames in global priority tiers. Requested-version sources win; each lower fallback version is nearest-first and fill-only. Same-tier new-schema-over-legacy behavior is preserved. The fixed-point regression covers both the engine query and raw table view.
2. **Legacy physical-backend identity** (`492053dfe`): legacy paths now retain their physical backend separately from inferred operation family. Mismatched `vllm -> trtllm` and unknown `futurelib` paths remain primary-only, with Rust and Python regressions covering both cases.
3. **rc20 engine parity** (`65c654283`): added an rc20 TRT-LLM WideEP compile/engine case that consumes the rc10 `trtllm_alltoall` fallback, with append-only context and generation goldens produced by `pin_goldens.py`.

Validation on the final rebuilt extension:
- Rust library: 505 passed, 1 ignored
- Compile/engine parity: 71 passed
- Focused source-ordering, table-view, and rc20 regressions: 59 passed
- Ruff, Rust formatting, and `git diff --check`: clean

CI is running on the updated head. @Yiming992, could you please re-review `65c654283` when you have a chance?

Yiming992 (approved):
Re-reviewed the response commits through `65c65428`. All three requested changes are resolved.

## Standards

No findings. The new rc20 compile/engine parity case, pin-script population, and append-only Rust-produced goldens satisfy the repository's parity requirements; all 53 pre-existing frozen references remain unchanged.

## Spec

No findings. Global priority is preserved across all four MoE A2A basenames with same-tier new-schema precedence and fill-only lower tiers; legacy paths retain their physical backend and fail closed for `vllm -> trtllm` and unknown-backend mismatches; the rc20 WideEP case pins its rc10 all-to-all fallback.

Validation:

- Source-resolution Rust tests: 20 passed
- MoE A2A Rust tests: 14 passed
- Cross-format engine/table-view regression: 1 passed
- Python reuse-ordering tests: 32 passed
- TRT-LLM WideEP parity cases: 2 passed
- Cargo formatting, Ruff lint/format, and `git diff --check`: clean
- Current remote head unchanged; 24/24 reported checks successful

Approving this revision.

Arsene12358 (approved):
Re-reviewed the response commits through `65c65428`. All three requested changes are resolved.

## Standards

No findings. The new rc20 compile/engine parity case, pin-script population, and append-only Rust-produced goldens satisfy the repository's parity requirements; all 53 pre-existing frozen references remain unchanged.

## Spec

No findings. Global priority is preserved across all four MoE A2A basenames with same-tier new-schema precedence and fill-only lower tiers; legacy paths retain their physical backend and fail closed for `vllm -> trtllm` and unknown-backend mismatches; the rc20 WideEP case pins its rc10 all-to-all fallback.

Validation:

- Source-resolution Rust tests: 20 passed
- MoE A2A Rust tests: 14 passed
- Cross-format engine/table-view regression: 1 passed
- Python reuse-ordering tests: 32 passed
- TRT-LLM WideEP parity cases: 2 passed
- Cargo formatting, Ruff lint/format, and `git diff --check`: clean
- Current remote head unchanged; 24/24 reported checks successful

Approving this revision.

PR #1475, opened 2026-08-04
feat(collector): add the FPM forward-pass collection workflow

liyuanzhe1991:
> [!IMPORTANT]
> Last of the stack. #1473 and #1474 are merged; this branch now includes
> current `main` plus the Collector remediation and Generator follow-ups.

## Overview

Add the FPM (ForwardPassMetrics) collection workflow
(`python -m collector.fpm_forward`; `collect.py --ops fpm_forward` remains a compatibility entry point): capability-driven planning, Kubernetes
execution through the public Generator API, a collector-owned in-pod
collection runtime, strict native-artifact validation, and idempotent parquet
publication — the collector half of #1365 restructured per the decoupling
proposal, plus #1430's multi-document manifest validation, per-resource
apply/delete verification, Grove pod discovery, and `--fpm-orchestrator`
(co-authored-by etshen; happy to re-cut as cherry-picks before undrafting).

Decoupling highlights:

- **`fpm_exec.sh` owns every collection judgement in-pod** (formerly split
  between the wrapper and the generated run.sh): rank-0 etcd, runtime
  preflight, completion gate, schema-v2 result checker, DP completion
  barrier, follower exit classification (probing the collector's own etcd),
  and engine process-group termination — all driven by the `FPM_*` variables
  the rendered `fpm_env.sh` exports, with the baseline exit-code table
  preserved.
- The runner selects pods exclusively through the collector-owned cell label
  (no workload-kind fork), reads manifests only through the contract's
  enumerated fields, and declares engine-limit suppression via
  `preserve_engine_limits`.
- Planning/validation/publication semantics follow #1365: frozen plan
  identity, resumable checkpoints, verified transfer, salvage, strict
  schema-v2 validation, schema-v6 parquet publication.

## Validation

- Whitelist `K8sConfig.extra_env` as a declared deployment input (#1365's
  known issue 4): user entries resolve before backend-policy and
  Collector-owned identities, name collisions fail closed, and contract
  variables remain rejected at render time. Live-motivated by forcing
  `VLLM_FLASHINFER_ALLREDUCE_BACKEND` on multinode B200, where the engine's
  auto-selected `mnnvl` allreduce hangs without an NVSwitch fabric.
- Post-review hardening: `--resume` no longer aborts when a passed cell's raw
  artifacts were pruned after publication (classified warning + skip, with a
  red/green regression test), and every in-pod results-path operation routes
  through the contract's `FPM_RESULTS_DIR`.

- Remediation validation: repository-wide Ruff passed; the integrated FPM contract/generator/collector slice passed **381 tests**; and the pre-integration full unit run passed **4,033 tests** with 10 skipped. Behavior tests for the moved runtime logic reuse the baseline's input vectors against the real `fpm_exec.sh`.
- Merge simulation: the three-PR stack merges into latest `main` with zero
  conflicts; the merged tree is byte-identical to the integration branch that
  passed the live runs below.
- Live (4×B200, GLM-5.2-NVFP4, tep tp=4):
  - smoke: preflight passed, native self-benchmark 4/4 points (schema v2,
    complete/valid), verified transfer, cleanup with zero leaked resources;
  - formal collection: 2 cells passed, `fpm_forward_perf.parquet` with
    10,281 rows (8,913 prefill + 1,368 decode) — the measurement grid is
    **identical point-for-point** to a formal run of the coupled #1365+#1430
    baseline (only-new: 0, only-old: 0), 102 context-clamped duplicates
    consolidated (same count as baseline), per-point latency correlation
    0.9992.

## User-approved design decisions

Recorded here because they change collection/publication semantics (each was
an explicit product decision by the maintainer, not a reviewer/implementer
choice):

- **Observe-don't-predict admission**: plan-time rejection is limited to the
  deterministic size-versus-capacity memory predicate. Structural validator
  errors and unresolved estimates remain runnable so execution records and
  automatically retests the observed outcome.
- **Partial publication** (`--fpm-publish-partial`, default off): a plan with
  failed cells may publish its passed cells; missing cells are recorded in
  the checkpoint **and** as a `campaign_incomplete` error so the run still
  exits nonzero. Execution policy is not part of plan identity.
- **First-publisher-wins**: republishing a cell_id under a different run
  identity skips that cell whole (sealed rows are never overwritten or
  mixed) while non-overlapping cells still land; skips are returned to the
  caller and logged loudly.
- **Schema v6 identity columns**: `moe_backend`/`attention_backend` default
  to `"auto"` when unpinned; `enable_wideep`/`enable_eplb` are real booleans
  defaulting to `false`. Pinned values must be plumbed to the engine and
  verified via resolved-config markers, else the plan fails closed.

## Summary by CodeRabbit

* **New Features**
  * Added FPM forward benchmarking with model capability detection, topology selection, memory-aware admission, smoke runs, resumable execution, and formal Parquet publication.
  * Added Kubernetes support for single-node and multinode deployments using LWS or Grove orchestration.
  * Generated deployments now include `k8s_deploy.yaml`, `fpm_env.sh`, and a launch-focused `run.sh`.
  * Added schema-v2 result validation, provenance tracking, artifact checks, runtime coordination, retries, and recovery.

* **Documentation**
  * Updated deployment, generator, CLI, and collector guides with workflow, orchestration, artifact, and configuration details.


collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1600 @@
...
+        smoke=smoke,
+    )
+    # The Generator owns deployment rendering, but its naive serving defaults
+    # intentionally synthesize max batch/token/sequence limits from an SLA.
+    # Native self-benchmarking must instead observe the limits resolved by the
+    # target vLLM image, so declare preserve_engine_limits: the Generator
+    # strips those optional engine-policy fields and keeps its rule plugin
+    # from reintroducing them; topology, dtype, model identity, and Pod
+    # rendering continue through the normal typed request path.
+    params = build_naive_generator_params(
```
Ethan-ES, line 1075: `--fpm-model-config` is resolved and frozen into the plan, but this render path passes only `plan.model_path`. `build_naive_generator_params()` then reloads model metadata from that path before applying overrides. For a private or pod-only checkpoint this fails before artifacts are rendered; when the path is reachable, it can also use facts different from the frozen plan. Could the Generator entry point accept the frozen resolved model config/facts and have this call pass `plan.capability.model_config`? Please add an integration test covering explicit model config plus `_render_cell()`.

Ethan-ES (commented):
One Generator integration gap remains for the explicitly frozen model-config workflow.

collector/fpm_forward/database.py
```diff
@@ -0,0 +1,456 @@
...
+        total_kv = int(point["total_kv_read_tokens"])
+        rows.append(
+            {
+                "cell_id": cell.cell_id,
+                "model_path": plan.model_path,
+                "system": plan.system,
+                "backend": plan.backend,
+                "backend_version": backend_version,
+                "weight_quantization": cell.weight_quantization,
+                "gemm_quant_mode": cell.gemm_quant_mode or cell.weight_quantization,
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — the substitution is deleted: `aggregate_cell` reads `cell.gemm_quant_mode` directly, and `FPMCell` now requires the field at construction (see the planner thread), so an unresolved value can no longer be published under a different `_ROW_KEY` coordinate.

collector/fpm_forward/database.py
```diff
@@ -0,0 +1,456 @@
...
+                "total_kv_read_tokens",
+                "partition_policy",
+                *_RUN_IDENTITY_FIELDS,
+            }
+            if not required.issubset(table.column_names):
+                raise ValueError(
+                    "existing FPM database predates the attempt-bound schema-v5 contract; "
+                    f"publish to a clean destination: {parquet_path}"
+                )
+            merged.extend(table.to_pylist())
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — the gate now requires `{*_ROW_KEY, *_RUN_IDENTITY_FIELDS}` (self-maintaining as `_ROW_KEY` grows) and the error names the missing columns. A legacy parquet missing `weight_quantization` now fails with that `ValueError` instead of a bare `KeyError`; covered by a new test.

collector/fpm_forward/native_artifact.py
```diff
@@ -0,0 +1,344 @@
...
+    if phase == "decode":
+        if total_prefill != 0:
+            raise ValueError(f"native decode point carries prefill tokens: {point}")
+        return {
+            "num_prefill_requests": 0,
+            "sum_prefill_tokens": 0,
+            "sum_prefill_kv_tokens": 0,
+            "num_decode_requests": batch,
+            "sum_decode_kv_tokens": total_kv,
+        }
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — prefill points now require `total_prefill_tokens >= batch_size` and decode points `total_kv_read_tokens >= batch_size`, as proposed; both floors are unit-tested.

collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1587 @@
...
+                formal_rows.extend(
+                    aggregate_cell(
+                        plan,
+                        cell,
+                        root / "cells" / cell.cell_id,
+                        expected_attempt_id=_required_attempt_id(entry, cell.cell_id),
+                    )
+                )
+            systems_root = Path(database_root).expanduser().resolve() if database_root else None
+            parquet_path, metadata_path = write_formal_database(plan, formal_rows, systems_root=systems_root)
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — formal publication is now additionally gated on full-plan coverage, so a deliberate partial run classifies as the existing `campaign_incomplete` error instead of a false `database` failure; a 2-cell / `cell_limit=1` regression test covers it. (The CLI already rejects `--limit` without `--smoke`, so this was reachable only for programmatic callers — but the honest classification matters there too.)

collector/fpm_forward/runtime/fpm_exec.sh
```diff
@@ -0,0 +1,353 @@
...
+  data_dir=/tmp/fpm-forward-etcd
+  rm -rf "${data_dir}"
+  etcd \
+    --data-dir "${data_dir}" \
+    --listen-client-urls http://0.0.0.0:2379 \
+    --advertise-client-urls "${etcd_endpoint}" \
+    --listen-peer-urls http://127.0.0.1:2380 \
+    >/results/etcd.log 2>&1 &
+  etcd_pid=$!
+fi
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — the leader's readiness loop now checks `kill -0 $etcd_pid` on every iteration; if etcd dies it waits for the exit status, tails the last 20 lines of `/results/etcd.log` to stderr, and exits 1 immediately instead of burning the 120 s probe budget. Covered by a test that kills the stub etcd and asserts fast failure with the log evidence.

collector/fpm_forward/runtime/preflight.py
```diff
@@ -0,0 +1,58 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Fail before model loading unless the image provides PR11509 native FPM."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+from dynamo.vllm.instrumented_scheduler import BenchmarkPoint, InstrumentedScheduler
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — the runtime import now happens inside `main()` under `try/except ImportError`; on failure the audit artifact is written with `status=failed` and the `import_error` before the same 'Provide a compatible Dynamo image' RuntimeError is raised, so an incompatible image still fails loudly *and* leaves the audit evidence.

docs/dynamo_deployment_guide.md
```diff
@@ -488,28 +488,34 @@ Use `--generator-set K8sConfig.<field>=value` (or place the same keys inside `--
...
+
+Each collection still starts a new engine and reloads the model. The collector's staged runtime stops result-producing engines after their local completion gate, refuses to overwrite existing benchmark outputs, and coordinates headless followers and final cleanup. By default `/results` is backed by Pod-local `emptyDir`, and those results disappear when the Pod is deleted; a matching user-provided volume and mount are preserved. Persistent engines and reuse of a GPU-resident model are outside the V1 scope. A target cluster needs the controller selected by `K8sConfig.fpm_orchestrator`; any scheduler and queue required by that cluster must be supplied through the generic Pod-spec and label overlays. GB200 configurations that enable MNNVL also include the generated `ComputeDomain` in `k8s_deploy.yaml`. See the [CLI User Guide](cli_user_guide.md#fpm-v1-resource-workload-and-run-script) for the full input example.
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in d22f1959 (the generator commit in this stack) — the deployment guide now states the condition explicitly: multinode GB200 configurations emit a `ComputeDomain`; single-node output never does.

docs/generator_overview.md
```diff
@@ -173,42 +173,48 @@ You can use the generator in two ways: AIConfigurator CLI or standalone (code/CL
...
 
-The FPM collector or agent remains responsible for staging the complete runtime bundle—scheduler code, cases, capacity, run-spec, runtime contracts, and validators—on every Pod; starting `run.sh` concurrently on all Pods in an LWS; coordinating exit status; strictly validating, downloading, aggregating, and recording evidence for the results; and cleaning up the workload. For a single-node Pod, the basic execution flow is:
+The FPM collector or agent remains responsible for staging the case manifest, Scheduler adapter, preflight, optional pinned runtime overlay, and wrapper on every Pod; starting `run.sh` concurrently on all workload Pods; coordinating exit status; strictly validating, downloading, aggregating, and recording evidence for the results; and cleaning up the workload. For a single-node Pod, the basic execution flow is:
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in d22f1959 — the stale sentence was replaced; the doc now consistently says the generator's `run.sh` launches the engine while the collector's `fpm_exec.sh` owns the in-pod collection runtime, matching lines 215-217.

src/aiconfigurator/generator/naive.py
```diff
@@ -524,4 +549,10 @@ def build_naive_generator_params(
         params["generator_dynamo_version"] = effective_dynamo_version
     _drop_empty_worker_roles(params)
 
+    if preserve_engine_limits:
+        for role_params in params.get("params", {}).values():
+            for key in _ENGINE_LIMIT_KEYS:
+                role_params.pop(key, None)
+        params["preserve_engine_limits"] = True
+
```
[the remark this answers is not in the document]
liyuanzhe1991: The FPM render path is fully guarded: FPM renders the vLLM backend only, and every engine-limit assignment in `vllm.rule` (`max_batch_size`, `max_num_tokens`, `max_seq_len`) sits under `when not (preserve_engine_limits or false)`. d22f1959 additionally adds `test_preserve_engine_limits_strips_disagg_roles` so a regression in role coverage would be caught.

The residual unguarded assignments (`cuda_graph_batch_sizes` / `tokens_per_block` in `trtllm.rule` and `sglang.rule`, and the TRT-LLM multimodal `encode` limits) are pre-existing upstream rule content — this stack's diff does not touch the rule files, and no FPM render reaches those backends. Guarding them changes non-FPM TRT-LLM/SGLang rendering behavior, so it belongs in a separate upstream change rather than this stack.

collector/fpm_forward/model_capability.py
```diff
@@ -0,0 +1,184 @@
...
+            "source_kind": self.source_kind,
+            "sha256": self.sha256,
+            "payload": self.payload,
+        }
+
+
+def _aic_cached_model_config_path(model_path: str) -> Path | None:
+    root = Path(__file__).resolve().parents[2]
+    cached = root / "src" / "aiconfigurator" / "model_configs" / f"{model_path.replace('/', '--')}_config.json"
+    return cached if cached.exists() else None
```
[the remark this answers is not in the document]
liyuanzhe1991: We verified this one and believe no change is needed. The lookup is a read-only probe into the repo-bundled `model_configs/` directory, whose filenames were generated with the same `replace('/', '--')` convention (it is also Hugging Face Hub's own on-disk cache mangling, `models--org--name`). The mapping is injective over every id that can exist on the Hub: `replace('/', '--')` collides only for ids containing a literal `--`, and HF rejects repo ids containing `--` (or `..`) both client-side (`huggingface_hub` validators) and server-side. Path traversal is impossible because the replacement removes all slashes, and the collector never writes to this directory, so there is no poisoning vector. Switching to a collision-free key would break every lookup against the shipped filenames (an aic-core-owned naming convention) for no reachable benefit. Happy to add a defensive shape guard (return `None` for non-HF-id-shaped strings, falling through to explicit resolution) if you'd like belt-and-braces here.

collector/fpm_forward/model_capability.py
```diff
@@ -0,0 +1,184 @@
...
+        return path
+    if explicit or path.is_absolute() or raw_path.startswith(("./", "../", "~")):
+        raise FileNotFoundError(f"model config does not exist: {path}")
+    return None
+
+
+def _load_file_config(path: Path, *, source_kind: str, cached: bool = False) -> ResolvedModelConfig:
+    payload, _ = _load_json_mapping(path, description="model config")
+    payload = _attach_sibling_quant_config(payload, path, cached=cached)
+    return ResolvedModelConfig(
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — an empty base payload is now rejected in `_load_file_config` before the sibling-quant merge and in `_download_model_config` before `_attach_hf_quant_config`, each with a clear `ValueError`; both paths are unit-tested.

collector/fpm_forward/native_artifact.py
```diff
@@ -0,0 +1,344 @@
...
+
+    assert canonical is not None
+    return str(canonical["runtime"]["backend_version"]), str(canonical["attempt_id"])
+
+
+def _expected_scheduled(point: dict[str, Any]) -> dict[str, int]:
+    phase = str(point["point_type"])
+    batch = int(point["batch_size"])
+    total_prefill = int(point["total_prefill_tokens"])
+    total_kv = int(point["total_kv_read_tokens"])
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — a `_require_int` helper (rejects float/bool/str) now validates `batch_size`, `total_prefill_tokens`, `total_kv_read_tokens`, and `benchmark_id`, so a noncompliant or tampered artifact fails validation instead of being silently truncated into a database row; parametrized tests cover float/bool/str.

collector/fpm_forward/planner.py
```diff
@@ -0,0 +1,472 @@
...
+@dataclass(frozen=True, slots=True)
+class FPMCell:
+    cell_id: str
+    workload_kind: str
+    topology: ParallelTopology
+    weight_quantization: str
+    kv_cache_dtype: str
+    backend_policy: BackendPolicy
+    parallel_strategy: str = "unspecified"
+    gemm_quant_mode: str | None = None
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — `FPMCell.gemm_quant_mode` is now a required field resolved once at cell construction, and `to_dict` emits it directly; both substitution fallbacks (planner and database) are deleted. Plan hashes are stable because production always populated the same value the fallback produced.

collector/fpm_forward/planner.py
```diff
@@ -0,0 +1,472 @@
...
+            fmha_quant_mode=capability.dtype.fmha_by_kv_dtype[kv_cache_dtype],
+            comm_quant_mode=capability.dtype.comm_quant_mode,
+            fmha_resolution=capability.dtype.fmha_resolution_by_kv_dtype[kv_cache_dtype],
+        )
+        for phase in ("prefill", "decode")
+        for topology in topologies
+        for kv_cache_dtype in capability.dtype.kv_cache_dtypes
+        if (topology, kv_cache_dtype) in runnable_dtype_pairs
+        for policy in policies
+    )
```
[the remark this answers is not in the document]
liyuanzhe1991: We traced every empty-population path and each already fails loudly before a plan exists: (1) `_validate_filters` (topology.py) rejects filters exceeding `--fpm-max-gpus` — the `--fpm-tp-sizes 32` on a 16-GPU plan example dies here with exit 2; (2) `enumerate_fpm_topologies` raises when filters/presets admit no topology; (3) `filter_memory_infeasible_topologies` logs the dropped-N-of-N warning and then raises when admission rejects everything. Given a non-empty admitted set, the comprehension produces >= 2 x len(topologies) cells by construction (each admitted topology retains >= 1 runnable dtype pair, policies is a constant 1-tuple, phases constant 2), and partial per-dtype drops are warning-logged. All of these surface as exit-2 CLI errors in both `--plan-only` and run paths; only a hand-built plan bypassing the planner could reach the runner with zero cells, and no such producer exists. We'd rather not assert a state the planner cannot produce, but can add the one-line invariant check if you feel strongly.

collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1600 @@
...
+    resolved_env: dict[str, dict[str, str]] = {}
+    for item in [*policy_env, *env]:
+        if not isinstance(item, dict) or not isinstance(item.get("name"), str):
+            raise TypeError("K8sConfig.extra_env entries must be {name, value} mappings")
+        name = item["name"]
+        existing = resolved_env.get(name)
+        if existing is not None and existing != item:
+            raise ValueError(f"conflicting FPM environment value for {name}")
+        resolved_env[name] = copy.deepcopy(item)
+    merged.setdefault("K8sConfig", {})["extra_env"] = list(resolved_env.values())
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — `extra_env` is now part of the deployment-input whitelist (`_DEPLOYMENT_K8S_FIELDS`) with shape validation, and `_render_cell` threads the operator's base `extra_env` into the conflict-checked resolution (base env, then policy env, then collector-owned env, with duplicate-key detection), so it is honored instead of silently dropped.

collector/fpm_forward/capabilities.py
```diff
@@ -0,0 +1,346 @@
...
+                fmha_resolution_by_kv[kv_name] = f"aic_data_fallback_from_{inferred_fmha}"
+        else:
+            raise ValueError(
+                f"AIC database has no joint fmha evidence for ({inferred_fmha!r} fmha, "
+                f"{kv_name!r} kv) on {context_op!r} and no bfloat16 slice to transfer from; "
+                f"jointly supported fmha modes: {sorted(joint_modes)}"
+            )
+    primary_kv = resolved_kv[0]
+    fmha = fmha_by_kv[primary_kv]
+    fmha_resolution = fmha_resolution_by_kv[primary_kv]
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — `resolve_model_capability` now rejects an empty KV-dtype request up front with an actionable message ("use 'auto' for the checkpoint-native dtype"), before any config or database resolution, so neither downstream module can see an empty tuple. (The CLI already maps an empty flag to `('auto',)`; this closes the programmatic path.)

collector/fpm_forward/memory_admission.py
```diff
@@ -0,0 +1,209 @@
...
+        if non_kv < 0 or capacity <= 0:
+            raise ValueError(f"invalid AIC memory estimate: non_kv={non_kv}, capacity={capacity}")
+    except Exception as error:
+        return DTypeMemoryEstimate(
+            kv_cache_dtype=kv_cache_dtype,
+            disposition="unknown",
+            estimated_non_kv_bytes=None,
+            gpu_capacity_bytes=None,
+            reason=f"AIC memory estimate unavailable: {type(error).__name__}: {error}",
+        )
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — the `fmha_by_kv_dtype[kv_cache_dtype]` lookup is hoisted above the `try` and re-raised as a capability-invariant `ValueError`, so planner-owned data fails closed while genuine estimator failures keep the intentional fail-open 'unknown' path; unit-tested.

collector/fpm_forward/runner.py
liyuanzhe1991: Fixed in d22f1959 (generator) + 4d2f9748 (collector). `build_naive_generator_params` now accepts a `model_config` parameter; when provided there is zero filesystem/network model resolution — weight estimation and architecture/MoE detection both read the frozen config. `_render_cell` passes `plan.capability.model_config.parsed_payload()` unconditionally, so rendering is a pure function of the frozen plan. Tests: a generator unit test poisons `get_model_config_from_model_path` to raise and still renders from the frozen config, and an integration test builds a plan with a nonexistent `model_path` plus an explicit `--fpm-model-config`, calls `_render_cell`, and asserts the three artifacts render with `ModelConfig.is_moe` matching the frozen capability.

collector/collect.py
```diff
@@ -1621,7 +1638,30 @@ def main():
...
+
+    add_fpm_arguments(parser)
+    # Do not import Generator code on ordinary op-level collector invocations.
+    # FPM is explicit-only, so a lightweight argv pre-scan is sufficient to
+    # expose its shared Generator/Kubernetes flags without changing the legacy
+    # import path or startup requirements.
+    import sys
+
+    if any(value in {FPM_FORWARD_OP, f"--ops={FPM_FORWARD_OP}"} for value in sys.argv):
+        add_fpm_generator_arguments(parser)
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — the argv pre-scan is deleted; `add_fpm_generator_arguments` is registered unconditionally (it lives in `fpm_forward.config`, so no generator import), and `reject_fpm_arguments_without_fpm` now also covers the deployment-input dests so they stay explicit-only via a clean `parser.error`. `--op=fpm_forward` (abbreviated) now works end to end, and `--ops gemm --namespace foo` exits 2 with the explicit message instead of a traceback; both are regression-tested.

src/aiconfigurator/generator/builders/fpm_builder.py
```diff
@@ -386,18 +510,83 @@ def _benchmark_wait_timeout_seconds(args: list[str]) -> int:
...
+    command: list[str],
+    env: list[tuple[str, str]],
 ) -> str:
     lines = [
         "#!/usr/bin/env bash",
         "set -Eeuo pipefail",
         "",
+        "# fpm_env.sh owns every render-time collection fact (topology, rank and",
+        "# leader discovery, benchmark identity); discovery failures exit 2 here.",
+        'source "$(dirname "${BASH_SOURCE[0]}")/fpm_env.sh"',
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in d22f1959 — the source line is rendered from `FPM_ENV_FILENAME`, so the constant and the artifact key can no longer drift.

tests/unit/collector/test_fpm_exec.py
```diff
@@ -0,0 +1,849 @@
...
+            shadow_python3=True,
+            teardown_seconds=3,
+        )
+
+        completed = _run(staged, timeout=30)
+    finally:
+        listener.close()
+
+    assert completed.returncode == 7
+    assert "Headless engine exited (status 7) while the leader was still alive" in completed.stderr
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in 4d2f9748 — the tests now allocate free ports at staging time (`_free_port()` redraws when the decimal collides with a literal being replaced), every replacement is count-asserted, and all tolerance paths (the 2379 skip and both 'someone else already listens' fallbacks) are removed, so the suite is order- and parallel-safe and can no longer false-pass against a foreign listener.

tests/unit/generator/test_fpm_artifacts.py
```diff
@@ -202,25 +182,317 @@ def _main_container(pod: dict) -> dict:
...
+        assignments.extend(tokens[1:])
+    return assignments
+
+
+def _export_value(script: str, name: str) -> str:
+    prefix = f"{name}="
+    for assignment in _export_assignments(script):
+        if assignment.startswith(prefix):
+            return assignment[len(prefix) :]
     raise AssertionError(f"missing export for {name}")
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in d22f1959 — `_export_value` now collects every export of the name and fails on duplicates as well as on missing exports, so a later duplicate can no longer hide behind the first.

tests/unit/generator/test_fpm_artifacts.py
```diff
@@ -1375,3 +1446,48 @@ def test_fpm_rejects_non_list_extra_cli_args(invalid):
...
+
+def test_fpm_multinode_fusion_guard_merges_into_existing_compilation_config():
+    # The base fixture already carries the collector's compilation config
+    # (cudagraph capture policy); the guard must merge into that single flag
+    # rather than appending a second occurrence.
+    script = _render(_multinode_params())[FPM_RUN_SCRIPT_FILENAME]
+
+    assert script.count("--compilation-config") == 1
+    assert '"cudagraph_mode":"FULL"' in script.replace("'", "").replace(" ", "") or "cudagraph_mode" in script
+    assert '"fuse_allreduce_rms":false' in script
```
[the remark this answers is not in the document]
liyuanzhe1991: Fixed in d22f1959 — the `or` clause is gone; the test now asserts `cudagraph_mode` is present in the rendered script.

liyuanzhe1991:
**Added in f32edb60 — keep engine progress floods out of the kubectl exec stream.**

During live merge-test smokes we lost two runs (tp8, then tp4) to the same failure: the engine was healthy, but ~45s into the flashinfer autotune phase — 8 TP workers each emitting dozens of tqdm progress-bar updates per second, all tee'd into the `kubectl exec` stream — the proxied exec session was killed remotely (`command terminated with exit code 137`, output truncated, no engine error, no pod event). A rerun that hit the warm autotune cache (no bars, quiet stream) passed on the same node with the same rendered artifacts.

`fpm_exec.sh` now filters tqdm-style bar lines out of the *streamed* copy of engine stdout/stderr only; `/results/engine.stdout.log` / `engine.stderr.log` still record the complete output for salvage. A single awk per stream writes both copies, so a dead stream reader can never SIGPIPE the file writer (macOS tee lacks `-p`, and the unit suite executes this script on macOS). Covered by `test_engine_progress_bars_are_kept_out_of_the_exec_stream`.

liyuanzhe1991:
**Correction to the hardening rationale above.** Follow-up forensics attributed the two exit-137 kills to cluster-side scheduling-policy enforcement reclaiming the pods — not to stream volume; the autotune-phase timing was coincidental (that phase dominates any cold start's first minutes).

The filter itself stays, on its own merits: it keeps multi-worker progress-bar floods out of the proxied exec channel and makes the failure output the collector quotes actually readable. The in-tree comment and test docstring were reworded to the accurate rationale in bc3334a5.

liyuanzhe1991:
**Teardown hardening in 01fc89a5 — background-delete escalation for stalled foreground cascades.**

Live finding from the queue-scheduled experiments: a controller that keeps reconciling children of a terminating parent can stall `kubectl delete --cascade=foreground` forever. We hit this with LWS v0.6.0 — while the workload was still unscheduled, the controller kept recreating its child StatefulSet against the GC for 40+ minutes, so every teardown (and every retry, which collides with the deterministically-named zombie) failed the post-delete verification.

`cleanup()` now: foreground cascade as before → any resource that survives it gets re-deleted with `--cascade=background` (the parent disappears immediately, the controller loses its reconcile target, the GC collects the orphans) → a bounded convergence poll re-verifies every owned resource and pod. The fail-closed contract is unchanged: no finalizer stripping, and anything that still remains raises the same loud error. Covered by two new unit tests (escalation path, and fail-closed non-convergence).

collector/collect.py
Arsene12358, line 1: Can we set up a FPM specific collection entry point instead of using the collect.py shared by op-level collectors?

Arsene12358 (changes requested):

collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1802 @@
...
+#     production steady-state decode has no such per-step hash (requests stay
+#     resident). Disabling is the serving-faithful choice here.
+_FPM_VLLM_DECODE_FAKE_KV_ARGS = ("--no-enable-prefix-caching",)
+# Must stay in sync with the engine's _kvwarm_warm_eligible strategy set.
+# pure_tp added 2026-08-20: the "moe_tp balanced by construction" immunity
+# assumption only covers cross-GPU balance; expert-activation dispersion is
+# still content-driven (same-pod fake/warm pairing measured -15.3% mid-band,
+# full-grid re-collection 10.43%->4.80%). Requires the engine image with the
+# matching _kvwarm_warm_eligible change (gc-warmtp-20260820 or later).
+_FPM_KVWARM_STRATEGIES = frozenset({"tep", "dep", "pure_tp"})
```
jasonqinzhou: **[P1] Reject images that do not implement the `pure_tp` warm contract.** This line now leaves prefix caching enabled for every `pure_tp` decode cell because the new engine is expected to warm it, but compatibility is never enforced: `runtime/preflight.py` still checks only the original PR11509 fields/methods, and `validate_native_collection()` deliberately accepts a missing `kvwarm` block as legacy. I reproduced a preflight pass with no `_kvwarm_warm_eligible` capability, while a legacy artifact is publishable. Against an older but otherwise compatible image, `pure_tp` stays on the fake-KV protocol while prefix caching is on, putting the full-context block hashing back inside the measured step (the code documents 26.5 -> 121 ms). Please make this capability part of preflight/native validation (or select the flag from a verified runtime contract) and reject missing/mismatched KV-warm metadata before publication; add an older-image/legacy-artifact regression for `pure_tp`.

collector/fpm_forward/database.py
```diff
@@ -0,0 +1,519 @@
...
+    kvwarm_meta = collection.kvwarm_meta
+
+    def _kv_seed_regime(point: dict[str, Any], phase: str) -> str:
+        if phase == "prefill":
+            return "n/a"
+        if kvwarm_meta is None:
+            return "legacy"
+        skip_reason = kvwarm_meta.get("skip_reason")
+        if skip_reason:
+            return f"skip:{skip_reason}"
```
jasonqinzhou, line 177: **[P1] Do not treat every KV-warm skip as a valid topology regime.** This unconditional override maps `prefix_caching_disabled` to `skip:prefix_caching_disabled` even when every point is marked `kvwarm_fake_fallback`. I reproduced that exact artifact and both rows published under the `skip:` value; a modeling loader that excludes only `fake_fallback`, as this change intends, will retain the 2-3.7x biased fallback rows. Only explicitly approved warm-ineligible reasons such as `moe_tp_balanced_by_construction` should override the point markers. Please validate the `warm_eligible`/`skip_reason` combination and either reject non-benign skips or classify their rows as `fake_fallback`; cover `prefix_caching_disabled` and unknown reasons.

collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1802 @@
...
+def _configured_sampling_metadata(
+    plan: FPMCollectionPlan,
+    cell: FPMCell,
+    *,
+    smoke: bool,
+) -> dict[str, int | str]:
+    if cell.workload_kind != "prefill":
+        # Auditable measurement condition: decode engines run with prefix
+        # caching pinned off (see _cell_generator_overrides).
+        return {"decode_prefix_caching": "disabled"}
```
jasonqinzhou: **[P1] Record the decode policy that was actually rendered.** Current code makes decode prefix caching regime-conditional: `tep`/`dep`/`pure_tp` leave it enabled for KV warm-up, while fake-KV `single`/`tp` cells add `--no-enable-prefix-caching`. However, `_configured_sampling_metadata()` still returns `decode_prefix_caching: disabled` for every decode cell, and both new-attempt records and passed-cell metadata refreshes use that helper. A direct current-head probe rendered `pure_tp` without the disable flag while recording `disabled`. Please derive this field from the same resolved policy/arguments (and fail closed on a conflicting policy override), then add regressions showing both warm and fake-KV cells record the flag they actually render.

src/aiconfigurator/generator/builders/fpm_builder.py
```diff
@@ -957,6 +958,23 @@ def _lower_resources(
...
+    # fleets (127.9 allocatable cores on the H200 pool) with daemonset
+    # headroom; 64Gi/GPU matches the validated F-arm pod.
+    limits = lowered.setdefault("limits", limits)
+    requests = lowered.setdefault("requests", {})
+    if "cpu" not in limits and "cpu" not in requests:
+        limits["cpu"] = str(gpu_limit * 14)
+        requests["cpu"] = str(gpu_limit * 14)
+    if "memory" not in limits and "memory" not in requests:
+        limits["memory"] = f"{gpu_limit * 64}Gi"
+        requests["memory"] = f"{gpu_limit * 64}Gi"
```
jasonqinzhou: **[P1] Keep request-only resource overlays valid.** The deployment guide documents one-sided `mainContainer.resources` overrides and says explicit values always win, but these defaults are installed before `_merge_container_resources()` applies that overlay. The documented request-only `memory: 448Gi` example still renders `requests.memory=448Gi` beside `limits.memory=256Gi` for a four-GPU Pod, so Kubernetes rejects the workload because the request exceeds the limit. The overlay test passes because it never asserts the opposite section. Please apply or inspect the overlay before filling defaults, and mirror or validate one-sided CPU/memory overrides so requests remain <= limits and the Guaranteed-QoS claim stays accurate; add a regression for the documented request-only overlay.

collector/fpm_forward/memory_admission.py
```diff
@@ -0,0 +1,303 @@
...
+                max_new_tokens=max_new_tokens,
+            )
+            for kv_cache_dtype in capability.dtype.kv_cache_dtypes
+        )
+        dispositions = {estimate.disposition for estimate in estimates}
+        if "admitted" in dispositions:
+            disposition = "admitted"
+            reason = "at least one requested KV dtype fits the configured max-new-token envelope"
+            admitted.append(topology)
+        elif strict_admission and "predicted_invalid" in dispositions and "unknown" not in dispositions:
```
jasonqinzhou: **[P1] Keep structural validator failures observable.** The diagnostic split labels this cause accurately, but the Collector policy permits generation-time filtering only for a size-vs-capacity memory predicate; every other queued case must execute or raise so the failure is recorded and automatically retested. With the default `--fpm-strict-admission true`, this branch still drops topologies when an AIC `ValueError` contains `Invalid quantized MoE configuration`, producing no cell failure record and permanently coupling coverage to mutable exception text. Please leave these estimates unknown/runnable and record the observed runtime failure, unless the repository's Collector policy is explicitly changed by its owners; update the default-path regression to assert the topology remains in the plan.

collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1802 @@
...
+            if isinstance(checkpoint["cells"].get(cell.cell_id), dict)
+            and checkpoint["cells"][cell.cell_id].get("status") == "passed"
+        ]
+        missing_cells = [cell.cell_id for cell in plan.cells if cell not in publishable_cells]
+        partial = bool(missing_cells)
+        try:
+            if partial and not publish_partial:
+                raise ValueError("internal: partial publication reached without --fpm-publish-partial")
+            formal_rows = []
+            for cell in publishable_cells:
```
jasonqinzhou, line 1791: **[P1] Do not republish a completed formal database from pruned raw artifacts.** Passed cells correctly skip rerun and tolerate a failed metadata refresh, but every non-smoke full-plan resume still re-aggregates all raw artifacts here. If operators prune those artifacts after successful publication, the existing checkpoint database status of `passed` is ignored, aggregation fails, and the completed campaign is overwritten as `formal_database_failed`. The existing pruned-raw regression uses `smoke=True` and bypasses this branch. Please make a previously committed database terminal after validating its recorded outputs, or retain enough committed rows to republish, and add the same regression in non-smoke mode.

collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1802 @@
...
+        backend_name=plan.backend,
+        mode="agg",
+        generator_dynamo_version=overrides.get("generator_dynamo_version"),
+        generator_overrides=overrides,
+        preserve_engine_limits=True,
+        # Render is a pure function of the frozen plan: model metadata comes
+        # from the plan's resolved config evidence, never from re-resolving
+        # plan.model_path — which may identify a checkpoint reachable only
+        # inside the cluster (--fpm-model-config campaigns).
+        model_config=plan.capability.model_config.parsed_payload(),
```
jasonqinzhou, line 1136: **[P1] Keep bootstrap templates renderable for unregistered architectures.** The frozen-config handoff added for the existing render thread still fails for the bootstrap support level: capability resolution deliberately admits an unknown architecture as `bootstrap_template`, but `ResolvedModelConfig.parsed_payload()` calls `_parse_hf_config_json()`, which raises whenever the architecture is absent from `ARCHITECTURE_TO_MODEL_FAMILY`. The same `NewMoeDsaForCausalLM` config accepted by planning still raises here before Generator render or Kubernetes submission. Please preserve the frozen raw config through an architecture-agnostic sizing/parser path, similar to the raw-config fallback in `naive.py`, and add a `_render_cell` regression for an unregistered dense or MoE config.

collector/fpm_forward/runner.py
```diff
@@ -0,0 +1,1802 @@
...
+            checkpoint["cells"][cell.cell_id]["duration_seconds"] = round(time.monotonic() - cell_started, 3)
+            _atomic_json(checkpoint_path, checkpoint)
+    # R16 §3 run manifest: the machine-readable timing map for the speed-up
+    # campaign. Collector-observable segments land now; engine-internal
+    # phases (kvwarm/seeding splits) are declared interface fields sourced
+    # from the engine artifact once dynamo-fpm phase instrumentation exists.
+    manifest_cells: dict[str, Any] = {}
+    run_total = 0.0
+    for cell in target_cells:
+        entry = checkpoint["cells"].get(cell.cell_id)
```
jasonqinzhou: **[P2] Keep timing manifests bound to the measured attempt.** This reconstructs the manifest from persistent checkpoint entries rather than attempts executed by this invocation, and the emitted cell record omits `attempt_id`. A no-op `--resume` still rewrites `run-manifest.json` with the prior attempt's positive `run_total_s` and phase map, while a cell that fails after render/scheduling emits `collector_phase_seconds=null` because `phase_marks` is saved only on success. The speed-up aggregator cannot distinguish current, historical, or partial timing. Please accumulate attempt-scoped manifest rows, including `attempt_id` and run timestamps/status, persist completed phase marks on failure, and cover both a no-op resume and a failed cell.

collector/README.md
```diff
@@ -116,6 +116,39 @@ capabilities.py                        — generation-time capability/denylist f
...
+pass its real local metadata explicitly with
+`--fpm-model-config /path/to/config.json` (a directory containing `config.json`
+is also accepted). The resolved config snapshot and SHA-256 are frozen into the
+collection plan.
+
+Generator and Collector share the Dynamo-native benchmark result schema
+version through the rendered `fpm_env.sh`; the staged `fpm_exec.sh` validates
+that contract in-pod without a Collector-side text patch. Formal
+`fpm_forward_perf.parquet` output
+uses metadata schema v5 and records `collector_attempt_id`, `runtime_run_id`,
```
jasonqinzhou: **[P2] Update the documented FPM publication contract.** This section still says schema v5 and that rows from a different attempt are rejected for an existing cell. The current code writes schema v6, skips the conflicting cell under first-publisher-wins while merging non-overlapping cells, adds opt-in partial publication plus default strict admission, and now publishes the additive `kv_seed_regime` field with legacy/null compatibility. Please document those current behaviors and the checkpoint's missing/skipped-cell evidence so operators and consumers can tell exactly what was and was not published.

jasonqinzhou (changes requested):

## What this PR does

This PR adds an end-to-end whole-forward FPM collection campaign with capability-driven planning, Generator-backed Kubernetes rendering, an in-pod benchmark runtime, resumable checkpoints, per-run timing, and schema-v6 parquet publication. The current delta expands KV warm-up to pure-TP decode and publishes a per-row kv_seed_regime derived from the engine artifact, including compatibility handling for older parquet rows.

The delta thoughtfully preserves older parquet rows during additive schema merges, checks the shared KV-warm regime across DP ranks, and adds focused synthetic coverage for real-KV, fake-fallback, topology-skip, legacy, and prefill cases.

## Why this score

The current head is 4.0/10 with REQUEST_CHANGES. The new per-row protocol evidence is valuable, but the delta introduces two silent data-quality paths: pure-TP now assumes an unverified engine-image capability, and any cell-level KV-warm skip is labeled as an acceptable skip even when the points are the biased fake-KV fallback. Five retained P1 issues also remain in checkpoint provenance, resource rendering, Collector admission policy, formal resume, and bootstrap rendering; attempt-scoped timing and publication documentation remain incomplete.

collector/fpm_forward/runner.py
jasonqinzhou: Addressed in `20ac99031`. The warm-capable strategies now share one contract table; runtime preflight requires `_kvwarm_warm_eligible`, and decode publication for `pure_tp`/`tep`/`dep` rejects legacy, disabled, missing, or mismatched KV-warm evidence. Regression coverage includes an older image without the capability and legacy artifacts.

collector/fpm_forward/runner.py
jasonqinzhou: Addressed in `20ac99031`. Decode prefix caching is now derived from the same resolved strategy policy used to render engine arguments and checkpoint metadata; warm strategies record enabled, fake-KV strategies record disabled, and conflicting overrides fail closed. Both policy branches have regression coverage.

collector/fpm_forward/database.py
jasonqinzhou: Addressed in `20ac99031`. Only the explicitly approved `moe_tp_balanced_by_construction` skip remains a benign `skip:` regime. `prefix_caching_disabled` and unknown/non-benign skips can no longer launder fake-KV rows as topology-valid evidence; the classifier and regression cases cover both paths.

src/aiconfigurator/generator/builders/fpm_builder.py
jasonqinzhou: Addressed in `9806a3fc8`. Container overlays are applied before defaults are completed, and a one-sided request override is mirrored/validated so requests cannot exceed limits while preserving Guaranteed QoS. The documented request-only memory overlay now renders matching request and limit values and is regression-tested.

collector/fpm_forward/memory_admission.py
jasonqinzhou: Addressed in `48e34db9c`. Structural estimator/validator errors are no longer plan-time rejection criteria; they remain `unknown` and runnable so execution records the observation. Only the permitted size-versus-capacity predicate can reject a topology, with default-path regression coverage.

collector/fpm_forward/runner.py
jasonqinzhou: Addressed in `48e34db9c`. A completed formal database is now terminal on resume after validating its schema-v6 metadata, digest, required columns, row count, and complete-plan checkpoint evidence. The non-smoke pruned-raw resume regression verifies that committed output is retained without reaggregation.

collector/fpm_forward/runner.py
jasonqinzhou: Addressed in `9806a3fc8`. Unknown bootstrap architectures now normalize the frozen payload through the architecture-agnostic `NaiveKVCacheEstimator` path using the effective `text_config`, rather than re-entering the registered-family parser. `_render_cell` regressions cover both dense and MoE unregistered configs.

collector/fpm_forward/runner.py
jasonqinzhou: Addressed in `48e34db9c`. Run-manifest schema v2 contains only attempts from the current invocation, including `attempt_id`, timestamps, status, and completed phase marks on failures. A no-op resume emits `attempts=[]`, `cells={}`, and zero run time; failed-attempt timing is also covered.

collector/README.md
jasonqinzhou: Addressed in `5eb0d63cb`. The Collector documentation now describes schema v6, first-publisher-wins behavior, opt-in partial publication, `kv_seed_regime` legacy/null compatibility, missing/skipped evidence, the observe-don't-predict admission rule, and attempt-scoped manifest v2.

collector/fpm_forward/runner.py
jasonqinzhou: Verified and extended in `9806a3fc8`. The frozen model-config handoff remains the render source, and bootstrap configs with unregistered architectures now take the architecture-agnostic sizing/normalization path instead of failing in the registered-family parser. Dense and MoE `_render_cell` regressions cover nonexistent/private model paths with frozen payloads.

collector/collect.py
jasonqinzhou: Addressed in `5eb0d63cb`. FPM now has a dedicated `python -m collector.fpm_forward` entry point (`cli.py` + `__main__.py`) while retaining `collect.py --ops fpm_forward` for compatibility. CLI-isolation tests verify both entry points without importing unrelated op-level collectors.

jasonqinzhou (approved):

Arsene12358 (approved):

PR #1477, opened 2026-08-04
fix(sdk): count cp in disagg rate-match GPU accounting

### LINKED ISSUE #1476: bug(sdk): disagg rate matching counts a cp-parallel prefill worker as pp*tp*dp GPUs (cp missing) — inflated tokens/s/gpu and impossible replica math
### Summary

Disagg rate matching counts a prefill worker's GPUs as `pp * tp * dp` — **context parallelism (cp) is missing** — so a `cp8` prefill worker is accounted as **1 GPU** instead of 8. Every derived metric on the composed row is wrong: `num_total_gpus`, `seq/s/gpu`, `tokens/s/gpu` (inflated up to ~8x per prefill worker), the final ranking (rows are ranked by the inflated `tokens/s/gpu`), replica math, and the CLI report (which also omits cp from the `(p)parallel` string, making the bug invisible in the output).

### Affected sites

- `src/aiconfigurator/sdk/sweep.py` — `_rate_match_dict`: `prefill_gpus = p["pp"] * p["tp"] * p["dp"]` (decode line has the same shape, harmless today since cp is prefill-only)
- `src/aiconfigurator/sdk/picking.py` — `_build_disagg_summary_dict`: same arithmetic (the two are kept in sync by the rate-match parity test, so both are wrong together)
- CLI report: `(p)parallel` renders as `tp1pp1` for a `cp8` worker, and `gpus/replica` math shows `2x1` for two 8-GPU workers

Note the *candidate* rows are correct (`num_total_gpus=8` for a cp8 worker), and `sweep_disagg`'s `_match_workers` argmax reads `p_worker["num_total_gpus"]`, so the worker-count matching itself uses the right GPU counts — only the composed row dict, the cross-category ranking, and the report are wrong.

### Repro

Qwen3-VL-8B, h200_sxm, sglang 0.5.14, `--isl 2000 --osl 200 --ttft 150 --tpot 20 --image-height 1344 --image-width 1344 --num-images 2`, with an EPD-enabled build (PR #1340 branch, `--enable-epd`):

- Prefill candidate `tp1pp1dp1etp1ep1cp8`: ttft=23.5 ms, seq/s=42.5, `num_total_gpus=8` (correct in the candidate frame)
- Winning disagg row reports: `gpus/replica 10 (=2x1+4x1+4x1(e))`, `(p)parallel tp1pp1`, 1179 tokens/s/gpu, "3 replicas in 32 GPUs"
- Actual deployment: 2 prefill workers x **8 GPUs (cp8)** + 4 decode + 4 encode = **24 GPUs per replica**; 3 replicas need **72 GPUs**, not 30
- Cross-check with `Task.run_single_disagg` on the same shape (which goes through the same `_rate_match_dict` but was fed tp1 without cp): 239 tokens/s/gpu — a ~5x mismatch against the sweep row

### Why it went unnoticed

The bug predates EPD (present at 240e66aed in both sweep.py and picking.py), but for VL disagg sweeps the colocated encoder latency pushed cp-prefill rows over tight TTFT filters, so they rarely won. With encoder disaggregation (PR #1340), the language-only prefill removes that penalty and cp rows become selectable — the inflated `tokens/s/gpu` then dominates the ranking (we saw fake +431% "EPD benefit" from exactly this). Any non-VL workload where cp prefill candidates pass the TTFT filter should be able to hit it on main as well.

### Suggested fix

Include cp in the composed-row GPU accounting in both `_rate_match_dict` and `picking._build_disagg_summary_dict` (e.g. `prefill_gpus = p["pp"] * p["tp"] * p["dp"] * p.get("cp", 1)`, or better, reuse the candidate row's `num_total_gpus` instead of re-deriving), update the rate-match parity test, and render cp in the report's `(p)parallel` / `gpus/replica` strings.

### Workaround

Disable cp candidates (`prefill_cp_candidates: [1]` / patch `task_v2._default_cp_list_for` to return `[1]`) until fixed.

tianhaox:
#### Overview:

Fixes #1476: the disagg rate-match composition counted a cp-parallel prefill worker as `pp*tp*dp` GPUs — **cp was missing** — so a cp8 worker was accounted as one GPU. This inflated `num_total_gpus`-normalized metrics (`seq/s/gpu`, `tokens/s/gpu`) up to ~8x, mis-ranked cp rows to the top of the disagg results, and reported replica math the cluster cannot satisfy (e.g. "3 replicas in 32 GPUs" for a shape that needs 72).

The `pp*tp*dp` derivation dates from the original picking implementation (#386 — correct at the time, cp didn't exist). CP arrived in #1247 and updated the parallel enumeration and `ModelConfig.total_gpus_per_worker`, but not these compose sites or the report.

#### Details:

Rather than patching `cp` into each product (which leaves the next parallelism dimension exposed to the same landmine), the worker GPU footprint becomes a single definition:

- **`picking.worker_gpus(row)`** — per-worker GPU count for a `ColumnsStatic`-style row: prefers the row's own `num_total_gpus` (the authoritative value backends stamp from `total_gpus_per_worker`, which already includes cp) and falls back to the dims product for partial dicts predating the column. Both `sweep._rate_match_dict` and `picking._build_disagg_summary_dict` now call it, so their arithmetic cannot drift.
- **`picking.WORKER_GPU_DIMS = ("tp", "pp", "dp", "cp")`** — the one list of dimensions that multiply into a worker's footprint. The CLI report derives `(p)`/`(d)` worker GPU totals from it (a future dimension lands in the table automatically), and now renders cp when > 1: `(p)parallel tp1pp1cp8`, `(p)gpus/worker 8 (=1x1x8)`, `gpus/replica (=2x8+4x1)`. The report previously hid cp entirely — which is what kept the inflation invisible.

Note the *candidate* frames and `sweep_disagg`'s `_match_workers` argmax already used the correct per-worker `num_total_gpus`; only the composed row dict, the cross-category ranking, and the CLI report were wrong.

**Tests**: regression pins for the cp8 accounting (`num_total_gpus == 2*8 + 4*1` + normalized throughput), the missing-`cp`-key default, sweep/picking parity with cp, and `worker_gpus` authoritative-first semantics (incl. NaN fallback). All 309 sweep/picking/cli unit tests pass.

- `src/aiconfigurator/sdk/picking.py` — `WORKER_GPU_DIMS` + `worker_gpus()`, and `_build_disagg_summary_dict` consuming it; `sweep._rate_match_dict` mirrors.
- `tests/unit/sdk/sweep/test_rate_match_parity.py::test_rate_match_counts_cp_in_gpu_accounting` — the pinned repro.

#### Related Issues:

- Closes #1476
- Context: found while validating EPD sweep output on #1340 (whose language-only prefill makes cp candidates selectable under tight TTFT, surfacing the inflation); the bug itself predates that PR and is reachable on main.
- Out of scope, follow-up candidate: `generator/module_bridge.py` has five sibling `tp*pp*dp` products; whether cp ever reaches the generator's task_config is a separate contract question.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected GPU totals and per-GPU performance metrics for configurations using context parallelism.
  - Improved worker and replica summaries to accurately reflect prefill and decode resources.
  - Preserved existing reporting behavior when context parallelism is disabled or set to one.
  - Updated parallelism labels and GPU counts in disaggregated worker reports.

- **Tests**
  - Added coverage for context-parallel GPU accounting, default handling, authoritative totals, and consistency across performance summaries.

### DIFF
--- a/src/aiconfigurator/cli/report_and_save.py
+++ b/src/aiconfigurator/cli/report_and_save.py
@@ -24,7 +24,7 @@ from aiconfigurator.generator.request import from_legacy_params
 from aiconfigurator.logging_utils import _cli_bold, _cli_underline
 from aiconfigurator.sdk import pareto_analysis
 from aiconfigurator.sdk.pareto_analysis import draw_pareto_to_string
-from aiconfigurator.sdk.picking import WORKER_GPU_DIMS
+from aiconfigurator.sdk.picking import WORKER_GPU_DIMS, parallel_dim
 from aiconfigurator.sdk.task_v2 import Task
 from aiconfigurator.sdk.utils import safe_mkdir
 
@@ -87,7 +87,7 @@ def _composed_worker_gpus(row: dict, role: str) -> int:
     """
     gpus = 1
     for dim in WORKER_GPU_DIMS:
-        gpus *= int(row.get(f"({role}){dim}") or 1)
+        gpus *= parallel_dim(row.get(f"({role}){dim}"))
     return gpus
 
 
@@ -251,7 +251,7 @@ def _plot_worker_setup_table(
             # the cp display factor stays hidden when cp=1 for readability.
             p_gpus = _composed_worker_gpus(row, "p")
             d_gpus = _composed_worker_gpus(row, "d")
-            p_cp = int(row.get("(p)cp") or 1)
+            p_cp = parallel_dim(row.get("(p)cp"))
             p_cp_label = f"cp{_cli_underline(str(p_cp))}" if p_cp > 1 else ""
             p_cp_factor = f"x{_cli_underline(str(p_cp))}" if p_cp > 1 else ""
             if is_moe:
--- a/src/aiconfigurator/cli/report_and_save.py
+++ b/src/aiconfigurator/cli/report_and_save.py
@@ -24,6 +24,7 @@ from aiconfigurator.generator.request import from_legacy_params
 from aiconfigurator.logging_utils import _cli_bold, _cli_underline
 from aiconfigurator.sdk import pareto_analysis
 from aiconfigurator.sdk.pareto_analysis import draw_pareto_to_string
+from aiconfigurator.sdk.picking import WORKER_GPU_DIMS
 from aiconfigurator.sdk.task_v2 import Task
 from aiconfigurator.sdk.utils import safe_mkdir
 
@@ -75,6 +76,19 @@ def _check_power_data_available(best_configs: dict[str, pd.DataFrame], threshold
 
 def _format_power(power_w: object) -> str:
     return "" if pd.isna(power_w) else f"{float(power_w):.1f}W"
+
+
+def _composed_worker_gpus(row: dict, role: str) -> int:
+    """Per-worker GPU count from a composed disagg row's ``(p)``/``(d)`` columns.
+
+    Iterates :data:`aiconfigurator.sdk.picking.WORKER_GPU_DIMS` so every
+    parallelism dimension lands here automatically; columns a row predates
+    (e.g. ``(d)cp``) default to 1.
+    """
+    gpus = 1
+    for dim in WORKER_GPU_DIMS:
+        gpus *= int(row.get(f"({role}){dim}") or 1)
+    return gpus
 
 
 def _plot_worker_setup_table(
@@ -232,11 +246,20 @@ def _plot_worker_setup_table(
         for i, row in enumerate(top_configs.to_dict("records")):
             display_total_gpus = row["total_gpus_used"] if "replicas_needed" in row else total_gpus
             display_concurrency = row["concurrency"] * row["replicas"]
+            # Worker GPU totals come from the shared dims list so a new
+            # parallelism dimension cannot be silently dropped again (#1476);
+            # the cp display factor stays hidden when cp=1 for readability.
+            p_gpus = _composed_worker_gpus(row, "p")
+            d_gpus = _composed_worker_gpus(row, "d")
+            p_cp = int(row.get("(p)cp") or 1)
+            p_cp_label = f"cp{_cli_underline(str(p_cp))}" if p_cp > 1 else ""
+            p_cp_factor = f"x{_cli_underline(str(p_cp))}" if p_cp > 1 else ""
             if is_moe:
                 p_parallel = (
                     f"tp{_cli_underline(str(row['(p)tp']))}"
                     f"pp{_cli_underline(str(row['(p)pp']))}"
                     f"dp{_cli_underline(str(row['(p)dp']))}"
+                    f"{p_cp_label}"
                     f"etp{row['(p)moe_tp']}ep{row['(p)moe_ep']}"
                 )
                 d_parallel = (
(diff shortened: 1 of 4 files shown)


src/aiconfigurator/cli/report_and_save.py
```diff
@@ -71,6 +72,19 @@ def _check_power_data_available(best_configs: dict[str, pd.DataFrame], threshold
...
+def _composed_worker_gpus(row: dict, role: str) -> int:
+    """Per-worker GPU count from a composed disagg row's ``(p)``/``(d)`` columns.
+
+    Iterates :data:`aiconfigurator.sdk.picking.WORKER_GPU_DIMS` so every
+    parallelism dimension lands here automatically; columns a row predates
+    (e.g. ``(d)cp``) default to 1.
+    """
+    gpus = 1
+    for dim in WORKER_GPU_DIMS:
+        gpus *= int(row.get(f"({role}){dim}") or 1)
```
jasonqinzhou: **[P2] Treat schema-filled NaN dimensions as absent.** `or 1` does not normalize `NaN` because `bool(float("nan"))` is true, so this helper and the `p_cp = int(...)` conversion below raise `ValueError` for legacy/partial `ColumnsDisagg` rows materialized with the current schema. On this exact head, composing schema-materialized legacy `ColumnsStatic` rows with valid `num_total_gpus` correctly produced `num_total_gpus=6` but left `(p)cp=NaN`; `_plot_worker_setup_table` then crashed here. Please normalize missing/null/non-finite dimensions to 1 in this CLI path and in `worker_gpus`' fallback, and make the compatibility regression materialize rows through `ColumnsStatic`/`ColumnsDisagg` so it covers pandas' NaN filling.

jasonqinzhou (approved):
**COMMENT · 8.5/10 · high confidence**

## What this PR does

This PR corrects disaggregated rate matching so context-parallel workers contribute their full GPU footprint to replica totals and per-GPU metrics, centralizes worker sizing around backend-stamped totals, and exposes CP-aware accounting in CLI deployment summaries.

The new authoritative-first worker_gpus helper keeps the sweep and picking implementations aligned, and the focused CP8, parity, and fallback tests directly capture the original mis-ranking failure.

## Why this score

8.5/10, COMMENT. The current head fixes the prior blocking numerical regression: the formerly failing inference-session test and exact-head GitHub unit job now pass, and CP8 totals are correctly normalized. One non-blocking compatibility issue remains: pandas schema materialization turns omitted CP fields into truthy NaN values, so the CLI's new integer conversion still crashes on the legacy/partial-row path that the PR says defaults to 1.

src/aiconfigurator/cli/report_and_save.py
tianhaox: Fixed in cc742db8c (and rebased onto main to clear the conflict). Added `picking.parallel_dim()` as the single normalizer — missing / None / NaN / non-positive → default — and routed all three consumption points through it: `worker_gpus`' fallback product, both compose sites' `"(p)cp"` emission (so composed rows never carry NaN in that column to begin with), and the CLI table's `_composed_worker_gpus` / `p_cp` conversions. The regression now materializes legacy rows through `ColumnsStatic`/`ColumnsDisagg` exactly as you described (asserts `cp` NaN-fills, composes, then renders the table on a row whose `(p)cp` is NaN-filled) — it crashed with ValueError before the fix and pins `6 (=1x4+1x2)` after. Thanks for the catch; ironically the same `NaN or 1` truthiness trap I'd flagged elsewhere.

tests/unit/sdk/sweep/test_rate_match_parity.py
```diff
@@ -159,3 +159,120 @@ def test_rate_match_missing_power_w_defaults_to_zero():
...
+    def test_nan_num_total_gpus_falls_back(self):
+        from aiconfigurator.sdk.picking import worker_gpus
+
+        assert worker_gpus({"tp": 2, "pp": 1, "dp": 1, "num_total_gpus": float("nan")}) == 2
+
+    def test_nan_dim_in_fallback_counts_as_one(self):
+        from aiconfigurator.sdk.picking import worker_gpus
+
+        # Schema-materialized legacy rows NaN-fill dims they predate.
+        assert worker_gpus({"tp": 2, "pp": 1, "dp": 1, "cp": float("nan")}) == 2
```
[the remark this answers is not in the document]
tianhaox: Added in dbc334216 — parametrized None / 0 / negative for both the dims fallback (`cp`) and the authoritative `num_total_gpus` branch. 24/24 passing.

PR #1478, opened 2026-08-04
feat: price intra-batch prefill imbalance from two work columns

YijiaZhao:
#### Overview:

A prefill batch is priced today from its per-request means, so two batches with
the same token totals and different spreads get the same estimate. They do not
cost the same: attention work is quadratic in the new-token count, and with a
sparse indexer the cost also depends on which requests sit above `topk`.
Measured on GLM-5-NVFP4 (B200, TP8), the gap reaches 17% of step time — a
uniform batch at `b=16 avg_s=2048 avg_p=1024` runs 1368.3 ms, while a batch
with the same totals and every row within 3% of the mean runs 1481.1 ms.

This adds the correction: per average point, a price for each of two work
columns, and a narrow rule for when it may be applied.

#### Details:

**The model.** The label is a difference against the equal-length batch of the
same cell, with both token totals conserved exactly, so every term that depends
on the totals alone cancels and what remains is the cost of the spread:

    y = T_batch - T_uniform = c_idx * x_idx + c_mla * x_mla

Two columns, grouped by the **geometry** the work follows rather than by which
kernel runs it. `x_mla` counts attention pairs, capped at `topk`, for every row.
`x_idx` counts the scoring trapezoid `s*p + s^2/2` of only those rows that cross
`topk`.

That second choice is the one worth reading the docstring for, because it is
deliberate and it is not a claim about the hardware. vLLM's
`sparse_attn_indexer` has no short-circuit: it scores the whole sequence and
runs the top-k selection on every row, padding with -1 when there are fewer
candidates than `topk`. Charging every row's scan to `x_idx` was tried and does
not work — within a purely unsaturated segment the scan and the attention are
then the *same trapezoid*, the two columns are exactly collinear, and only their
sum is determined. So `c_idx` is the price of **crossing** the bound, over and
above what the same row would have cost staying below it, and a short row's
indexer scan is priced inside `c_mla` where no measurement here can separate it
from that row's attention.

An earlier revision split attention by kernel path instead (capped vs uncapped)
for three coefficients. Measured, those two columns sit 170–177° apart across 15
cells with the same sign in 15 of 15 — collinear, so the split bought nothing,
and it left saturated cells with a dense price no segment could isolate (9 of 18
rejected outright).

**Solving is staged, not least-squares.** The regime decides which columns are
live, so a pure saturated segment isolates `c_idx` in one rung, a pure
unsaturated one isolates `c_mla`, and a mixed rung closes the rest. A cell
expresses exactly two of the three regimes. Algebraically, one pure rung plus
one mixed rung can determine both prices when the pure segment survives. The
planner nevertheless retains up to three rungs per usable segment to span the
range, make the pure-segment median robust, and leave residuals where the
candidate budget permits. At an unsaturated average point whose mixed segment
must carry both unknowns, three usable mixed candidates are required before the
segment is retained. Residuals are reported, never used to reject, and a segment
with exactly as many surviving rungs as unknowns is solved rather than refused
for having no residual to show.

How many cells that leaves standing is worth stating plainly: 17 of 45 in the
larger run, 11 of 23 in the smaller. The dominant reason for the rest is not the
fit but the engine -- 31 of the 40 rejections are cells where no label cleared
the step's own latency spread. At a cell whose average request is short the
prefill step sits at a flat ~240 ms whatever the spread, and a batch carrying
twice the modelled attention work moves the clock by 0.07 ms. Those are
measurements of an effect the step does not have; a cell that reports it is
telling you it is overhead-bound, not that the model is wrong. An earlier
revision, which demanded one extra rung per unknown on top of this, left 5 of 33.

Prices are solved per cell and not shared across cells, because they are not
constant. Over 45 cells, `c_idx` has a median of 1.09 ms per M reads (18 cells
price it, range 0.14–3.05) and `c_mla` a median of 6.13 ms/M (14 cells, range
2.13–44.18). Different cells run at different MFU, so a spread in the prices is
expected rather than a sign the fit is wrong; what it does rule out is fitting
one global pair. Only 14 cells price both columns — a purely unsaturated cell
prices nothing for crossing `topk` because none of its batches can cross it, and
that is a property of the cell, not a failure.

**Applying a price is gated.** Each rule is a fix for a measured failure, not a
precaution:

- `delta_idx >= 10 M` reads or `delta_mla >= 2 M`. Absolute, not a fraction of
  the cell's work: a relative gate divides a real 200 ms delta by a large cell's
  large denominator and discards it while keeping a 10 ms delta in a small one.
  Each threshold is the machine's 3-sigma step-time jitter (9.0 ms) over that
  column's median price.
- An unpriced column must be carrying no work. `None` for a column usually means
  it is identically zero across the cell, where the missing price multiplies
  zero and the correction is already whole. An unpriced column that *does* move
  work is refused: dropping it silently leaves the priced column to explain that
  work too, which is how a partial correction overshoots (measured max error
  87.6% -> 207.8%, worse than not correcting at all).
- No negative price, in the cell or at either end of an interpolation. A price
  is a marginal cost and cannot be below zero; a negative one means the fit is
  not describing the hardware, usually a mismeasured uniform batch dragging the
  whole column down.
- Interpolation only between two calibrated batch sizes bracketing the query at
  the same average point. Single-ended extrapolation and inverse-distance
  weighting were tried and produced every regression we measured — the worst
  turning a 0.30% estimate into 11.50%, on batches that were already fine.

Only the batch-size axis carries a price, and the reason is structural rather
than statistical: along it `avg_s + avg_p` is unchanged, so every request keeps
its kernel path and the cell keeps expressing the same regimes. Along `avg_s` a
row can cross `topk`, and the two ends of the interpolation are then not
answering the same question.

**Results** (GLM-5-NVFP4, 258 imbalanced batches over two independent runs,
bucketed by how wrong the uniform estimate was; median absolute error within
each bucket, through the shipped gates):

| uniform-estimate error | n | before | after |
|---|---|---|---|
| 0–2%   | 108 | 0.49% | 0.48% |
| 2–5%   | 34  | 3.49% | 2.36% |
| 5–10%  | 35  | 7.08% | 5.11% |
| 10–25% | 46  | 16.39% | 5.06% |
| >25%   | 35  | 43.87% | 38.97% |

Overall median 3.74% -> 1.09%. 89 of the 258 batches are corrected at all; the
rest are left alone by the delta gate, which is the intent — the 0–2% row barely
moves.

The `>25%` row is carried by an engine-level pathology rather than a modelling
gap, and it is worth separating out. Half of that bucket (17 of 35) sits on
cells where `avg_p` lands exactly on `topk`. There the *uniform* batch itself
misbehaves: at `b=2 avg_s=4096 avg_p=2048` it measures 875.6 ms with a CV of
0.02% while its neighbours at `avg_p=1984` and `avg_p=2112` measure 544.3 and
414.5 ms — a deterministic, reproducible spike, not noise, and the subtrahend of
every label in the cell. Excluding those cells:

| uniform-estimate error | n | before | after |
|---|---|---|---|
| 0–2%   | 106 | 0.48% | 0.45% |
| 2–5%   | 27  | 3.52% | 2.02% |
| 5–10%  | 30  | 7.08% | 2.53% |
| 10–25% | 36  | 16.39% | 3.85% |
| >25%   | 18  | 37.17% | 1.45% |

Median 2.31% -> 0.74% over the remaining 217 batches, 7 made worse. The
`avg_p == topk` boundary is left as it is: characterising it belongs with the
attention kernel, not here.

**What this PR does not do: apply the correction.** Nothing in the estimator
calls `predict_delta` yet, and that is a missing API rather than missing wiring.
Every prefill entry point today takes means only -- `predict_prefill_latency(bs,
isl, prefix)`, `run_static(batch_size=..., isl=..., prefix=...)` -- so there is
no argument through which a caller could describe the spread this module prices.
`run_static` already carries a scalar `seq_imbalance_correction_scale`, which is
the natural place for the result to land once a call site can supply the per-
request shapes; the compiled engine would need the same. Landing the model,
the calibration planner and the solver first keeps that follow-up reviewable on
its own, and lets the Dynamo-side collection start against a fixed definition of
the columns.

`sdk/work_delta/planner.py` — start at the module docstring, which is where the
two-column basis is defined and where the argument for excluding sub-`topk` rows
from `x_idx` lives. Everything else assumes it.

Then `sdk/work_delta/solver.py` — `solve_cell` for the staging, `predict_delta`
for the gates. Both a cell's own prices and an interpolated pair go through the
same `predict_delta`, so a carried price is held to what a calibrated one is.

`sdk/work_delta/field.py` (128 lines) — the whole application policy, and the
shortest file to disagree with.

`tests/unit/sdk/test_work_delta.py` — 33 behavioural tests, named for the
property rather than the function.

#### Related Issues:

- Relates to AIC-1668 (Intra-Batch Sequence-Length Heterogeneity)

---

**Not covered here.** The calibration batches are produced by a matching change
on the Dynamo side (`--benchmark-collect-imbalanced`), which is a separate PR;
without it this module has no data to fit. Coverage is also honest-but-thin: the larger run measured 45 cells, of which
17 survived the fit and application gates, and only a handful of average points
carry two calibrated batch sizes, so most queries away from a measured cell decline rather
than interpolate. One residual is unexplained — `b=16 avg_s=2048 avg_p=1024`
runs 113 ms above the model with no row near the `topk` boundary.


aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
```diff
@@ -0,0 +1,408 @@
...
+    The two MLA columns share one threshold: they are two paths through the
+    same kernel and a batch's rows split between them, so gating them
+    separately would fail a batch whose rows divide evenly -- which is the
+    batch with the largest deviation, not the smallest.
+    """
+    return (
+        abs(x_idx) >= MIN_ABS_DELTA_IDX_M * 1e6,
+        abs(x_mla_sparse) >= MIN_ABS_DELTA_MLA_M * 1e6,
+        abs(x_mla_dense) >= MIN_ABS_DELTA_MLA_M * 1e6,
+    )
```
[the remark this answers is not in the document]
YijiaZhao: Fixed. The even-split case is real — at 1.5M per column neither side clears a 2M gate although the batch moved 3M — so the two MLA columns are now summed once and the result gates both:

```python
mla_moved = abs(x_mla_sparse) + abs(x_mla_dense)
mla_passes = mla_moved >= MIN_ABS_DELTA_MLA_M * 1e6
return (abs(x_idx) >= MIN_ABS_DELTA_IDX_M * 1e6, mla_passes, mla_passes)
```

The docstring now carries that arithmetic so the next reader does not have to rediscover it.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
```diff
@@ -0,0 +1,408 @@
...
+    Every column is priced with the coefficient for the kernel that column's
+    rows actually run on. A cell that did not fit a coefficient contributes
+    nothing through it rather than borrowing another one's.
+    """
+    if not any(column_gate(x_idx, x_mla_sparse, x_mla_dense)):
+        return 0.0
+    delta = (fit.c_idx or 0.0) * x_idx + (fit.c_mla_sparse or 0.0) * x_mla_sparse + (fit.c_mla_mha or 0.0) * x_mla_dense
+    if noise > 0.0 and abs(delta) < MIN_LABEL_SNR * noise:
+        return 0.0
+    return delta
```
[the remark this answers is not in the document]
YijiaZhao: Fixed. `predict_delta` now rejects a negative price itself rather than relying on `CoefficientField` to have filtered the cell out. You are right that the function is public and reachable without the field.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
```diff
@@ -0,0 +1,408 @@
...
+        if live:
+            solved = _solve_n(mixed, tuple(col[n] for n in live), lambda p: p.y - known(p))
+            if solved is None:
+                fit.rejected.append(
+                    f"{len(mixed)} mixed rungs cannot close {len(live)} "
+                    "remaining prices: too few, or their columns are parallel"
+                )
+            else:
+                for name, value in zip(live, solved, strict=True):
+                    setattr(fit, name, value)
```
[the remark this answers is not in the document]
YijiaZhao: This one came from a stale docstring, which I have rewritten.

`c_mla_mha` is no longer fitted once across the unsaturated cells and handed to the rest as a known constant. Each cell owns all three of its prices. Measurement did not support the shared-constant theory: at a cell whose average request is long the dense column tops out at `(b - 1) * topk^2 / 2` while the cell's own work grows with `s_bar^2`, so the same nominal price buys a very different share of the step, and the fitted ratio varies by orders of magnitude between cells.

A saturated cell therefore has no dense rung and leaves `c_mla_mha` unset, which is correct — that cell's batches never take the dense path, so the column is identically zero and its price is never multiplied by anything.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
```diff
@@ -0,0 +1,408 @@
...
+    # own latency spread. Reported, not silent: a cell that drops most of its
+    # batches here is telling you its step is overhead-bound.
+    below_noise: int = 0
+
+    @property
+    def accepted(self) -> bool:
+        # Residuals are diagnostics. A cell whose segments each isolated their
+        # own price is determined, however large the scatter of a segment that
+        # happened to carry a spare rung.
+        return not self.rejected
```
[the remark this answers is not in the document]
YijiaZhao: This is deliberate, and reverting it would undo the change that made the module usable.

Collection affords one rung per pure segment and two for mixed — exactly enough to determine three prices and not one rung more. A pure segment solved from a single rung is exactly determined and has no residual to report; `residuals` is only populated where a segment happened to carry a spare rung. Gating on it therefore rejects cells for having been *lucky enough to be over-sampled*, not for being wrong.

Measured on the calibration grid: with the residual gate at 0.10, 5 of 33 cells survived and no average point had two calibrated batch sizes, so `CoefficientField`'s bracketing rule had nothing to bracket. Without it, 45 of 45 solve. Residuals are still recorded and worth reading; they are diagnostics, not admission criteria.

The protection you are reaching for is there, just expressed differently: a fit whose subtrahend was mismeasured shows up as a negative price, and both `CoefficientField` and (as of the latest commit) `predict_delta` decline those.

aic-core/src/aiconfigurator_core/sdk/work_delta/planner.py
```diff
@@ -0,0 +1,496 @@
...
+    """
+    b = len(rows)
+    x_idx = sum(idx_work(s, p, topk) for s, p in rows if runs_sparse(s, p, topk))
+    x_sp = sum(mla_work(s, p, topk) for s, p in rows if runs_sparse(s, p, topk))
+    x_dn = sum(mla_work(s, p, topk) for s, p in rows if not runs_sparse(s, p, topk))
+    if runs_sparse(s_bar, p_bar, topk):
+        x_idx -= b * idx_work(s_bar, p_bar, topk)
+        x_sp -= b * mla_work(s_bar, p_bar, topk)
+    else:
+        x_dn -= b * mla_work(s_bar, p_bar, topk)
```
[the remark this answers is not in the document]
YijiaZhao: The columns are not "which requests ran the indexer" — they are "which kernel path this row's work should be priced at", and the subtrahend is credited to whichever columns the *uniform* batch runs on. Those are different questions and the second is the one the model needs.

An unsaturated row does execute the indexer, and its cost is real; it is priced through `c_mla_mha`, which is the whole cost of the dense path at that cell rather than an attention-only term. Splitting it out would need a fourth column that no segment can isolate: a pure unsaturated segment moves indexer and dense attention together in fixed proportion, so the two are collinear by construction and only their sum is identifiable.

The consistency you are checking does hold. `work_columns` subtracts `b * idx_work(s_bar, p_bar)` when the average point is sparse and `b * mla_work(s_bar, p_bar)` when it is dense, so `c_idx * x_idx + c_mla_sparse * x_mla_sparse + c_mla_mha * x_mla_dense` reconstructs `T_batch - T_uniform` with the same accounting on both sides.

aic-core/src/aiconfigurator_core/sdk/work_delta/field.py
```diff
@@ -0,0 +1,126 @@
...
+class CoefficientField:
+    """The accepted per-cell fits, queryable at batch sizes between them."""
+
+    def __init__(self, fits: dict[tuple[int, int, int], CellFit], topk: int):
+        # Only accepted fits enter. A fit whose residual exceeded tolerance is
+        # worth reporting and worth keeping in the calibration record, but
+        # using it is not a lesser form of correcting -- one such cell, whose
+        # uniform batch had been mismeasured, produced coefficients that drove
+        # a 1709.8 ms estimate negative.
+        self._fits = {k: v for k, v in fits.items() if v.accepted}
```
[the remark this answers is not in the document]
YijiaZhao: Agreed in principle, and out of scope here.

A coefficient field is calibrated against one deployment — one model, one parallelism, one machine — and is queried by that deployment. `num_heads` is constant across every cell in a field, so adding it to the key would make the key wider without making it more selective.

It becomes load-bearing the moment fields are persisted and shared, which is where the head count stops being implicit. That is worth doing with the serialization format rather than ahead of it, so the bucket is checked at the boundary where a mismatched field can actually arrive.

aic-core/rust/aiconfigurator-core/src/fpm/metrics.rs
```diff
@@ -37,6 +37,31 @@ pub struct ScheduledRequestMetrics {
...
+    /// `var_decode_kv_tokens` on the decode side).
+    ///
+    /// Optional beyond the Dynamo v1 wire schema: emitters that do not report
+    /// it leave it at `0.0`, which makes the intra-batch work-delta term vanish
+    /// and reproduces the pre-existing uniform-batch estimate exactly. Paired
+    /// with `var_prefill_length` (variance of the FULL prompt length `s + p`)
+    /// it determines the trapezoid work delta
+    /// `(n / 2) * (var_prefill_length - var_prefill_kv_tokens)`.
+    #[serde(default)]
+    pub var_prefill_kv_tokens: f64,
```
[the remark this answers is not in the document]
YijiaZhao: This file is a stale model and should not have been in this PR.

`work_delta.rs` still carries the earlier single-scalar `beta_at(depth)` formulation, which the Python side replaced with the three kernel-path prices this PR describes. Reviewing it against the new model is not useful, and shipping both would leave the two engines disagreeing.

I am pulling the Rust side out and porting it separately once the Python contract settles.

aic-core/rust/aiconfigurator-core/src/lib.rs
```diff
@@ -44,6 +45,10 @@ pub use fpm::{
 // `engine/runtime.rs`) keep resolving after the types moved into `fpm`.
 pub(crate) use fpm::validate_forward_pass_metrics;
 pub use fpm::{ForwardPassMetrics, QueuedRequestMetrics, ScheduledRequestMetrics, FPM_VERSION};
+// Intra-batch prefill work delta: prices the gap between a heterogeneous
+// prefill batch and the uniform collapse `forward_pass_time_ms` estimates it
+// as. Uncalibrated by default, in which case every estimate is unchanged.
+pub use work_delta::{PrefillDeltaModel, PrefillDeltaPoint, PrefillShape};
```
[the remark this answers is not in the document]
YijiaZhao: This file is a stale model and should not have been in this PR.

`work_delta.rs` still carries the earlier single-scalar `beta_at(depth)` formulation, which the Python side replaced with the three kernel-path prices this PR describes. Reviewing it against the new model is not useful, and shipping both would leave the two engines disagreeing.

I am pulling the Rust side out and porting it separately once the Python contract settles.

aic-core/rust/aiconfigurator-core/src/work_delta.rs
```diff
@@ -0,0 +1,901 @@
...
+        if !(uniform_work > 0.0) {
+            return None;
+        }
+        // Var(L) − Var(p) = Var(s) + 2 Cov(s, p); halved and scaled by the
+        // batch size this is Σ trap(s, p) − n · trap(s̄, p̄).
+        let work_delta =
+            n * (scheduled.var_prefill_length - scheduled.var_prefill_kv_tokens) / 2.0;
+        if !work_delta.is_finite() {
+            return None;
+        }
```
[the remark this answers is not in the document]
YijiaZhao: This file is a stale model and should not have been in this PR.

`work_delta.rs` still carries the earlier single-scalar `beta_at(depth)` formulation, which the Python side replaced with the three kernel-path prices this PR describes. Reviewing it against the new model is not useful, and shipping both would leave the two engines disagreeing.

I am pulling the Rust side out and porting it separately once the Python contract settles.

aic-core/rust/aiconfigurator-core/src/work_delta.rs
```diff
@@ -0,0 +1,901 @@
...
+            return Err(CalibrationRejection::AnchorNotUniform);
+        }
+        let relative = rung_shape.relative_work_delta();
+        if relative.abs() < MIN_RELATIVE_RUNG_DELTA {
+            return Err(CalibrationRejection::RungHasNoDelta);
+        }
+        let beta = ((self.rung_wall_ms - self.anchor_wall_ms) / self.anchor_attention_ms) / relative;
+        if !beta.is_finite() {
+            return Err(CalibrationRejection::InvalidAttentionEstimate);
+        }
```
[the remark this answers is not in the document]
YijiaZhao: This file is a stale model and should not have been in this PR.

`work_delta.rs` still carries the earlier single-scalar `beta_at(depth)` formulation, which the Python side replaced with the three kernel-path prices this PR describes. Reviewing it against the new model is not useful, and shipping both would leave the two engines disagreeing.

I am pulling the Rust side out and porting it separately once the Python contract settles.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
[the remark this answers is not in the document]
YijiaZhao: You are right and my earlier reply was wrong. I claimed a saturated cell's dense column is identically zero; it is not. At a saturated average point `segments_for()` yields `(MIXED, SAT)`, the mixed rungs hold their short rows below `topk`, and `work_columns()` books those into `x_mla_dense`. Measured on the calibration grid, all 18 saturated cells have a nonzero dense column, up to 31.44M reads at `b=16 avg_s=4096`.

The consequence is exactly the one you predicted. With `c_mla_mha` unset, the mixed stage carries two unknowns and `_solve_n` refuses the fit:

```
b=2  avg_s=4096 avg_p=1024   2 mixed rungs cannot close 3 remaining prices
b=3  avg_s=4096 avg_p=1024   1 mixed rung  cannot close 3 remaining prices
b=5  avg_s=2048 avg_p=1024   1 mixed rung  cannot close 3 remaining prices
b=13 avg_s=4096 avg_p=1024   1 mixed rung  cannot close 3 remaining prices
```

Nine of the eighteen are rejected outright; the rest solve only because they happened to draw a third mixed rung, and a cell-local dense price from mixed rungs alone is not the price the model claims to be applying.

Keeping this open. The fix is a dense-price phase with an explicit contract rather than a caller convention — either the batch-size-wide fit passed into `solve_cell`, or the saturated plan carrying its own unsaturated rung so the cell stays self-contained. Which one is right depends on whether the dense price is stable across batch size, which I can answer from the collected grid before committing to an interface. I would rather settle that than ship a signature I have to change again.

aic-core/src/aiconfigurator_core/sdk/work_delta/field.py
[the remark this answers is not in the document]
YijiaZhao: No follow-up issue needed — this is already tracked on our side as part of the work-delta serialization work, and I would rather not have the same item in two trackers.

aic-core/rust/aiconfigurator-core/src/fpm/metrics.rs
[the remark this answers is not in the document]
YijiaZhao: No follow-up issue needed — the Rust port is already tracked on our side.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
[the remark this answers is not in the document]
YijiaZhao: Resolved by collapsing the attention split, not by threading a dense price through the solver.

You were right that a saturated cell's mixed rungs take the dense path — I measured it and all 18 saturated cells have a nonzero dense column. But the fix was upstream of the solver: the split into capped and uncapped attention should not have been there.

The original design had two features, and its stated test for grouping was geometry rather than kernel: *"x_idx packs MQA and topk — legitimate because they share the same trapezoid — and x_mla is the pairs delta."* Splitting attention by kernel path violates that test, because both paths count attention pairs. The cap is already expressed in how many pairs a row reads; a second coefficient adds nothing to express.

The measurements say the same thing twice over. Across 15 cells the two attention prices were the same sign every time, at a ratio of 1.0-3.4 (median 1.65), and the two columns sat at 170-177 degrees — near-antiparallel, because a mixed rung moving a row from one path to the other moves both in lockstep under the conserved totals. Only their sum was ever determined; the solver was splitting one identifiable quantity along a fixed direction and reporting it as two prices.

With one attention column the finding does not arise: a saturated cell's mixed segment has one unknown left after its pure segment fixes `c_idx`, and the two rungs the planner emits leave a residual. Rescored on the calibration grid, the two-price model is also the better one — median 3.62% -> 0.64% against 0.79% for three, with the 2-5% bucket at 0.25% against 2.43%.

`solve_cell` keeps its signature. No caller convention, no cross-cell parameter.

aic-core/src/aiconfigurator_core/sdk/work_delta/planner.py
```diff
@@ -0,0 +1,506 @@
...
+    or below the bound read all of their keys; longer ones are capped.
+    """
+    return s + p > topk
+
+
+def work_columns(rows: list[tuple[int, int]], s_bar: float, p_bar: float, topk: int) -> tuple[float, float]:
+    """``(x_idx, x_mla)`` against the cell's uniform batch.
+
+    Two columns, grouped by the GEOMETRY the work follows rather than by which
+    kernel runs it. The indexer scans the full trapezoid ``s*p + s^2/2`` whether
```
jasonqinzhou: **[P2] Make the two-column basis match the implementation.** This new contract says the indexer scans the full trapezoid whether or not top-k discards anything, but `work_columns()` still excludes every unsaturated row from `x_idx` and subtracts the uniform indexer only when the average is saturated. For `rows=[(96, 0), (32, 0)]` around `(s_bar=64, p_bar=0, topk=128)`, the raw `idx_work` deviation is 1024 while `work_columns()` returns `(0, 1024)`. That can be a valid identifiable basis only if `c_mla` intentionally absorbs unsaturated indexer work, but the planner and solver docs instead call `c_idx` the indexer price and still say short requests bypass it. Please define one exact basis, make `idx_work`, `work_columns`, and `solve_cell` describe it consistently, and add pure-unsaturated, saturated, mixed, and minimum-plan-to-solve behavioral tests.

aic-core/src/aiconfigurator_core/sdk/work_delta/planner.py
```diff
@@ -0,0 +1,506 @@
...
+    regime: Regime
+    m: int
+    short_len: int
+    rows: list[tuple[int, int]]
+    x_idx: float
+    x_mla: float
+
+    @property
+    def columns(self) -> tuple[float, float, float]:
+        return self.x_idx, self.x_mla
```
jasonqinzhou, line 408: **[P2] Finish the public two-column conversion.** This return now has two values, but `CalibrationBatch.columns` is still annotated as `tuple[float, float, float]` and the class docstring still promises three work columns. The planner and solver module docs still discuss three prices and a global dense fit, `Interpolated` still calls itself a coefficient triple, and the live PR title/body still advertise `c_mla_sparse`, `c_mla_mha`, and the removed Rust implementation. Please update the public types and documentation to the current two-column Python-only contract and add a small shape/behavior regression rather than relying only on import coverage.

jasonqinzhou (commented):
**COMMENT · 7.4/10 · high confidence**

## What this PR does

This PR adds Python-side calibration planning, coefficient solving and lookup, and dual-namespace SDK exports for pricing intra-batch prefill imbalance. The latest commit replaces the unidentifiable sparse/dense attention split with a two-column cell-local model; collection and runtime integration remain separate work.

The latest commit responds substantively to the solver feedback by measuring the attention-column collinearity and simplifying the planner, solver, and field together instead of adding a fragile cross-cell coefficient.

## Why this score

The current head is 7.4/10 with COMMENT: the two-column refactor removes the prior blocking three-price solver rejection, all targeted checks and 19 live CI checks pass, and the import surface remains sound. Two P2s remain: the stated two-column accounting basis still conflicts with what work_columns computes and has no behavioral tests, and the public return annotation, module documentation, and PR description still advertise the removed three-column contract.

aic-core/src/aiconfigurator_core/sdk/work_delta/planner.py
```diff
@@ -0,0 +1,506 @@
...
+    their sum was determined. Splitting them also left saturated average points
+    with a dense price no segment could isolate, which rejected nine of eighteen
+    such cells outright.
+
+    The subtrahend goes to whichever columns the UNIFORM batch runs on -- decided
+    by the average point's own length, not by the calibration batch's regime --
+    so an unsaturated average point contributes nothing to the indexer column.
+    """
+    b = len(rows)
+    x_idx = sum(idx_work(s, p, topk) for s, p in rows if runs_sparse(s, p, topk))
```
tianhaox, line 181: **Blocking (needs a decision, not necessarily a code change):** this line charges indexer work only for rows with `runs_sparse(s, p, topk)` — i.e. zero for rows at or below the bound. But `idx_work`'s docstring above says the opposite: vLLM's `sparse_attn_indexer` has *no* short-circuit, scores the whole sequence regardless, and "charging zero here made every request below the bound look free, which is exactly the requests a mixed batch is full of". `work_columns`'s own docstring agrees ("scans the full trapezoid … whether or not the top-k selection had anything to discard"). The code implements exactly the behavior the docstring calls a fixed bug.

This matters beyond documentation, because the solver's identification story depends on which side is true. Stage 1 relies on `x_idx ≡ 0` for unsaturated segments — that only holds for the *gated* version. If the ungated physics is right, an unsat row's indexer deviation and its attention deviation are the same trapezoid shape (near-collinear columns), so the "c_mla" pinned from unsat segments actually absorbs `c_idx + c_mla`, and handing that price to the mixed stage misprices the capped rows' attention work.

Which is authoritative — the gate or the docstring? If the gate is right (e.g. it's a deliberate reparameterization: below-bound indexer cost folded into `c_mla` because both follow the trapezoid), say so here explicitly and fix the two docstrings. If the docstring is right, the staging needs rework and the measured results should be re-derived.

aic-core/src/aiconfigurator_core/sdk/work_delta/planner.py
```diff
@@ -0,0 +1,506 @@
...
+
+    regime: Regime
+    m: int
+    short_len: int
+    rows: list[tuple[int, int]]
+    x_idx: float
+    x_mla: float
+
+    @property
+    def columns(self) -> tuple[float, float, float]:
```
tianhaox: Annotation says `tuple[float, float, float]` but this returns a 2-tuple. Part of a wider sweep worth doing: the three-column revision's language survives in several places — `planner.py` L21 ("Three prices, not two" followed by a two-term equation), `solver.py` L11 ("Three prices, because…" over the same two-term equation), `field.py`'s `Interpolated` ("A coefficient triple" with two fields). Please grep for "three"/"triple" and reconcile with the two-column model.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
```diff
@@ -0,0 +1,385 @@
...
+# column's measured price, so below it the correction is smaller than the
+# spread it would have to be verified against.
+#
+#   indexer  0.96 ms per M reads (0.943-0.999 across 12 cells, near-constant)
+#   MLA      ~5 ms per M, five times dearer
+#   jitter   9.0 ms median 3-sigma   ->  9.0/0.96 = 9.4 -> 10 M,  9.0/5 -> 2 M
+#
+# Re-derive both when the model, parallelism or hardware changes: the formula
+# transfers, these two numbers do not.
+MIN_ABS_DELTA_IDX_M = 10.0
```
tianhaox, line 91: Three generations of comments are stacked on top of these two constants and they contradict each other:

1. L83–98 describes a *relative* gate ("as a fraction of the cell's own total work") and concludes "Left at 0 (off)" — but no such constant exists, and the values below are 10.0 / 2.0. Stale; delete.
2. L99–101 is a Chinese internal note referencing "slide 6", which upstream readers can't see. Untranslated internal notes shouldn't land in the repo — fold whatever it adds into the English block or drop it.
3. L102+ is the current absolute-gate rationale, which is good — keep only this one.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
```diff
@@ -0,0 +1,385 @@
...
+        for r in range(n):
+            if r == col:
+                continue
+            factor = aug[r][col]
+            aug[r] = [v - factor * w for v, w in zip(aug[r], aug[col], strict=True)]
+    # Undo the normalisation so the prices are in the original units.
+    return tuple(aug[i][n] / norms[i] for i in range(n))
+
+
+def _solve_two(points, col_a, col_b, target) -> tuple[float, float] | None:
```
tianhaox: `_solve_two` is never called — `solve_cell` handles the two-unknown case with `_solve_n`. Dead code, and it takes `MAX_COLLINEARITY` down with it (that constant is used nowhere else, so the collinearity rejection it documents doesn't actually run; `_solve_n` guards via `MIN_GRAM_PIVOT` instead). Either delete both, or if the explicit collinearity bound is the intended guard for the 2×2 case, call this instead of `_solve_n` there.

aic-core/src/aiconfigurator_core/sdk/work_delta/field.py
```diff
@@ -0,0 +1,125 @@
...
+        weight = (b - b_lo) / (b_hi - b_lo)
+
+        def blend(name: str) -> float | None:
+            a, c = getattr(lo, name), getattr(hi, name)
+            if a is None or c is None:
+                return None
+            return a + (c - a) * weight
+
+        values = {name: blend(name) for name in COLUMNS}
+        if all(v is None for v in values.values()):
```
tianhaox: Two issues here.

(a) This guard allows a *partial* result: one column `None`, the other live (one end of the bracket solved only one price). `predict_delta`'s own docstring says corrections are all-or-nothing because a partial correction "leaves the surviving column to explain the whole delta and overshoots (measured max error 87.6% -> 207.8%)". Shouldn't this be `if any(v is None …): return None`?

(b) Bigger: nothing can consume an `Interpolated`. The only application function, `predict_delta`, takes a `CellFit`, and the absolute column gates + SNR gate live only there — so any future caller of the interpolation path either can't apply the coefficients or reimplements the gates by hand (and the negative-price veto is the only gate this class applies). Suggest making `predict_delta` accept the coefficient pair (or `Interpolated`) so the exact-hit and interpolated paths share one gated application. Since no caller exists yet this is API design rather than a live bug — but it's exactly the seam the integration PR will hit.

tianhaox (changes requested):
Thanks — the design notes in this PR are unusually good: every gate carries a measured failure as its justification, and "decline and fall back to the uniform estimate" is the right default. Requesting changes on three things, plus some smaller cleanups inline.

**1. The code contradicts its own docstrings on whether the indexer is gated** (inline at `planner.py` `work_columns`). This is the one I'd like resolved before anything else, because the solver's staged identification depends on which side is true.

**2. The PR title and body describe a previous revision of the code.** The title says "three kernel-path columns" and the body's model equation has `c_mla_sparse` / `c_mla_mha`, but the code has two columns (`c_idx`, `c_mla`) — `work_columns`'s docstring even explains why the three-column split was abandoned (near-antiparallel, only the sum determined). The body also directs reviewers to `rust/aiconfigurator-core/src/work_delta.rs`, which is not in this diff (9 files, all Python). Please either push the Rust port or drop it from the description, and update title/body to match the two-column model. The same stale "three prices" language survives inside the code (inline comment).

**3. No unit tests.** ~1,100 lines of numeric logic land with only the import-contract registration. Everything here is a dependency-free pure function, so the important invariants are cheap to pin:
- planner: every emitted batch conserves both totals exactly (`sum(s) == b*s_bar`, `sum(p) == b*p_bar`), satisfies `admits(rows, regime, topk)`, respects `max_model_len` and `MIN_NEW_TOKENS`;
- solver: `solve_cell` recovers known prices from synthetic measurements in each staging path (pure-sat + mixed, pure-unsat + mixed, mixed-only via `_solve_n`), and rejects the ill-conditioned cases it claims to;
- `predict_delta`: negative-price veto, absolute column gates, SNR gate — each on/off;
- field: exact hit / bracket / decline, and the negative-price veto at both ends.

For a module whose entire value is "the numbers are right", merging without these means every future refactor is unverifiable.

**Integration status, for other readers:** nothing consumes this module yet — there is no caller in the estimation path, calibration data comes from a separate Dynamo-side PR (`--benchmark-collect-imbalanced`), and how fits are persisted/ingested is not defined here. That's fine for a first PR, but worth one sentence in the body stating the intended wiring so nobody assumes estimates already include this correction.

Minor: `solver.py`'s module docstring says collection affords "one rung per pure segment and two for mixed", while the planner requires two rungs per pure segment (`_RUNGS_NEEDED`) and picks `max(needed, 3)` — align the story with the constants.

aic-core/src/aiconfigurator_core/sdk/work_delta/planner.py
YijiaZhao: Decided in favour of the gate, and the docstrings now say so explicitly — thank you for forcing the choice, the ambiguity was real and I had it backwards in prose.

The gate is authoritative, and it is a deliberate reparameterisation rather than a claim about the hardware. Your collinearity argument is exactly why: with the ungated column, a pure unsaturated segment has every row's scan and every row's attention on the *same* trapezoid, so the two columns are exactly collinear there and only their sum is determined. The gated column puts that sum where it is identifiable. So:

- `c_idx` is the price of a row **crossing** `topk`, over and above what the same row would have cost below it — not the price of running the indexer.
- `c_mla` is the whole cost of a row that stays below `topk`: its attention **and** its indexer scan, which no measurement here separates.

Naming `c_idx` after the indexer is shorthand, and the module docstring now says that in as many words, under a heading `WHICH ROWS ENTER x_idx`. `idx_work` keeps the ungated formula and its docstring now points at `work_columns` for the separate question of which rows are charged; `work_columns` and `solve_cell` describe the same basis.

Your `[(96, 0), (32, 0)]` example is in the test suite as `test_only_rows_crossing_the_bound_enter_the_scoring_column`, asserting `x_idx == 0` *and* that both rows have nonzero `idx_work` — so the two facts are pinned together and a future refactor cannot quietly reconcile them the wrong way.

Behavioural tests added as asked: pure-unsaturated, pure-saturated, mixed, and minimum-plan-to-solve, plus the regime boundary and the gates. 31 in `tests/unit/sdk/test_work_delta.py`.

aic-core/src/aiconfigurator_core/sdk/work_delta/planner.py
YijiaZhao: Decided in favour of the gate, and the docstrings now say so explicitly — thank you for forcing the choice, the ambiguity was real and I had it backwards in prose.

The gate is authoritative, and it is a deliberate reparameterisation rather than a claim about the hardware. Your collinearity argument is exactly why: with the ungated column, a pure unsaturated segment has every row's scan and every row's attention on the *same* trapezoid, so the two columns are exactly collinear there and only their sum is determined. The gated column puts that sum where it is identifiable. So:

- `c_idx` is the price of a row **crossing** `topk`, over and above what the same row would have cost below it — not the price of running the indexer.
- `c_mla` is the whole cost of a row that stays below `topk`: its attention **and** its indexer scan, which no measurement here separates.

Naming `c_idx` after the indexer is shorthand, and the module docstring now says that in as many words, under a heading `WHICH ROWS ENTER x_idx`. `idx_work` keeps the ungated formula and its docstring now points at `work_columns` for the separate question of which rows are charged; `work_columns` and `solve_cell` describe the same basis.

Your `[(96, 0), (32, 0)]` example is in the test suite as `test_only_rows_crossing_the_bound_enter_the_scoring_column`, asserting `x_idx == 0` *and* that both rows have nonzero `idx_work` — so the two facts are pinned together and a future refactor cannot quietly reconcile them the wrong way.

Behavioural tests added as asked: pure-unsaturated, pure-saturated, mixed, and minimum-plan-to-solve, plus the regime boundary and the gates. 31 in `tests/unit/sdk/test_work_delta.py`.

aic-core/src/aiconfigurator_core/sdk/work_delta/planner.py
YijiaZhao: All done.

- `CalibrationBatch.columns` is `tuple[float, float]`, docstring says "two work columns".
- `planner.py` and `solver.py` module docs describe two prices; the global dense fit is gone, and the docstring now says why it was dropped — the ratio varies by orders of magnitude between cells because the capped column tops out at `(b-1)·topk²/2` while the cell's own work grows with `s_bar²`.
- `Interpolated` is "A price pair".
- PR title and body rewritten: `c_mla_sparse` / `c_mla_mha` and the Rust implementation are gone.

While rewriting the body I found two things in it that were false under the two-column fit and fixed those too: it claimed `c_idx` was "0.943–0.999 across 12 cells, effectively constant" (measured, 4 of 18 cells fall in that window; median 1.09, range 0.14–3.05), and its results table had two rows transposed. The same stale figures were in a `solver.py` comment and are now the measured ones.

Regression tests rather than import coverage: shape is asserted directly (`test_a_plan_carries_two_columns`), and behaviour around it — `test_the_minimum_plan_solves_both_prices` runs a real `plan_cell` output through `solve_cell` and checks both prices come back.

aic-core/src/aiconfigurator_core/sdk/work_delta/planner.py
YijiaZhao: All done.

- `CalibrationBatch.columns` is `tuple[float, float]`, docstring says "two work columns".
- `planner.py` and `solver.py` module docs describe two prices; the global dense fit is gone, and the docstring now says why it was dropped — the ratio varies by orders of magnitude between cells because the capped column tops out at `(b-1)·topk²/2` while the cell's own work grows with `s_bar²`.
- `Interpolated` is "A price pair".
- PR title and body rewritten: `c_mla_sparse` / `c_mla_mha` and the Rust implementation are gone.

While rewriting the body I found two things in it that were false under the two-column fit and fixed those too: it claimed `c_idx` was "0.943–0.999 across 12 cells, effectively constant" (measured, 4 of 18 cells fall in that window; median 1.09, range 0.14–3.05), and its results table had two rows transposed. The same stale figures were in a `solver.py` comment and are now the measured ones.

Regression tests rather than import coverage: shape is asserted directly (`test_a_plan_carries_two_columns`), and behaviour around it — `test_the_minimum_plan_solves_both_prices` runs a real `plan_cell` output through `solve_cell` and checks both prices come back.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
YijiaZhao: All three, as described.

1. Deleted. The constant it documented no longer existed, so it was 15 lines of rationale with nothing under it.
2. Deleted.
3. Kept, and corrected — the figures in it were still the three-column ones. Now: `c_idx` median 1.09 ms/M over 18 cells (0.14–3.05), `c_mla` median 6.13 over 14 (2.13–44.18), 3σ jitter 9.0 ms, giving 9.0/1.09 → 10 M and 9.0/6.13 → 2 M. The comment now also says the gate is sized from the *median* and that a cell far from it gets a gate that is loose or tight for it — the intended trade, since a per-cell gate would be derived from the very fit it guards.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
YijiaZhao: Deleted, both. `MIN_GRAM_PIVOT` inside `_solve_n` is the real guard and the only one that ever ran, so keeping `MAX_COLLINEARITY` around would have documented a rejection that does not happen.

aic-core/src/aiconfigurator_core/sdk/work_delta/field.py
YijiaZhao: (a) and (b) both done, and (b) turned out to be the more useful of the two.

`predict_delta` now takes a `Prices` protocol — anything carrying `c_idx` and `c_mla` — which both `CellFit` and `Interpolated` satisfy. The exact-hit and interpolated paths go through one gated application, so a carried price is held to exactly what a calibrated one is. That closes the seam you point at.

On (a): I first made it strict `if any(v is None): return None`, and the measured error got **worse** — median 0.64% → 0.80% on one run, 0.86% → 1.26% on the other. The reason is that a purely unsaturated cell leaves `c_idx` unsolved *because* every one of its batches has `x_idx ≡ 0`, so the missing price multiplies zero and the correction was already whole; refusing it threw away good corrections.

So the rule is now the one the overshoot argument actually supports: **an unpriced column is refused only when that column moves work**, at the same threshold as the magnitude gate. In `field`, a column neither end priced is carried through as `None` and left for `predict_delta` to judge; a column only *one* end priced still declines the query, since that is single-ended extrapolation of a real price. Both cases are pinned by tests.

aic-core/src/aiconfigurator_core/sdk/work_delta/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+    CellPlan,
+    Regime,
+    admits,
+    classify,
+    idx_work,
+    mla_work,
+    plan_cell,
+    segments_for,
+    work_deltas,
+)
```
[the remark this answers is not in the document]
YijiaZhao: Fixed in 65741a5 — thank you, and apologies for leaving this one without a reply while the others got answered.

`__init__.py` now imports and exports `work_columns`, the name `planner.py` actually defines, so the package imports. The same commit added the `src/aiconfigurator/sdk/work_delta/` facade, which re-exports the canonical package rather than restating its symbol list, so the two cannot drift apart the way this pair had.

The import contract test was extended to cover the new modules in 55e39f3, which is what would have caught this: it imports every public SDK package, so a name that does not exist fails there rather than at a user's first import.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
```diff
@@ -0,0 +1,371 @@
...
+Two prices, because a prefill batch is charged on two different shapes: the
+candidate-scoring trapezoid a row presents by crossing ``topk``, and the
+attention pairs every row actually reads. See :mod:`planner` for what each
+column does and does not include -- in particular, a sub-``topk`` row's indexer
+scan is priced through ``c_mla``, not ``c_idx``.
+
+Nothing here fits both at once. The order is forced by what each segment
+can move, and each step subtracts what the previous one already fixed:
+
+    1. c_mla  from the unsaturated segments. Those batches short-circuit the
```
jasonqinzhou: **[P2] Finish reconciling the two-column contract.** The revised opening now correctly says a sub-`topk` indexer scan is folded into `c_mla`, but the staged narrative immediately below still says unsaturated batches "short-circuit the indexer", treats `c_mla` as a dense-kernel price fitted per batch size rather than per cell, and says residuals above a tolerance invalidate a cell even though `CellFit.accepted` ignores residuals. `planner._knobs` likewise still talks about "two attention columns", and the documented/PR-body budget of one pure plus two mixed batches does not match `plan_cell`: `(8, 4096, 1024, 2048)` emits two mixed and three saturated batches. This is the contract the Dynamo collector must consume, so please update the remaining narrative and batch-budget claim to the current per-cell two-column behavior (or change the planner if the smaller budget is intended).

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
```diff
@@ -0,0 +1,371 @@
...
+    """
+    n = len(cols)
+    # n, not n+1. With exactly n rungs the system is square, so it reproduces
+    # its labels whatever the prices are and leaves no residual to read. That is
+    # a reason to report no residual, not a reason to refuse: residuals are
+    # diagnostics here and never reject a fit, and collection cannot afford a
+    # spare rung -- the planner emits two mixed rungs because two is what a
+    # mixed segment needs. Demanding a third rejected 2 cells outright whose
+    # prices were fully determined.
+    if len(points) < n:
```
jasonqinzhou, line 220: **[P2] Pin the square mixed-only solve with a regression.** This commit changes `_solve_n` from rejecting `len(points) == n` to accepting it, which is what lets two mixed rungs determine both prices when no pure segment survives. None of the 31 new tests reaches the `fit.c_idx is None and fit.c_mla is None` branch: `test_the_minimum_plan_solves_both_prices` gets `c_idx` from a pure saturated rung and only exercises the one-dimensional closure. A direct two-measurement probe currently recovers known prices `(3.0, 7.0)`, so please codify that case plus an almost-parallel rejection to protect this guard and `MIN_GRAM_PIVOT`.

jasonqinzhou (approved):
**COMMENT · 8.5/10 · high confidence**

## What this PR does

This PR adds a Python-side work-delta package that plans conserved-total calibration batches, solves per-cell two-column prices, interpolates accepted fits across batch size, and exposes the API through both SDK namespaces. The current commit resolves the prior basis ambiguity, adds 31 behavioral tests, and makes exact and interpolated prices share the same application gates.

The update engages the earlier feedback directly: it explains why the gated scoring column is an identifiable reparameterization, turns the reviewer's concrete unsaturated example into a regression test, and unifies the calibrated and interpolated application path behind one protocol.

## Why this score

The current head is 8.5/10 with COMMENT: the two prior runtime/contract concerns are substantively fixed, targeted tests and adversarial probes pass, all 19 live checks are green, and no blocking correctness issue remains. Two focused P2s remain: the newly accepted square mixed-only solver path is not covered by the committed tests, and several solver/planner narratives still describe the removed model and a calibration-batch budget that does not match plan_cell's output.

tianhaox (approved):

tianhaox:
Re-reviewed at 730daceb6. **All five of my round-1 comments are genuinely resolved**, and several are resolved better than what I asked for:

1. **Indexer gate vs docstring (my blocking):** resolved by decision, and the new basis writeup in `planner.py` (module doc + `idx_work` + `work_columns`) is exactly what was needed — `c_idx` redefined as "the cost of crossing the bound", the sub-`topk` scan explicitly folded into `c_mla` with the collinearity argument spelled out. The identification story is now self-consistent. Thanks for making the call explicit instead of just patching prose.
2. **Three-column sweep:** title, body, `columns` annotation, `Interpolated` all done — with one leftover, below.
3. **Stacked comment generations over `MIN_ABS_DELTA_*`:** deleted, and the surviving block re-derived with the two-column figures plus an honest note on the per-cell spread. Good.
4. **`_solve_two` / `MAX_COLLINEARITY`:** gone.
5. **`Prices` protocol + one gated application path:** this closed the seam properly, and the blend semantics (one-end-priced → decline; both-`None` → carried through and re-checked by gate 1) are the right split. The tests covering each gate are appreciated.

The 31 tests and the green CI address my third blocking. I'm clearing my REQUEST_CHANGES once the two open P2s from @jasonqinzhou are handled — I verified both against this head and they are real:

- **solver.py module doc (P2, seconding):** the staged narrative at L20–26 still describes the previous revision — unsaturated batches "short-circuit the indexer" (contradicts the new `planner.py` contract), `c_mla` "fitted per batch size rather than per cell -- it is a property of the dense kernel" (contradicts both the code and `planner.py`'s "Each cell owns both of its prices"), and the closing paragraph still references "the tolerance" although `RESIDUAL_TOLERANCE` was deleted in this same commit.
- **Square mixed-only solve untested (P2, seconding):** confirmed none of the 31 tests reaches the `fit.c_idx is None and fit.c_mla is None` → `_solve_n` branch — `test_mixed_rungs_close_the_price_the_pure_segment_left_open` carries a pure rung, and `test_the_minimum_plan_solves_both_prices` is a saturated average point with a pure sat rung. The branch this commit deliberately relaxed (`len(points) < n` instead of `<= n`) is exactly the one nothing pins.

Two new items from this pass:

**(A) The rung-count prose drifted again, in three places that now disagree.** `planner.py` L486–488 was edited to say a mixed segment at an unsaturated average point "needs two", but L531 still reads `needed = 3 if (regime == MIXED and not avg_is_sat)`; meanwhile `_solve_n`'s new comment says "the planner emits two mixed rungs because two is what a mixed segment needs", while the planner actually picks `max(needed, 3)`. I think the design is coherent — plan three for noise margin, solve with two survivors — but the three comments each tell a different story. Please make them tell that one.

**(B) Non-blocking question on interpolation magnitude.** The new `field.py` docstring honestly reports that `c_mla` spans **20.7x across batch sizes at one average point** — which retires the old "prices are near-constant along b" justification. Bracketing still guarantees the two ends are commensurable, but a *linear* blend in `b` between ends 20x apart is close to arbitrary in magnitude at mid-bracket, and the gates catch sign and smallness, not a wrong-magnitude positive price. Two questions: did the 258-batch validation actually exercise the interpolated path (non-exact-hit queries), and if so how did mid-bracket points fare? And was blending in log-space considered, given the spread is multiplicative? If the validation covered this, a sentence saying so in the docstring would close it; if it didn't, that's worth knowing before the integration PR leans on interpolation.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
YijiaZhao: Done in de3feb1e. All four narrative remnants are reconciled with the code:

- **"short-circuit the indexer"** is gone. Stage 1 now says what the code relies on: in an unsaturated segment no row crosses `topk`, so `x_idx` is identically zero *by the column definition* and the segment pins `c_mla` on its own — the whole below-bound price, a row's attention together with its own indexer scan (vLLM runs the scan either way; the basis just prices it through `c_mla`).
- **Per cell, not per batch size**: the stage-1 note now states `c_mla` is solved per cell like every other price, and records why the shared-kernel-property version was rejected (the ratio varies by orders of magnitude between cells).
- **Residuals**: the module tail and `solve_cell` now say explicitly that a large residual is a diagnostic to read, never a gate — `CellFit.accepted` is structural only. The `field.py` entry comment was aligned to the same contract (negative-price fits are caught at query time by `at`, not filtered on entry).
- **`_knobs` wording**: "attention columns" → "work columns" (both places).

On the budget: I kept the planner and changed the words. `plan_cell` emits up to three rungs per segment (`_pick_rungs(..., max(needed, 3))`, floors 2, or 3 for a mixed segment at an unsaturated average point), and your `(8, 4096, 1024, 2048)` example — two mixed + three saturated — is exactly that. The shipped validation (258 batches, the medians in the PR body) was collected on this budget, so the docstrings now describe it instead of shrinking the plan to match stale prose; the algebraic minimum (one pure + two mixed) remains stated as the floor, with the extra rungs feeding the pure-segment median and the readable residual.

aic-core/src/aiconfigurator_core/sdk/work_delta/solver.py
YijiaZhao: Done in de3feb1e — both regressions added:

- `test_two_mixed_rungs_alone_carry_both_prices`: an unsaturated average point with no pure segment, two mixed rungs built from real row constructions (prefix concentrated on one long row vs spread across the short rows). The prefix placement is what buys the conditioning — it moves the `x_idx:x_mla` ratio between the rungs — so the square `_solve_n` recovers the planted `(3.0, 7.0)` exactly through the `len(points) == n` path your comment points at.
- `test_nearly_parallel_mixed_rungs_are_refused`: two mixed rungs varying only row lengths at zero prefix, whose column ratios come out almost identical. The normalised Gram pivot falls below `MIN_GRAM_PIVOT`, the fit is refused, and both prices stay `None` — pinning the guard itself, not just the happy path.

(The second fixture is the shape a naive two-rung probe naturally produces, which is a nice property: the test documents why the well-conditioned fixture needs the prefix trick.)

PR #1488, opened 2026-08-05
fix(sdk): make the dense-attention conflict log diagnosable (kernel_source, version, full key)

Arsene12358:
## Summary

Companion to #1487, completing [AIC-1663](https://linear.app/nvidia/issue/AIC-1663) item 2's second box — split into its own aic-core-scoped PR per the module-boundary rule (the collector-side invariant test rides #1487).

The dense-attention loaders strip `kernel_source` from the row key, so two backends colliding on one shape key are resolved first-wins by parquet row order — and the old conflict message omitted `kernel_source` entirely (plus one index dimension per phase: `b` in context, `s` in generation), making such a collision undiagnosable from logs. The message now names the full key, the dropped row's `kernel_source`, and its `version` (both via `row.get(...)`, `None`-safe on pre-column tables).

Deliberately **stays at debug level**: first-wins is load-bearing for the declared-reuse and earlier-version merge channels (`_build_op_sources` appends them after the primary), whose rows are *expected* to conflict on every shared key — warn level would spam every database load. The new comment documents this, and what a same-version dual-`kernel_source` conflict would mean. Log/comment-only change; no control-flow or data change, so no Rust parity action (the Rust loader is silent first-wins by design).

## Verification

`ruff format --check` + `ruff check` clean; `pytest tests/unit -k "attention and not collector"`: 94 passed, 5 skipped.

Part of AIC-1663.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced attention CSV conflict logs with kernel source and version details.
  * Conflict handling continues to preserve the first row encountered while clearly identifying later discarded entries.

### DIFF
--- a/aic-core/src/aiconfigurator_core/sdk/operations/attention.py
+++ b/aic-core/src/aiconfigurator_core/sdk/operations/attention.py
@@ -1229,11 +1229,18 @@ def load_context_attention_data(context_attention_file):
         kv_cache_dtype = common.KVCacheQuantMode[kv_cache_dtype]
 
         try:
-            # Check for conflict
+            # Check for conflict. First-wins is load-bearing: earlier-version /
+            # declared-reuse sources are appended after the primary and are
+            # EXPECTED to conflict on shared keys, so this stays debug-level
+            # (warn would spam every load). kernel_source is stripped from the
+            # key, so the message must name it — a same-version conflict with
+            # two kernel_source values means two backends collided on one shape
+            # key and the loser was dropped by row order (AIC-1663).
             context_attention_data[quant_mode][kv_cache_dtype][kv_n][head_size][window_size][n][s][b]
             logger.debug(
                 f"value conflict in context attention data: {quant_mode} {kv_cache_dtype} "
-                f"{head_size} {window_size} {kv_n} {n} {s}"
+                f"{head_size} {window_size} {kv_n} {n} {s} {b} — dropping later row "
+                f"(kernel_source={row.get('kernel_source')}, version={row.get('version')})"
             )
         except KeyError:
             # Store all three values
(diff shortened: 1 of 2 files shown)


aic-core/src/aiconfigurator_core/sdk/operations/attention.py
```diff
@@ -1229,11 +1229,18 @@ def load_context_attention_data(context_attention_file):
         kv_cache_dtype = common.KVCacheQuantMode[kv_cache_dtype]
 
         try:
-            # Check for conflict
+            # Check for conflict. First-wins is load-bearing: earlier-version /
+            # declared-reuse sources are appended after the primary and are
+            # EXPECTED to conflict on shared keys, so this stays debug-level
```
jasonqinzhou: **[P2] Distinguish expected reuse overlap from a real backend collision.** Keeping every duplicate at `DEBUG` avoids reuse spam, but it also leaves the unsafe case described below invisible at the normal INFO CLI level. `_read_filtered_rows()` flattens primary, declared-reuse, fallback, and cross-backend rows, and the stored leaf retains only latency/power/energy, so by the time this branch runs the retained row's provenance is gone. A same-version probe with `winner_backend` followed by `dropped_backend` kept the winner's latency, emitted nothing at INFO, and at DEBUG named only `dropped_backend`; the log therefore cannot identify the two backends that collided. Please track the first row's `(kernel_source, version)` per full key, keep expected reuse overlaps at DEBUG, but warn for same-version/different-kernel collisions while naming both kept and dropped sources. Add equivalent `caplog` coverage for the context and generation loaders.

jasonqinzhou (approved):
**COMMENT · 8.4/10 · high confidence**

## What this PR does

This PR makes first-wins dense-attention row conflicts easier to inspect by adding the complete context/generation lookup key plus the dropped row's kernel_source and version to the SDK loader's debug messages, while documenting why ordinary shared-layer reuse conflicts remain debug-level.

The patch correctly restores the missing batch dimension for context, the post-step sequence dimension for generation, and uses row.get(...) so legacy tables without kernel_source or version remain safe.

## Why this score

8.4/10, COMMENT. The patch is small, behavior-preserving, clean under Ruff, green across the 94-test attention selection, and backed by all passing exact-head CI. One non-blocking observability gap remains: expected reuse overlaps and unsafe same-version backend collisions still share the same suppressed DEBUG path, and the message records only the dropped row, so it cannot identify both backends involved in the collision described by the new comment.

aic-core/src/aiconfigurator_core/sdk/operations/attention.py
jasonqinzhou: Fixed in `45927839`: both loaders now retain the kept row's `(kernel_source, version)` per full key, keep cross-version/legacy reuse overlap at `DEBUG`, and warn for same-version/different-`kernel_source` collisions while naming both kept and dropped provenance. Added equivalent context and generation `caplog` coverage; the local attention selection passes (96 passed, 5 skipped).

jasonqinzhou (approved):
Approved at `45927839`. The follow-up cleanly separates expected reuse overlap from same-version backend collisions, preserves first-wins behavior, logs both kept and dropped provenance, and adds equivalent context/generation coverage. The review thread is resolved; all 19 exact-head checks are green.

PR #1491, opened 2026-08-05
perf(core): optimize MoE perf-table queries

jthomson04:
## Summary

- build hierarchical categorical/numeric indices when MoE perf tables load, replacing per-query full-table scans and query-string construction
- store MoE token-latency data as immutable sorted curves, replacing per-query generic interpolation-tree construction
- convert ordinary and low-latency MoE, SGLang DeepEP/WideEP, TRT-LLM WideEP, MegaMoE, projected dispatch, and mHC callers without adding a compatibility path or changing public APIs
- share one direct curve-owning MoE index across ordinary, SGLang WideEP, and TRT-LLM WideEP paths, and retain raw DeepEP-normal dispatch points instead of multiplying them into six curves per SMS bucket

## Details

The lookup index preserves existing distribution precedence and fallback rules, duplicate handling, quantization selection, and error behavior. Successful queries borrow their loaded key directly; detailed key construction is retained only on error paths.

`TokenCurve` performs exact lookup, adjacent-point interpolation, and boundary util-hold directly over one load-time sorted boxed slice. Generic multidimensional interpolation remains available for unrelated operators.

## Validation

Correctness:

- `cargo test -p aiconfigurator-core --lib`: 309 passed
- `cargo test -p aiconfigurator-core --all-features --lib`: 316 passed
- `cargo check -p aiconfigurator-core-public-api`
- relevant non-Torch Python MoE, WideEP, shared-layer, compile-engine, and engine-step tests: 108 passed
- rebuilt-extension compile-engine and engine-step parity corpus: 376 passed
- `test_moe_dispatch.py` could not be collected because the local development environment does not include its optional Torch dependency; the corresponding Rust dispatch tests passed
- `git diff --check`
- final-tree Clippy has 64 existing warnings versus 65 at the pre-fix head; the `token_curve.rs` warning is gone and the review fix adds no new warning
- targeted `rustfmt --check` over the Rust files changed by this PR

Focused tests cover indexed selection/fallback/duplicates plus exact, interpolated, singleton, and held-boundary curves across ordinary and low-latency MoE, SGLang DeepEP/WideEP, TRT-LLM WideEP, MegaMoE, dispatch projections, and mHC.

Performance (uncommitted harness, internal sampling only):

| Group | Before | After | Improvement |
| --- | ---: | ---: | ---: |
| 4,096-shape distribution lookup | 16,719 ns | 14 ns | 1,194.21x |
| Combined 12-curve MoE corpus | 1,160 ns | 54 ns | 21.48x / 95.34% lower |

| Engine-step case | Before | After | Result |
| --- | ---: | ---: | ---: |
| SILICON, Kimi K2.5 vLLM | 88.920 us | 60.903 us | 31.51% faster |
| EMPIRICAL, Kimi K2.5 vLLM | 32,496.620 us | 31,924.127 us | 1.76% faster |
| HYBRID miss, Astra DeepSeek V4 Flash SGLang | 122.885 us | 46.002 us | 62.56% faster |

All 12 prediction values in every engine-step case were exactly equal between revisions.

The review-fix lookup point is 14 ns versus 196 ns at the pre-fix head, and the combined curve point is 54 ns versus 73 ns, so the follow-up introduces no material lookup regression.

A final 59.998-second, 12-worker steady-state Astra profile captured 5,870 samples with no loss. The former `MoeTable::resolve_distribution` hotspot (23.87% self) and old `query_token_curve` helper have no samples; the new `TokenCurve::query` is 0.41% self. Remaining generic `perf_interp::Node` construction is under unrelated collective-message interpolation, not any converted MoE token-curve path.

The optimized one-hour Astra input completes before a 60-second attachment can finish, so this structural confirmation used an uncommitted deterministic 2x extension of the same trace with shifted timestamps and hash remapping that preserves within-copy prefix relationships without cross-copy reuse. The older profile and candidate use different coherent AIC wire-schema revisions and are therefore used for structural hotspot confirmation, not as a direct end-to-end throughput comparison.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved performance-data lookup and interpolation across MoE, mHC, WideEP, and DeepEP configurations.
  * Added more consistent token-based performance estimates between measured points and beyond measured ranges.
  * Improved resolution when requested distribution data is unavailable through appropriate fallback behavior.
* **Reliability**
  * Preserved existing validation, error handling, kernel selection, and boundary behavior.
  * Added coverage for interpolation, fallback resolution, duplicate data, and edge cases.

### DIFF
--- a/aic-core/rust/aiconfigurator-core/src/perf_database/dsv4_megamoe.rs
+++ b/aic-core/rust/aiconfigurator-core/src/perf_database/dsv4_megamoe.rs
@@ -43,9 +43,9 @@ use std::collections::BTreeMap;
 use std::path::PathBuf;
 use std::sync::OnceLock;
 
+use super::token_curve::TokenCurve;
 use crate::common::enums::MoeQuantMode;
 use crate::common::error::AicError;
-use super::token_curve::TokenCurve;
 use crate::perf_database::parquet_loader::PerfReader;
 
 pub struct Dsv4MegaMoeTable {
@@ -215,7 +215,11 @@ fn load_module_parquet(path: &PathBuf) -> Result<Dsv4MegaMoeGrids, AicError> {
     for row in reader.rows()? {
         let row = row?;
         for (col, expected, error) in [
-            (used_cuda_graph_col, true, "DSv4 MegaMoE perf row was not collected with CUDA Graph"),
+            (
+                used_cuda_graph_col,
+                true,
+                "DSv4 MegaMoE perf row was not collected with CUDA Graph",
+            ),
             (
                 includes_gate_topk_col,
                 false,
@@ -228,7 +232,10 @@ fn load_module_parquet(path: &PathBuf) -> Result<Dsv4MegaMoeGrids, AicError> {
             ),
         ] {
             if row.bool(col)? != expected {
-                return Err(AicError::PerfDatabase(format!("{error}: {}", path.display())));
+                return Err(AicError::PerfDatabase(format!(
+                    "{error}: {}",
+                    path.display()
+                )));
             }
         }
         let phase = row.str_owned(phase_col)?;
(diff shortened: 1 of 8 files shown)


jthomson04:
@coderabbitai Follow-up on the public-API advisory: perf_database is declared as a private module at the crate root, so wideep_moe and its grid/key records are not part of the supported external Rust API. cargo check -p aiconfigurator-core-public-api passes at 977cbee. The follow-up therefore makes the internal grid/key records private without a compatibility shim. I also made the dispatch non-empty invariant unconditional at construction, with diagnostic first/last expectations.

tedzhouhk (approved):
great work!

PR #1495, opened 2026-08-05
perf(core): cache successful GEMM queries

jthomson04:
## Summary

- add a private, per-`GemmTable` `quick_cache::sync::Cache` using its modified CLOCK-Pro/S3-FIFO policy, with 32,768 total unit-weight capacity split evenly across 16 shards (2,048 entries per shard); skew can therefore evict before global occupancy reaches 32,768
- use 16 internal shards rather than the host-core-derived default, keeping construction cost and per-table memory predictable on large CPU hosts
- use Quick Cache as the pattern for hot, high-cardinality scalar-query memoization; DSA's small, unbounded map instead caches a handful of lazily loaded table objects
- key successful GEMM table results by normalized quant and `(m, n, k)`, so `fp8_static` and `fp8` share the raw lookup
- preserve error precedence and cache only finite successful table results; errors, fallbacks, provenance, compute-scale, and scale-matrix results are unchanged
- keep miss resolution outside cache synchronization and allow benign duplicate concurrent misses

No public API, prediction, fallback, or provenance behavior changes.

## Validation

Correctness:

- `cargo check -p aiconfigurator-core`
- `cargo test -p aiconfigurator-core gemm_query_cache` — 4 passed, including the exact production 16-shard/2,048-entry-per-shard skew case and a forced four-way duplicate-miss case (4 resolutions, 1 cached entry, then a resolver-free hit)
- `cargo test -p aiconfigurator-core` — 314 passed
- `cargo test -p aiconfigurator-core --features embed-python` — 321 unit tests and 3 integration tests passed
- deterministic 12-worker DynoSim reports were byte-identical after removing only `wall_time_ms`, `processed_tokens_per_s`, and `processed_output_tokens_per_s`; normalized SHA-256: `9a9fc6f636917b1bd84c5c97450299e08e5ca28167875cb1729a87a2f55a51e3`
- `git diff --check`
- exact `6169ef4` versus `ead965e` release safety check: no out-of-line `resolve_gemm_query_cached` symbol, identical corpus digests, repeated-hit latency 12.95% lower, and unique-cold latency unchanged within 0.002%

`cargo clippy -p aiconfigurator-core --all-targets -- -D warnings` remains blocked by the current `main` baseline (64 library and 70 test-target diagnostics). `cargo fmt --all -- --check` likewise reproduces the repository's existing formatting drift.

Performance, exact `main` baseline `3f05337`, ordinary release with `RUSTFLAGS="-C target-cpu=x86-64-v3"`:

| Benchmark | Baseline | Candidate | Result |
| --- | ---: | ---: | ---: |
| repeated GEMM queries | 376.793 ns/query | 26.820 ns/query | 14.05x faster |
| unique cold GEMM queries | 1,596.714 ns/query | 1,682.506 ns/query | 5.37% slower |
| SILICON engine-step corpus | 61.593 us | 25.328 us | 58.88% faster |
| EMPIRICAL engine-step corpus | 32,327.515 us | 32,041.709 us | 0.88% faster |
| Astra HYBRID-miss engine-step corpus | 46.899 us | 26.024 us | 44.51% faster |

Quick Cache is 3.51x faster than the prior Moka candidate on repeated hits and reduces its cold-query overhead by 17.26%. It exceeds the repeated-query and engine-step targets, misses the synthetic <=3% cold-regression target by 2.37 percentage points, and produces bit-identical prediction vectors.

Matched DynoSim point using Dynamo `e322c8d5ce0649c86c9235dac5e3243007fd6a85`, 12-worker aggregated Astra SGLang/B200, CPU 0, candidate then baseline, no warm-ups:

| Revision | Wall time | Processed tokens/s | Requests | Processed tokens |
| --- | ---: | ---: | ---: | ---: |
| baseline | 48.495 s | 26,609,074 | 93,953 | 1,290,416,132 |
| candidate | 42.418 s | 30,421,081 | 93,953 | 1,290,416,132 |

That is 14.33% higher processed-token throughput and 12.53% lower wall time in this point comparison. Against the deterministic Moka candidate, Quick Cache improved throughput by another 0.62%. The host has no isolated CPU; preflight showed background host load but no competing process on CPU 0.

The deterministic baseline, Moka, and Quick Cache reports all completed 93,953 requests with identical token totals and were byte-identical after normalization.

## Summary by CodeRabbit

* **Performance**
  * Improved repeated matrix multiplication configuration queries through bounded caching.
  * Reduced redundant computation for frequently requested matrix shapes and quantization settings.
  * Cache behavior avoids retaining failed or invalid results and remains responsive under concurrent access.

### DIFF
--- a/aic-core/rust/aiconfigurator-core/src/perf_database/gemm.rs
+++ b/aic-core/rust/aiconfigurator-core/src/perf_database/gemm.rs
@@ -45,10 +45,14 @@ struct GemmQueryKey {
     k: u32,
 }
 
-fn gemm_query_cache(capacity: usize) -> Cache<GemmQueryKey, f64> {
+// Quick Cache is the intended pattern for hot, high-cardinality scalar-query
+// memoization in the perf database. DSA's `sparse` map is deliberately
+// different: it holds a small, unbounded set of lazily loaded table objects,
+// rather than memoizing high-volume scalar results.
+fn gemm_query_cache() -> Cache<GemmQueryKey, f64> {
     let options = OptionsBuilder::new()
-        .estimated_items_capacity(capacity)
-        .weight_capacity(capacity as u64)
+        .estimated_items_capacity(GEMM_QUERY_CACHE_CAPACITY)
+        .weight_capacity(GEMM_QUERY_CACHE_CAPACITY as u64)
         .shards(GEMM_QUERY_CACHE_SHARDS)
         .build()
         .expect("valid static GEMM query cache options");
@@ -92,7 +96,12 @@ pub struct GemmTable {
     gemm: OnceLock<Result<GemmEngineGrids, AicError>>,
     compute_scale: OnceLock<Result<TwoDGrids, AicError>>,
     scale_matrix: OnceLock<Result<TwoDGrids, AicError>>,
+    /// Successful finite scalar GEMM queries, keyed by normalized quant and
+    /// shape and bounded independently for each shard.
     query_cache: OnceLock<Cache<GemmQueryKey, f64>>,
+    // These test-only counters intentionally measure probes at this call site.
+    // Quick Cache's optional internal statistics cannot distinguish a missed
+    // probe from `GemmTable::query` never reaching the cache at all.
     #[cfg(test)]
     query_cache_hits: AtomicUsize,
     #[cfg(test)]
@@ -177,15 +186,15 @@ impl GemmTable {
         // data miss when the quant's table also happens to be uncollected.
         let spec = &self.system_spec;
         let tc_flops = quant_tc_flops(spec, lookup_quant.mapping())?;
+        // Keep the cache probe below FLOPS resolution, even on hits. Hoisting
+        // it would change MissingSystemFlops-over-PerfDatabase error precedence.
         let key = GemmQueryKey {
             quant: lookup_quant,
             m,
             n,
             k,
         };
-        let cache = self
-            .query_cache
-            .get_or_init(|| gemm_query_cache(GEMM_QUERY_CACHE_CAPACITY));
+        let cache = self.query_cache.get_or_init(gemm_query_cache);
         if let Some(value) = cache.get(&key) {
             #[cfg(test)]
             self.query_cache_hits.fetch_add(1, Ordering::Relaxed);
@@ -208,6 +217,8 @@ impl GemmTable {
         };
         let cfg = gemm_engine_config(&sol);
         let value = index.resolve(&cfg, &[m as f64, n as f64, k as f64])?;
+        // Preserve resolver behavior by returning non-finite results, but do
+        // not make them sticky cache hits for later queries.
         if value.is_finite() {
             cache.insert(key, value);
         }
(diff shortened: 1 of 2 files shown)


Arsene12358 (approved):
Approving — I went looking for a merge blocker and could not find one. I re-verified the claims independently rather than relying on the PR-body numbers.

**Correctness**

- The cache key `(normalized quant, m, n, k)` is complete: downstream of the probe the result is a pure function of the key plus per-table immutable state (`system_spec` fixed at construction, `tc_flops` derived from spec + normalized quant, grids behind `OnceLock`). The cache is per-`GemmTable`, and `GemmTable` is neither `Clone` nor serialized, so there is no aliasing path to divergent cache state.
- Sharing the raw lookup between `fp8` and `fp8_static` is parity-safe because `normalize_fp8_static_quant` already runs before every lookup on `main`, so the two were the same computation at this level to begin with; the `fp8_static`-only compute_scale / scale_matrix paths are separate methods and untouched.
- Probe placement below `quant_tc_flops` preserves the MissingSystemFlops-over-PerfDatabase error precedence, and since errors and non-finite values never enter the cache there is no path to a sticky failure or sticky NaN. A hit skipping `load_gemm()` is safe because a hit can only exist if an identical query already succeeded and the `OnceLock`-backed grids cannot later regress.

**Independent verification**

- `cargo test -p aiconfigurator-core` on the PR head: 314 passed locally (rustc 1.96.0, macOS arm64), including the 5 new cache tests and the Python-v2 parity anchor.
- Cross-branch bit-exactness: I appended an identical env-gated dump test to `gemm.rs` on base `3f05337` and on the PR head, dumping `f64::to_bits()` (or the full error string) for 9 quant modes × 37 m-values × 13×13 (n, k), plus the fp8/fp8_static compute_scale and scale_matrix side tables, in two passes — 114,478 rows per branch, 56,277 distinct keys, deliberately above the 32,768-entry capacity so insert→evict→re-miss all get exercised. The dumps are byte-identical across branches (identical SHA-256), pass 2 (cache hits) is byte-identical to pass 1 within the PR, and all ~53k error rows (including every uncollected quant) match text-for-text.
- `cargo clippy --all-targets` produces identical diagnostics on main and on the PR — the clippy drift noted in the description is fully pre-existing.

**Non-blocking notes**

1. `Cargo.lock` isn't updated for `quick_cache`, but that lockfile has been stale since #1322 (it already lacks bincode/parquet/pyo3) and no CI job builds with `--locked`, so this is a repo-level follow-up (refresh + enforce `--locked`, or drop the file) rather than something to fix here.
2. The disclosed +5.37% synthetic cold-query cost against the ≤3% target reads as the right trade given every realistic corpus improves — appreciated the transparency about missing the target.
3. Per-shard eviction skew is disclosed and pinned by `gemm_query_cache_enforces_production_per_shard_bound`; its worst case is recomputation, never a wrong value.

Nice, careful PR — the probe-placement comment and the errors-never-cached semantics made this straightforward to verify.

aic-core/rust/aiconfigurator-core/src/perf_database/gemm.rs
```diff
@@ -727,6 +820,141 @@ mod tests {
...
+            .collect();
+        let values: Vec<_> = handles
+            .into_iter()
+            .map(|handle| handle.join().unwrap())
+            .collect();
+
+        assert!(values
+            .windows(2)
+            .all(|pair| pair[0].to_bits() == pair[1].to_bits()));
+        assert_eq!(table.query_cache_stats().entries, 1);
```
jasonqinzhou: **[P2] Make this test prove the duplicate-miss path.** The barrier only synchronizes threads before `query()`. One worker can still resolve and insert before the others call `cache.get()`, and the current assertions (identical values and `entries == 1`) pass in both that case and a single-flight/serialized implementation. Since this PR explicitly preserves benign concurrent duplicate misses to keep resolution outside cache synchronization, please add a test-only hook or barrier after the miss probe and assert `query_cache_stats().misses == THREADS` while retaining the single-entry assertion.

jasonqinzhou (commented):
**COMMENT · 8.8/10 · high confidence**

## What this PR does

This PR adds a private bounded Quick Cache to each Rust GemmTable so repeated normalized GEMM latency queries avoid rerunning interpolation while preserving existing error, fallback, provenance, compute-scale, and scale-matrix behavior.

The implementation carefully keeps FLOPS error precedence ahead of cache hits, shares fp8_static and fp8 entries only after normalization, excludes errors and non-finite values, and documents both shard-capacity tradeoffs and measured end-to-end performance.

## Why this score

8.8/10 with COMMENT: the production cache integration is sound and the focused and full Rust unit suites pass, with no blocking correctness issue found. The one P2 asks the new concurrency regression to force and measure duplicate misses; its current assertions would also pass if only one thread missed or miss resolution became serialized.

jasonqinzhou (approved):

PR #1499, opened 2026-08-06
fix: Qwen3.5 modeling and collection gaps (GDN hybrid)

wenpengw-nv:
#### Overview:

  Correctness fixes for Qwen3.5 (hybrid GDN + full-attention, dense & MoE), from per-op timeline alignment (vLLM 0.24.0 / SGLang 0.5.14) and review. Modeling + collector only — no refreshed perf data (see data note).

  #### Details:

  **SDK modeling:**
  - Per-op composition vs serving: qkv GEMM missed the output gate (2x q slice, ~44% under); `in_proj_ba` is its own GEMM (`2*num_v_heads`); shared expert = gated-SiLU + scalar`hidden -> 1` gate; per-head q/k RMSNorm now priced (`use_qk_norm=True`).
  - Hybrid memory: token-linear KV only on full-attention layers (was all layers — 4x over on 35B-A3B); GDN layers charge a constant per-request FP32 state.
  - GDN lookup: TP-local head counts, **exact-or-SOL** in Python and the Rust twin (nearest-shape borrowing could serve one SKU from another's rows as silicon). SOL formulas: conv width `2K+V`, FP32 state.
  - MoE collectives: vLLM runs no pre-dispatch collective at `attention_dp == 1` — gated behind new `MoEDispatch(attn_ar_modeled=...)`, default False keeps every other model's behavior (vLLM branch only). SGLang default path omits pre-dispatch at dp == 1. SGLang DeepEP tables store the full dispatch+combine round trip, so composing pre+post double-charged every layer — now one dispatch op (`SGLangEPMOEModel` precedent), with the shared expert replicated (`tp_size=1`), not TP-sharded.
  - Generation: vLLM decode runs shared/routed experts on parallel streams — now an `OverlapOp` (deepseek precedent) plus a `shared_merge` element-wise op. `moe_backend/is_context/sms` forwarded; EPLB stays unmodeled; `cp_size > 1` rejected.
  - `ENGINE_SPEC_SCHEMA_VERSION` 6 -> 7 (`MoEDispatchOp` gained a positional bincode field).

  **Collector (GDN):**
  - Serving parity: conv metadata precomputed outside the timed span (was a D2H sync + numpy repack per call); FP32 recurrent state; SGLang `dt_bias` dtype; conv state in the serving SD layout + transposed view (the kernels take explicit strides).
  - Timing domain: graph replay everywhere (the eager prefill conv timed the launch envelope, ~4.9x high; replay matches production to -0.3%), `repeat_n=10` kernels per replay (decode conv ~2x high at repeat_n=1).
  - Cases: GDN expands TP-local geometries over declared `tensor_parallel_sizes` (TP16 on MoE SKUs); exact below-grid GEMM widths (scalar gate n=1, b/a widths per TP).

  **Data dependency (follow-up collection PR):** packaged `gdn_perf` tables predate the TP-local lookup — tp>1 queries fall back to SOL, and vLLM tables carry the old calibration even at tp=1. The new GEMM widths / TP16 geometries have no rows yet (SILICON reports missing data; HYBRID works). Re-collect with the target system's existing gdn parquet removed first.

  **Scope:** Qwen3.5 text tower only (vision encoder not modeled); TRT-LLM not validated — its dispatch branch keeps the legacy proxy AR and its GDN collector is untouched.

  - `aic-core/src/aiconfigurator_core/sdk/models/qwen35.py` — op composition, memory overrides
  - `aic-core/src/aiconfigurator_core/sdk/operations/moe.py` — `attn_ar_modeled`, the only cross-model surface (default preserves existing behavior)
  - `aic-core/src/aiconfigurator_core/sdk/operations/mamba.py` + `aic-core/rust/aiconfigurator-core/src/perf_database/state_space.rs` — exact-or-SOL parity
  - `collector/vllm/collect_gdn.py`, `collector/sglang/collect_gdn.py` — serving parity, timing domain
  - `collector/case_generator.py`, `collector/cases/models/Qwen3_5*_cases.yaml` — case expansion

  - Relates to GitHub issue: #xxx

### DIFF
--- a/aic-core/rust/aiconfigurator-core/src/config.rs
+++ b/aic-core/rust/aiconfigurator-core/src/config.rs
@@ -34,7 +34,9 @@ pub const ENGINE_CONFIG_SCHEMA_VERSION: u32 = 1;
 // - 6: `Kda` op variant appended (Kimi-K3). Claimed version 5 on its own
 //   branch concurrently with #1460, so the merge renumbered it (same
 //   precedent as the v3/v4 collision above).
-pub const ENGINE_SPEC_SCHEMA_VERSION: u32 = 6;
+// - 7: `MoEDispatchOp` gained `attn_ar_modeled` — a bincode op-layout
+//   change (same class as v5).
+pub const ENGINE_SPEC_SCHEMA_VERSION: u32 = 7;
 
 /// Static engine identity and setup information carried by an
 /// [`crate::engine::spec::EngineSpec`].
--- a/aic-core/rust/aiconfigurator-core/src/engine/spec.rs
+++ b/aic-core/rust/aiconfigurator-core/src/engine/spec.rs
@@ -327,6 +327,7 @@ mod tests {
             is_context: false,
             sms: 12,
             scale_num_tokens: 1,
+            attn_ar_modeled: false,
         }
     }
 
--- a/aic-core/rust/aiconfigurator-core/src/operators/mamba.rs
+++ b/aic-core/rust/aiconfigurator-core/src/operators/mamba.rs
@@ -165,7 +165,8 @@ impl GdnOp {
         let hk = self.head_k_dim as f64;
         let nv = self.num_v_heads as f64;
         let hv = self.head_v_dim as f64;
-        let conv_channels = nk * hk + nv * hv;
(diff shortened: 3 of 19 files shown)


wenpengw-nv:
Superseded pr #1503

PR #1503, opened 2026-08-07
fix: Qwen3.5 modeling and collection gaps (GDN hybrid)

wenpengw-nv:
#### Overview:

  Correctness fixes for Qwen3.5 (hybrid GDN + full-attention, dense & MoE), from per-op timeline alignment (vLLM 0.24.0 / SGLang 0.5.14) and review. Also collected data on vllm and sglang.

  #### Details:

  **SDK modeling:**
  - Per-op composition vs serving: qkv GEMM missed the output gate (2x q slice, ~44% under); `in_proj_ba` is its own GEMM (`2*num_v_heads`); shared expert = gated-SiLU + scalar`hidden -> 1` gate; per-head q/k RMSNorm now priced (`use_qk_norm=True`).
  - Hybrid memory: token-linear KV only on full-attention layers (was all layers — 4x over on 35B-A3B); GDN layers charge a constant per-request FP32 state.
  - GDN lookup: TP-local head counts, **exact-or-SOL** in Python and the Rust twin (nearest-shape borrowing could serve one SKU from another's rows as silicon). SOL formulas: conv width `2K+V`, FP32 state.
  - MoE collectives: vLLM runs no pre-dispatch collective at `attention_dp == 1` — gated behind new `MoEDispatch(attn_ar_modeled=...)`, default False keeps every other model's behavior (vLLM branch only). SGLang default path omits pre-dispatch at dp == 1. SGLang DeepEP tables store the full dispatch+combine round trip, so composing pre+post double-charged every layer — now one dispatch op (`SGLangEPMOEModel` precedent), with the shared expert replicated (`tp_size=1`), not TP-sharded.
  - Generation: vLLM decode runs shared/routed experts on parallel streams — now an `OverlapOp` (deepseek precedent) plus a `shared_merge` element-wise op. `moe_backend/is_context/sms` forwarded; EPLB stays unmodeled; `cp_size > 1` rejected. TODO: vLLM stops overlapping above 256 tokens/rank — always-overlapped is optimistic there.
  - Guards: `cp_size > 1` and `moe_backend='megamoe'` rejected (sglang MegaMoE serves DeepSeek-V4 only; letting it through priced the attention AR twice).
  - Below-grid GEMM shapes (scalar gate n=1, ba widths at high TP) degrade to SOL in SILICON via an opt-in `below_grid_sol` flag; quant misses stay strict, HYBRID unchanged, exact rows win once collected. Rust mirrors the rule.
  - `ENGINE_SPEC_SCHEMA_VERSION` 6 -> 8 (v7: `MoEDispatchOp.attn_ar_modeled`; v8: `GemmOp.below_grid_sol`).

  **Collector (GDN):**
  - Serving parity: conv metadata precomputed outside the timed span (was a D2H sync + numpy repack per call); FP32 recurrent state; SGLang `dt_bias` dtype; conv state in the serving SD layout + transposed view (the kernels take explicit strides).
  - Timing domain: graph replay everywhere (the eager prefill conv timed the launch envelope, ~4.9x high; replay matches production to -0.3%), `repeat_n=10` kernels per replay (decode conv ~2x high at repeat_n=1).
  - Cases: GDN expands TP-local geometries over declared `tensor_parallel_sizes` (TP16 on MoE SKUs); exact below-grid GEMM widths (scalar gate n=1, b/a widths per TP).

   **Data**: the GDN re-collection ships in this PR (`d69affee`) as a wholesale delete-then-write replacement (8 systems × {vLLM 0.24.0, SGLang 0.5.14}); GDN serves silicon at every declared TP on both backends, with no stale-row window. `collection_meta.yaml` carries provenance and reviewed failure tails; `check_collector_data.py` passes.

  Remaining data-conditional: the declared exact-width GEMM rows (gate n=1, GDN ba) are uncollected — those two ops take the explicit `below_grid_sol` SOL degrade until rows land.TRT-LLM is out of scope end-to-end (not validated; legacy proxy AR; old GDN rows).

**Scope:** Qwen3.5 text tower only (vision encoder not modeled); TRT-LLM not validated — its dispatch branch keeps the legacy proxy AR and its GDN collector is untouched.

  - `aic-core/src/aiconfigurator_core/sdk/models/qwen35.py` — op composition, memory overrides
  - `aic-core/src/aiconfigurator_core/sdk/operations/moe.py` — `attn_ar_modeled`, the only cross-model surface (default preserves existing behavior)
  - `aic-core/src/aiconfigurator_core/sdk/operations/mamba.py` + `aic-core/rust/aiconfigurator-core/src/perf_database/state_space.rs` — exact-or-SOL parity
  - `collector/vllm/collect_gdn.py`, `collector/sglang/collect_gdn.py` — serving parity, timing domain
  - `collector/case_generator.py`, `collector/cases/models/Qwen3_5*_cases.yaml` — case expansion

  - Relates to GitHub issue: #xxx

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive Qwen3.5 and Qwen3.5 MoE modeling across supported tensor-parallel configurations.
  - Added backend-aware attention, expert dispatch, shared-expert, and memory accounting.
  - Added optional SOL estimates for GEMM shapes outside measured performance grids.
  - Expanded GEMM and GDN benchmark coverage across Qwen3.5 variants.

- **Bug Fixes**
  - Improved GDN estimates for packed projections and recurrent-state traffic.
  - Prevented inaccurate nearest-shape fallback results.
  - Corrected communication and attention reduction cost calculations.

- **Improvements**
  - Aligned benchmark behavior with serving implementations.
  - Updated engine specification compatibility for new modeling options.


Arsene12358 (changes requested):
## Review: modeling verified correct — one hard blocker on the data path

The per-op corrections here are real, and I verified them independently against serving source at the pinned versions (vLLM v0.24.0, SGLang v0.5.14, transformers main): the 2× gated q-slice (sigmoid before o_proj), unconditional per-head q/k RMSNorm, the qkvz (2K+2V) + ba (2·num_v_heads) GEMM split exactly as both frameworks merge the checkpoint's four Linears, conv over 2K+V with z excluded, fp32 SSM / bf16 conv state (vLLM adopts the checkpoint's `mamba_ssm_dtype` via `models/config.py:603-616`), the Qwen2-MoE-style scalar shared-expert gate, vLLM's no-collective-at-dp==1 pre-dispatch (`moe_runner.py:719-753`), the sglang StandardDispatcher behavior, the DeepEP one-round-trip convention, and the decode overlap — including the family nuance that sglang 0.5.14 enables CUDA shared-expert *fusion* for qwen3_next but not qwen3_5, which makes the OverlapOp right for this class. The exact-or-SOL lookup change and its Rust mirror are rule-for-rule identical with the pinned tests inverted correctly. Unit suite: 3,035 passed. Nice work — this is the most thoroughly cited collector change I've reviewed in this repo.

### Hard blocker: every Qwen3.5-MoE SILICON run fails on this branch, all backends

The recomposed shared expert adds `{prefix}_shared_expert_gate_gemm` with **n=1** to every MoE layer on every backend, no packaged GEMM table has n=1 (base grid min 32; the new case widths are declared but uncollected), and `GEMM.query` **raises** on a below-grid miss (`gemm.py:618`) instead of degrading like `GDNKernel` does. Reproduced with the support matrix's own pinned commands:

| Command (currently PASS in the matrix) | main `e3c6d55` | this branch |
|---|---|---|
| `Qwen3.5-35B-A3B · gb200 · sglang 0.5.14 · SILICON · 32 GPU` | agg 2,851 results · best 32,551 tok/s/gpu | **agg 0, disagg 0** — `GEMM perf data not available… n=1, k=2048` at every parallelism |
| `Qwen3.5-35B-A3B · b200_sxm · trtllm 1.3.0rc20 · SILICON · 32 GPU` | PASS | **agg 0, disagg 0** — identical n=1 failure |
| same, `--database-mode HYBRID` | — | works (3,105 results) — matches your data note |

The data note frames this as "SILICON reports missing data," but it's a total-failure flip of currently-green support-matrix rows on merge day — and the follow-up collection plan can't heal **TRT-LLM**, whose collector/data are explicitly out of scope while its compositions also carry the n=1 gate. The GDN-side staleness (tp>1 → SOL until re-collect) degrades gracefully and is a fine interim; the GEMM side is not.

**Suggested fix (small, self-healing, consistent with this PR's own exact-or-SOL philosophy):** an explicit opt-in on the op for below-grid degradation —

```python
# qwen35.py
ops.GEMM(f"{prefix}_shared_expert_gate_gemm", count, 1, h,
         common.GEMMQuantMode.bfloat16, scale_num_tokens=..., below_grid="sol")

# gemm.py — scoped, everything else keeps the strict raise
if n < grid_min_n and self._below_grid:
    return PerformanceResult(get_sol()[0], source="sol")
```

mirrored in `operators/gemm.rs` with an oracle anchor per the parity rule, plus a Python+Rust test pinning "below-grid + flag → SOL; without flag → raise." Exact rows win automatically once the collected n=1/ba widths land. (Alternatives — landing the collection PR first extended to trtllm GEMM widths, or moving the shared-expert recomposition into the data PR — also resolve it, but heavier and the query stays brittle for the next model.)

### Important: one missing serving citation (collector rule)

`collector/sglang/collect_gdn.py:399` — the `dt_bias` float32→model-dtype change is **correct** (sglang creates `dt_bias` without an explicit dtype, `qwen3_5.py:237-239 @0.5.14`, and the kernel upcasts on load, `fused_sigmoid_gating_recurrent.py:154-160`) but per the metadata-parity rule every changed hand-constructed input needs its citation comment. One line fixes it.

### Soft notes (no change requested in this PR)

- **Schema v7 vs the in-flight large-EP cascade**: both touch EP dispatch surfaces; the v5/v6 renumber-at-merge precedent applies — coordinate merge order.
- **GDN state reservation counts one request** (kimi_k3 single-elastic-pool precedent — fine to stay consistent). At 397B scale that's ≈189 MB/request; at concurrency 64, ~12 GB of unreserved state. Worth a follow-up issue for batch-aware accounting; it will matter for the AIC-1149 alignment work.
- **vLLM decode overlap** is modeled unconditionally while serving overlaps only ≤256 tokens/rank (env default) — already comment-documented; consider a batch clamp when sweeps explore very deep decode batches.

Happy to re-review as soon as the below-grid path lands — everything else is mergeable as-is.

wenpengw-nv:
@Arsene12358 Thanks for the review— both addressed in c16beaa6.

  **Below-grid SOL:** added below_grid_sol=True (boolean) on the n=1 gate GEMM and the GDN in_proj_ba GEMM (n=2·nv/tp is also below grid at tp≥16). Degrades only on a SILICON shape  miss — quant misses still raise, HYBRID verified unchanged, exact rows win once collected.
  
  **dt_bias**: citation added, re-verified against v0.5.14.

  **Note two commits**: After your snapshot for the re-review: 6b0a56fc (post-AR after merge; sglang fused gate removes n=1 from its default path) and ef645f08 (decode conv update from causal_conv1d_triton).

collector/vllm/collect_gdn.py
```diff
@@ -179,12 +210,10 @@ def run_conv1d(_conv_input=conv_input, _conv_state=conv_state):
...
-                repeat_n=1,
-                # vLLM 0.24's packed prefill convolution performs a metadata
-                # copy that CUDA graph capture rejects. Every SM90 full-run
-                # point therefore used eager execution; make that method
-                # explicit instead of silently falling back at runtime.
-                use_cuda_graph=False,
+                # Production launches this op eagerly but back-to-back in a deep
+                # queue, so the row must hold GPU service time; a sync-bounded
+                # eager loop would time the launch envelope instead.
+                repeat_n=10,
```
tianhaox, line 216: **Blocking (needs an answer / a mechanism, not necessarily code in this PR): stale-data lifecycle.**

At tp=1 the old packaged `gdn_perf` rows key **identically** to the new lookup (global heads == local heads), but they were measured under the old timing domain this hunk retires (eager prefill conv timing the launch envelope, ~4.9x high by your own numbers; `repeat_n=1` decode conv ~2x high) and the old layouts/dtypes. After this PR merges, those rows keep serving as **SILICON** under the new composition — miscalibrated numbers labeled as measured.

The PR body says the follow-up collection must "remove the target system's existing gdn parquet first", but nothing enforces that. Could you:
1. make the data PR hard-delete the stale gdn rows for affected systems (or version the op so stale rows structurally cannot serve — cf. the collector op-version-management design), and
2. state the intended merge order and the interim blast radius — in the window between this PR and the data PR, Qwen3.5 SILICON at tp=1 is known-wrong on vLLM; will the daily support matrix surface that or silently absorb it?

aic-core/src/aiconfigurator_core/sdk/models/qwen35.py
```diff
@@ -192,8 +351,11 @@ def _ffn_context_ops(
                 ops_list.append(
                     ops.GEMM(f"{prefix}_router_gemm", count, cfg.num_experts, h, common.GEMMQuantMode.bfloat16)
                 )
-            ops_list.extend(
-                [
+            # StandardDispatcher (sglang default MoE path) has no pre-dispatch
+            # collective when attention_dp == 1; with DP the LayerCommunicator
+            # pre-MLP gather is priced by the dispatch op.
+            if not (self._backend_name == "sglang" and self.config.moe_backend is None and attn_dp == 1):
```
tianhaox, line 363: **Blocking: the `moe_backend is None` gate leaves `megamoe` with the double-counted attention AR this PR fixes elsewhere.**

CLI accepts `--moe-backend megamoe`, and this model doesn't reject it. Verified empirically on this branch (sglang, megamoe, tp=4, attn_dp=1): the op list contains **both** `*_moe_pre_dispatch` and the explicit `*_gdn_ar`/`*_full_ar`. The sglang non-DeepEP dispatch branch charges an all-reduce at attn_tp>1 (`operations/moe.py:1305-1307`), so the attention all-reduce is priced twice — exactly the error this PR corrects for the vLLM and sglang-default paths, silently surviving on the megamoe path.

Either reject megamoe explicitly for Qwen3.5 (same pattern as the `cp_size > 1` guard above), or — if megamoe also routes through the StandardDispatcher, with a citation — widen the gate to `!= "deepep_moe"`. Same applies to the twin gate in `_ffn_generation_ops`.

aic-core/src/aiconfigurator_core/sdk/models/qwen35.py
```diff
@@ -385,37 +561,74 @@ def _ffn_generation_ops(
...
+                    is_context=False,
+                    moe_backend=self.config.moe_backend,
+                    enable_eplb=False,
+                )
+            )
+            shared_ops = (
+                self._shared_expert_ops(prefix, count, h, tp, gemm_q, cfg) if cfg.shared_expert_inter_size > 0 else []
+            )
+            # vLLM and sglang decode run shared and routed experts on
+            # parallel CUDA streams (vLLM only up to 256 tokens per rank —
```
tianhaox: **Non-blocking: the 256-tokens/rank overlap gate is real and reachable.**

Throughput sweeps routinely put decode above 256 tokens per rank, where vLLM disables the dual-stream shared-expert overlap but this model still prices it overlapped — a systematic optimistic bias exactly in the large-batch region that decides max-throughput configs. `OverlapOp.query` sees `x`, so the gate could be applied at query time; at minimum leave a TODO stating the bias direction.

tianhaox (changes requested):
Reviewed the full diff against the serving-alignment claims and the collector governance rules; also ran the PR's touched unit tests locally (100 passed; `test_moe_dispatch.py` skipped for a local missing-torch reason and statically checked instead).

The direction and execution quality here are high: the exact-or-SOL GDN lookup kills a real silicon-mislabeling bug, the Rust/Python parity is pinned with a Python oracle test, the citation density matches the collector "framework truth" rule, and the case work follows the declare-never-filter pattern exactly (fail-loud validation, physical-tuple dedup, correct layer ownership). The composition follows the `SGLangEPMOEModel` precedent faithfully (verified: it composes only the pre op for DeepEP, `models/moe.py:544/636`), and `scale_num_tokens` is only consumed by the sglang DeepEP branch (`moe.py:1257`), so the context `scale_num_tokens=tp` is safe for the other backends.

Two blocking items (inline), one non-blocking modeling note (inline), plus:

**PR description is stale.** The body says `ENGINE_SPEC_SCHEMA_VERSION` 6 -> 7, but the code goes 6 -> **8** — the `below_grid_sol` bump (and the whole below-grid-SOL feature) isn't mentioned in the description at all. Please update it; the description is what future archaeology reads.

**Governance note for maintainers.** This PR is a declared coordinated producer+consumer change (GDN row semantics switch to TP-local geometry, with the SDK/Rust consumer in the same PR) and it extends the case-generator mechanism (`model_case_values.gemm` expansion, GDN TP expansion). Per `.claude/rules/collector/layer_permissions.md`, both classes of change require explicit human approval — the PR declares them properly, so approval should be an explicit act of this review rather than a default. The implementations themselves conform to `case_authoring.md`.

**Minor:**
- Below-grid SOL results return `energy=0.0` — negligible for these tiny ops, but it is an undercount in the energy ledger; worth a one-line comment.
- The TRT-LLM path keeps the legacy attention-AR double-count (`attn_ar_modeled=True` is passed but the trtllm branch ignores it). The PR declares TRT-LLM not validated — please also state this known error in the model docstring so it isn't later mistaken for fixed.

aic-core/src/aiconfigurator_core/sdk/models/qwen35.py
wenpengw-nv: Agreed on reachability and bias direction. Left a TODO at the composition site stating both

aic-core/src/aiconfigurator_core/sdk/models/qwen35.py
wenpengw-nv: Fixed in 5a1834df, option 1: Qwen35Model.__init__ now rejects moe_backend='megamoe'  (same pattern as the cp guard).

collector/vllm/collect_gdn.py
wenpengw-nv: The data PR will re-collect GDN for the affected systems and replace the existing gdn_perf.parquet wholesale (delete-then-write, not append). Merge order: this PR first, the data PR right after. In the interim window Qwen3.5 SILICON tp=1 on vLLM keeps serving the old conv rows and the support matrix will not flag it (the rows stay structurally valid) — the window is declared as known-miscalibrated in the PR description.

wenpengw-nv: Update: the re-collection is done and **folded into this PR** (`d69affee`), so both asks dissolve — the hard delete happens here (wholesale delete-then-write of `gdn_perf.parquet`, 8 systems × {vLLM 0.24.0, SGLang 0.5.14},  rows each with the full TP-local grids), and there is no interim window for the support matrix to absorb.

Arsene12358 (approved):
## Re-review: all findings resolved — APPROVE

Following up on my 8/7 CHANGES_REQUESTED — the fix wave + data refresh resolve everything, and I re-verified each item empirically.

**H1 (SILICON hard-failure) — fixed as specified and verified on both previously-failing paths.** `below_grid_sol` is exactly the scoped opt-in suggested (SILICON-only degradation, exact rows win, quant-mode misses stay strict, HYBRID untouched), with the Rust mirror, schema v8 on both sides, and the public-api pin. Empirical: the support-matrix commands that returned **zero configs** on the original branch now complete — gb200/sglang/0.5.14 SILICON: 2,954 agg + 96 disagg results; b200/trtllm/1.3.0rc20 SILICON (exercises the below-grid SOL path, since trtllm keeps the scalar-gate GEMM): 2,242 agg + 87 disagg.

**I1 (dt_bias citation) — added verbatim.**

**The v3 data refresh is the strongest part.** TP-local GDN grids for 8 systems × both collector pins with full provenance (collector ref/hash, case-plan hash, image digest, honest failure notes in collection_meta). Independent silicon validation from our side: against InferenceX Qwen3.5-397B-FP8 disagg baselines on gb300 (15 points), the decode-step (TPOT) prediction error goes **main +80% → modeling-only +35% (GDN at SOL) → modeling+your data +40%** — the ~+4pp giveback when real TP-local rows replaced the SOL floor matches the physically expected direction and magnitude, i.e. the new rows land where they should. The remaining ~40% is attributed outside this PR's scope (step-composition bias across silicon-backed families + attention kernel-source selection — analysis follow-ups on our side).

Also reviewed in the wave: own-lane precedence (physical kernel lanes over cross-backend donor rows — sensible, exact-or-SOL preserved), the sglang fused shared-gate refinement, serial dispatch around the decode overlap, the DP-attention AR fold, and the megamoe rejection with documented bounds — all with test updates. Unit suite: 3,035 passed (one local failure was my stale rust-extension build; `uv sync --reinstall-package aiconfigurator-core` resolves it — noting for other reviewers).

Two carried notes, no action needed here: (1) schema is now v8 — the renumber-at-merge coordination with the in-flight large-EP cascade matters even more; (2) the gdn state-dtype axis question (serving configs override `mamba_ssm_dtype` to bf16 in the wild while collectors pin the checkpoint's fp32) remains a follow-up worth an owner decision.

Nice turnaround — this is mergeable from my side.

collector/vllm/collect_gdn.py
tianhaox: Verified at d69affee: `gdn_perf.parquet` is a wholesale replacement across all 8 systems × {vLLM 0.24.0, SGLang 0.5.14} — spot-checked b200_sxm/vllm: 9100 rows (== collection_meta), all 18 TP-local (nk, nv) geometries present, and only physical kernel lanes (`chunk_gated_delta_rule_flashinfer`, `fused_recurrent_gated_delta_rule_packed_decode`, conv ops) — no legacy logical rows from the old timing domain survive. The provenance block (image digest, collector/case-plan hashes, per-table error notes) is a nice upgrade too. With the data folded in, there is no interim window; this ask is fully resolved.

One non-blocking nit: `collector_ref: unknown` in the new metas — worth stamping the collector git ref in future campaigns. Also reviewed the own-physical-lane precedence flip this required (mamba.py / state_space.rs): symmetric in both engines, ambiguity still fails closed even when a logical row exists, and the donor-fill pins were consciously re-pinned. Resolving.

aic-core/src/aiconfigurator_core/sdk/models/qwen35.py
tianhaox: Verified at 5a1834df: `__init__` rejects `moe_backend='megamoe'` with the DeepSeek-V4-only rationale, matching the cp guard pattern, and `test_qwen35_rejects_megamoe_backend` pins it. Resolving.

tianhaox (approved):

PR #1505, opened 2026-08-08
fix(recommend): scale TP/EP/DP candidates when escalating GPU budget

natoscott:
## Overview

`cli_recommend` has an escalation loop that doubles the GPU budget when a model doesn't fit in a single node. The loop correctly scaled `total_gpus` and `agg_num_gpu_candidates`, but left `agg_tp_candidates`, `agg_moe_tp_candidates`, `agg_moe_ep_candidates`, and `agg_dp_candidates` frozen at the values set by `build_default_tasks` — always capped at `[1, 2, 4, 8]` regardless of budget.

As a result, large MoE models that require TP/EP > 8 to fit in GPU memory (e.g. Kimi-K3 on H200 SXM, which needs moe_ep ≥ 32) raised `InsufficientMemoryError` at every budget level up to 64 GPUs, even though configs at the larger parallelism sizes would have fit and been valid.

## Details

**Root cause** — `_build_recommend_tasks` in `src/aiconfigurator/cli/api.py`:

```python
# Before: only these two were updated on escalation
"total_gpus": total_gpus,
"agg_num_gpu_candidates": num_gpu_candidates,
# TP/EP/DP candidates were never touched → always [1, 2, 4, 8]
```

**Fix** — add a local `_extend` helper that appends missing powers-of-two up to the new budget, leaving dimensions already pinned to `[1]` (non-MoE `moe_ep`/`moe_tp`) unchanged. Applied to all parallelism dimensions for both agg and disagg workers.

**Dimensions scaled:** `agg_tp_candidates`, `agg_dp_candidates`, `agg_moe_tp_candidates`, `agg_moe_ep_candidates` (and disagg prefill/decode equivalents).

**Dimensions intentionally not scaled:** `agg_pp_candidates` (PP is not used in recommend mode), `agg_cp_candidates` (CP is for long-context within a request, not model fitting).

**Non-MoE models** are unaffected: their `moe_ep`/`moe_tp` candidates are `[1]` and `_extend` leaves them untouched.

## Test plan

- [x] Unit tests: `TestBuildRecommendTasksScaling` covers base budget, MoE scaling, non-MoE pinned dims, intermediate budgets, disagg scaling, and DP pinned-at-one behaviour
- [x] Existing `TestCLIRecommendUnit` fixtures updated with new required fields; all pass
- [x] New e2e test: `test_recommend_multi_node_moe_returns_results` — asserts Kimi-K3 on H200 returns at least one config with >8 GPUs (regression guard for the escalation path)
- [x] New e2e test: `test_recommend_single_node_dense_model` — baseline guard that small dense models still resolve within one node
- [x] Manually verified: `cli_recommend("moonshotai/Kimi-K3", "h200_sxm", backend="vllm", target_concurrency=16)` now returns tp=16/moe_tp=16 at 16 GPUs; was `NoFeasibleConfigError` before

## Summary by CodeRabbit

* **Improvements**
  * Recommendations now recalculate parallelism options for the requested GPU budget.
  * Improved support for larger multi-node deployments and disaggregated workloads.
  * Recommendations continue to respect backend, topology, and hardware constraints.

* **Bug Fixes**
  * Improved latency predictions for supported multi-node model configurations.
  * Ensured smaller models remain within a single-node GPU budget.

* **Tests**
  * Added coverage for GPU scaling, candidate recalculation, and multi-node recommendations.

### DIFF
--- a/src/aiconfigurator/cli/api.py
+++ b/src/aiconfigurator/cli/api.py
@@ -375,23 +375,22 @@ _MAX_GPUS_PER_WORKER = 64
 
 
 def _build_recommend_tasks(base_tasks: dict, total_gpus: int) -> dict:
-    """Scale GPU candidates for recommend mode."""
-    import dataclasses
+    """Rebuild recommend tasks at a new GPU budget.
 
-    num_gpu_candidates = [1 << i for i in range(total_gpus.bit_length()) if (1 << i) <= total_gpus]
+    Resets all candidate fields (those ending in ``_candidates`` whose
+    dataclass default is None) to None, then replaces ``total_gpus``. This
+    causes Task's ``__post_init__`` → ``_resolve_agg_search`` /
+    ``_resolve_disagg_search`` to recompute every parallelism dimension at the
+    new budget, preserving all backend-specific and topology-specific caps
+    (e.g. deepep_moe intra-node EP limits, non-MoE pinned-to-1 dims)
+    without any manual extension logic.
+    """
+    import dataclasses
 
     rebuilt = {}
     for name, task in base_tasks.items():
-        overrides = {
-            "total_gpus": total_gpus,
-            "agg_num_gpu_candidates": num_gpu_candidates,
-        }
-        if task.serving_mode == "disagg":
-            overrides.update(
-                prefill_num_gpu_candidates=num_gpu_candidates,
-                decode_num_gpu_candidates=num_gpu_candidates,
-            )
-        rebuilt[name] = dataclasses.replace(task, **overrides)
+        reset = {f.name: None for f in dataclasses.fields(task) if f.name.endswith("_candidates") and f.default is None}
+        rebuilt[name] = dataclasses.replace(task, total_gpus=total_gpus, **reset)
     return rebuilt
 
 
--- a/src/aiconfigurator/cli/api.py
+++ b/src/aiconfigurator/cli/api.py
@@ -385,23 +385,22 @@ _DSPARK_DEFAULT_ACCEPTANCE = 0.8
 
 
 def _build_recommend_tasks(base_tasks: dict, total_gpus: int) -> dict:
-    """Scale GPU candidates for recommend mode."""
-    import dataclasses
+    """Rebuild recommend tasks at a new GPU budget.
 
-    num_gpu_candidates = [1 << i for i in range(total_gpus.bit_length()) if (1 << i) <= total_gpus]
+    Resets all candidate fields (those ending in ``_candidates`` whose
+    dataclass default is None) to None, then replaces ``total_gpus``. This
+    causes Task's ``__post_init__`` → ``_resolve_agg_search`` /
+    ``_resolve_disagg_search`` to recompute every parallelism dimension at the
+    new budget, preserving all backend-specific and topology-specific caps
+    (e.g. deepep_moe intra-node EP limits, non-MoE pinned-to-1 dims)
+    without any manual extension logic.
+    """
+    import dataclasses
 
     rebuilt = {}
     for name, task in base_tasks.items():
-        overrides = {
-            "total_gpus": total_gpus,
-            "agg_num_gpu_candidates": num_gpu_candidates,
-        }
-        if task.serving_mode == "disagg":
-            overrides.update(
-                prefill_num_gpu_candidates=num_gpu_candidates,
-                decode_num_gpu_candidates=num_gpu_candidates,
-            )
-        rebuilt[name] = dataclasses.replace(task, **overrides)
+        reset = {f.name: None for f in dataclasses.fields(task) if f.name.endswith("_candidates") and f.default is None}
+        rebuilt[name] = dataclasses.replace(task, total_gpus=total_gpus, **reset)
     return rebuilt
 
 
--- a/src/aiconfigurator/cli/api.py
+++ b/src/aiconfigurator/cli/api.py
@@ -44,7 +44,7 @@ from aiconfigurator.sdk.speculative import (
 from aiconfigurator.sdk.speculative import (
     normalize_speculative_decoding as _normalize_nextn,
 )
-from aiconfigurator.sdk.task_v2 import Task
+from aiconfigurator.sdk.task_v2 import TASK_REBUILD_FIELDS, Task
 
 # Default per-phase latency-correction scales for single-point disagg estimates.
 # (Migrated from the legacy V1 task module; same values as Task's field defaults.)
@@ -387,19 +387,18 @@ _DSPARK_DEFAULT_ACCEPTANCE = 0.8
 def _build_recommend_tasks(base_tasks: dict, total_gpus: int) -> dict:
     """Rebuild recommend tasks at a new GPU budget.
 
-    Resets all candidate fields (those ending in ``_candidates`` whose
-    dataclass default is None) to None, then replaces ``total_gpus``. This
-    causes Task's ``__post_init__`` → ``_resolve_agg_search`` /
-    ``_resolve_disagg_search`` to recompute every parallelism dimension at the
-    new budget, preserving all backend-specific and topology-specific caps
-    (e.g. deepep_moe intra-node EP limits, non-MoE pinned-to-1 dims)
-    without any manual extension logic.
+    Resets every field listed in ``TASK_REBUILD_FIELDS`` to None, then
+    replaces ``total_gpus``.  This causes Task's ``__post_init__`` →
+    ``_resolve_agg_search`` / ``_resolve_disagg_search`` to recompute every
+    parallelism dimension at the new budget, preserving all backend-specific
+    and topology-specific caps (e.g. deepep_moe intra-node EP limits,
+    non-MoE pinned-to-1 dims) without any manual extension logic.
     """
     import dataclasses
 
     rebuilt = {}
     for name, task in base_tasks.items():
-        reset = {f.name: None for f in dataclasses.fields(task) if f.name.endswith("_candidates") and f.default is None}
+        reset = {f.name: None for f in dataclasses.fields(task) if f.name in TASK_REBUILD_FIELDS}
         rebuilt[name] = dataclasses.replace(task, total_gpus=total_gpus, **reset)
     return rebuilt
 
@@ -565,15 +564,36 @@ def cli_recommend(
         moe_backend=moe_backend,
     )
 
-    # Build escalation sequence: gpus_per_node, *2, *4, *8, capped at _MAX_GPUS_PER_WORKER.
-    budgets = [gpus_per_node]
+    # Build escalation sequence: *2, *4, *8 from gpus_per_node, capped at _MAX_GPUS_PER_WORKER.
+    # Attempt 0 uses base_tasks directly (already built at gpus_per_node — no rebuild needed).
+    escalation_budgets = []
     budget = gpus_per_node * 2
     while budget <= _MAX_GPUS_PER_WORKER:
-        budgets.append(budget)
+        escalation_budgets.append(budget)
         budget *= 2
 
-    result = None
-    for attempt, gpu_budget in enumerate(budgets):
+    result = _execute_and_wrap_result(
+        base_tasks,
+        mode="default",
+        top_n=top_n,
+        strict_sla=strict_sla,
+        target_request_rate=target_request_rate,
+        target_concurrency=target_concurrency,
+        parallel_experiments=True,
+    )
+    retriable = [
+        name
+        for name, outcome in result.outcomes.items()
+        if outcome.error is not None and is_gpu_retriable(outcome.error)
+    ]
+    for attempt, gpu_budget in enumerate(escalation_budgets):
+        if not retriable:
+            break
+        logger.info(
+            "Recommend: %s may fit at larger budget; escalating to %d GPUs.",
+            ", ".join(retriable),
+            gpu_budget,
+        )
         tasks = _build_recommend_tasks(base_tasks, gpu_budget)
         result = _execute_and_wrap_result(
             tasks,
@@ -589,13 +609,6 @@ def cli_recommend(
             for name, outcome in result.outcomes.items()
             if outcome.error is not None and is_gpu_retriable(outcome.error)
         ]
-        if not retriable or attempt == len(budgets) - 1:
-            break
-        logger.info(
-            "Recommend: %s may fit at larger budget; escalating to %d.",
-            ", ".join(retriable),
-            budgets[attempt + 1],
-        )
 
     if not result.best_configs:
         raise NoFeasibleConfigError("No feasible GPU configuration found for the given load target.")
--- a/tests/e2e/cli/test_cli_recommend.py
+++ b/tests/e2e/cli/test_cli_recommend.py
@@ -14,12 +14,11 @@ pytestmark = [pytest.mark.e2e, pytest.mark.build]
 
 
 def test_recommend_multi_node_moe_returns_results():
-    """cli_recommend escalates to multi-node for Kimi-K3 on B200.
+    """cli_recommend returns a multi-node config for Kimi-K3 on B200.
 
     Kimi-K3 (large MoE) does not fit in a single B200 node (8x 192 GB).
-    The recommend escalation must scale TP/EP candidates beyond the single-node
-    limit and return at least one feasible configuration.  Uses real silicon
-    data (no HuggingFace credentials required).
+    cli_recommend must return at least one feasible configuration with
+    num_total_gpus > 8.  Uses real silicon data.
     """
     result = cli_recommend(
         model_path="moonshotai/Kimi-K3",
--- a/tests/e2e/cli/test_cli_recommend.py
+++ b/tests/e2e/cli/test_cli_recommend.py
@@ -1,0 +1,65 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+End-to-end tests for cli_recommend, covering the GPU-budget escalation path
+needed for large models that exceed a single node's memory.
+"""
+
+import pytest
+
+from aiconfigurator.cli.api import cli_recommend
+
+pytestmark = pytest.mark.e2e
+
+
+def test_recommend_multi_node_moe_returns_results():
+    """cli_recommend finds valid configs for DeepSeek-V3 on H200.
+
+    DeepSeek-V3 (671B MoE) does not fit in a single H200 node (8x 80 GB).
+    The recommend escalation must scale TP/EP candidates beyond the single-node
+    limit and return at least one feasible configuration.  Uses a hermetic local
+    config (no HuggingFace credentials required).
+    """
+    result = cli_recommend(
+        model_path="deepseek-ai/DeepSeek-V3",
+        system="h200_sxm",
+        backend="vllm",
+        isl=4000,
+        osl=1000,
+        target_concurrency=16,
+        database_mode="HYBRID",
+    )
+
+    assert result.chosen_exp is not None
+    best = result.best_configs.get(result.chosen_exp)
+    assert best is not None and not best.empty, "Expected at least one recommended config"
+
+    top = best.iloc[0]
+    # Model requires more than one H200 node (8 GPUs) to fit in memory.
+    assert top["num_total_gpus"] > 8, f"Expected multi-node config (>8 GPUs), got {top['num_total_gpus']}"
+    # At least one parallelism dimension must exceed single-node capacity,
+    # confirming TP/EP candidates were actually scaled during escalation.
+    assert max(top["tp"], top["moe_tp"], top["moe_ep"]) > 8
+    # Sanity: predicted latencies are positive and finite.
+    assert top["ttft"] > 0
+    assert top["tpot"] > 0
+
+
+def test_recommend_single_node_dense_model():
+    """cli_recommend finds configs for a small dense model within one node."""
+    result = cli_recommend(
+        model_path="meta-llama/Llama-3.1-8B-Instruct",
+        system="h200_sxm",
+        backend="vllm",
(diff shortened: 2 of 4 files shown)


tianhaox (changes requested):
Thanks for this — the root-cause diagnosis is correct and I verified the supporting pieces: the candidate lists in `_resolve_agg_search` indeed never depended on `total_gpus`, so the escalation loop was only ever enlarging `*_num_gpu_candidates`; and multi-node comm *is* modeled (the NCCL empirical path clamps utilization to the largest collected rank count and takes SOL at the real rank count with inter-node bandwidth via `SystemSpec.get_p2p_bandwidth`), so extending TP/EP beyond a node is directionally sound. The vLLM Kimi-K3 case this targets is a real fix.

That said, one design issue makes me want a change before merge.

## Blocking: `_extend` cannot distinguish budget caps from capability caps

The `max(candidates) <= 1` gate only protects dimensions pinned to `[1]`. But several default lists have caps **> 1 that encode backend/topology capability**, not budget, and `_extend` erases them:

1. **sglang `deepep_moe` (non-wideep)** — `task_v2.py` sets `agg_moe_ep_candidates=[1,2,4,8]` with the comment *"Intra-node DeepEP (ep 1-8, NVLink): EP-only"*. `_extend` pushes this to 64. Cross-node EP is the wideep/internode kernel path; the intra-node DeepEP perf data and deployment config don't cover ep > 8.
2. **sglang `megamoe`** — `_sglang_megamoe_parallel_lists` is explicitly rack-NVL aware: on systems without rack NVL (e.g. H200) `ep_list=[8]`, i.e. exactly one full node. Since max=8 > 1, `_extend` grows it to `[8,16,32,64]`, silently deleting the `has_rack_nvl` gate.
3. **trtllm wideep** — tp is deliberately capped at `[1,2,4,8]` while dp/moe_ep go to 64 ("attention stays intra-node, scale-out goes through EP"). `_extend` pushes tp to 64 as well.

All three paths are reachable from `cli_recommend` (`backend`/`moe_backend` are public parameters), and large MoE models are exactly the ones that trigger escalation. `enumerate_parallel_config`'s backend filters don't catch this (they filter moe_tp>1 combos etc., not EP magnitude), so the sweep would price cross-node configs with intra-node kernel data and can surface non-deployable configs as recommendations.

Two ways to fix, either is fine by me:

- **Conservative:** restrict `_extend` to the paths where extension is semantically safe — the plain fused paths (vllm, trtllm non-wideep) that this PR actually validated — and leave sglang `deepep_moe`/`megamoe` lists and the wideep tp dimension untouched.
- **Cleaner but crosses into the SDK:** thread the escalated budget into `build_default_tasks` / `_resolve_agg_search` so the owner of candidate-list policy derives the lists per budget, instead of the CLI layer second-guessing which lists are extendable. That needs maintainer buy-in on the API change, so I'd understand deferring it.

## Non-blocking

- **e2e test rides the HYBRID fallback, not real data** — `test_recommend_multi_node_moe_returns_results` uses Kimi-K3 + vllm on `h200_sxm`, but Hopper has no vllm KDA/K3 op data (`linear_attention/vllm` there only contains GDN/Qwen; K3 vllm data was collected on Blackwell). The test passes via the HYBRID modeling fallback, which makes it sensitive to unrelated fallback-behavior changes, and `ttft > 0` doesn't validate realism. Suggest `b200_sxm` instead: real K3 vllm data, same 8 GPUs/node, and the model still needs >1 node — strictly better as a regression guard.
- **Unit tests assert on implementation detail** — `_run` patches `dataclasses.replace` and inspects kwargs, so all eight tests break if `_build_recommend_tasks` ever stops using `replace`, and the `MagicMock` task won't catch field renames. Acceptable, just noting the coupling.
- **FakeTask duplication** — five identical copies each gained the same four lines; a shared fixture would help. Pre-existing debt, not this PR's fault.
- Dense-model tp also gets extended (e.g. to 64 on Hopper); cross-node dense TP will basically never win on SOL inter-node comm, so it's harmless but adds sweep time. Could exclude non-MoE extreme values in `_extend` if you touch it anyway.
- `test_pp_and_cp_candidates_not_scaled` documenting the pp/cp exclusion as an assertion is a nice touch.

src/aiconfigurator/cli/api.py
```diff
@@ -354,16 +354,37 @@ def _build_recommend_tasks(base_tasks: dict, total_gpus: int) -> dict:
...
+
     rebuilt = {}
     for name, task in base_tasks.items():
         overrides = {
             "total_gpus": total_gpus,
             "agg_num_gpu_candidates": num_gpu_candidates,
+            "agg_tp_candidates": _extend(task.agg_tp_candidates),
+            "agg_dp_candidates": _extend(task.agg_dp_candidates),
+            "agg_moe_tp_candidates": _extend(task.agg_moe_tp_candidates),
+            "agg_moe_ep_candidates": _extend(task.agg_moe_ep_candidates),
```
jasonqinzhou: **P1 — Preserve backend-specific topology caps during escalation**

`_extend` treats every multi-valued candidate list as a lower bound, but these lists also encode supported topology caps. For example, `Task._resolve_agg_search` defines SGLang `moe_backend="deepep_moe"` with `enable_wideep=False` as intra-node DeepEP and caps EP at `[1, 2, 4, 8]`. Rebuilding that real task at budget 64 changes the list to `[1, 2, 4, 8, 16, 32, 64]`, and `iter_parallel()` accepts EP 16/32/64. The same helper also extends role-specific WideEP prefill caps. As a result, `cli_recommend` can make an unsupported cross-node topology look feasible and return it. Please derive expansion from backend/role capability (or at minimum preserve the intra-node DeepEP and other explicit caps), and cover the real `Task` paths rather than only uniform mocks.

tests/e2e/cli/test_cli_recommend.py
```diff
@@ -0,0 +1,64 @@
...
+    assert max(top["tp"], top["moe_tp"], top["moe_ep"]) > 8
+    # Sanity: predicted latencies are positive and finite.
+    assert top["ttft"] > 0
+    assert top["tpot"] > 0
+
+
+def test_recommend_single_node_dense_model():
+    """cli_recommend finds configs for a small dense model within one node."""
+    result = cli_recommend(
+        model_path="meta-llama/Llama-3.1-8B-Instruct",
```
jasonqinzhou, line 52: **P1 — Keep this E2E independent of gated Hugging Face access**

This identifier has no packaged config and is gated. Running the new test at this PR head without Hugging Face credentials fails before recommendation with `HuggingFaceDownloadError: HTTP Error 401`, while the repository already ships a suitable config for `meta-llama/Meta-Llama-3.1-8B`. Please use that hermetic identifier (or a local fixture) so contributors and CI do not need a private token to execute the new test.

```suggestion
        model_path="meta-llama/Meta-Llama-3.1-8B",
```

jasonqinzhou (changes requested):

## What this PR does

This PR expands recommend-mode TP, DP, and MoE parallelism candidates as the GPU budget escalates for aggregate and disaggregated tasks, with unit and end-to-end regression coverage.

The regression is well isolated, and the tests thoughtfully cover intermediate budgets, pinned non-MoE dimensions, both disaggregated worker roles, and the user-visible Kimi-K3 failure path.

## Why this score

The score and REQUEST_CHANGES recommendation reflect two confirmed blockers: generic candidate expansion discards backend- and role-specific topology caps and can return unsupported configurations, while the new dense end-to-end test requires gated Hugging Face access and fails in an uncredentialed environment.

src/aiconfigurator/cli/api.py
natoscott: Done.

tests/e2e/cli/test_cli_recommend.py
natoscott: Fixed.

tests/unit/cli/test_cli_api.py
```diff
@@ -780,3 +780,63 @@ def test_disagg_estimate_honors_explicit_free_gpu_memory_fraction():
...
+    def test_rebuilds_all_tasks(self):
+        """Every task in the dict is rebuilt at the new budget."""
+        from aiconfigurator.cli.api import _build_recommend_tasks
+
+        task = self._make_real_task(total_gpus=8)
+        tasks = {"agg": task, "disagg": task}
+        result = _build_recommend_tasks(tasks, 32)
+        assert set(result.keys()) == {"agg", "disagg"}
+        assert result["agg"].total_gpus == 32
+        assert result["disagg"].total_gpus == 32
```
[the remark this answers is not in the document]
natoscott: Done.

natoscott:
@jasonqinzhou @tianhaox PTAL, I think the issues are all resolved here now.

tianhaox:
**Review summary** — The rewrite direction (delegating candidate recomputation to `Task.__post_init__` instead of manually extending GPU lists) is sound, and the `f.default is None` guard correctly excludes `afd_max_candidates` / `afd_prefill_max_candidates` today. However, escalation is still defeated for disagg by a non-candidate resolved field, one backend path regresses versus the old explicit GPU list, and the PR's own regression tests are largely inert (deselected in CI, dead fake fields, agg-only column assumptions). Findings below, most severe first.

---

### Correctness

**1. `max_gpu_per_replica` survives the rebuild, so disagg escalation is still defeated** — `src/aiconfigurator/cli/api.py:392`

The reset only clears `*_candidates` fields, but `Task.__post_init__` also clamps `max_gpu_per_replica` in place to `min(total_gpus, ...)` (task_v2.py:1599-1600). After the base task is built at the single-node budget, that field holds a concrete node-sized value (e.g. 8), and `dataclasses.replace` carries it into every escalated rebuild — the `if is None` re-default is skipped, `min(64, 8) = 8`, and `_replica_num_gpu_list` (task_v2.py:2480-2488) filters out every multi-node replica.

Failure scenario: `cli_recommend` on an 8-GPU/node system for a model whose smallest feasible disagg replica needs >8 GPUs (e.g. DeepSeek-V3) — every escalation attempt (16/32/64) repeats the identical infeasible disagg search before `NoFeasibleConfigError`, contradicting the new docstring's "recompute every parallelism dimension at the new budget". `max_gpu_per_replica` (or any other `__post_init__`-resolved non-candidate field) needs to be reset too.

*(Pre-existing: the old code carried this field through `dataclasses.replace` the same way, so the bug predates this PR — flagging it here because the PR's stated goal is to rebuild tasks correctly at the new budget, and this field still defeats that for disagg.)*

**2. Possible regression vs the deleted explicit `num_gpu` list for sglang MoE without large-EP perf coverage** — `src/aiconfigurator/cli/api.py:393`

The old code set `agg/prefill/decode_num_gpu_candidates = [1,2,...,64]` while preserving tp/dp lists, letting `enumerate_parallel_config` reach e.g. 32-GPU fused workers on sglang's fused path. The new code re-resolves from the fused ladder `x = [1,2,4,8,16]` (task_v2.py:1645/1657), and the >16 wide ladder is gated on `_large_ep_eps` coverage (task_v2.py:1698). With no coverage rows for the model shape, budgets 16/32/64 re-run an identical ≤16-GPU search and `cli_recommend` raises `NoFeasibleConfigError` where it previously returned a config.

### Efficiency

**3. Attempt 0 rebuilds tasks at the budget they were just built with, and dense-model escalation re-runs identical sweeps** — `src/aiconfigurator/cli/api.py:556`

`budgets[0] == gpus_per_node ==` the `total_gpus` the base tasks were built with, yet `_build_recommend_tasks` re-runs full `__post_init__` resolution per task for an identical result (attempt 0 could just use `base_tasks`). Also, dense ladders are fixed at `[1,2,4,8]` independent of `total_gpus` (task_v2.py:1620-1628), so a dense model that OOMs at 8 GPUs triggers 3 extra full agg+disagg sweeps at 16/32/64 with identical candidates — minutes of wasted CPU and misleading "escalating to N" logs before `NoFeasibleConfigError`.

*(Partially pre-existing: the escalation loop already rebuilt at attempt 0 before this PR; what's new is that each rebuild now runs full `__post_init__` resolution, so the waste is more expensive.)*

### Design (suggestion)

**4. The `endswith('_candidates') and default is None` heuristic encodes Task-internal knowledge at the wrong layer** — `src/aiconfigurator/cli/api.py:392`

task_v2.py:1526-1533 already enumerates `f'{role}_{dim}_candidates'` inside `_resolve_search_space`. Fragilities: (a) forwarding `encoder_tp` through `cli_recommend` (as `cli_default` already does) would make escalation silently wipe a user's encoder-TP pin; (b) changing `afd_max_candidates` to None-means-unlimited would null an int config field; (c) a future `*_candidates` field with a `default_factory` silently escapes the reset, reintroducing the stale-candidate bug. A Task-side `rebuilt_at(total_gpus)` method or an explicit `CANDIDATE_FIELDS` constant next to `_resolve_search_space` would fail loudly instead.

### Tests

**5. The only end-to-end escalation guard never runs in CI, and times out locally** — `tests/e2e/cli/test_cli_recommend.py:13`

CI runs pytest with `-m "unit or build"` for all matrix suites including `tests/e2e` (build-test.yml:174-179), but this file carries only `pytestmark = pytest.mark.e2e` (repo convention adds `pytest.mark.build` for CI-run e2e tests, cf. test_cli_build_default.py:9) — both tests are always deselected, so a future re-break of candidate scaling ships with green CI. Meanwhile pytest.ini has `testpaths = tests` and `timeout = 120`, so a plain local `pytest` collects the multi-minute DeepSeek-V3 8→16→32→64 sweep and fails on timeout.

**6. e2e assertion reads agg-only columns; KeyError whenever disagg wins** — `tests/e2e/cli/test_cli_recommend.py:43`

The assertion reads `tp/moe_tp/moe_ep` from `best_configs[chosen_exp]`, but `cli_recommend` also sweeps disagg whose rows use `(p)tp/(d)tp/(p)moe_ep` — no plain `tp` column. The test currently passes only because disagg is infeasible for DeepSeek-V3, which is an accident of finding #1; fixing #1 (or any perf-DB update that makes disagg feasible) breaks this test.

**7. Test model ID triggers a live fetch of a gated HF repo** — `tests/e2e/cli/test_cli_recommend.py:52`

`meta-llama/Llama-3.1-8B-Instruct` is not in the bundled model-config cache (the bundled ID is `meta-llama/Meta-Llama-3.1-8B` — note `Meta-` prefix, no `-Instruct`), so the test falls through to `_download_hf_config`, contradicting the module's "no HuggingFace credentials required" claim. Offline CI or any machine without an accepted-license HF token gets a download error. Suggest using the bundled ID.

**8. The ~28 candidate fields added to the FakeTask fakes are dead code — the escalation unit tests no longer exercise candidate reset at all** — `tests/unit/cli/test_cli_api.py:560`

The added fields use `field(default_factory=...)`, so `f.default is dataclasses.MISSING` (not `None`) — they match neither the new reset predicate nor any other read, and FakeTask has no `__post_init__`. `reset == {}` for every FakeTask; only `total_gpus` is replaced. If the reset predicate regresses (suffix typo, or a real Task candidate field gains a non-None default), all four escalation unit tests stay green while escalation silently reuses stale candidates — the exact bug this PR fixes. Either drop the unused fields, or make the fakes' candidate fields default to `None` and assert they get reset.

**9. Five copy-pasted nested FakeTask dataclasses updated in lockstep** — `tests/unit/cli/test_cli_api.py:553`

FakeTask is defined at lines 553, 594, 632, 675, 765 and this PR applies the same field additions to each. One module-level FakeTask with defaults covers all five call sites; as is, the next Task shape change requires five synchronized edits and a missed copy silently diverges.

**10. (minor) `TestBuildRecommendTasksScaling` uses real perf-DB-backed Task constructions for trivial assertions, and `test_rebuilds_all_tasks` puts an agg-mode task under the `'disagg'` key** — `tests/unit/cli/test_cli_api.py:892`

`test_budget_is_forwarded` rebuilds real DeepSeek-V3 Tasks at 3 budgets just to assert `total_gpus` forwarding, and `{'agg': task, 'disagg': task}` reuses the same agg task for both keys, testing only dict iteration. A cheap None-default fake proves the same forwarding/iteration; keep the real Tasks for the two sentinel recomputation tests.

---

Blocking before merge, in my view: **#1**, **#5**, and **#8** — together they mean the fix is ineffective for disagg and no test in the PR would ever detect that.

natoscott:
@tianhaox thanks for extensive review!  All items listed have been addressed now.

tianhaox:
**Round-3 verification** (head 61d124f85) — I re-verified every item from my previous review empirically: rebuilt the Rust extension at the PR head, ran the suites, and cross-checked behavior against `main` (805a14ed5). **All 10 items are genuinely resolved.** One new must-fix surfaced in the SDK extension code, plus a few notes. Details below.

## Previous items — all confirmed fixed

1. **`max_gpu_per_replica` reset** ✔ — `TASK_REBUILD_FIELDS` covers the four non-candidate resolved fields; measured the escalated disagg task at budgets 16/32/64 → cap now 16/32/64, and the rebuilt task matches a fresh `Task(total_gpus=B)` field-by-field (sole exception noted below).
2. **sglang fused regression** ✔ — fused `num_gpu` extension restores 32/64-wide fused workers on sglang.
3. **Attempt-0 rebuild** ✔ — base attempt uses `base_tasks` directly, and dense ladders now extend with the budget.
4. **Reset predicate at the wrong layer** ✔ — the explicit `TASK_REBUILD_FIELDS` constant living next to the resolution code in task_v2 is exactly the right shape.
5. **Tests deselected in CI** ✔ — `build` marker added; I confirmed "Build and Test (e2e)" at this head actually executed the new tests, and they run in ~11s/~9s locally, well under the timeouts.
6. **Disagg column KeyError** ✔ — dual column-name sets via `.get()`.
7. **Gated HF model** ✔ — bundled `Meta-Llama-3.1-8B`; passes with no HF token in the environment.
8. **Dead fake fields** ✔ — and the sentinel tests on real Tasks now discriminate: I ran the new `TestBuildRecommendTasksScaling` against main and 5/6 fail there. That's a real regression guard.
9. **FakeTask duplication** ✔ — consolidated into `_FakeRecommendTask`.
10. **Real-Task trivial assertions** ✔.

## New must-fix

**The disagg `num_gpu` extension overwrites user-supplied candidate lists** — `task_v2.py`, the new block at the end of `_resolve_disagg_search`:

```python
_current = getattr(self, _field)
if _current is not None:
    setattr(self, _field, sorted(set(_current) | _extras))
```

By this point in `__post_init__` the field is always non-None (defaults already filled), so the extension applies to explicit user input too. Measured: `Task(..., total_gpus=64, prefill_num_gpu_candidates=[4, 8])` resolves to `[4, 8, 32, 64]` — a user's exp-YAML pin is silently widened. This violates the user-supplied-wins contract that `_resolve_search_space` itself documents, and it's asymmetric: the agg-side twin (`fused["num_gpu"]` flowing through `_set`) respects pins correctly. Fix is small — fold the extension into the default derivation like the agg side, or gate it on the `defaulted` set that `_resolve_search_space` already computes.

## Non-blocking notes

- **Rebuild marks inferred quant modes as user-explicit.** The rebuilt task carries the base task's *resolved* `fmha/kvcache_quant_mode` values, so `__post_init__` sets `_fmha_explicit=True` and skips `_apply_fmha_data_fallback` — freezing the budget-8 fallback decision even though that decision is search-space-dependent by design (it's why the resolution is split in two). Values are identical today, so this is second-order; a comment on `TASK_REBUILD_FIELDS` acknowledging the residue would do.
- **The multi-node e2e passes on main too** — on b200 the data-gated large-EP ladder already reaches multi-node, so `test_recommend_multi_node_moe_returns_results` is a smoke test of the escalation path rather than a regression guard (the unit tests carry that role now, which is fine). Suggest softening its docstring ("must scale TP/EP candidates beyond the single-node limit") to match.
- **Default-mode behavior change, for the record:** dense models with `total_gpus > 16` now sweep cross-node TP in *every* mode, not just recommend (h200 dense: max width 8 → 32/64, +2..+12 configs per backend). MoE default-mode is measurably unchanged — I enumerated `iter_parallel` on main vs this head across vllm/sglang/trtllm at budgets 8/32/64 and the MoE sets are identical, because the per-dimension caps of 16 plus backend filters block wider fused combos on vllm/trtllm. Which also means the code comment "combining them (e.g. tp=8, dp=4) already reaches 32 GPUs" holds for sglang only; on vllm (`moe_tp>1 && moe_ep>1` blocked) and trtllm (`tp>1 && dp>1` blocked) the fused `num_gpu` extension is currently a no-op. Worth a maintainer ack since it's an SDK-wide change, but the blast radius is small and the dense widening is what makes dense escalation work at all.
- Trivial: the assertion message in `test_max_gpu_per_replica_reset_on_escalation` references `_REBUILD_RESET_FIELDS` (actual name: `TASK_REBUILD_FIELDS`), and the `_FakeRecommendTask` docstring says real Tasks' candidate fields use `default_factory` (they default to `None`).

I also re-checked the PR's motivating case: `cli_recommend("moonshotai/Kimi-K3", "h200_sxm", backend="vllm", ...)` solves at 16 GPUs on this head — and on main as well, since the SDK ladders landed there in the meantime. So this PR's standing value is the policy-to-SDK consolidation plus the disagg replica-cap fix, both of which are real.

Once the disagg pin-overwrite is fixed I'm happy for this to merge.

jasonqinzhou (approved):

natoscott:
@tianhaox does the recent changes look OK to you?  Many thanks!

tianhaox:
Verified at cc689f5b1 (rebuilt the Rust extension for this head since the rebase moved it):

- **Pin overwrite fixed** — `Task(..., total_gpus=64, prefill_num_gpu_candidates=[4, 8])` now resolves to `[4, 8]`; pinning one role leaves the other's defaulted list extending to `[..., 32, 64]` as intended; unpinned lists still extend.
- **Escalation intact** — rebuilt disagg tasks at budget 64 still reach `num_gpu` 64 / replica cap 64, and the rebuilt task matches a fresh `Task(total_gpus=B)` field-for-field (only the previously-noted `_fmha_explicit` residue differs).
- Unit 29/29, e2e 2/2 locally; CI green at this head. Trivial doc/message fixes confirmed.

Thanks for the thorough follow-through — LGTM.

tianhaox (approved):

PR #1506, opened 2026-08-08
feat(recommend): auto-enable DSPARK speculative decoding for Kimi-K3

natoscott:
## Overview

`cli_recommend` (and the `/recommend` REST endpoint) under-estimated Kimi-K3 throughput by ~18x and could over-provision GPUs as a result.

Kimi-K3 always runs with DSPARK speculative decoding in production (a standalone 5-layer draft model proposing `nextn=7` tokens per step). Because the draft model is a separate checkpoint rather than inline MTP layers, `nextn="auto"` correctly returns 0 and speculative decoding was never modeled in the recommend path.

## Details

**Root cause** — `nextn_accepted` has no backend equivalent (it was deliberately removed from the aic-core engine spec in PR #1405). It is purely AIC's throughput planning assumption with no user-facing backend config. Requiring callers to know about DSPARK internals to get accurate sizing estimates is the wrong API contract for a procurement tool.

**Fix** — Add `resolve_dspark_nextn` to `config_builders.py` as a peer of `resolve_nextn_auto`. For DSPARK architectures it returns `(nextn, nextn_accepted)` — the architectural block size (a fixed constant of the draft model, maps to `speculative_config.num_speculative_tokens` in the backend) paired with a conservative throughput planning assumption (0.8 × block_size). Wire it into `cli_recommend` only; `cli_estimate` and `cli_default` retain strict explicit-nextn_accepted behaviour.

**New constant** — `DSPARK_NEXTN: dict[str, int]` added to `aic-core/sdk/common.py` alongside `DSPARK_ARCHITECTURES`. Block sizes are fixed architectural constants of the draft model, not user-configurable.

**Throughput impact for Kimi-K3 on H200:**
| Mode | Predicted tokens/s |
|---|---|
| Before (nextn=0, no DSPARK) | ~33 tok/s |
| After (nextn=7, nextn_accepted=5.6) | ~598 tok/s |

## Test plan

- [x] `TestResolveDsparkNextn` in `tests/unit/sdk/test_speculative.py` — covers non-DSPARK returns None, Kimi-K3 returns correct block size and acceptance, empty path raises, fetch failure returns None
- [x] `TestCLIRecommendUnit::test_dspark_nextn_auto_enabled` — verifies nextn/nextn_accepted are set automatically
- [x] `TestCLIRecommendUnit::test_dspark_explicit_nextn_accepted_not_overridden` — user-supplied nextn_accepted is preserved
- [x] `TestCLIRecommendUnit::test_non_dspark_model_unaffected` — non-DSPARK models unchanged
- [x] `ruff format` + `ruff check` clean on all changed files

## Summary by CodeRabbit

* **New Features**
  * Added automatic DSPARK speculative-decoding configuration for supported models.
  * Recommendation commands now determine an appropriate draft block size automatically by default.
  * Explicit configuration values and opt-out behavior remain supported.
  * Improved recommendation rebuilding to recalculate GPU parallelism when resource limits change.

* **Bug Fixes**
  * Improved handling of unsupported models, missing metadata, and invalid model paths during configuration.

### DIFF
--- a/aic-core/src/aiconfigurator_core/sdk/config_builders.py
+++ b/aic-core/src/aiconfigurator_core/sdk/config_builders.py
@@ -92,6 +92,41 @@ def resolve_nextn_auto(model_path: str) -> int:
     return int(cfg.get("num_nextn_predict_layers") or 0)
 
 
+# Conservative nextn_accepted fraction for DSPARK recommend mode.
+# nextn_accepted has no backend equivalent — it is AIC's throughput planning
+# assumption (average accepted tokens per step as a fraction of the block size).
+# 0.8 is conservative for a well-aligned draft model.
+_DSPARK_DEFAULT_ACCEPTANCE = 0.8
+
+
+def resolve_dspark_nextn(model_path: str) -> tuple[int, float] | None:
+    """Resolve DSPARK speculative parameters for the recommend/sizing path.
+
+    DSPARK architectures use a standalone trained draft model whose block size
+    is a fixed architectural constant — not stored in the main checkpoint, so
+    ``nextn='auto'`` always returns 0 for these models.
+
+    Returns ``(nextn, nextn_accepted)`` when the model is a DSPARK architecture,
+    where ``nextn`` is the architectural block size and ``nextn_accepted`` is a
+    conservative throughput planning assumption (no backend equivalent).
+    Returns ``None`` when the model is not DSPARK or the config cannot be fetched.
+    Raises ``ValueError`` when ``model_path`` is empty, matching ``resolve_nextn_auto``.
+    """
+    from aiconfigurator_core.sdk.common import DSPARK_NEXTN
+    from aiconfigurator_core.sdk.utils import get_model_config_from_model_path
+
+    if not model_path:
+        raise ValueError("resolve_dspark_nextn requires a model path.")
+    try:
+        info = get_model_config_from_model_path(model_path)
+        block_size = DSPARK_NEXTN.get(info.get("architecture", ""), 0)
+        if block_size:
+            return block_size, block_size * _DSPARK_DEFAULT_ACCEPTANCE
+    except Exception:
+        pass
+    return None
+
+
 def apply_nextn(
     model_config: ModelConfig,
     nextn: int | None,
--- a/src/aiconfigurator/cli/api.py
+++ b/src/aiconfigurator/cli/api.py
@@ -26,6 +26,7 @@ from aiconfigurator.cli.report_and_save import save_results
 from aiconfigurator.sdk.config import ModelConfig
 from aiconfigurator.sdk.config_builders import apply_nextn as _apply_nextn
 from aiconfigurator.sdk.config_builders import build_model_config as _build_model_config
+from aiconfigurator.sdk.config_builders import resolve_dspark_nextn as _resolve_dspark_nextn
 from aiconfigurator.sdk.config_builders import resolve_nextn_auto as _resolve_nextn_auto
 from aiconfigurator.sdk.errors import ExperimentOutcome, NoFeasibleConfigError, is_gpu_retriable
 from aiconfigurator.sdk.models import (
@@ -372,6 +373,13 @@ def cli_default(
 
 _MAX_GPUS_PER_WORKER = 64
 
+# Conservative nextn_accepted fraction used when recommend auto-enables DSPARK.
+# nextn_accepted has no backend equivalent — it is AIC's throughput modeling
+# assumption (accepted tokens per step as a fraction of the block size).
+# Users who have empirical acceptance data can override via cli_recommend's
+# nextn_accepted parameter. 0.8 is conservative for a well-aligned draft model.
+_DSPARK_DEFAULT_ACCEPTANCE = 0.8
+
 
 def _build_recommend_tasks(base_tasks: dict, total_gpus: int) -> dict:
     """Scale GPU candidates for recommend mode."""
@@ -513,6 +521,17 @@ def cli_recommend(
     # nextn="auto" resolves the draft depth from the checkpoint first.
     if nextn == "auto":
         nextn = _resolve_nextn_auto(model_path)
+
+    # DSPARK architectures (e.g. Kimi-K3) always run with a standalone draft
+    # model in production. Their block size is an architectural constant absent
+    # from the checkpoint, so nextn="auto" returns 0. Auto-enable here so that
+    # recommend produces accurate throughput estimates without requiring callers
+    # to know about DSPARK internals.
+    if nextn == 0 and (dspark := _resolve_dspark_nextn(model_path)):
+        nextn, dspark_accepted = dspark
+        if nextn_accepted is None:
+            nextn_accepted = dspark_accepted
+
     nextn, nextn_accepted = _normalize_nextn(nextn, nextn_accepted)
 
     base_tasks = build_default_tasks(
--- a/tests/unit/cli/test_cli_api.py
+++ b/tests/unit/cli/test_cli_api.py
@@ -540,6 +540,93 @@ class TestCLIRecommendUnit:
         assert captured_kwargs["enable_wideep"] is True
         assert captured_kwargs["moe_backend"] == "deepep_moe"
 
+    def test_dspark_nextn_auto_enabled(self, monkeypatch):
+        """cli_recommend auto-enables DSPARK via the nextn='auto' → 0 → DSPARK path.
+
+        Exercises the realistic production flow: nextn='auto' calls
+        _resolve_nextn_auto which returns 0 for DSPARK models (block size is not
+        in the checkpoint), then _resolve_dspark_nextn fills in the architectural
+        block size and planning acceptance.
+        """
+        import aiconfigurator.cli.api as api
+
+        captured = {}
+
+        def fake_build_default_tasks(**kwargs):
+            captured.update(kwargs)
+            return {}
+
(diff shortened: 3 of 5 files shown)


tianhaox (commented):
The direction is right: `kimi_k3.py` already fully models DSPARK (its docstring states "`nextn` is the dspark block size", with the 5-layer draft-model ops and the (nextn+1) verify width), but the recommend path could never reach it because `nextn="auto"` reads the main checkpoint and always resolves to 0. Auto-enabling in recommend is a sensible product decision, and the happy path is well tested. Two issues below I'd consider blocking, plus a few smaller ones.

## Blocking

### 1. Explicit `nextn=0` is silently overridden, with no opt-out

The trigger condition is `nextn == 0` (`src/aiconfigurator/cli/api.py:501`), but `cli_recommend`'s default is `0`, not `"auto"` (api.py:400). So "caller didn't specify" and "caller explicitly asked for no speculative decoding" are indistinguishable: a user sizing Kimi-K3 *without* DSPARK (e.g. to quantify its benefit) passes `nextn=0` and gets silently rewritten to `nextn=7` — and there is no parameter that can turn this off. It also contradicts the spirit of this PR's own test (`test_dspark_explicit_nextn_accepted_not_overridden`): explicit acceptance is preserved, but explicit nextn is not.

Suggestion: move the DSPARK fallback inside the `nextn == "auto"` branch (auto → checkpoint returns 0 → consult the DSPARK table), and change recommend's default to `"auto"`. That keeps `auto` meaning "figure out the right speculative config for me" while `0` recovers its "explicitly disabled" meaning. The docstring "Default is 0 (disabled)" (api.py:448) needs updating either way — it's currently wrong for DSPARK models.

### 2. The acceptance assumption moved into aic-core, undoing #1405's layering

In the same file, `validate_nextn`'s docstring says: "The aic-core layer owns only the compute-side draft depth. Accepted-token progress is modeled by the upper prediction layer and therefore is intentionally not part of this helper." And this PR's own description notes `nextn_accepted` was deliberately removed from the core engine spec in #1405. Yet `_DSPARK_DEFAULT_ACCEPTANCE` and the acceptance return value now live in core's `config_builders.resolve_dspark_nextn` (config_builders.py:97,122).

Suggestion: have the core resolver return only the block size (`int | None` from the `DSPARK_NEXTN` table), and keep the 0.8 planning assumption in `src/aiconfigurator/cli/api.py` — which already defines that constant (see below), suggesting an earlier iteration had exactly this shape.

## Non-blocking

3. **`_DSPARK_DEFAULT_ACCEPTANCE = 0.8` at api.py:356 is dead code.** It's never used (the value comes from `resolve_dspark_nextn`) and duplicates config_builders.py:97. After the refactor in (2) it becomes the live copy; otherwise it should be deleted.

4. **`except Exception: pass` swallows failures silently** (config_builders.py:123). If the Kimi-K3 config fetch fails, recommend silently falls back to `nextn=0` — reintroducing the very 18x underestimate this PR exists to fix, with no signal. At minimum log a warning; the broad except also hides genuine programming errors.

5. **The 0.8 acceptance needs a stated basis.** The comment claims "0.8 is conservative for a well-aligned draft model", but for a 7-token block, `accepted = 0.8 × 7 = 5.6` means ~6.6 tokens land per step. If 0.8 were a *per-token* acceptance with prefix-contiguous acceptance, the expected accepted length is `p(1−p⁷)/(1−p) ≈ 3.2`, not 5.6 — roughly a 60% throughput difference, which for a procurement tool means under-provisioning. If there's measured Kimi-K3 DSPARK acceptance-length data supporting 5.6, please cite it in the comment; if not, "conservative" should go. Relatedly, the 33 → 598 tok/s claim is 18x, above the `nextn+1 = 8x` per-step ceiling — likely superlinear from the tpot-SLA-feasible config space shifting, but worth a sentence confirming it's not a modeling artifact.

6. **Tests:** once (1) is fixed, please add an "explicit `nextn=0` stays 0" case. The three CLI tests share most of their scaffolding and could be a single parametrized test (optional).

## What's done well

- Wired into `cli_recommend` only; `cli_estimate`/`cli_default` keep strict behaviour — well-scoped.
- `DSPARK_NEXTN` sits next to the existing `DSPARK_ARCHITECTURES` with a clear note on why it isn't user-configurable.
- `get_model_config_from_model_path` is `@cache`d, so the extra lookup for non-DSPARK models costs nothing in practice.
- Both core-side and CLI-side tests exist, and the monkeypatch points (module-attribute level) are correct.

aic-core/src/aiconfigurator_core/sdk/config_builders.py
```diff
@@ -90,6 +90,41 @@ def resolve_nextn_auto(model_path: str) -> int:
...
+    from aiconfigurator_core.sdk.common import DSPARK_NEXTN
+    from aiconfigurator_core.sdk.utils import get_model_config_from_model_path
+
+    if not model_path:
+        raise ValueError("resolve_dspark_nextn requires a model path.")
+    try:
+        info = get_model_config_from_model_path(model_path)
+        block_size = DSPARK_NEXTN.get(info.get("architecture", ""), 0)
+        if block_size:
+            return block_size, block_size * _DSPARK_DEFAULT_ACCEPTANCE
```
jasonqinzhou, line 124: **[P1] Keep acceptance workload-scoped instead of architecture-scoped.** The block size is an architectural fact, but `nextn_accepted` is the prompt/workload-dependent progress assumption that the upper prediction layer applies to service metrics. Returning `7 * 0.8 = 5.6` here makes every Kimi-K3 recommend run divide decode work by the same 6.6 tokens/iteration; if a deployment's actual acceptance is lower, the procurement result overstates throughput and can underprovision GPUs. This also crosses the aic-core boundary that intentionally owns compute depth only. Please auto-resolve the fixed block size separately and keep acceptance explicit/measured, or introduce a clearly labeled estimated scenario backed by backend/workload-specific evidence rather than treating one 0.8 value as universally conservative.

src/aiconfigurator/cli/api.py
```diff
@@ -484,6 +492,17 @@ def cli_recommend(
     # nextn="auto" resolves the draft depth from the checkpoint first.
     if nextn == "auto":
         nextn = _resolve_nextn_auto(model_path)
+
+    # DSPARK architectures (e.g. Kimi-K3) always run with a standalone draft
+    # model in production. Their block size is an architectural constant absent
+    # from the checkpoint, so nextn="auto" returns 0. Auto-enable here so that
+    # recommend produces accurate throughput estimates without requiring callers
+    # to know about DSPARK internals.
+    if nextn == 0 and (dspark := _resolve_dspark_nextn(model_path)):
```
jasonqinzhou, line 530: **[P1] Preserve the explicit `nextn=0` opt-out.** This condition cannot distinguish an omitted/default value from a caller deliberately passing zero, so `cli_recommend(..., nextn=0)` now forwards `nextn=7, nextn_accepted=5.6`. That contradicts the public API and recommend help text (`0` is documented as disabled) and leaves no way to size a Kimi-K3 deployment that does not load DSPARK; such a run can be underprovisioned by modeling speculative throughput that will not exist. Please distinguish automatic/default selection from explicit zero (and update the recommend API/CLI documentation for the chosen default), with a regression test proving the explicit disable path remains disabled.

jasonqinzhou (changes requested):

## What this PR does

This PR adds Kimi-K3 DSPARK block-size metadata and makes cli_recommend automatically model the standalone draft checkpoint with nextn=7 and an accepted-token estimate, so GPU sizing includes speculative-decoding throughput.

The block-size resolution is separated cleanly from checkpoint MTP resolution, and the focused tests cover unsupported architectures, metadata failures, explicit acceptance preservation, non-DSPARK behavior, and the previously identified auto-path gap.

## Why this score

The 6.5 score and REQUEST_CHANGES recommendation reflect two procurement-safety regressions despite clean implementation mechanics and passing focused tests: the change turns a workload-dependent acceptance measurement into a fixed architecture constant, and it overwrites the existing explicit nextn=0 disable contract. Either behavior can materially overstate throughput and underprovision deployments that do not achieve or use the assumed DSPARK behavior.

natoscott:
(accidentally closed when moving from PR #1505 base to main branch based)

aic-core/src/aiconfigurator_core/sdk/config_builders.py
natoscott: Done.

src/aiconfigurator/cli/api.py
natoscott: Done and test added.

tests/unit/cli/test_cli_api.py
```diff
@@ -520,6 +520,91 @@ def fake_execute(tasks, mode, **kwargs):
...
+                lambda cap, api: abs(cap["nextn_accepted"] - 7 * api._DSPARK_DEFAULT_ACCEPTANCE) < 1e-9,
+                id="dspark-auto-enable",
+            ),
+            pytest.param(
+                7,
+                {"nextn_accepted": 3.0},  # caller supplies explicit acceptance
+                7,
+                lambda cap, _: cap["nextn_accepted"] == 3.0,
+                id="dspark-explicit-acceptance-preserved",
+            ),
```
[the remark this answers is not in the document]
natoscott: Done.

natoscott:
@jasonqinzhou @tianhaox PTAL, I think the issues are all resolved here as well now.

tianhaox (approved):

jasonqinzhou (approved):

PR #1507, opened 2026-08-08
feat: MiniMax-M3 MSA silicon support — collectors (trtllm/vllm/sglang), SDK tables, multi-platform data

tianhaox:
## Release at a Glance

MiniMax-M3 sparse attention (MSA) goes from HYBRID-estimated to **measured SILICON support**:

- **New MSA module collectors for all three backends** (`msa_context_module` / `msa_generation_module`), benchmarking the full M3 sparse-attention module through each framework's own serving construction path (serving-parity citations at the pinned versions).
- **`b200_sxm` gains M3 MSA tables for all three backends** (8x NVIDIA B200 collection).
- **SDK (Python + Rust) MSA silicon tables**: SILICON resolves on the measured grid; HYBRID prefers silicon and falls back to the pre-existing DSA xop transfer only on a typed miss. Rust port is parity-tested.
- New `msa` catalog family: sglang msa pins the official v0.5.16 image (first release with M3) and trtllm msa pins 1.3.0rc23 (first release with the fmha_sm100 backend) while every other family stays on the stock pins.

### Customer impact

MSA-module-accurate MiniMax-M3 modeling on `b200_sxm` for trtllm/vllm/sglang. Previously M3 was xop-estimated only (utilization borrowed from GLM-5 DSA), which measurement shows underestimates long-context prefill and large-batch decode by an order of magnitude.

Cross-framework findings from the new B200 data (same GPU, same harness):

| point | trtllm rc23 (fmha_sm100) | vLLM 0.24.0 (fmha_sm100) | sglang 0.5.16 (Triton) |
|---|---|---|---|
| ctx b=2 s=512 h=64 | 4.76 ms | **0.33 ms** | 0.69 ms |
| ctx b=1 s=16384 h=64 | 15.9 ms | **4.5 ms** | 13.7 ms |
| gen b=48 kv=10000 h=8 | 0.396 ms | **0.063 ms** | 0.076 ms |
| gen b=256 kv=65536 h=8 | 7.37 ms | **0.68 ms** | 0.796 ms |

Decode rows in this table are hand-measured spot checks (b=48 is not on the collection grid). The sglang b=256 kv=65536 cell originally read 0.23 ms, which matched no measurement of that shape (nearest table row is b=256 step=1024 = 0.2429); the corrected 0.796 ms is a direct rerun (2026-08-08).

Root-cause debugging of the trtllm gap (full-index-K-pool copy on decode, host-bound eager prefill — both fixed on upstream main post-rc23) and the sglang prefill gap (Triton reference kernel at ~5.7% of B200 peak; the official image lacks the fmha_sm100 wheel) is recorded in the collection_meta files and tracked for re-collection in #1509.

vLLM and TRT-LLM rc23 run the **same fmha_sm100 kernel package**; the 3-14x gap is integration-layer overhead (trtllm's `implementation="msa"` is flagged prototype upstream).

## Backend images (runtime versions verified in-container)

- trtllm `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc23` (msa family; digest pinned in `framework_manifest.yaml`); rc20 remains the default for every other family
- vllm `vllm/vllm-openai:v0.24.0`
- sglang `lmsysorg/sglang:v0.5.16` (msa family; stock 0.5.14 elsewhere)

## What's in the change

- Collectors: `collector/{trtllm,vllm,sglang}/collect_msa_module.py` + registry entries + `msa` catalog family + manifest family pins; M3 case plan gains the MoE shape row (6144/3072/128x4, DeepSeekV3-style sigmoid routing) and msa op activation.
- SDK: `msa_context_module_perf`/`msa_generation_module_perf` tables (keyed like DSA modules), Python + Rust loaders/queries; silicon-first HYBRID; bundled M3 config enriched to full HF checkpoint fields.
- Data: b200_sxm msa (context+generation rows) — vllm 11712+8832, trtllm 5070+3864, sglang 26933+1477.

## Collector details worth review attention

- trtllm rc23 `implementation="msa"`: new flat-metadata contract handled by a Metadata-class probe (never version sniffing); KV page size 128 for the msa path (`py_executor_creator.py:427@1.3.0rc23`); the rc19/rc20 Triton contract keeps working.
- KVCacheManagerV2 dummy-request pool headroom (INDEX_KEY pages share the max_tokens budget; `is_gen=True` for decode dummies).
- Offline-config shims for the M3 model_type in each framework (construction-only; kernels untouched).
- TP-shard emulation follows each framework's real index-head sharding (trtllm replicates; vLLM rides the KV-head shard; sglang uses idx_head_tp_size=min(tp,4)) — cited in each collector.

## Known gaps / accepted failures (documented in each collection_meta.yaml)

- `num_heads=128` cases fail by construction on trtllm (`num_index_heads % num_kv_heads` framework check; serving-unreachable shard) — FIXME(kernel-limit) at the call site.
- b200_sxm trtllm: 54 bare AssertionErrors clustered at s<=32/b>=64 shapes (0.9% of plan; serving-parity audit pending).
- sglang B200 data is the Triton path: the official v0.5.16 image does not bundle the fmha_sm100 wheel, so that IS the image's serving truth; re-collect when a bundled image exists.
- Full-model M3 estimates: this PR ships the msa AND moe rows (MoE on all 8 platforms, see the matrix below); gemm/comm serve from the platforms' existing tables.
- trtllm B200 data uses `implementation="msa"`; serving DEFAULTS to the slower Triton reference unless the knob is set.
- Registry maturity markers: the MSA ops carry `unverified_sms=(121,)` in all three registries (the withdrawn SM89/SM120 sglang cells are expressed as R matrix cells, not markers). H100/H200 (SM90) MSA tables ARE included in this PR.

## Validation

- `pytest tests/unit/collector tests/unit/sdk/database -m unit`: 1077 passed.
- Rust: `cargo test -p aiconfigurator-core` 302 passed; engine-step parity suite 321 passed.
- SILICON grid-point queries match independently measured smoke values within ~1%.

## Summary by CodeRabbit

* **New Features**
  * Added native MSA performance data and queries for MiniMax M3 context and generation workloads.
  * Added controlled fallback behavior across SILICON, HYBRID, and EMPIRICAL modes.
  * Added MSA benchmarking support for TensorRT-LLM, vLLM, and SGLang.
  * Expanded MiniMax M3 configuration, runtime coverage, and framework-specific collection support.

* **Documentation**
  * Updated MSA performance-table and model behavior documentation.

* **Tests**
  * Added coverage for interpolation, lookup priority, fallback handling, schemas, and missing data.

---

## Supported matrix (2026-08-20 — exact, per cell)

Legend: **S** = own SILICON table (measured, serving-path) · **H** = typed SILICON miss, HYBRID empirical transfer (provenance-tagged) · **R** = explicitly rejected cell (typed `EmpiricalNotImplementedError`; never a silent fallback).

**MSA module — BF16 checkpoint (`MiniMaxAI/MiniMax-M3`)**

| platform | trtllm 1.3.0rc23 | vllm 0.24.0 | sglang 0.5.16 |
|---|---|---|---|
| b200_sxm / b300_sxm / gb200 / gb300 | **S** (fmha_sm100; generated deployments prescribe `sparse_attention_config.implementation: msa` on both entry points) | **S** (fmha_sm100 impls) | **S** (Triton sparse) |
| h100_sxm / h200_sxm | **S** (Triton reference) | **S** (Triton impls) | **S** (Triton sparse) |
| l40s | **S** (Triton reference) | **S** (Triton impls) | **R** — SGLang v0.5.16 has no CC-8.9 branch; default serving cannot initialize (tables withdrawn) |
| rtx_pro_6000_server | **R** — no table yet (collection-pool availability) and no DSA xop donor; regression test pins the typed contract | **S** (Triton impls) | **R** — no CC-12 branch, same as l40s |

**MSA module — NVFP4 checkpoint (`nvidia/MiniMax-M3-NVFP4`)** — its MXFP8 projections price as `gemm=fp8_block`:

| platform | trtllm | vllm | sglang |
|---|---|---|---|
| b200/b300/gb200/gb300/h100/h200 | **S** (fp8_block gemm tier collected) | **S** (fp8_block tier collected) | **H** — SGLang's quantized-M3 flow (MXFP8) is unsupported in serving; tables are bf16-gemm only by declaration, and the DSA donor makes the HYBRID transfer resolve |
| l40s | **S** (fp8_block tier collected) | **S** (fp8_block tier collected) | **R** — tables withdrawn (no CC-8.9 branch) AND no DSA xop donor: SILICON raises `PerfDataNotAvailableError`, HYBRID raises `EmpiricalNotImplementedError`; regression test pins both |
| rtx_pro_6000_server | **R** (as BF16 row) | **R** — the fp8_block tier fails classified on SM120 (DeepGEMM `layout.hpp:59`) AND no DSA xop donor, so both modes fail typed (regression-pinned); re-collect when vLLM fixes DeepGEMM on SM120 | **R** (as BF16 row) |

**MoE (M3 geometry, both checkpoints)**: BF16 rows on all 8 platforms x 3 frameworks. NVFP4 shipped rows exist only where the checkpoint's nvfp4 kernels exist (`min_sm: 100` capability floor): **5/8 platforms for trtllm and vllm** (b200/b300/gb200/gb300/rtx) and **4/8 for sglang** (rtx: flashinfer_cutlass `select_experts` NotImplemented on SM120 — framework gap). On SM89/SM90 platforms (h100/h200/l40s) the nvfp4 quant mode is capability-gated out at plan time — no NVFP4 rows ship there; any estimate on those platforms comes from the MoE quant transfer ladder, which is transfer behavior, not shipped data.

Every **H**/**R** cell above is a framework or infrastructure gap documented in the affected `collection_meta.yaml`; none is an omission of collectable data. The rtx trtllm MSA table lands as a follow-up when the collection pool recovers (per-family rc23 access is anchored by a single mla-family reuse marker; other families serve rc20 via the resolver's native sibling inheritance).


aic-core/src/aiconfigurator_core/systems/data/b200_sxm/msa/trtllm/1.3.0rc23/collection_meta.yaml
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# 8x B200 campaign via internal collection CI (2026-08-08).
+# kernel_source = msa_fmha_sm100: collected with implementation="msa"
+# (the SM100-family true-sparse path; serving DEFAULTS to the slower
+# "triton" reference unless the deployment sets the knob —
+# llm_args.MiniMaxM3SparseAttentionConfig.implementation@1.3.0rc23).
```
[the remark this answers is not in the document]
tianhaox: Agreed this is a real contract question, and it was a deliberate scope call rather than an oversight. Rationale, recorded in this leaf's collection_meta.yaml: on SM100/103 the `implementation="msa"` (fmha_sm100) path is the performance path the sparse config exists for — the Triton reference path on those parts measures far off SOL and is not a sensible production configuration, while `kernel_source` in every persisted row records which kernel actually ran. Encoding `implementation` as a first-class axis of the persisted-data/query contract is a producer+consumer contract change (perf schema + SDK loaders + Rust parity engine), which per the collector module-boundary rules needs its own coordinated change rather than riding a data PR. Proposed follow-up: add the axis (or an explicit meta gate) once the next TRT-LLM rc stabilizes the MSA path — that rc bump already requires re-collection here (the rc23 measurement caveats are documented in this meta), so the axis can land with data that carries it from day one.

collector/trtllm/collect_msa_module.py
```diff
@@ -0,0 +1,1073 @@
...
+    # 1.3.0rc23 added implementation: Literal["triton","msa"] (default
+    # "triton"; llm_args.py MiniMaxM3SparseAttentionConfig). The "msa"
+    # (fmha_sm100) kernels are the performance path on the SM100 family and
+    # hard-require it (sparse/minimax_m3/msa_availability.py:ensure_msa_available
+    # raises off SM100/103), so collect "msa" there and the Triton reference
+    # elsewhere. kernel_source records which one actually ran. Field-presence
+    # probe keeps rc19/rc20 (no such field, pydantic strict) working.
+    if "implementation" in getattr(MiniMaxM3SparseAttentionConfig, "model_fields", {}):
+        sparse_kwargs["implementation"] = "msa" if get_sm_version() in (100, 103) else "triton"
+    sparse_attention_config = MiniMaxM3SparseAttentionConfig(**sparse_kwargs)
```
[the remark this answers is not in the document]
tianhaox: The override is intentional and owner-authorized (recorded at the construction site with the rc23 citations, including `llm_args.py` defaulting to `"triton"`): serving's *default* is indeed Triton, but the collected configuration on SM100/103 is the explicit `implementation="msa"` serving knob — the fmha_sm100 path this config field was added for, hard-gated by `ensure_msa_available` to those SMs. This mirrors the user-facing escape-hatch pattern the collector rules allow when a framework default cannot represent the production-intent path, and `kernel_source` records ground truth per row (`msa_fmha_sm100` vs `default`), so the data never claims to be the Triton path. The default-Triton-vs-msa selection question is the same contract issue as your comment on the collection meta — see the reply there for the proposed follow-up.

collector/trtllm/collect_msa_module.py
```diff
@@ -0,0 +1,1073 @@
...
+    sm = get_sm_version()
+
+    gemm_types = ["bfloat16"]
+    if sm >= 89:
+        gemm_types.append("fp8_block")
+    if sm >= 100:
+        gemm_types.append("nvfp4")
+
+    attn_combos = [("bfloat16", "bfloat16")]
+    return [(c, kv, g) for g in gemm_types for c, kv in attn_combos]
```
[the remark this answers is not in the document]
tianhaox: Half of this is done in this push: the sglang fp8_block combos are now scoped via `attention_types: [mla, dsa]` in `cases/base_ops/mla_module.yaml` and the in-collector `spec.gemm_type != "bfloat16"` filter is deleted — the msa query now yields exactly the combos the collector runs, with golden case counts unchanged. For trtllm/vllm the MSA grids stay in-collector for this PR: each combo there carries per-framework serving citations (fp8-KV handling and nvfp4 floors differ per framework), and migrating the declaration source after the shipped tables were collected under the current expansion couples a mechanism change into a data PR. Follow-up proposed: declare msa-scoped combos in the trtllm/vllm sections and verify a zero-diff expansion before switching the collectors over.

collector/vllm/collect_msa_module.py
```diff
@@ -0,0 +1,1010 @@
...
+    prefix_len = int(prefix_len) if is_context else 0
+
+    if is_context:
+        batch_spec = BatchSpec(
+            seq_lens=[prefix_len + seq_len] * batch_size,
+            query_lens=[seq_len] * batch_size,
+        )
+    else:
+        batch_spec = BatchSpec(
+            seq_lens=[seq_len] * batch_size,
```
jasonqinzhou: **[P1] Measure the KV length recorded in the row.** `get_generation_test_cases()` defines this argument as `kv_cache_len`, and the row later writes `isl=1, step=seq_len` (loader key `s=seq_len+1`). But `BatchSpec(seq_lens=[seq_len], query_lens=[1])` makes `num_computed_tokens=seq_len-1` in `create_common_attn_metadata`, and the current token is placed at `seq_len-1`; the measured total length is therefore `seq_len`, one below the persisted coordinate. Please construct generation metadata with `seq_len` cached tokens and the current token at position `seq_len`, then recollect the vLLM generation parquets.

collector/trtllm/collect_msa_module.py
```diff
@@ -0,0 +1,1116 @@
...
+    if is_context:
+        max_seq = prefix_len + seq_len + 1
+        total_tokens = seq_len * batch_size
+        seq_len_q = seq_len
+        kv_cache_len = prefix_len
+    else:
+        max_seq = seq_len + 1
+        total_tokens = batch_size
+        seq_len_q = 1
+        kv_cache_len = seq_len - 1
```
jasonqinzhou: **[P1] Keep TensorRT-LLM generation rows on the declared KV-length coordinate.** The generation getter labels `seq_len` as `kv_cache_len`, but this subtracts one while the persisted row writes `isl=1, step=seq_len`; the loader places the measurement at total `s=seq_len+1` even though the runtime sees `seq_len-1` cached tokens plus one query, total `seq_len`. Please use `seq_len` cached tokens with consistent dummy-request and current-position metadata, then recollect the TensorRT-LLM generation parquets.

collector/vllm/collect_msa_module.py
```diff
@@ -0,0 +1,1010 @@
...
+        isl = 1
+        step = seq_len
+
+    op_name = f"msa_{phase}_module"
+
+    # Ground truth: the impl class the framework's own dispatch selected at
+    # construction (select_main_impl_cls — Triton off-SM100, MSA on SM100).
+    kernel_source = type(attn_module.impl).__name__
+
+    log_perf(
```
jasonqinzhou: **[P1] Fail the case when persistence fails.** `log_perf()` returns `False` after lock exhaustion or a write/fsync exception, but this call discards the result. The worker then returns normally and the checkpoint can advance even though no MSA row exists. Please mirror the TensorRT-LLM and SGLang paths by raising when `log_perf(...)` is false, and add a failure-path test so a dropped row cannot be reported as a successful case.

collector/vllm/collect_msa_module.py
```diff
@@ -0,0 +1,1010 @@
...
+        raise
+
+    # 6. Benchmark. CUDA graph capture is mandatory (allow_graph_fail=False):
+    # the M3 Triton kernels are capture-safe (no host syncs; the builders
+    # declare AttentionCGSupport.UNIFORM_BATCH, common/sparse_attention.py
+    # :161-167, common/indexer.py:214-219@v0.24.0), verified on SM90.
+    def kernel_func():
+        attn_module.forward(positions, hidden_states)
+
+    with benchmark_with_power(
```
jasonqinzhou: **[P2] Guarantee cleanup after every case.** `_cleanup()` runs for dry-run exceptions and at the successful tail, but an exception from `benchmark_with_power`, result extraction, persistence, or reporting bypasses it. `collect.py` reuses the worker for later cases, so the vLLM workspace singleton and GPU allocations can stay pinned and turn one failure into misleading follow-on OOMs. Please put the case body under an outer `try/finally` that always resets the workspace and CUDA allocations.

jasonqinzhou (dismissed):
**REQUEST_CHANGES · 5.1/10 · high confidence**

## What this PR does

This PR replaces MiniMax-M3 MSA's DSA cross-op estimate with native silicon modeling. It adds serving-path collectors for TensorRT-LLM 1.3.0rc23, vLLM 0.24.0, and SGLang 0.5.16; Python and Rust MSA table loading with silicon-first HYBRID fallback; MiniMax-M3 and NVFP4 model declarations; Dynamo 1.3 routing; and MSA or MoE data across B200, B300, GB200, GB300, H100, H200, L40S, and RTX Pro 6000 systems.

The collectors contain unusually detailed serving-parity citations and explicit guards for architecture, precision, kernel selection, CUDA-graph coverage, and persistence. The Python and Rust parity coverage plus the per-collection metadata make this large hardware-data contribution substantially easier to audit.

## Why this score

The current head is 5.1/10 with REQUEST_CHANGES. The SDK path and focused unit coverage are strong, but the generation collectors for both TensorRT-LLM and vLLM benchmark one fewer cached token than their persisted table coordinate claims, so those generation tables are systematically mis-keyed. The vLLM path can also report a case successful after log_perf fails, and the branch omits the required backend-facts update for 150 new slices. Resource cleanup is not guaranteed after later vLLM failures. The branch additionally fails Ruff and format, PR-title validation, and DCO, although all test, build, parity, wheel, prediction, chart, and parquet checks have passed.

### P1 — Land the backend-facts registry update for the new MSA slices

**[P1] Land the backend-facts registry update.** `python3 tools/perf_database/backend_facts.py --check` fails with 150 drift differences, and six new labels are absent from `collector/kernel_source_backends.yaml`: the three SGLang MSA labels, `msa_fmha_sm100`, and both vLLM M3 implementation labels. Please trace and add the label mappings, regenerate `collector/op_backend_facts.yaml`, and make the fail-closed collector-data check pass before merging the new tables.

collector/vllm/collect_msa_module.py
tianhaox: Confirmed and fixed. `create_common_attn_metadata` derives `num_computed_tokens = seq_lens - query_lens`, so `seq_lens=[seq_len]` cached `seq_len-1` tokens while the row keyed `s = isl + step = seq_len + 1` — exactly the one-token skew you describe. The generation spec now passes `seq_lens=[seq_len + 1]` with the current token at position `seq_len` (measured total == keyed coordinate), with the contract documented at the construction site. Instead of recollecting, the 15 shipped trtllm/vllm generation parquets were re-keyed `step -> step - 1` onto their measured coordinates: each measurement is a valid point whose label was wrong, so the shift restores label truth without re-measurement (verified: no key collisions, min step 2 -> 1; per-leaf collection_meta notes added). New collections land on the corrected construction. Note the same construction pattern exists in the vLLM DSA module collector (`collect_mla_module.py` uses the identical gen BatchSpec and row math), so the shipped DSA module generation tables carry the same skew — happy to file the follow-up for that family separately since its tables predate this PR.

collector/trtllm/collect_msa_module.py
tianhaox: Confirmed and fixed. `kv_cache_len = seq_len - 1` cached one token short of the persisted coordinate. Now: `kv_cache_len = seq_len`, dummy-request `token_nums = [seq_len + 1]` (past + current, per `add_dummy_requests` semantics), decode position `seq_len`, and `max_seq = seq_len + 2` keeps the prior +1 headroom above the new total. Shipped trtllm generation parquets re-keyed `step -> step - 1` onto their measured coordinates (same rationale as the vLLM thread: valid measurements, wrong labels; meta notes added per leaf) — re-measurement is planned anyway for the next TRT-LLM rc (the rc23 measurement caveats documented in the collection metas), and those recollections will land on the corrected construction.

collector/vllm/collect_msa_module.py
tianhaox: Fixed: the row write now goes through `_persist_msa_row`, which raises when `log_perf` returns False (mirroring the TRT-LLM/SGLang paths), so a dropped row fails the case instead of advancing the checkpoint. Added the requested failure-path unit test (`test_vllm_msa_persist_row_raises_when_log_perf_fails`) covering both the success passthrough and the raise.

collector/vllm/collect_msa_module.py
tianhaox: Fixed: the case body (dry run -> benchmark -> persistence -> reporting) now runs under an outer try/finally that always releases the WorkspaceManager singleton and CUDA allocations (`_cleanup()`), so a failure in any later stage can no longer pin the workspace and cascade into follow-on OOMs for the worker's subsequent cases.

tianhaox:
@jasonqinzhou Thanks for the review — all items addressed as of ec39c42b:

- **Generation KV coordinate (both P1s)**: confirmed real for trtllm and vLLM (sglang was already correct). Construction fixed to cache `seq_len` tokens with the current token at position `seq_len`; the 15 shipped trtllm/vllm generation parquets were re-keyed `step -> step-1` onto their measured coordinates (valid measurements, wrong labels — details in the two threads), with per-leaf collection_meta notes. The same pattern exists in the pre-existing vLLM DSA module collector/tables; proposed as a separate follow-up.
- **backend-facts registry**: landed earlier — the six MSA kernel_source labels are classified in kernel_source_backends.yaml with serving citations and the registry is regenerated; `backend_facts.py --check` passes (3,269 slices), as does `check_collector_data.py`.
- **vLLM persistence**: `log_perf` failures now raise (with a failure-path unit test), and the case body runs under try/finally that always releases the workspace/CUDA allocations.
- **Ruff/format, PR title, DCO**: all fixed earlier; CI was 23/23 green before this push and the branch is re-merged with latest main.

Local gates on the new head: unit 3261, cargo 420, parity 449, ruff/format clean, both data checks OK.

aic-core/rust/aiconfigurator-core/src/operators/msa.rs
```diff
@@ -72,41 +81,63 @@ impl MsaModuleOp {
             DatabaseMode::Sol | DatabaseMode::SolFull => {
                 Ok(PerformanceResult::new(sol * self.scale_factor, Source::Sol))
             }
-            DatabaseMode::Silicon => Err(AicError::PerfDatabase(
-                "MSA has no silicon data; use HYBRID or EMPIRICAL.".to_string(),
-            )),
+            DatabaseMode::Silicon => self
+                .silicon_context(db, batch_size, s, prefix)
+                .map(|latency| PerformanceResult::new(latency * self.scale_factor, Source::Silicon)),
             DatabaseMode::Hybrid | DatabaseMode::Empirical => {
```
Arsene12358: **P1 — EMPIRICAL is taking the SILICON path.** README and CLI define EMPIRICAL as SOL+empirical for all, but this branch makes EMPIRICAL silicon-first. Reproduced on B200/vLLM M3: SILICON, HYBRID, and EMPIRICAL all returned 12.618 ms with source=<code>silicon</code>. Please route EMPIRICAL directly to the XOP context/generation paths and add source/provenance assertions.

collector/sglang/collect_msa_module.py
```diff
@@ -0,0 +1,1485 @@
...
+    # per-token req_to_token slots and constrain only block_size_k
+    # (kernels/ops/attention/minimax_sparse/prefill/topk_sparse.py:279-292).
+    #
+    # Dispatch-not-skip: this changes HOW the case is constructed (a
+    # ServerArgs knob), never WHETHER it runs; kernel_source still records
+    # the sparse path the backend actually selected. Smoke-validated on
+    # SM120; SM89 evidenced by the l40s full-run failure log above; SM121
+    # shares the identical source-level gap (major-12 predicates) and stays
+    # registry-gated (unverified_sms) until validated.
+    attention_backend = "triton" if get_sm_version() in (89, 120, 121) else None
```
Arsene12358: **P1 — these tables do not represent the deployment AIC models.** This forces opt-in Triton after documenting that framework-selected FlashInfer crashes, while Task defaults to FlashInfer, rejects Triton, and the MSA lookup has no attention-backend key. AIC consequently returns silicon data for a default deployment that cannot initialize. Either select and emit Triton end-to-end with a keyed runtime identity, or do not advertise these system/backend slices.

collector/trtllm/collect_msa_module.py
```diff
@@ -0,0 +1,1115 @@
...
+    )
+    # 1.3.0rc23 added implementation: Literal["triton","msa"] (default
+    # "triton"; llm_args.py MiniMaxM3SparseAttentionConfig). The "msa"
+    # (fmha_sm100) kernels are the performance path on the SM100 family and
+    # hard-require it (sparse/minimax_m3/msa_availability.py:ensure_msa_available
+    # raises off SM100/103), so collect "msa" there and the Triton reference
+    # elsewhere. kernel_source records which one actually ran. Field-presence
+    # probe keeps rc19/rc20 (no such field, pydantic strict) working.
+    if "implementation" in getattr(MiniMaxM3SparseAttentionConfig, "model_fields", {}):
+        sparse_kwargs["implementation"] = "msa" if get_sm_version() in (100, 103) else "triton"
```
Arsene12358, line 366: **P1 — this collects a non-default implementation without a consumer key.** rc23 serving defaults to Triton, but all SM100/103 rows are collected with implementation=msa and the PR reports a material performance gap. ModelConfig and the generator have no matching knob, and MsaTable cannot distinguish implementations. Collect the default path or carry this setting through row identity, lookup, SDK validation, and generated deployment configuration.

collector/cases/models/MiniMaxM3ForCausalLM_cases.yaml
```diff
@@ -6,8 +6,96 @@ architecture: MiniMaxM3ForCausalLM
...
+      sglang_moe_has_correction_bias: true
+      framework_quantization:
+        trtllm:
+          allowed_modes: [nvfp4]
+        sglang:
+          allowed_modes: [nvfp4]
+        vllm:
+          allowed_modes: [nvfp4]
+  mla_module:
+    - model_path: MiniMaxAI/MiniMax-M3
```
Arsene12358, line 85: **P1 — the advertised NVFP4 artifact cannot populate MSA cases.** The top-level model_paths entry activates both MSA ops, but this sole MSA row names only BF16; exact NVFP4 targeting returns zero specs for all three backends. The real checkpoint requests gemm=fp8_block while every SGLang MSA table is BF16-only. Add a checkpoint-specific row and matching data where supported, reject unsupported cells explicitly, and add a targeted non-empty plus end-to-end lookup test.

src/aiconfigurator/generator/config/backend_version_matrix.yaml
```diff
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 matrix:
+  1.3.0:
+    vllm: "0.24.0"
+    sglang: "0.5.16"
+    trtllm: "1.3.0rc23"
```
Arsene12358, line 8: **P1 — the RTX/TRT-LLM HYBRID fallback is unreachable at this selected version.** RTX has no 1.3.0rc23 database root or reuse marker, so the CLI exits at the exact-version gate before MSA can produce a typed miss and use XOP. Reproduced with the documented RTX/TRT-LLM HYBRID command; only rc10, rc15, and rc20 are available. Add a valid rc23 root/reuse declaration or ship rc23 data, with an exact CLI regression test.

collector/sglang/collect_msa_module.py
```diff
@@ -0,0 +1,1485 @@
...
+        )
+
+        # Post-load drops against real runtime limits (logged, mirroring
+        # collect_mla_module): actual KV-pool capacity and the runtime
+        # chunk size (serving forms one extend batch per chunk).
+        page_size = kv_pool_page_size(model_runner)
+        capacity = kv_pool_capacity_tokens(model_runner)
+        if capacity is not None:
+            before = len(shapes)
+            shapes = [
```
Arsene12358: **Critical — grouped work is being removed after queueing.** The outer task carries placeholders, the worker reconstructs the inner grid, and this post-load filter silently changes the workset, so task IDs, resume checkpoints, and failure records do not bind the shapes that may produce rows. Build and memory-filter the exact manifest in get_*_test_cases(), log drops there, attach it to the outer task, and execute or raise for every retained shape.

collector/trtllm/collect_msa_module.py
```diff
@@ -0,0 +1,1115 @@
...
+        Triton softmax in the cache dtype (sparse/minimax_m3/backend.py
+        _sparse_gqa_masked@1.3.0rc20), and the index-K side cache is
+        allocated bf16 (cache_manager._torch_dtype_for_index_cache
+        falls back to bf16 for non-fp16/fp32 main dtypes). No fp8-KV
+        combo is declared until the framework grows one.
+    """
+    sm = get_sm_version()
+
+    gemm_types = ["bfloat16"]
+    if sm >= 89:
```
Arsene12358: **Important — this duplicates and disagrees with YAML precision policy.** At SM89 this collector adds FP8-block GEMM although get_mla_module_precision_specs declares BF16 only; the vLLM collector independently adds undeclared FP8-KV tuples. Consume the YAML-backed precision specs directly for backend, phase, SM, and attention_type=msa, and test equality between declared and emitted combinations.

Arsene12358 (changes requested):
Requesting changes on reviewed head <code>bdf3ac160258ab7c56bbb43ae02b9fc16d367961</code>. I found six distinct merge blockers after accounting for overlap between the Standards and Spec axes.

## Spec — 3 blockers

1. **EMPIRICAL incorrectly returns SILICON data.** <code>MsaModuleOp</code> groups <code>DatabaseMode::Empirical</code> with <code>Hybrid</code>, although README and CLI define EMPIRICAL as SOL+empirical for all. A B200/vLLM M3 probe returned the identical <code>12.618 ms, source=silicon</code> for SILICON, HYBRID, and EMPIRICAL. Route EMPIRICAL directly through the XOP path and add source/provenance tests for context and generation.

2. **The advertised NVFP4 coverage is not delivered.** An exact <code>COLLECTOR_MODEL_PATH=nvidia/MiniMax-M3-NVFP4</code> run produces zero MSA specs for TRT-LLM, vLLM, and SGLang because the only <code>mla_module</code> row names the BF16 artifact. The SDK resolves the checkpoint to <code>gemm=fp8_block</code>, while all eight SGLang MSA table pairs contain only <code>bfloat16</code>; SILICON misses and HYBRID silently uses empirical data. Add the checkpoint-specific declaration and matching data where serving works, explicitly reject unsupported cells, narrow the support claim, and add real-checkpoint end-to-end matrix tests.

3. **The stated RTX/TRT-LLM HYBRID fallback is unreachable.** Dynamo 1.3 selects TRT-LLM <code>1.3.0rc23</code>, but RTX has no rc23 database root or reuse marker. The CLI exits at the exact-version gate before an MSA typed miss can fall back; only rc10, rc15, and rc20 are available. Add a valid rc23 root/reuse declaration or ship the rc23 database, then test this exact CLI path.

## Standards — 4 blockers

1. **Critical: SGLang benchmarks a substituted backend.** On SM89/120/121 the collector forces opt-in <code>attention_backend=triton</code> after documenting that framework-selected FlashInfer crashes. This violates the repository requirement to measure framework-selected serving truth and never replace a failing selected path. It is also functionally unsafe because the SDK defaults to FlashInfer, rejects <code>triton</code>, and the MSA key has no attention-backend axis.

2. **Critical: grouped SGLang work is not bound to its inner shapes and silently loses queued work.** The getter queues outer placeholders, the worker rebuilds the grid, and the post-load capacity check removes shapes. Move memory filtering into <code>get_*_test_cases()</code>, attach the exact retained inner manifest to the task, and execute or raise for every retained shape.

3. **Important: the declared NVFP4 MSA plan completes with zero cases.** The planned MSA ops have no logged capability or memory drop explaining the empty expansion, which violates the collector population contract. This is the same underlying coverage defect as Spec finding 2.

4. **Important: TRT-LLM and vLLM reconstruct YAML precision policy in collector code.** Their hardcoded SM formulas already disagree with <code>get_mla_module_precision_specs</code> on SM89: TRT-LLM adds undeclared FP8-block GEMM cases, and vLLM adds undeclared FP8-KV cases. Consume the YAML-backed precision specs directly and add parity coverage.

## Cross-module runtime consistency

TRT-LLM SM100/103 tables are collected exclusively with opt-in <code>implementation="msa"</code>, while rc23 serving defaults to <code>triton</code>; the PR reports a material 3–14× gap. <code>MsaTable</code> has no implementation key and ModelConfig/generator has no matching knob, so default deployments unconditionally receive latency from a different implementation. Either collect the default path or carry the implementation selection end-to-end through collection, lookup, SDK validation, and generated deployment configuration.

The focused existing tests pass, but they do not cover checkpoint-targeted non-empty population, exact platform/version construction, runtime-path identity, or EMPIRICAL source provenance. Please add those guards with the fixes.

aic-core/rust/aiconfigurator-core/src/operators/msa.rs
tianhaox: Confirmed and fixed (d845dbf6 lineage): the MSA op grouped Empirical with Hybrid — EMPIRICAL now routes straight to the XOP transfer (matching the DSA op and the README/CLI mode contract), with a Rust oracle test pinning the full mode/source matrix for context and generation (silicon-injected: SILICON/HYBRID -> Source::Silicon; EMPIRICAL -> Source::Empirical with xop provenance and a value distinct from the injected silicon point).

collector/sglang/collect_msa_module.py
tianhaox: Resolved by withdrawal (owner decision): the l40s/rtx SGLang MSA tables are removed from the PR, the collector escape hatch is dropped (attention_backend stays framework-selected; CC-8.9/12 cases now fail classified — the honest unsupported signal, with the crash chain documented at the construction site), registry/docstring claims narrowed, and the backend-facts registry regenerated. Re-collect when SGLang grows the CC-8.9/12 branches in its M3 server-args handler.

collector/cases/models/MiniMaxM3ForCausalLM_cases.yaml
tianhaox: Per owner decision the implementation="msa" selection will be carried end-to-end: the generator will emit the knob for MiniMax-M3 x trtllm on the SM100 family so the collected rows represent exactly the prescribed deployment (no new data axis needed — the prescription resolves the cell). That plumbing is generator-module work landing in the next push on this PR; the collection-side authorization and citations stay as documented.

src/aiconfigurator/generator/config/backend_version_matrix.yaml
tianhaox: Fixed: rtx now ships per-family 1.3.0rc23 reuse markers (rc20 donors, 8 families; comm deliberately excluded — topology-bound per the R3 data-check rule, and the msa family carries none). The exact-version gate passes and the M3 MSA cell fails with a typed EmpiricalNotImplemented — an explicitly rejected cell rather than a version-gate exit (there is no DSA xop donor on rtx/trtllm, so the earlier "falls back to HYBRID" claim was wrong and has been corrected in the registry comment). Regression test added: the rc23 view loads and the M3 query raises the typed error.

collector/trtllm/collect_msa_module.py
tianhaox: Fixed: both MSA precision grids are now DECLARED in cases/base_ops/mla_module.yaml (attention_types [msa]; the pre-existing MLA/DSA combos are scoped accordingly) and both collectors consume get_mla_module_precision_specs. The per-SM yield is verified identical to the previous in-code grids (no data-identity change), and a declared-vs-emitted equality test covers trtllm/vllm across SM 89/90/100/103/120 for both phases. On your SM89 example specifically: the declared trtllm MSA policy is bf16 + fp8_block(min_sm 89), bf16-KV only; vLLM pairs every gemm tier with both KV dtypes with no SM floor on fp8-KV (serving citation on the combo).

collector/trtllm/collect_msa_module.py
tianhaox: Same disposition as the collection-meta thread: owner decision is to carry implementation="msa" end-to-end via generator emission for M3 x trtllm x SM100-family (next push). The rows then describe the prescribed deployment; MsaTable needs no implementation axis because the generator always prescribes the collected configuration for those cells.

collector/sglang/collect_msa_module.py
tianhaox: Acknowledged as a real structural issue. Interim hardening already landed: the post-load chunk filter was replaced by per-shape execute-or-raise (a shape past the resolved chunk raises a classified error instead of vanishing), and the raised decode budget is now a dedicated MSA-scoped knob (generation.msa_max_tokens) so the queued grid is deterministic per SM. The full fix you describe — building and memory-filtering the exact inner manifest in get_*_test_cases(), attaching it to the outer task so task IDs / resume checkpoints / failure records bind the retained shapes, and executing-or-raising each — is agreed and in progress as the next commit on this PR (it reshapes the sglang task tuples, so it lands together with its population/resume tests).

src/aiconfigurator/generator/config/backend_version_matrix.yaml
tianhaox: Correction to the above after an owner review of the mechanism: the source resolver already inherits strictly-earlier same-backend sibling tables per family, so the per-family markers were over-declared. Now exactly ONE rc23 reuse marker ships (the mla family, matching its existing rc15->rc10 precedent) — it anchors the rc23 version root for the exact-version gate; every other family serves its rc20 tables through native sibling inheritance with per-source provenance. Verified: the rc23 view loads, a gemm query resolves via inheritance, and the M3 MSA cell still fails with the typed rejection (regression test unchanged).

collector/sglang/collect_msa_module.py
tianhaox: Implemented in 6f70bac1: the getter now resolves each case's EXACT (b, s, prefix) manifest at plan time (sweep budgets + env pins + the sanctioned generation-time memory filter with counted drop logs; empty-manifest groups are logged and never queued) and attaches it to the case tuple — task IDs, resume checkpoints, and failure records bind the retained shapes. The worker executes exactly the manifest or raises per shape: the post-load KV-capacity filter is now a per-shape classified raise, and a worker invoked without a manifest raises instead of re-deriving the grid. The full-sweep CLI resolves the same manifest so reproductions run the planned shapes. Tests: deterministic binding (case manifest == re-resolved plan), budget adherence, and the no-rederive contract.

collector/trtllm/collect_msa_module.py
tianhaox: Implemented in ad2bbfa3: the module bridge sets ModelConfig.msa_sparse_implementation (keyed on the checkpoint ARCHITECTURE MiniMaxM3ForCausalLM and the system sm_version fact — 100/103 only, trtllm only), a trtllm rule maps it to a worker param, and a new versioned extra_engine_args.1.3.0rc23 template renders sparse_attention_config.implementation into the engine config. Generated M3 deployments on the SM100 family therefore run exactly the configuration the perf rows represent; MsaTable needs no implementation axis because the prescription resolves the cell. Rendering tests (block present for M3+SM100/103, absent for other models/SMs/backends) and bridge quadrant tests added.

tianhaox:
@Arsene12358 All six blockers are now resolved as of ad2bbfa3 — thank you for the unusually precise review. Disposition:

**Spec**
1. *EMPIRICAL returns silicon* — fixed: `Empirical` routes straight to the XOP transfer (matching the DSA op and the README/CLI contract); Rust oracle test pins the full mode/source matrix for context+generation.
2. *NVFP4 coverage not delivered* — the artifact now aliases onto the canonical MSA row (the module benchmark is artifact-insensitive: dummy weights, identical sparse geometry, precision as explicit sweep axes), so exact targeting populates non-empty plans for all three backends; per-backend end-to-end lane tests added (trtllm/vllm fp8_block SILICON hit; sglang typed SILICON miss + HYBRID transfer, claim narrowed).
3. *RTX/TRT-LLM rc23 unreachable* — one rc23 reuse marker (mla family, per its rc15→rc10 precedent) anchors the version root; other families serve rc20 via the resolver's native sibling inheritance. The M3 MSA cell now fails with a typed EmpiricalNotImplemented — an explicitly rejected cell (there is no DSA xop donor on rtx/trtllm; the earlier "falls back to HYBRID" claim was wrong and is corrected), with a CLI-path regression test.

**Standards**
1. *SGLang substituted backend* — resolved by withdrawal (owner decision): the l40s/rtx sglang MSA tables are removed, the escape hatch dropped (framework-selected default; CC-8.9/12 cases fail classified), claims narrowed, facts registry regenerated.
2. *Grouped work not bound* — the getter now resolves each case's EXACT inner-shape manifest at plan time (budgets + env pins + the sanctioned generation-time memory filter with counted drops) and attaches it to the case tuple; the worker executes exactly the manifest or raises per shape (post-load capacity check is a per-shape classified raise); binding/budget/no-rederive tests added.
3. *NVFP4 zero-case plan* — same fix as Spec-2, with empty-manifest groups logged and never silently queued.
4. *YAML precision policy* — both MSA grids are declared in mla_module.yaml (attention_types [msa]; MLA/DSA combos scoped) and both collectors consume get_mla_module_precision_specs; per-SM yield verified identical to the previous in-code grids, with a declared-vs-emitted equality test across SM 89/90/100/103/120.

**Cross-module** — the implementation selection is now carried end-to-end: the bridge prescribes `implementation=msa` for MiniMaxM3ForCausalLM on sm_version 100/103 (trtllm only), a rule maps it, and the new rc23 engine template renders `sparse_attention_config.implementation`, so default generated deployments run exactly the configuration the tables represent.

Full unit tree passes under xdist (3704), plus cargo, parity, ruff, and both data gates.

collector/trtllm/collect_msa_module.py
Arsene12358: Rechecked at `ad2bbfa3`: this fixes the optimized `cli default` path, but the public `cli generate` path still emits a deployment that does not match the measured rows.

Reproduction:

```bash
tmp=$(mktemp -d)
.venv/bin/aiconfigurator cli generate \
  --model-path MiniMaxAI/MiniMax-M3 \
  --total-gpus 8 \
  --system b200_sxm \
  --backend trtllm \
  --config-template-version 1.3.0rc23 \
  --save-dir "$tmp" \
  --no-color
rg -n 'msa_sparse_implementation|sparse_attention_config|implementation' "$tmp"
```

The command succeeds, but `rg` returns no matches. `_msa_sparse_implementation()` is called only from `task_config_to_generator_config()` in `module_bridge.py`; `build_naive_generator_params()` independently constructs `ModelConfig` in `naive.py:545-549` without the prescription. As a result, `cli generate` leaves TRT-LLM on its Triton default while the corresponding SM100/103 perf rows were measured with `implementation="msa"`.

| Current | Required |
|---|---|
| The optimized bridge derives the MSA prescription; naive generation bypasses it. | Put the architecture/backend/SM prescription behind one shared helper used by both optimized and naive generation, preserving explicit user-override precedence. |
| Tests inject the field into the renderer and test the bridge helper separately. | Add an end-to-end `cli generate` regression that inspects the emitted engine config, with positive SM100/103 MiniMax-M3 × TRT-LLM cases and negative non-M3, non-TRT-LLM, and non-SM100-family cases. |

This remains a P1 deployment/data-identity blocker until both supported generator entry points prescribe the implementation represented by the tables.

collector/cases/models/MiniMaxM3ForCausalLM_cases.yaml
Arsene12358: Rechecked at `ad2bbfa3`: the alias fixes the zero-case population defect, but it does not substantiate the PR body's claim that MSA + MoE tables ship for “8 platforms × 3 frameworks, both checkpoints.”

An exact B200/SGLang/NVFP4/SILICON CLI run produced no configuration and reported 193,500 unanswered batch points. Direct `fp8_block` MSA queries show context and generation misses for every SGLang system in the advertised matrix and for all three RTX backends. The committed `test_nvfp4_checkpoint_lane_resolution_per_backend` explicitly expects SGLang SILICON to miss and HYBRID to estimate, so the test currently codifies behavior that contradicts the release claim.

| Current | Required |
|---|---|
| The PR advertises complete 8 × 3 × 2 measured coverage while tests accept missing SILICON cells. | Either collect the missing checkpoint/backend/system lanes, or replace the blanket claim with an explicit support matrix that marks SILICON misses and empirical-only/unsupported cells. |
| Coverage checks exercise one B200 context point and accept SGLang fallback. | For every cell that remains advertised, add both context and generation SILICON lookups plus an exact checkpoint CLI regression. Unsupported cells should be explicitly rejected and omitted from measured-coverage claims. |

The zero-population portion is resolved; the advertised NVFP4 support contract remains a merge blocker.

src/aiconfigurator/generator/config/backend_version_matrix.yaml
Arsene12358: The rc23 version-root behavior is fixed at `ad2bbfa3`: the exact RTX/TRT-LLM command now loads `1.3.0rc23` and reaches the intended `EmpiricalNotImplementedError` because no DSA donor exists. Two acceptance gaps remain.

First, the PR body still says RTX/TRT-LLM MSA “fall[s] back to the HYBRID path” and that RTX SGLang MSA is included. Runtime deliberately rejects the TRT-LLM cell, and the SGLang tables were withdrawn. Second, `test_rtx_trtllm_rc23_loads_and_m3_is_explicitly_rejected` invokes `_engine_query` directly and catches generic `Exception`; it neither asserts the typed exception class nor covers the exact CLI path requested in the original finding.

| Current | Required |
|---|---|
| Release text promises fallback and included SGLang data. | State that RTX/TRT-LLM and RTX/SGLang MSA are unavailable in this PR, and that HYBRID produces a typed rejection because no empirical donor exists. |
| A direct op test catches `Exception` by message. | Assert `EmpiricalNotImplementedError` explicitly and add an exact CLI regression proving rc23 passes the version gate before reaching that rejection. |

The original runtime defect is repaired, but the thread should remain open until the public contract and its regression coverage match the implemented behavior.

Arsene12358 (changes requested):
Requesting changes on current head `ad2bbfa3`. The earlier EMPIRICAL routing, SGLang withdrawal, grouped-manifest, and YAML precision findings are resolved. CI is green, but three reviewed contracts remain incomplete.

1. `cli generate` still bypasses the new MiniMax-M3/TRT-LLM/SM100-family prescription. `cli default` emits `sparse_attention_config.implementation: msa`; the public naive generator emits no implementation block and therefore deploys the TRT-LLM default that the perf rows do not represent.
2. The PR still advertises MSA + MoE support for 8 platforms × 3 frameworks × both checkpoints, while SGLang NVFP4 SILICON misses across the matrix and all tested RTX NVFP4 MSA paths miss. The added test explicitly accepts one of those missing cells via HYBRID estimation.
3. RTX rc23 now loads and reaches the intended typed no-DSA rejection, but the PR body still promises HYBRID fallback and included SGLang MSA data. Its regression test catches generic `Exception` through a direct op query rather than protecting the exact CLI/typed contract.

Requested fixes are detailed in the existing threads: share the MSA implementation prescription across optimized and naive generator entry points with an end-to-end `cli generate` test; either ship the advertised NVFP4 cells or publish the exact supported matrix with unsupported cells explicitly identified; and align the RTX release text plus regression test with the implemented typed rejection.

tianhaox:
@Arsene12358 All three items from 4972622548 are addressed as of 1758d2df:

1. **Naive entry point** — the prescription decision moved to `generator/utils.msa_sparse_implementation`, shared by the module bridge AND `build_naive_generator_params`, so `cli generate` deployments for MiniMax-M3 on the SM100 family now carry `sparse_attention_config.implementation: msa` identically to the optimized path (naive-path test: b200 emits, h200 does not).
2. **Support claim** — the PR body now carries an exact per-cell matrix (S = own silicon / H = typed miss + HYBRID transfer / R = explicitly rejected) for both checkpoints across all 8 platforms x 3 frameworks, including the NVFP4 fp8_block-lane behavior you probed: sglang = H everywhere (its quantized-M3 flow is MXFP8, unsupported in serving; tables bf16-gemm by declaration), rtx/vllm = H (DeepGEMM SM120 classified failures), rtx/trtllm = R. Every H/R cell maps to a documented framework or infrastructure gap.
3. **RTX release text + regression** — the body no longer promises a HYBRID fallback or SGLang MSA data on rtx/l40s, and the regression test pins the exact typed contract: the miss crosses the engine FFI as `EmpiricalNotImplementedError` (no generic-Exception catch).

Full unit tree under xdist: 3705 green.

collector/trtllm/collect_msa_module.py
Yiming992: Rechecked at `c1f932ae`: the deployment/data-identity defect is fixed. The prescription now comes from the shared `generator.utils.msa_sparse_implementation` helper and is consumed by `build_naive_generator_params`. Exact public `cli generate` probes for BF16 and NVFP4 MiniMax-M3 on B200/TRT-LLM rc23 emitted both `generator_config.yaml: msa_sparse_implementation: msa` and `agg_config.yaml: sparse_attention_config.implementation: msa`; H200, non-TRT-LLM, and non-M3 negative quadrants omitted the setting.

The committed tests cover the previously broken naive-builder handoff and semantic artifact rendering. A full CLI/golden regression would be useful defense in depth, but its absence is not a merge blocker now that the behavior and both relevant seams are covered. Resolving this thread.

collector/cases/models/MiniMaxM3ForCausalLM_cases.yaml
Yiming992: Rechecked at `c1f932ae`: the S/H/R format is the right way to narrow the claim, but the current matrix still misclassifies two NVFP4 MSA cells and overstates the shipped NVFP4 MoE data.

Direct context and generation probes classify both of these cells as **R**, not **H**:

- `l40s × sglang × NVFP4`: SILICON raises `PerfDataNotAvailableError`; HYBRID raises `EmpiricalNotImplementedError` because there is no DSA donor.
- `rtx_pro_6000_server × vllm × NVFP4`: SILICON raises `PerfDataNotAvailableError`; HYBRID raises `EmpiricalNotImplementedError` for the same no-donor condition. The leaf metadata also records that every `fp8_block` cell failed.

The exact M3-geometry parquet scan (`hidden=6144`, `inter=3072`, `topk=4`, `experts=128`, `moe_dtype=nvfp4`) also finds no NVFP4 MoE rows on H100, H200, or L40S for any backend. Shipped-row coverage is therefore 5/8 platforms for TRT-LLM, 5/8 for vLLM, and 4/8 for SGLang—not 8/8, 8/8, and 7/8. Transfer behavior can be described separately, but it should not be called shipped rows.

| Current | Required |
|---|---|
| The combined NVFP4 row labels SGLang **H** through L40S, and the RTX/vLLM cell **H**. | Split out L40S and label `l40s × sglang` **R**; label `rtx_pro_6000_server × vllm` **R**. |
| The MoE sentence claims NVFP4 rows on 8/8 TRT-LLM, 8/8 vLLM, and 7/8 SGLang platforms. | State the actual shipped-row coverage—5/8, 5/8, and 4/8—or provide a per-platform MoE matrix that distinguishes rows from transfer/rejection. |
| “Known gaps” says SM90 MSA data is absent and SM103/120/121 remain unverified. | Remove or update this stale line: H100/H200 MSA tables are present, and the current TRT-LLM/vLLM registries mark only SM121 unverified. |

The minimum merge fix is an accurate PR description; collecting new data is not required if unsupported and transfer-only cells are labeled honestly. A table-driven context+generation regression for the corrected R cells would be useful follow-up protection. This remains the sole merge blocker.

src/aiconfigurator/generator/config/backend_version_matrix.yaml
Yiming992: Rechecked at `c1f932ae`: the rc23 version-root and typed-rejection contract are fixed. The exact RTX/TRT-LLM command now loads `1.3.0rc23` and reaches `EmpiricalNotImplementedError` instead of failing at version selection, and the committed regression now asserts that exact exception type rather than catching generic `Exception`. The PR body also classifies the cell as **R** instead of promising HYBRID success.

An exact CLI-wrapper regression would improve integration coverage, but the direct regression already exercises version-root loading and the typed engine result, and the public CLI path has been manually verified. That remaining test gap is non-blocking. Resolving this thread.

Arsene12358:
Review update at `c1f932ae`: the `cli generate` MSA-prescription defect and the RTX/TRT-LLM rc23 version/typed-rejection defect are functionally fixed, so I have resolved those two threads. The missing full CLI-wrapper tests are useful follow-up coverage, not merge blockers.

One merge blocker remains: the revised NVFP4 support description still labels two rejected cells as HYBRID-capable and overstates shipped NVFP4 MoE rows. I posted the verified cell classifications, actual row counts, and minimum description-only fix in the [remaining review thread](https://github.com/ai-dynamo/aiconfigurator/pull/1507#discussion_r3813371142).

collector/cases/models/MiniMaxM3ForCausalLM_cases.yaml
tianhaox: All three points verified locally and confirmed correct — fixed at `19639ee0` (tests) plus a PR-description update:

1. **Both cells reclassified R.** My probes reproduce your classification exactly: `l40s × sglang` and `rtx_pro_6000_server × vllm` raise `PerfDataNotAvailableError` in SILICON and `EmpiricalNotImplementedError` ("no DSA util to transfer from") in HYBRID, for context and generation alike. The matrix now splits l40s out of the combined row and labels both cells **R** with the typed-failure detail. I also re-probed the surviving **H** cells (h100/h200 × sglang HYBRID) to confirm they still resolve through the DSA donor.
2. **MoE sentence corrected to shipped rows.** Parquet scan at the exact M3 geometry matches your counts: NVFP4 rows on 5/8 platforms for trtllm and vllm (b200/b300/gb200/gb300/rtx) and 4/8 for sglang — the `min_sm: 100` capability floor means no NVFP4 rows can exist on h100/h200/l40s. The description now states 5/8, 5/8, 4/8 and explicitly separates transfer-ladder behavior from shipped data.
3. **Stale known-gaps line replaced** with the current truth: H100/H200 (SM90) MSA tables are included; registry markers are `unverified_sms=(121,)` for trtllm/vllm and `(120,)` for sglang.

Also took your follow-up suggestion now rather than later: `19639ee0` adds a table-driven regression (`test_rejected_msa_cells_raise_typed_errors`) parametrized over the two corrected R cells, pinning both typed exceptions in both modes for context and generation.

Arsene12358: Rechecked at `19639ee0`: the substantive support-contract blocker is resolved. The NVFP4 matrix now correctly classifies `l40s × sglang` and `rtx_pro_6000_server × vllm` as **R**, the shipped MoE coverage is accurately separated from transfer behavior, and `test_rejected_msa_cells_raise_typed_errors` covers both cells, SILICON/HYBRID, and context/generation with the exact typed exceptions. The focused regression passes and all PR CI is green. Resolving this thread.

Three PR-description cleanups remain non-blocking: SGLang's MSA registry marker is `unverified_sms=(121,)`, not `(120,)`; the “msa-only/needs moe” sentence should acknowledge the shipped MoE rows; and the current B200 counts are vLLM `11712+8832`, TRT-LLM `5070+3864`, and SGLang `26933+1477`.

Arsene12358 (approved):
Approved at `19639ee0`. All previously blocking functional and support-contract findings are resolved: both generator entry points prescribe the measured TRT-LLM MSA implementation, database-mode semantics and RTX rc23 typed rejection are correct, grouped collection work is bound at plan time, and the NVFP4 S/H/R matrix now matches the verified context and generation behavior. The new typed-error regression passes, and all required CI checks are green.

The remaining PR-description inconsistencies noted in the resolved thread are non-blocking editorial cleanup; they do not change or overstate the executable support matrix.

PR #1510, opened 2026-08-09
feat(sdk): self-contained result rows + closed-loop latency refinement (report + opt-in SLA)

tianhaox:
## What

Three pieces; default results unchanged (the third is opt-in):

**1. Result rows become self-contained.** `run_agg` records the operating-point step timings it already computes into the summary row (`ColumnsAgg`: `mix_step_ms`, `genonly_step_ms`, `prefill_step_ms`, `num_mix_steps`, `num_genonly_steps`). Disagg rows record `(p)prefill_step_ms` — the raw solo context latency of one prefill worker, captured *before* the autoscale TTFT pre-correction.

**2. A post-processing refinement tier** (`sdk/closed_loop_ttft.py`):

```python
out = refine_closed_loop_latency(df)   # df in, df out — consumes only the row's own columns
# adds ttft_refined / tpot_refined / throughput_refined; legacy columns untouched
```

Under the hood: a deterministic replay of the scheduler's own budget arithmetic at pass granularity — the fused continuous-batching calendar for agg rows, the P/D tandem (round-robin router, whole-prompt prefill batches, decode-attach handoff) for disagg rows. No RNG, no fitted constants, ~ms per row. The three refined columns are one consistent timeline (they satisfy the closed-loop identity `C/X = TTFT + (osl−1)·TPOT`), so consume them as a set.

**3. Opt-in refined-SLA enforcement in the sweep** (`Task(refined_sla=True)`, default off): `sweep_agg`/`sweep_disagg` compare the TTFT/TPOT targets against the refined steady closed-loop values instead of the fixed-factor-corrected ones. Agg candidate points gate on the solo chunked-prefill lower bound and are then priced; disagg drops the 1.8 prefill gate to the solo bound (a true necessary condition) and knee-tunes the top rate-matched combinations per decode-parallel category — a binary search for the largest compliant concurrency (never above the decode ladder's memory-validated point), with the decode step iterated to its occupancy-consistent value on the decode ladder (Little's law fixed point). Priced rows are repriced from the tandem, retiring the flat rate-match derate for refined rows. Post-processing helpers `filter_closed_loop_sla` / `pick_under_closed_loop_sla` cover the same semantics outside the pipeline. Pricing is memoized; overhead on an 8-GPU Qwen3-32B sweep: agg +~2s, disagg ~2min.

## Usage & cost

**Default: everything off.** Without touching anything, the only change is bookkeeping — result rows carry the extra step-timing columns; sweep results, SLA filtering, picking and runtimes are byte-identical to main (locked by the parity/unit suites).

**Tier 1 — post-processing (call it when you want it):**

```python
out  = refine_closed_loop_latency(df)                    # adds *_refined columns
ok   = filter_closed_loop_sla(df, ttft_ms=500)           # post-filter on refined values
best = pick_under_closed_loop_sla(df, ttft_ms=500)       # re-pick per deployment
```

Cost: ~ms per row (pure-Python deterministic replay; no DB or engine calls).

**Tier 2 — in-sweep enforcement:** `Task(refined_sla=True)` — or `refined_sla: true` in an experiment YAML (maps through `Task.from_yaml`). Deliberately NOT applied to `pick_autoscale` (the dynamo-planner profiling flow: its 1.8 is autoscaling headroom, a different semantic) or AFD frames (pass through unpriced pending their own validation). Measured on an 8-GPU Qwen3-32B sweep (isl4096/osl256, ttft 2000ms / tpot 30ms, H20 perf data, warm DB):

All warm-process numbers (cold adds a one-time ~3-6s perf-DB load either way); pricing is memoized process-wide, so repeated sweeps over the same operating points pay it once:

| sweep | default | refined_sla=True |
|---|---|---|
| agg, single SLA point | ~0.5s | +<1s |
| agg, pareto tpot sweep | ~1s | +~3.6s first run, +0.2s cached |
| disagg, single SLA point | ~0.4s | ~3.2s |
| disagg, pareto tpot sweep | ~2.9s | ~58s |
| single-point estimate (`run_single_agg`/`run_single_disagg`) | ms | +ms (headline repriced to the closed-loop values; `run_single_disagg`'s default output previously reported the raw solo TTFT with no correction at all) |

The disagg pareto cost is the concurrency knee search: each latency-target pair prices its own rate-matched candidates through the tandem (~1.8k pricings at 4–235ms each; operating-point results are memoized). Single-SLA-point tasks — the common "give me a deployment for this target" call — stay at seconds.

## Why

The heuristic TTFT corrections on result rows are clamped/fitted and miss the real shape of closed-loop queueing — which is not even monotone in concurrency (replacement prefill chunks stretch passes; past the knee, larger decode batches eat the cycle and TTFT recedes). A recursion reproduces this because both effects emerge from iterating the budget arithmetic; a multiplier cannot. Derivation and scope notes: `docs/design/closed_loop_refinement.md`.

## Validation (vs the full queueing-model evaluator, identical timing; Qwen3-32B, H20, isl4096/osl256)

| arm | result |
|---|---|
| agg estimator, exact timing callables, C=2..64 | TTFT/TPOT/throughput within **0.2%** (bit-identical ≤C=8) |
| agg estimator fed only recorded row scalars | within **~3%** over the same sweep |
| disagg tandem vs disagg evaluator (1P1D/2P1D/2P2D, serial & batched prefill, shallow→deep queueing) | **bit-identical** TTFT & throughput at all nine points |

## Validation against real serving

Same estimators, nine closed-loop operating points against `trtllm-serve` 1.3.0rc20 (Qwen3-32B tp4, H20, chunked prefill ON, isl 1024/4096/8192 x C 4/16/48; predictions computed from the perf DB before the measurements ran):

| isl/osl | C | throughput est vs real (d%) | TTFT p99 d% | TPOT d% |
|---|---|---|---|---|
| 1024/128 | 4 | 2.68 vs 2.57 (+4.4%) | +2.4% | -13.5% |
| 1024/128 | 16 | 5.07 vs 4.81 (+5.3%) | +8.3% | -10.8% |
| 1024/128 | 48 | 6.16 vs 3.97 (+55%) | +3.8% | -49% |
| 4096/256 | 4 | 0.95 vs 0.91 (+4.5%) | -0.3% | -2.6% |
| 4096/256 | 16 | 1.39 vs 1.39 (+0.1%) | +29.5% | -4.5% |
| 4096/256 | 48 | 1.61 vs 1.29 (+24.3%) | -0.3% | -18.5% |
| 8192/128 | 4 | 0.69 vs 0.70 (-0.6%) | +2.4% | +2.7% |
| 8192/128 | 16 | 0.78 vs 0.81 (-4.2%) | +22.6% | +5.7% |
| 8192/128 | 48 | 0.81 vs 0.82 (-0.5%) | +30.6% | +7.5% |

Reading: throughput within 5.3% on 7/9 points; the two misses (1024 C48, 4096/256 C48) trace to perf-DB timing in deep-churn regimes (real mixed passes cost up to ~2x the DB pricing; the same gap appears if you price those steps standalone), not to the recursion — refined TTFT errs on the conservative side there. The measured closed-loop TTFT correction factor (steady TTFT / solo no-queue TTFT) spans **2.8-14.4x** across this small grid and is non-monotone in concurrency — which is why the refinement computes it from the operating point instead of applying any fixed multiplier.

## 8xH20 A/B arbitration: legacy pick vs refined pick

For the same task (8 GPUs, isl4096/osl256, TTFT<=2000ms, TPOT<=30ms) the legacy sweep picks 1P(tp4)+1D(tp4)@C5; `refined_sla=True` picks 3P(tp2)+1D(tp2)@C16. Both were deployed on 8xH20 (TRT-LLM native disagg, ctx unchunked bs1, gen bs64; predictions frozen before measuring; steady-state protocol: 20 generations/slot, first 5 and last 1 discarded — the synchronized first round costs 1.5-5x the steady TTFT and multi-worker round-robin needs ~5 generations to relax).

Arm A — the legacy pick's topology at three concurrencies the legacy model prices identically (solo x1.8 = 1077.5ms):

| C | real TTFT | refined pred | legacy pred |
|---|---|---|---|
| 5 | 629 | 737 (+17%) | 1077.5 (+71%) |
| 9 | 2828 | 3132 (+11%) | 1077.5 (-62%) |
| 13 | 5065 | 5526 (+9%) | 1077.5 (-79%) |

Arm B — the refined pick's topology:

| C | real TTFT | refined pred | real TPOT (pred) | real X (pred) |
|---|---|---|---|---|
| 16 (pick) | **1201 <= SLA** | 1596 (+33%, conservative) | 17.17 (17.62) | 2.867 (2.713) |
| 19 | 2239 — violates, as predicted | 2731 (+22%) | 17.21 | 2.863 |

Real-vs-real: the refined pick serves **91.8 vs 57.1 tok/s/gpu (+61%) under the same TTFT SLA** — the legacy 1.8x gate had pre-killed the tp2-prefill family (solo x1.8 = 2070 > 2000) whose true closed-loop TTFT at the knee is 1201ms. X(C16)=X(C19) confirms the knee plateau: concurrency past the knee buys zero throughput and doubles TTFT.

## Rust/Python parity

No parity-governed path is touched (`operations/**`, `perf_database.py`, `perf_interp/**`, `engine.py`, `rust_engine_step.py` all unchanged). The recorded scalars are the FFI-returned values — identical whichever engine step computed them — and the estimators live in the Python orchestration layer, which stays Python by the dedup plan. `test_rate_match_parity` extended coverage passes.

## Tests

1004 sdk unit tests green (13 new: estimator semantics incl. the closed-loop identity, P/D imbalance signatures, chunked-off admission, DataFrame refine with NaN tolerance, rate-match parity).

## Summary by CodeRabbit

- **New Features**
  - Added closed-loop latency estimation for aggregate and disaggregated serving configurations.
  - Added refined TTFT, TPOT, and throughput metrics based on recorded operating-point measurements.
  - Added modeling for batching, queueing, prefix caching, chunked prefill, worker scaling, and handoff delays.
  - Results now include detailed prefill, generation, mixed-step timing metrics, and step counts.

- **Bug Fixes**
  - Preserved raw prefill-step latency alongside corrected TTFT values.

- **Documentation**
  - Added guidance on closed-loop refinement, validation, and supported scenarios.

### DIFF
--- a/docs/design/closed_loop_refinement.md
+++ b/docs/design/closed_loop_refinement.md
@@ -58,26 +58,14 @@ H20, isl4096/osl256):
   and batched prefill, shallow to deep queueing): bit-identical TTFT and
   throughput at all nine validation points.
 
-Against the dynamo mocker (independent discrete-event serving simulation,
-wall clock, same perf-DB timing; 31 closed-loop operating points: 9 agg tp4
-+ 22 disagg across 1P1D/2P1D/1P2D, isl 1024-8192, osl 128-2048, C 4-48; the
-16 extended-disagg predictions were generated before their measurements):
-
-- throughput: within 1.3% on agg and the extended disagg grid (3.8% on the
-  first disagg matrix);
-- TTFT mean: within ~11% on 25 of 31 points; TPOT within ~10% everywhere;
-- structural cross-checks confirmed by the DES: prefill-saturated regime
-  shift (TTFT recedes past the knee while TPOT grows), and the closed-loop
-  conservation result that adding decode capacity to a prefill-bound tandem
-  leaves throughput unchanged while nearly doubling steady TTFT.
-
 Against real serving (trtllm-serve 1.3.0rc20, H20 tp4, chunked prefill,
-same 9 agg closed-loop points, predictions unchanged): throughput within
-5.3% on 7 of 9 points and TTFT p99 within a few percent at low/moderate
-concurrency; the residual splits cleanly into the structure component
-(bounded by the DES comparison above) and a perf-DB timing component that
-dominates only in deep-churn small-prompt corners (real mixed passes up to
-2x the DB pricing) — a timing-data issue, not an estimator issue.
+9 agg closed-loop points, isl 1024-8192, osl 128-256, C 4-48, predictions
+generated before the measurements): throughput within 5.3% on 7 of 9
+points and TTFT p99 within a few percent at low/moderate concurrency; the
+residual concentrates in deep-churn small-prompt corners where real mixed
+passes cost up to 2x the perf-DB pricing — a timing-data gap, not an
+estimator-structure one (refined TTFT errs on the conservative side
+there).
 
 Known error modes (measured, understood, not corrected for):
 
--- a/docs/design/closed_loop_refinement.md
+++ b/docs/design/closed_loop_refinement.md
@@ -89,13 +89,15 @@ estimator does not model. Open-loop arrivals, shape distributions,
 prefix-cache dynamics and latency *distributions* (p50/p99 structure,
 transients) belong to the full queueing-model evaluator work.
 
-## Relation to SLA filtering (non-goal here)
+## Relation to SLA filtering
 
-The refined columns are report-only: sweep, SLA-target filtering, pareto
-and picking all still read the legacy `ttft`/`tpot` columns, so results
-are unchanged by this PR. If refined values should ever drive SLA
-filtering, the wiring is: refine before the filter and compare
-`ttft_refined`/`tpot_refined` against the targets, behind an opt-in flag
-(default off) — the refined steady closed-loop TTFT at the row's
-operating point is the quantity an SLA on served latency actually
-constrains. Left to a follow-up change.
+The pipeline is untouched: sweep, SLA-target filtering, pareto and picking
+all still read the legacy `ttft`/`tpot` columns, so default results are
+unchanged by this PR. For consumers who want the SLA verdict on the
+values a fixed-concurrency benchmark would actually measure,
+`filter_closed_loop_sla(df, ttft_ms=..., tpot_ms=...)` is an approximate
+post-filter: it refines the rows and drops those whose refined values
+exceed the targets (rows it cannot price are kept — it only ever tightens
+the legacy verdict where it has evidence). Moving this comparison into
+the pipeline filter itself (replacing the legacy columns) is a product
+decision left to a follow-up.
--- a/docs/design/closed_loop_refinement.md
+++ b/docs/design/closed_loop_refinement.md
@@ -104,10 +104,23 @@ post-filtering a picked table can return empty even when a
 lower-concurrency point of the same deployment complies;
 `pick_under_closed_loop_sla(df, ...)` closes that gap — feed it the full
 per-operating-point summary (sweep with a loose TTFT target) and it
-re-picks each deployment's best surviving row, which is equivalent to
-enforcing the SLA at the original filter. Moving this comparison into
-the pipeline filter itself (replacing the legacy columns) is a product
-decision left to a follow-up. AFD frames pass through unpriced (refined
-NaN, legacy verdict kept): their closed-loop mapping — the mini tandem
-over `prefill_t_step`/`decode_t_step` — needs its own validation pass
-first.
+re-picks each deployment's best surviving row.
+
+For full in-pipeline enforcement, `Task(refined_sla=True)` (default off)
+makes the sweep itself compare the targets against the refined values:
+agg candidate points gate on the solo chunked-prefill lower bound (the
+static TTFT is NOT a bound — deep saturation moves cycle time into TPOT,
+so gating on it would discard compliant points) and are then priced;
+disagg drops the 1.8 prefill gate to the solo bound and keeps, per
+decode-parallel category, the best rate-matched combination whose
+refined values comply. Pricing is memoized on the operating-point
+inputs. Two effects observed on an 8-GPU Qwen3-32B sweep
+(isl4096/osl256, ttft 2000ms): points the fixed factor wrongly killed
+come back (disagg: a 3-worker tp2 prefill pool, solo x1.8 = 2070ms >
+target but refined 1145ms, +57% tokens/s/gpu over the legacy pick; agg:
+deep-saturation rows with static TTFT above target but refined below,
++6% on the top pick), and rate-matched combinations whose refined TTFT
+explodes under queueing are dropped. AFD frames pass through unpriced
+(refined NaN, legacy verdict kept): their closed-loop mapping — the mini
+tandem over `prefill_t_step`/`decode_t_step` — needs its own validation
+pass first.
--- a/docs/design/closed_loop_refinement.md
+++ b/docs/design/closed_loop_refinement.md
@@ -47,6 +47,14 @@ effects emerge from iterating the budget arithmetic.
 
 ## Validation
 
+The figures below are one-off measurements (August 2026, Qwen3-32B,
+H20 trtllm perf data, isl4096/osl256), recorded with their configuration
+rather than reproduced in CI; the evaluator comparisons ran against the
+gated evaluator on the queueing-model branch. The in-repo unit tests
+(`tests/unit/sdk/test_closed_loop_ttft.py`) lock the estimator semantics
+these results rely on (closed-loop identity, P/D imbalance signatures,
+admission rules, DataFrame refinement).
+
 Against the full queueing-model evaluator on identical timing (Qwen3-32B,
 H20, isl4096/osl256):
 
--- a/docs/design/closed_loop_refinement.md
+++ b/docs/design/closed_loop_refinement.md
@@ -1,0 +1,63 @@
+# Closed-loop latency refinement (post-processing)
+
+Sweep result rows carry `ttft`/`tpot` computed with per-operating-point
+heuristics (agg: a clamped queuing factor; disagg: a constant prefill
+pre-correction). This feature makes result rows **self-contained** and adds
+a post-processing step that replays the scheduler's own arithmetic to
+produce quantitatively refined latencies — without touching the prediction
+pipeline.
+
+## Recorded columns (additive)
+
+Agg rows (`ColumnsAgg`): `mix_step_ms`, `genonly_step_ms`,
+`prefill_step_ms`, `num_mix_steps`, `num_genonly_steps` — the
+operating-point step timings `run_agg` already computes (through the
+compiled engine step when routed there; the values are the FFI-returned
+scalars, identical under either engine, so no Rust-side change is needed
+and no parity-governed path is touched).
+
+Disagg rows (`ColumnsDisagg`): `(p)prefill_step_ms` — the raw solo context
+latency of one prefill worker, captured *before* the autoscale TTFT
+pre-correction is applied.
+
+## The estimators (`sdk/closed_loop_ttft.py`)
+
+`estimate_closed_loop_latency` replays the fused continuous-batching pass
+calendar (budget arithmetic per pass, admission-ordered chunking, closed
+loop with visibility delay) — a deterministic recursion, no RNG, no fitted
+constants, milliseconds per operating point. `estimate_disagg_closed_loop_latency`
+replays the P/D tandem (whole-prompt prefill batches behind a round-robin
+router, decode-attach handoff, per-iteration decode). Timing enters as the
+recorded row scalars.
+
+`refine_closed_loop_latency(df)` dispatches per row shape and returns a
+copy with additive `ttft_refined` / `tpot_refined` / `throughput_refined`
+columns. The three are one consistent timeline — they satisfy the
+closed-loop identity `C/X = TTFT + (osl-1) * TPOT` — so consume them as a
+set; mixing a refined column with a legacy one breaks that conservation.
+Rows that cannot be priced keep NaN.
+
+## Why a recursion instead of a formula
+
+Closed-loop TTFT under continuous batching is not monotone in concurrency
+(replacement prefill chunks stretch everyone's passes; past the knee,
+larger decode batches eat the cycle and TTFT recedes), and any closed-form
+multiplier misses that shape. The recursion reproduces it because both
+effects emerge from iterating the budget arithmetic.
+
+## Validation
+
+Against the full queueing-model evaluator on identical timing (Qwen3-32B,
+H20, isl4096/osl256):
+
+- agg estimator, exact timing callables: TTFT/TPOT/throughput within 0.2%
+  over C=2..64 (bit-identical up to C=8);
+- agg estimator fed only the recorded row scalars: within ~3% over the
+  same sweep (the linear-in-tokens chunk reconstruction);
+- disagg tandem estimator vs the disagg evaluator (1P1D/2P1D/2P2D, serial
+  and batched prefill, shallow to deep queueing): bit-identical TTFT and
+  throughput at all nine validation points.
+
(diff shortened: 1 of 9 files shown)


docs/design/closed_loop_refinement.md
```diff
@@ -0,0 +1,63 @@
...
+Against the full queueing-model evaluator on identical timing (Qwen3-32B,
+H20, isl4096/osl256):
+
+- agg estimator, exact timing callables: TTFT/TPOT/throughput within 0.2%
+  over C=2..64 (bit-identical up to C=8);
+- agg estimator fed only the recorded row scalars: within ~3% over the
+  same sweep (the linear-in-tokens chunk reconstruction);
+- disagg tandem estimator vs the disagg evaluator (1P1D/2P1D/2P2D, serial
+  and batched prefill, shallow to deep queueing): bit-identical TTFT and
+  throughput at all nine validation points.
```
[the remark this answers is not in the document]
tianhaox: Done in b6b069e1 — the validation figures are now marked as one-off measurements (August 2026, Qwen3-32B / H20 trtllm perf data, isl4096/osl256; evaluator comparisons against the gated evaluator on the queueing-model branch), and the section references the in-repo unit tests that lock the semantics these results rely on.

src/aiconfigurator/sdk/closed_loop_ttft.py
```diff
@@ -0,0 +1,358 @@
...
+    if concurrency < 1 or isl < 1 or osl < 1:
+        raise ValueError("concurrency, isl and osl must be >= 1")
+    if not 0 <= prefix < isl:
+        raise ValueError("prefix must be in [0, isl)")
+    eff_isl = max(1, isl - prefix)
+
+    # slot = [remaining_prefill, generated, arrival_ms, first_token_ms,
+    #         last_token_ms, gap_sum, gap_count, gaps_list_or_None]
+    def new_slot(arrival: float) -> list:
+        return [eff_isl, 0, arrival, -1.0, -1.0, []]
```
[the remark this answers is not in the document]
tianhaox: Fixed in b6b069e1 — comment now lists the six actual fields.

src/aiconfigurator/sdk/inference_session.py
```diff
@@ -270,6 +270,7 @@ def _get_disagg_summary_df(
         Get the disagg summary df based on prefill and decode summary df
         """
         prefill_dict = prefill_summary_df.iloc[0].to_dict()
+        prefill_dict["prefill_step_ms"] = prefill_dict["ttft"]  # raw solo, pre-correction
         prefill_dict["ttft"] = prefill_dict["ttft"] * _AUTOSCALE_TTFT_CORRECTION_FACTOR
```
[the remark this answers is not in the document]
tianhaox: Good catch — fixed in b6b069e1 at all four stash sites (both DisaggInferenceSession paths, picking.find_best_autoscale flow, and sweep._find_best_disagg_under_constraint): `prefill_step_ms` now subtracts the colocated `encoder_latency` so the estimator prices the raw context step. Text-only rows carry encoder_latency=0, so existing numbers are unchanged (full sdk suite green).

src/aiconfigurator/sdk/picking.py
```diff
@@ -119,6 +119,7 @@ def _build_disagg_summary_dict(
         "tokens/s/gpu": tokens_s_gpu,
         "tokens/s/user": decode_summary_dict["tokens/s/user"],
         "(p)seq/s/worker": prefill_summary_dict["seq/s"],
+        "(p)prefill_step_ms": prefill_summary_dict.get("prefill_step_ms"),
```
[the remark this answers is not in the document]
tianhaox: Partially adopted in b6b069e1: both builders now coerce to `float(...)` when the field is present, keeping the column numeric. The absent case intentionally stays `None` rather than `float("nan")` — the rate-match parity test compares the two builders by equality, and NaN != NaN would fail it; `refine_closed_loop_latency` already fail-opens (NaN refined columns) on rows it cannot price, so downstream behavior is the same either way.

src/aiconfigurator/sdk/closed_loop_ttft.py
```diff
@@ -0,0 +1,515 @@
...
+                osl=int(row["osl"]),
+                prefill_workers=int(row["(p)workers"]),
+                decode_workers=int(row["(d)workers"]),
+                prefill_step_ms=float(row["(p)prefill_step_ms"]),
+                decode_step_ms=float(row["tpot"]),
+                prefill_batch=int(row["(p)bs"]),
+                decode_max_seqs=int(row["(d)bs"]),
+            )
+        else:
+            chunk_ref = max(1, min(int(row["ctx_tokens"]), int(row["isl"]) - int(row.get("prefix", 0) or 0)))
```
jasonqinzhou, line 408: **[P1] Do not apply the text-only replay to multimodal rows.** `BaseBackend.run_agg` prices prefill/decode at `text_isl + img_ctx_tokens` but stores only the text `isl` in the result row, while `prefill_step_ms` intentionally excludes the encoder phase; the disagg row likewise stores raw context time after subtracting `encoder_latency`. Both branches here then ignore `encoder_latency`, and the agg branch also reconstructs the pass count from the text-only length. A focused probe shows that changing a row from `encoder_latency=0` to `500` leaves the refined triple identical, and `reprice_closed_loop_row` overwrites the legacy TTFT with that lower value. With `refined_sla=True` this can admit a VLM deployment whose actual TTFT exceeds the target. Please either record/model the visual-token and encoder stages for both row shapes, or return NaN and preserve the legacy verdict for multimodal rows until that replay is supported; add aggregate and disaggregated VLM regressions.

src/aiconfigurator/sdk/closed_loop_ttft.py
```diff
@@ -0,0 +1,515 @@
...
+            step = float(row["prefill_step_ms"])
+            t_gen = float(row["genonly_step_ms"])
+            r = estimate_closed_loop_latency(
+                concurrency=int(row["bs"]),
```
jasonqinzhou, line 412: **[P1] Preserve deployment-wide aggregate throughput scaling.** `run_agg` models one local batch `b` and then scales both `concurrency` and output throughput by `pp * attention_dp_size`. Passing only `row["bs"]` here correctly replays one local scheduler for latency, but the returned `throughput_rps` is then published by `refine_closed_loop_latency` / `reprice_closed_loop_row` as deployment-wide throughput without restoring that scale. For a synthetic row with `bs=4` and `concurrency=8`, the closed-loop identity holds for `4 / X` but fails by 2x for the row's advertised `8 / X`; `run_single_agg(refined_sla=True)` correspondingly halves `seq/s` and `tokens/s`. Please scale the replay throughput by `concurrency / bs` (and account for that factor in or after the cache) while leaving per-replica latency unchanged, with a PP/attention-DP regression.

src/aiconfigurator/sdk/closed_loop_ttft.py
```diff
@@ -0,0 +1,515 @@
...
+        raise RuntimeError("did not converge within max_passes")
+
+    window = now - (steady_start or 0.0)
+    q = lambda v, x: sorted(v)[min(len(v) - 1, max(0, round(x * len(v)) - 1))] if v else float("nan")
+    return {
+        "ttft_steady_mean": sum(ttfts) / len(ttfts) if ttfts else float("nan"),
+        "ttft_steady_p50": q(ttfts, 0.5),
+        "ttft_steady_p99": q(ttfts, 0.99),
+        "ttft_transient_max": transient_max,
+        "tpot_mean": sum(tpots) / len(tpots) if tpots else float("nan"),
```
jasonqinzhou, line 157: **[P2] Define no-decode TPOT for `osl=1` instead of returning NaN.** `osl=1` is a supported no-decode workload with existing backend coverage, but neither estimator records an inter-token gap, so both return `tpot_mean=NaN`. The row still has a finite refined TTFT, which makes `_sweep_one_parallel_agg` treat it as priced and then reject it because `NaN <= tpot_target` is false; `reprice_closed_loop_row` also turns the otherwise valid request latency into NaN because `NaN * 0` remains NaN. A focused probe reproduced both behaviors. Please give the no-decode case an explicit finite TPOT/request-latency convention (for example TPOT 0), mirror it in the tandem estimator, and add aggregate/disaggregated `osl=1` tests through the refined-SLA and reprice paths.

src/aiconfigurator/sdk/closed_loop_ttft.py
```diff
@@ -0,0 +1,515 @@
...
+            t_gen = float(row["genonly_step_ms"])
+            r = estimate_closed_loop_latency(
+                concurrency=int(row["bs"]),
+                isl=int(row["isl"]),
+                osl=int(row["osl"]),
+                # linear-in-tokens reconstruction of the recorded
+                # per-chunk prefill cost; decode priced at the recorded
+                # operating-point iteration time
+                prefill_ms=lambda b, i, p, _s=step, _c=chunk_ref: _s * (b * max(1, i - p)) / _c,
+                decode_ms=lambda b, c, _g=t_gen: _g,
```
jasonqinzhou, line 419: **[P2] Honor aggregate speculative-decoding progress in the replay.** When `nextn_accepted > 0`, `run_agg` reduces `decode_iterations` and `num_genonly_steps` using `decode_tokens_per_iteration`; `genonly_step_ms` remains the cost of one verification iteration. This replay always advances one output token per `genonly_step_ms` and neither reads the recorded step counts nor carries `decode_tokens_per_iteration`, so an MTP row with fewer scheduler steps prices identically to a non-speculative row with the same scalars. That overstates refined TPOT/latency and can discard valid `refined_sla` points or corrupt the single-point headline. Please record the accepted-token progress in the self-contained row and advance the replay consistently (or fail open for such rows), with a `SpeculativeDecodingProfile` regression.

jasonqinzhou (changes requested):

## What this PR does

This PR makes aggregate and disaggregated result rows carry their operating-point timing inputs, adds deterministic closed-loop replay for refined TTFT, TPOT, and throughput, and optionally uses those values for sweep SLA enforcement and single-point headlines.

The implementation is unusually thoughtful about keeping the default path unchanged: the timing columns are additive, the replay is deterministic and memoized, the design notes distinguish the post-processing and in-sweep tiers, and the author backed the model with semantic tests plus queueing-evaluator and real-serving comparisons.

## Why this score

The current head is 5.8/10 with REQUEST_CHANGES. The text-only single-replica path is well structured and its submitted tests and CI are green, but two confirmed P1 contract breaks make the opt-in results unsafe for supported configurations: multimodal rows are repriced without encoder latency or visual context tokens, and aggregate refined throughput loses the pp/attention-DP deployment scale. Two narrower P2 gaps also remain: osl=1 produces a NaN TPOT that rejects valid refined-SLA rows, and aggregate speculative-decoding progress is ignored by the replay.

docs/design/closed_loop_refinement.md
```diff
@@ -0,0 +1,137 @@
...
+`refine_closed_loop_latency(df)` dispatches per row shape and returns a
+copy with additive `ttft_refined` / `tpot_refined` / `throughput_refined`
+columns. The three are one consistent timeline — they satisfy the
+closed-loop identity `C/X = TTFT + (osl-1) * TPOT` — so consume them as a
+set; mixing a refined column with a legacy one breaks that conservation.
+Rows that cannot be priced keep NaN.
+
+## Why a recursion instead of a formula
+
+Closed-loop TTFT under continuous batching is not monotone in concurrency
```
Spycsh, line 42: Does “closed-loop” here mean an infinite request stream with a fixed concurrency/BS, where whenever one request finishes, a new request immediately arrives to keep the number of active requests constant? If so, the arrival times are derived from the simulated completion times rather than being explicit inputs—is that the intended workload model?

tianhaox: yes. the closed loop refers to the concurrency based workload where a concurrency N is kept. The open loop is more like request rate based. there's no fixed budget of the slots for the concurrent requests.

PR #1513, opened 2026-08-10
feat(support-matrix): add current NVIDIA NVFP4 variants

jasonqinzhou:
## Summary

- register 12 current NVIDIA NVFP4 checkpoints in the default support-matrix roster
- bundle their Hugging Face configs and quantization metadata for offline loading
- preserve current `main`'s model topology and backend behavior while adding per-operation mixed-precision modes
- model Qwen3.6-35B routed experts as weight-only `W4A16_NVFP4` end to end across Python metadata, Rust evaluation, engine wire, memory, admission, and transfer paths
- honor checkpoint exclusions for attention/projection, dense-FFN, and shared-expert operations in Qwen, Gemma, and Kimi families
- keep data-less W4A16 GEMMs on the policy-gated `XPROFILE` transfer path instead of hard-aliasing INT4 silicon tables
- register artifact-specific Collector cases with exact geometry and quantization modes
- regenerate the 12-model support-matrix slice against the Rust-only evaluator

Linear: [AIC-1709](https://linear.app/nvidia/issue/AIC-1709/expand-default-support-matrix-coverage-for-current-nvidia-nvfp4)

This change is still needed after the Python-to-Rust migration: current `main` contains none of these 12 model artifacts and does not define the distinct `w4a16_nvfp4` precision. The branch now includes `main` at `163f6625`, including the Rust single-evaluator migration and the latest CI/sanity-check changes. Conflict resolution kept the Rust-backed operation/query path from `main`; no retired Python evaluation math was reintroduced. Python changes in this PR are limited to checkpoint metadata, model graph construction, and compatibility wrappers around the Rust engine.

## Matrix impact

| Metric | Current `main` | With #1513 | Delta |
| --- | ---: | ---: | ---: |
| Default support-matrix models | 59 | 71 | +12 |
| Checked-in CSV rows | 11,008 | 13,840 | +2,832 |
| Fresh-generation combinations | 6,962 | 8,378 | +1,416 |
| Fresh aggregate/disaggregate rows | 13,924 | 16,756 | +2,832 |

Current discovery produces 118 combinations and 236 aggregate/disaggregate rows per added model. The checked-in delta contains exactly 2,832 additions, 0 removals, 0 retained-model row changes, and 0 duplicate row identities across 10 systems.

Final Rust-evaluator status distribution for the 2,832 new rows:

| Status | Rows |
| --- | ---: |
| `PASS` (measured silicon) | 500 |
| `HYBRID_PASS` (policy-gated transfer) | 162 |
| `HW_INCOMPATIBLE` | 1,068 |
| `FRAMEWORK_INCOMPATIBLE` | 146 |
| `FAIL` | 956 |

## Rust-migration integration

1. Merged current `main` and retained its Rust-backed GEMM/MoE operation wrappers and query behavior.
2. Kept both post-migration `nvfp4_wo` and this PR's distinct `w4a16_nvfp4` modes, with append-only Rust wire enums.
3. Ported mixed-precision checkpoint exclusions onto the refactored DeepSeek and Gemma model builders, including the shared large-EP MoE builder.
4. Preserved Rust-side W4A16 XPROFILE admission without aliasing INT4 silicon data.
5. Refreshed all new-model matrix evidence because the Rust evaluator and current backend/version inventory changed the old status distribution and added 96 rows.
6. Resolved the latest Collector add/add conflicts by retaining the PR's mixed-precision checkpoint metadata, adopting `main`'s newer backend declarations, keeping Nano and Lightning as distinct Nemotron artifacts, and avoiding duplicate/mislabeled DeepSeek module rows.

## Validation

- freshly generated and schema-validated all 2,832 new rows on 31 existing Brev CPU workers against the Rust-migrated production code
- verified exact 12-model coverage, 236 rows per model, unique row identities, valid status/source pairings, and zero field-level drift from the checked-in PR rows
- Rust library: 498 passed, 1 ignored; public API crate: 3 passed
- exact-head sentinel/selector sanity-check, NVFP4 model, model-config, and Collector validation: 359 passed on Brev
- support-matrix e2e contract: 3 passed, 1 environment-gated skip
- merged Collector roster validation passed with 6,606 MoE cases and 8 DeepSeek-V4 mHC cases
- Ruff, Python compile checks, and `git diff --check` passed
- GitHub unit, e2e, parity, public-API, prediction-regression, wheel, Collector-data, perf-sanity, policy, DCO, and review checks passed on final head `a2f986a5`

Approved by Yiming (`@Arsene12358`); all review threads are resolved.


jasonqinzhou:
@coderabbitai review

aic-core/src/aiconfigurator_core/sdk/models/helpers.py
```diff
@@ -343,7 +350,11 @@ def _infer_quant_modes_from_raw_config(raw_config: dict, architecture: str | Non
 
     # DeepSeek-V4 native checkpoints use MXFP4 routed-expert weights with MXFP8
     # activations, while non-expert weights remain FP8 block quantized.
-    if architecture == "DeepseekV4ForCausalLM" and str(raw_config.get("expert_dtype", "")).lower() == "fp4":
+    if (
+        architecture == "DeepseekV4ForCausalLM"
+        and str(raw_config.get("expert_dtype", "")).lower() == "fp4"
+        and overrides.get("moe_quant_mode") != common.MoEQuantMode.nvfp4
```
jasonqinzhou, line 374: **[P1] Preserve the FP8-block non-expert lane for these DSV4 checkpoints.** Both new NVIDIA V4 configs retain `quant_method: "fp8"`, `weight_block_size: [128, 128]`, and exclude `*.attn.*` plus `*.ffn.shared_experts.*` from the NVFP4 group; only `quantized_layers[*].ffn.experts` are explicitly NVFP4. However, `_collect_mixed_precision_layer_algos()` also consumes the broad `targets: ["Linear"]` group without applying `ignore`, so `get_model()` resolves both the global GEMM mode and the DSV4 module GEMM key to `nvfp4` instead of `fp8_block`. The checked-in matrix then has no PASS/HYBRID row for either 228-row NVIDIA DSV4 slice. Please preserve FP8-block for non-expert GEMMs, set only the routed MoE lane to NVFP4, add a regression covering both modes, and regenerate the affected matrix rows.

aic-core/src/aiconfigurator_core/sdk/models/helpers.py
```diff
@@ -185,8 +185,10 @@ def _normalize_mixed_precision_layer_algo(value: object) -> str | None:
         "fp8": "fp8",
         "e4m3": "fp8",
         "e5m2": "fp8",
+        "mxfp8": "mxfp8",
         "nvfp4": "nvfp4",
         "fp4": "nvfp4",
+        "w4a16_nvfp4": "nvfp4",
```
jasonqinzhou: **[P1] Do not collapse weight-only NVFP4 into W4A4.** `W4A16_NVFP4` means FP4 weights with 16-bit activations; the bundled Qwen3.6-27B `group_1` confirms this by declaring 4-bit weights and no `input_activations`, and both Qwen hf quant artifacts label their FFN targets with this algorithm. AIC's `GEMMQuantMode.nvfp4` instead has `compute=4` and `compute_dtype="fp4"`, so this alias makes those FFNs query W4A4 NVFP4 data and feeds the generated support matrix the wrong kernel and memory model. Please represent W4A16_NVFP4 with weight-only/BF16-compute semantics (including its scale overhead), or fail closed until that mode exists, then test the Qwen op modes and regenerate their matrix rows.

aic-core/src/aiconfigurator_core/sdk/models/qwen35.py
```diff
@@ -6,7 +6,65 @@
...
+                continue
+            target_name = str(target).lower()
+            if ".linear_attn." in target_name or ".self_attn." in target_name:
+                modes["projection"] = mode
+            elif "shared_expert" in target_name:
+                modes["shared_expert"] = mode
+            elif ".mlp." in target_name and not _is_routing_expert_target(target_name):
+                modes["dense_ffn"] = mode
+
+    if modes["projection"] != default:
```
jasonqinzhou: **[P2] Preserve an explicit `fp8_static` global override as well.** This equality check cannot distinguish the checkpoint-inferred projection mode from a caller that explicitly supplied the same enum. With `ModelConfig(gemm_quant_mode=GEMMQuantMode.fp8_static)`, both Qwen3.6 models still rewrite their dense/shared FFN GEMMs to NVFP4, even though the nearby contract and existing BF16 test say a caller-selected global mode wins. Please carry whether `gemm_quant_mode` was originally unset (or pass an explicitness flag) instead of inferring intent from equality, and extend the override regression to the `fp8_static` case.

aic-core/src/aiconfigurator_core/model_configs/nvidia--Qwen3.6-35B-A3B-NVFP4_config.json
```diff
@@ -0,0 +1,1492 @@
...
+          "num_bits": 4,
+          "type": "float",
+          "group_size": 16
+        },
+        "weights": {
+          "dynamic": false,
+          "num_bits": 4,
+          "type": "float",
+          "group_size": 16
+        },
```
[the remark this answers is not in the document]
jasonqinzhou: Verified against the canonical NVIDIA Qwen3.6-35B artifact: the checked-in config matches the current upstream file byte-for-byte. This PR now interprets the paired per-layer W4A16_NVFP4 declarations with a distinct weight-only/BF16-compute mode instead of mutating the vendored metadata. The final Qwen tests and regenerated matrix exercise that semantic path.

aic-core/src/aiconfigurator_core/model_configs/nvidia--Qwen3.6-35B-A3B-NVFP4_config.json
```diff
@@ -0,0 +1,1492 @@
...
+    },
+    "ignore": [
+      "mtp.layers.0*",
+      "mtp*"
+    ],
+    "quant_algo": "MIXED_PRECISION",
+    "producer": {
+      "name": "modelopt",
+      "version": "0.37.0"
+    },
```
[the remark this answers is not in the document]
jasonqinzhou: Verified against the canonical upstream pair: both checked-in files match the current Hugging Face artifacts byte-for-byte, including the producer-version mismatch. I am retaining canonical vendor provenance rather than guessing which producer string to rewrite locally; runtime behavior is covered independently by the loader and quant-mode tests.

aic-core/src/aiconfigurator_core/model_configs/nvidia--Qwen3.6-35B-A3B-NVFP4_hf_quant_config.json
```diff
@@ -0,0 +1,1050 @@
+{
+  "producer": {
+    "name": "modelopt",
+    "version": "0.44.0"
+  },
+  "quantization": {
+    "quant_algo": "MIXED_PRECISION",
+    "kv_cache_quant_algo": "FP8",
```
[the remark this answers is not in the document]
jasonqinzhou: Confirmed the loader merges the paired hf_quant_config and resolves kv_cache_quant_algo=FP8 for this checkpoint. The checked-in pair matches the current upstream artifacts byte-for-byte, so no speculative vendor-file normalization was made. Offline quant inference and KV-mode regression coverage pass.

aic-core/src/aiconfigurator_core/sdk/models/helpers.py
jasonqinzhou: Fixed. DeepSeek V4 now preserves fp8_block for non-expert/module GEMMs and applies NVFP4 only to explicitly declared routed experts. The mixed-precision regression covers both V4 checkpoints, and the affected matrix rows were regenerated.

aic-core/src/aiconfigurator_core/sdk/models/helpers.py
jasonqinzhou: Fixed completely. W4A16_NVFP4 now has a distinct scale-aware weight-only/BF16-compute profile in Python and Rust, distinct cache/SOL identities, and a distinct engine-wire dtype. Borrowed INT4 utilization is permitted only as empirical HYBRID evidence; SILICON fails closed without a measured W4A16 lane. Final regeneration intentionally reclassifies 148 prior PASS rows to HYBRID_PASS with no PASS-to-FAIL or HYBRID regressions.

aic-core/src/aiconfigurator_core/sdk/models/qwen35.py
jasonqinzhou: Fixed. Task and ModelConfig now carry explicit GEMM-mode provenance through dataclasses.replace(), so explicit fp8_static and BF16 overrides suppress checkpoint mixed-precision splitting. The final follow-up also preserves this provenance through AFD combined-with-PD prefill; focused Qwen/Task coverage passes.

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -1542,6 +1549,8 @@ def build_model_config(self, *, role: Literal["agg", "prefill", "decode"]) -> co
             wideep_num_slots=self.wideep_num_slots,
             forward_model=self.forward_model or "op_level",
         )
+        model_config._gemm_quant_mode_is_explicit = self._gemm_quant_mode_explicit_by_role.get(role, False)
+        return model_config
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 8571e87b. AFD combined-with-PD now records the internal prefill role as explicit when it inherits an explicit agg GEMM override (or has its own explicit override). The new Qwen3.6 AFD regression builds the prefill model and verifies the FFN remains fp8_static; 146 focused Qwen/Task tests pass.

aic-core/src/aiconfigurator_core/sdk/config.py
```diff
@@ -16,6 +16,9 @@ class ModelConfig:
     tp_size: int = 1
     pp_size: int = 1
     gemm_quant_mode: common.GEMMQuantMode | None = None
+    # Internal provenance carried through dataclasses.replace() during sweeps.
+    # None preserves the legacy direct-caller rule (non-None mode is explicit).
+    _gemm_quant_mode_is_explicit: bool | None = field(default=None, repr=False, compare=False)
```
jasonqinzhou: **[P1] Keep the internal provenance field from shifting ModelConfig positional arguments.** Adding `_gemm_quant_mode_is_explicit` as a normal dataclass field inserts it between `gemm_quant_mode` and `moe_quant_mode`. Existing positional callers now silently assign each subsequent quant mode to the wrong field. Please make this private field keyword-only so `dataclasses.replace()` still carries it while preserving the established positional constructor order, and add a positional-construction regression test.

jasonqinzhou: Fixed in `efa41be4`: `_gemm_quant_mode_is_explicit` is now a keyword-only dataclass field, which preserves the legacy positional constructor order while keeping the field in `dataclasses.replace()`. Added a positional-construction regression test covering GEMM/MoE/KV-cache/FMHA modes. Verification: all 52 NVFP4 variant tests pass; pinned Ruff lint/format and `git diff --check` pass.

Arsene12358 (changes requested):
Requesting changes because the current head is both unmergeable with `main` and incorrect for several of the mixed-precision paths this PR is intended to add. I reproduced each blocking item below on head `efa41be4324c842320daa71a454ee34d505eb8d8`.

### 1. [P0] Rebase and semantically integrate the Qwen3.5 fixes from #1503

GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, and `git merge-tree --write-tree --messages origin/main HEAD` exits 1 with content conflicts in `sdk/common.py` and `sdk/models/qwen35.py`. The `common.py` conflict is a roster union, but `qwen35.py` cannot be resolved with ours/theirs: current `main` added backend-aware dispatch, TP-local GDN geometry, a separate BA GEMM, output gating/QK norm, attention-AR folding, gated shared-expert overlap, and corrected KV/GDN memory, while this branch adds projection/dense/shared quant splits to the older topology.

The semantic difference is observable at TP4 on this head: Qwen3.5-35B has `context_gdn_in_proj_gemm.n=3584`, no BA GEMM, GDN heads `16/32`, `context_qkv_gemm.n=1536`, `use_qk_norm=False`, and 20,480 KV elements/token. Current `main` tests require `3072`, a BA GEMM with `n=16`, TP-local heads `4/8`, the gated Q slice with QK norm, and 5,120 KV elements/token plus constant GDN state. This head also accepts `moe_backend="megamoe"`, which current `main` rejects.

| Current head | Required before merge |
| --- | --- |
| Older Qwen3.5 graph plus the new quant variables; choosing this side resurrects #1503 correctness bugs. | Start from current `main`'s constructor and graph, retain `backend_name` and all validation/memory/dispatch behavior, then add `raw_config` and `allow_checkpoint_split`. Thread the projection mode through the main qkvz, BA, and output projections; the dense mode through dense FFNs; and the shared mode through shared gate-up/down. Union the Step-3.7 and NVFP4 roster entries in `common.py`. Update `context_gdn_shared_up_gemm` in the new test to main's `context_gdn_shared_gate_up_gemm`, run both `test_qwen35.py` and `test_nvfp4_variants.py`, and regenerate the checked-in matrices from the rebased graph. |

### 2. [P1] Model Qwen3.6-35B routed experts as W4A16, not W4A4

The bundled authoritative per-layer metadata declares every `*.mlp.experts` lane as `W4A16_NVFP4`, but `helpers.py:307-310` gives a broad `nvfp4` group priority and resolves `model_config.moe_quant_mode` to `MoEQuantMode.nvfp4`. Runtime confirms `context_gdn_moe=nvfp4` with `compute=4` while shared FFNs are correctly `w4a16_nvfp4` with `compute=1`. The checked-in H100/H200 rows consequently all report `HW_INCOMPATIBLE` because the model is treated as requiring FP4 activation compute, even though weight-only W4A16 uses the BF16 compute lane.

| Current head | Required before merge |
| --- | --- |
| `*.mlp.experts: W4A16_NVFP4` becomes `MoEQuantMode.nvfp4` (9/16-byte weights, FP4 activation compute). The coarse test at `test_nvfp4_variants.py:70-74` codifies that mismatch. | Add a scale-aware weight-only NVFP4 MoE mode (9/16-byte weights, BF16 compute) across Python, Rust, wire conversion, memory/admission, and transfer handling. Give explicit per-layer routed-expert metadata precedence over broad groups, assert the routed MoE operation's mode, and regenerate the Qwen3.6-35B rows. |

### 3. [P1] Honor checkpoint exclusions for attention, dense, and shared operations

The PR summary promises per-operation metadata accuracy, but several bundled artifacts are still modeled uniformly from the global `nvfp4` enum. For example, Qwen3.5-122B's metadata ignores every `linear_attn`/`self_attn` and `shared_expert` path, yet the runtime graph assigns NVFP4 to `context_gdn_in_proj_gemm`, `context_qkv_gemm`, and `context_gdn_shared_up_gemm`. Gemma-4-26B ignores its attention and MLP paths but models `context_swa_qkv_gemm` and `context_swa_shared_mlp_gate_up_gemm` as NVFP4. Kimi-K2.6 ignores `shared_experts*` but models `context_shared_gate_up_gemm` as NVFP4. The tests at lines 83-100 assert only coarse global modes and therefore miss these operation-level contradictions.

| Current head | Required before merge |
| --- | --- |
| Broad `config_groups`/global modes are applied to operation categories that `ignore`/`exclude_modules` explicitly leave unquantized. | Parse the exclusion patterns into family-specific projection, dense-FFN, and shared-expert modes, use BF16 for excluded operation categories in Qwen/Gemma/Kimi graphs, and add per-operation assertions for each affected family. Regenerate every affected matrix slice after the graph modes are corrected. |

### 4. [P1] Register quant-distinct collector artifacts as model case rows

All 12 IDs were appended only to top-level `model_paths`; none appears in `model_case_values` or `framework_specific_model_case_values`. A targeted `COLLECTOR_MODEL_PATH=nvidia/Qwen3.6-35B-A3B-NVFP4` probe returns zero attention, GDN, GEMM, and MoE model profiles. That contradicts `collector/cases/README.md` and `.claude/rules/collector/case_authoring.md`, which require a separate model row plus `framework_quantization` for a new checkpoint quant artifact; a top-level path is only registration and does not attach exact geometry/quant policy.

| Current head | Required before merge |
| --- | --- |
| New IDs exist in the architecture roster but targeted generation has no artifact-specific case row. | Add exact per-op `model_case_values` entries for the new artifacts, using `model_aliases` only for proven shape-only operations and per-row `model_paths` plus `framework_quantization.allowed_modes` for runtime/quant-sensitive operations. Add a targeted generation test proving each artifact resolves non-empty profiles with the intended geometry and quant lane. |

### 5. [P1] Do not bypass the GEMM transfer policy with a hard INT4 table alias

`docs/design/how_to_add_a_gemm_quant_mode.md` explicitly says a data-less GEMM mode needs the enum/profile lines and “No table normalization, no `validation_aliases`, no query-layer changes”; the policy-gated transfer ladder then chooses the reference, with weight-only modes borrowing BF16 under `XPROFILE`. This branch instead maps `w4a16_nvfp4` directly to the `int4_wo` table in Python and Rust.

I reproduced the behavioral regression with the comprehensive fixture: after adding an INT4 table and setting `transfer_policy="balanced"` (only `xshape`/`xquant`, no `xprofile`), a W4A16-NVFP4 HYBRID query still returns `source=empirical` with provenance only `empirical`. The hard alias therefore crosses the `(0.5625,1)` to `(0.5,1)` profile boundary while bypassing the policy that explicitly disallows cross-profile transfer.

| Current head | Required before merge |
| --- | --- |
| `_normalize_gemm_quant_mode_for_table(w4a16_nvfp4) -> int4_wo` makes an unmeasured cross-profile reference look like own-table empirical data and ignores `XPROFILE` admission. | Remove the W4A16 table normalization and the special query/cache branches, retain the enum plus utilization-profile entries, and let the existing compute-first transfer ladder select BF16 under `XPROFILE`. Add the W4A16 analogue of `test_xprofile_weight_only_borrows_bf16_not_fp8`, assert that `balanced` raises the typed miss, and extend the Rust transfer oracle/parity coverage. If an INT4 calibration alias is intentionally desired instead, it needs an explicit design-owner decision and must still be classified and gated as cross-profile transfer. |

### 6. [P1 coverage] The new direct GEMM regressions are not selected by CI

The unit workflow runs `pytest tests/unit -m "unit or build"`, but `tests/unit/sdk/database/test_gemm_op.py` has no module marker. `pytest --collect-only -q -m unit tests/unit/sdk/database/test_gemm_op.py` reports `23 deselected / 0 selected`, so the new normalization and W4A16 reference tests in that file did not run in the reported unit check.

| Current head | Required before merge |
| --- | --- |
| Critical new tests live in an unmarked module and are skipped by the CI selector. | Add `pytestmark = pytest.mark.unit` to the module (or move the transfer-policy coverage into the already marked `test_gemm_quant_transfer.py`) and verify the CI-shaped collection selects them. |

### Verification performed

- `git diff --check` and Ruff on all changed Python files passed.
- 191 focused Python tests passed when invoked directly.
- All 364 Rust library tests passed.
- All 23 added JSON artifacts parse.
- Against the historical PR base, the matrix has 2,736 new rows, zero removals/retained-row mutations/duplicates, exactly 228 rows for each of the 12 intended IDs, and the status totals stated in the PR body. Those rows still need regeneration after the correctness and rebase fixes above.

Please re-request review after rebasing and regenerating from the corrected operation graph.

jasonqinzhou:
Addressed all six requested changes and rebased onto current `main`:

1. Preserved #1503's current Qwen3.5 graph/dispatch/memory behavior while adding the projection, dense, shared, and routed-expert precision splits.
2. Added end-to-end `W4A16_NVFP4` routed-MoE handling for Qwen3.6-35B across Python, Rust, wire conversion, memory/admission, and transfer policy.
3. Applied checkpoint attention/dense/shared exclusions for Qwen, Gemma, and Kimi at the operation level.
4. Added artifact-specific Collector rows and targeted profile/mode tests.
5. Removed the W4A16-to-INT4 table alias and routed the data-less mode through policy-gated `XPROFILE` transfer with Rust/Python parity coverage.
6. Added the unit marker so the direct GEMM regressions are selected by CI.

I also fixed support-matrix classification so the known SILICON W4A16 MoE data gap retries through HYBRID while explicit SILICON mode remains strict. The committed matrices were regenerated from the corrected graph: 2,736 unique new rows across 12 models and 10 systems, with zero retained-row drift versus `main`.

The branch is conflict-free, all review threads are resolved, and the final required checks are green. @Arsene12358, please take another look when convenient.

Arsene12358 (changes requested):
Request changes on `db58f3533795090e78bd283395a52a99def69994`.

I independently verified that the rebase now merges cleanly and preserves current-main Qwen3.5 behavior. The routed W4A16 mode, attention/shared-expert exclusion handling, removal of the W4A16→INT4 table alias, and the missing unit marker are also fixed. Five merge blockers remain that the current tests do not exercise:

1. **Kimi K2.6/K2.7 and MiniMax M3 still omit their dense FFN layers.** The Kimi configs declare 61 layers, `first_k_dense_replace=1`, dense width 18,432, and layer 0 excluded from NVFP4; the runtime graph nevertheless emits shared/router/NVFP4-MoE operations with scale 61 and no dense op. MiniMax declares 60 layers, `dense_intermediate_size=12288`, and `moe_layer_freq=[0,0,0,1…]`, with those three dense MLPs explicitly MXFP8; [`minimax_m3.py`](https://github.com/ai-dynamo/aiconfigurator/blob/db58f3533795090e78bd283395a52a99def69994/aic-core/src/aiconfigurator_core/sdk/models/minimax_m3.py#L157-L209) emits shared/router/NVFP4-MoE operations with scale 60 and no dense op. Parse the layer partition and dense width from the raw config, emit Kimi as 1 BF16 dense + 60 MoE and MiniMax as 3 MXFP8 dense + 57 MoE in both phases, and assert operation names, widths, quant modes, and scale factors. Regenerate every affected matrix row.

2. **`nvidia/MiniMax-M3-NVFP4` has no viable default-matrix path.** An audit of all 228 new rows gives exactly 120 `FAIL`, 104 `HW_INCOMPATIBLE`, 4 `FRAMEWORK_INCOMPATIBLE`, and 0 `PASS`/`HYBRID_PASS`. Blackwell failures include missing NVFP4 MoE data for `{power_law_1.01, topk=4, experts=128, hidden=6144, inter=3072}`. This violates [`model-roster.md`](https://github.com/ai-dynamo/aiconfigurator/blob/db58f3533795090e78bd283395a52a99def69994/docs/support-matrix/model-roster.md#L21-L23), which requires at least one viable combination before roster inclusion. Add exact data or a justified policy-gated rescue and regenerate, with a regression requiring every new roster ID to have at least one `PASS`/`HYBRID_PASS`; otherwise keep the config bundled but remove MiniMax from `SupportMatrixHFModels` until a viable path exists.

3. **Qwen3.6-35B’s W4A16 MoE collector declaration expands to zero cases.** The model row allows only `w4a16_nvfp4` for SGLang/TRT-LLM/vLLM at [`Qwen3_5MoeForConditionalGeneration_cases.yaml:82`](https://github.com/ai-dynamo/aiconfigurator/blob/db58f3533795090e78bd283395a52a99def69994/collector/cases/models/Qwen3_5MoeForConditionalGeneration_cases.yaml#L82-L94), but no backend in `base_ops/moe.yaml` declares that mode. At SM100 the real backend-mode × model-mode intersection is `[]` for all three backends, yielding zero expanded cases (despite 117/117/42 available shapes). Add source-backed backend quant specs and collector support only where runnable, or remove the planned activation; test actual generated cases rather than only `moe_model_allows_quantization()`.

4. **MiniMax collector cases carry a conflicting/wrong architecture identity.** `nvidia/MiniMax-M3-NVFP4` appears in both MiniMax case files. Targeted plan resolution loads both and returns `model_architecture=None`; the only MoE row, located under [`MiniMaxM3ForCausalLM_cases.yaml`](https://github.com/ai-dynamo/aiconfigurator/blob/db58f3533795090e78bd283395a52a99def69994/collector/cases/models/MiniMaxM3ForCausalLM_cases.yaml#L5-L33), is emitted as `MiniMaxM3ForCausalLM`, while the bundled config declares `MiniMaxM3SparseForConditionalGeneration`. This violates the collector fail-loud identity rule and can poison persisted data. Remove the NVIDIA ID/row/activation from the CausalLM file, move them to the Sparse file, and assert exact one-file resolution plus matching plan and emitted-case architecture.

5. **The data-less W4A16 GEMM implementation violates its documented seam and omits the required exact oracles.** [`how_to_add_a_gemm_quant_mode.md`](https://github.com/ai-dynamo/aiconfigurator/blob/db58f3533795090e78bd283395a52a99def69994/docs/design/how_to_add_a_gemm_quant_mode.md#L7-L45) explicitly requires enum/parser + utilization-level additions with “no query-layer changes,” a cloned BF16-vs-FP8 Python template, and Rust values pinned at `1e-9`. The PR adds Python `_empirical_grid_key()` and splits Rust query/cache identity into `table_quant + modeled_quant`; these changes are unnecessary now that W4A16 retains its own table identity. The Python test only asserts `borrowed > 0`, and the Rust Python-oracle test has no W4A16 entries. Restore main’s FP8-static-only cache/query mechanism, retain the new enum/parser/util-level row, clone the full Python reference-selection assertion, and add the independently reproduced Rust oracles `m=512 → 0.036529976276703825` and `m=8192 → 0.5714972813924153` at `1e-9`.

Validation: clean `merge-tree`; exact matrix delta of 12 IDs × 228 = 2,736 rows with zero removals, retained-row changes, or duplicate identities; 403 focused Python tests pass after rebuilding the pinned Rust extension; 379 Rust workspace/public-API/integration tests pass; all 25 GEMM tests are selected by `-m unit`; `git diff --check` is clean; all 22 GitHub checks pass. Those checks currently miss the five issues above.

Arsene12358 (approved):
Approved at reviewer direction. This supersedes my prior changes-requested review.

PR #1514, opened 2026-08-11
fix(sdk): admit partial perf data before version fallback

YijiaZhao:
#### Overview:

  Fix perf-data fallback semantics for partially collected tables. Successfully collected rows from the requested framework version now remain the primary source, while older versions fill only missing shapes.

  #### Details:

  - Treat collection_meta.yaml entries with status: partial as incomplete coverage rather than invalid data.
  - Admit successfully collected rows from partial tables as the primary source.
  - Preserve first-wins source ordering so older versions only fill missing shapes.
  - Continue rejecting directories marked with legacy INCOMPLETE.txt, since that marker lacks table- or shape-level coverage information.
  - Add regression tests covering family-first and legacy layouts, partial-table admission, and legacy incomplete-directory rejection.
  - Verified with Ruff and the relevant SDK database tests:
      - 87 passed
      - 2 skipped

  Please start with:

  - aic-core/src/aiconfigurator_core/sdk/perf_database.py
      - _version_dir_state
      - _version_dir_unusable_for_request
      - PerfDatabase._build_op_sources

  - aic-core/src/aiconfigurator_core/sdk/operations/base.py
      - _version_dir_is_unusable
      - resolve_op_data_path

  Regression coverage is in:

  - tests/unit/sdk/database/test_reuse_ordering.py
  - tests/unit/sdk/database/test_dual_layout_discovery.py

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading and fallback behavior for partially populated collections.
  * Structured partial collections remain available for valid data retrieval and reuse.
  * Legacy incomplete markers continue to exclude affected directories safely.
  * Improved discovery and validation across legacy and family-based layouts.
  * Ensured valid fallback sources can be used when primary data is incomplete.

### DIFF
--- a/aic-core/src/aiconfigurator_core/sdk/perf_database.py
+++ b/aic-core/src/aiconfigurator_core/sdk/perf_database.py
@@ -405,11 +405,27 @@ def _load_collection_meta_yaml(path: str) -> dict:
     return raw
 
 
-def _collection_meta_has_partial_table(meta: dict) -> bool:
+def _collection_meta_partial_tables(meta: dict) -> frozenset[str]:
+    """Return tables whose collection coverage is incomplete.
+
+    ``status: partial`` is table/shape coverage metadata, not a statement that
+    the successfully collected rows are invalid.  Those rows remain eligible
+    as the primary source; older versions only fill coordinates they do not
+    contain (design §6.1/§6.2 first-wins semantics).
+    """
     tables = meta.get("tables")
     if not isinstance(tables, dict):
-        return False
-    return any(isinstance(table, dict) and table.get("status") == "partial" for table in tables.values())
+        return frozenset()
+    return frozenset(
+        name
+        for name, table in tables.items()
+        if isinstance(name, str) and isinstance(table, dict) and table.get("status") == "partial"
+    )
+
+
+def _collection_meta_has_partial_table(meta: dict) -> bool:
+    """Compatibility predicate for metadata/reporting callers."""
+    return bool(_collection_meta_partial_tables(meta))
 
 
 def _version_dir_state(version_path: str, *, data_dir: str | None = None) -> dict[str, object]:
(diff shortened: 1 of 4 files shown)


aic-core/src/aiconfigurator_core/sdk/perf_database.py
```diff
@@ -2261,23 +2288,19 @@ def _build_op_sources(
         op_file_basename = op_filename_enum.value
         records: list[dict[str, object]] = []
         primary_version_dir = os.path.dirname(primary_path)
-        if os.path.isfile(primary_path) and _version_dir_partial_for_request(
+        if os.path.isfile(primary_path) and _version_dir_unusable_for_request(
```
jasonqinzhou, line 2291: **[P1] Reject legacy-incomplete sources in every donor channel.** This usability check only covers the primary path. Each donor loop below accepts the result of `resolve_op_data_path` after only an existence check, and that resolver's final legacy-layout fallback does not inspect `INCOMPLETE.txt`. I reproduced a legacy-incomplete directory being admitted independently as `declared_reuse`, same-backend `fallback`, and `cross_backend` at this head. That violates the new whole-directory veto and can fill missing prediction shapes with rows the loader promises to reject. Please apply the same `_version_dir_unusable_for_request` check to every resolved donor before appending it, and add legacy-layout donor regressions for the affected channels. A one-click suggestion is not safe because the fix spans three source loops and their tests.

jasonqinzhou (changes requested):

## What this PR does

This PR changes SDK perf-data discovery and source admission so successfully collected rows from structured partial tables remain primary, with older versions filling only missing shapes, while legacy INCOMPLETE.txt directories remain whole-directory vetoes. It updates both family-first and legacy path handling and adds focused regression coverage for the new primary-source behavior.

The distinction between structured partial coverage and an unstructured legacy marker is modeled clearly, and the new family-layout and legacy-layout tests make the intended primary-before-fallback ordering easy to follow.

## Why this score

The current head is 6.5/10 with REQUEST_CHANGES. The structured-partial primary behavior is coherent and the relevant local suites pass, but the legacy whole-directory veto is applied only to the primary path: declared-reuse, same-backend fallback, and cross-backend donor loops still admit legacy-layout directories carrying INCOMPLETE.txt. Because those rows can silently fill prediction shapes that the PR promises to reject, this source-admission gap must be fixed before merge.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
YijiaZhao: Fixed in `36bac968`. The declared-reuse, same-backend fallback, and cross-backend loops now all call `_version_dir_unusable_for_request(...)` before admitting a resolved donor. Added legacy-layout `INCOMPLETE.txt` regressions for all three channels. Targeted database tests pass (`48 passed`), and Ruff/format checks pass.

YijiaZhao:
Addressed the remaining review feedback in `36bac968`:

- reject legacy `INCOMPLETE.txt` donors in declared-reuse, same-backend fallback, and cross-backend channels;
- add regressions for all three donor channels;
- add a real row-level first-wins test: an overlapping shape keeps the primary value, while a shape missing from primary is filled by fallback.

Validation:
- `ruff check` passed
- `ruff format --check` passed
- targeted database tests: `48 passed`

jasonqinzhou (approved):
**APPROVE · 9.5/10 · high confidence**

## What this PR does

This PR changes SDK perf-data discovery and source admission so rows from structured partial tables remain primary while older versions fill only missing shapes, and legacy INCOMPLETE.txt directories remain whole-directory vetoes. The current commit completes that policy across declared-reuse, same-backend fallback, and cross-backend donor paths and adds direct ordering and row-loading regressions.

The follow-up addresses the prior source-admission gap comprehensively: it applies the same request-scoped guard to all three donor channels, adds a focused regression for each channel, and strengthens the suite with an actual first-wins row-loading test.

## Why this score

The current head is 9.5/10 with APPROVE. The previous blocking finding is fixed: every resolved donor is now checked for the legacy whole-directory veto before admission, and all new and adjacent targeted tests pass along with the broader SDK database suite. No actionable current-head findings remain; the only outstanding uncertainty is five live CI contexts that were still running when this packet was written.

PR #1515, opened 2026-08-11
feat: Adding Laguna-S and XS to AIC

DanBlanaru:
#### Overview:

Adding Laguna-S FP8 on H200s to AIC perf measurements.

- closes GitHub issue: #xxx

## Summary by CodeRabbit

* **New Features**
  * Added support for Laguna S 2.1 FP8 and Laguna XS.2 FP8 models.
  * Added hybrid full- and sliding-window attention with dense and mixture-of-experts processing.
  * Added Laguna-specific quantization, routing, and KV-cache capacity support.
  * Added model configuration and inference workflow integration.

* **Bug Fixes**
  * Improved FP8 MoE benchmarking by using checkpoint-defined quantization settings when available.

* **Tests**
  * Added comprehensive coverage for Laguna model resolution, attention profiles, quantization, operation generation, and KV-cache behavior.


collector/cases/models/LagunaForCausalLM_cases.yaml
```diff
@@ -0,0 +1,50 @@
...
+      inter_size: 1024
+      topk: 10
+      num_experts: 256
+      # Block-FP8 requires the per-rank intermediate width to remain divisible
+      # by the checkpoint's 128-element quantization block.
+      max_tp_exclusive: 16
+      framework_quantization:
+        vllm:
+          allowed_modes:
+            - fp8_block
```
Arsene12358, line 37: **[Blocker]** This row leaks into sglang/trtllm cross-model sweeps: `framework_quantization` narrows only the backends it lists, and `framework_specific_op_cases` gates only model-scoped runs. Measured on this branch: vllm plans 36 Laguna cases (intended), sglang and trtllm plan **90 each** under the base ungated modes — with softmax/no-bias default routing, since those collectors have no laguna mapping and this row sets no `sglang_moe_*` fields. Routing is a launch parameter of the fused kernels those collectors invoke, `moe_perf` has no model/routing column, this shape tuple is unique to Laguna, and merges are append-missing — so the rows would be wrong, indistinguishable, and permanent. Same trap `KimiK3ForConditionalGeneration_cases.yaml` documents; same fix:

```suggestion
        vllm:
          allowed_modes:
            - fp8_block
        # Keep cross-model sglang/trtllm sweeps from planning Laguna under the
        # base ungated modes: no sglang/trtllm routing mapping or sglang_moe_*
        # fields exist yet, so leaked cases would benchmark default softmax
        # routing and persist wrong-identity rows at Laguna's unique shape keys
        # (Kimi-K3 precedent, review 2026-08-04).
        sglang:
          allowed_modes: []
        trtllm:
          allowed_modes: []
```

aic-core/src/aiconfigurator_core/sdk/models/laguna.py
```diff
@@ -0,0 +1,542 @@
...
+                kvcache_q,
+                window_size=dims["window_size"],
+                head_size=dims["head_size"],
+                use_qk_norm=cfg.use_qk_norm,
+            )
+        )
+        gate_ops = []
+        if cfg.gating:
+            gate_ops = [
+                ops.GEMM(f"{prefix}_attention_gate_gemm", count, dims["n_q_per_gpu"], self._hidden_size, gemm_q),
```
Arsene12358: At tp=8 this GEMM is n=6 for the global bucket (48/8) and n=9 for SWA (72/8) — below what the h200/vllm/0.24.0 GEMM table can answer (smallest collected n is 32), so every tp=8 candidate errors and is silently skipped by the sweep:

```
sweep_agg: error at tp=8 pp=1 dp=1 moe_tp=1 moe_ep=8, skipping
sweep_disagg/prefill: ... GEMM perf data not available for requested shape.
  quant_mode='bfloat16', m=4000, n=6, k=3072
```

Recommend modeling the per-head gate as `ElementWise` instead of a table GEMM: a hidden x n_q matvec per token is memory-bound on reading the hidden vector (norm-like cost), which removes the degenerate-width query class for every current and future tp. Alternatives in the review body (small-n GEMM collection, or documenting tp8 exclusion as a deliberate decision).

Arsene12358 (changes requested):
The core enablement looks solid and I verified it end-to-end on this branch: `aiconfigurator cli default --model-path poolside/Laguna-S-2.1-FP8 --total-gpus 8 --system h200_sxm --backend vllm --backend-version 0.24.0 --database-mode SILICON --isl 4000 --osl 1000 --ttft 2000 --tpot 50` completes in ~16 s with sane agg/disagg configs; the resolved quant defaults (gemm=bf16 / moe=fp8_block / kv=fp8) match the checkpoint's compressed-tensors `ignore` list (only routed experts are FP8); the window-capped KV math and its capacity inverse check out; and the fresh rows are exactly the declared shapes (714/840 attention = the three SWA TP shards; 972 moe = 36 cases x 27 token counts) with no key collision against the pre-existing topk-8 rows at 3072/1024.

Requesting changes on three findings that no CI check surfaces (deliberately not re-litigating the red checks — those are directly visible). Evidence below is from a local checkout of this branch (`uv sync --all-extras --all-groups`).

### 1. [Blocker] Laguna MoE geometry leaks into sglang/trtllm cross-model collection plans with a wrong routing identity

`framework_specific_op_cases: vllm: moe: cases: all` gates the op only for Laguna-scoped runs; cross-model getter sweeps still expand the `model_case_values.moe` row for every backend, and `framework_quantization` narrows only the backends it lists — unlisted backends fall through to the base ungated quant modes. Measured on this branch:

```
get_common_moe_test_cases(backend='vllm')   -> 36 Laguna cases   (intended)
get_common_moe_test_cases(backend='sglang') -> 90 Laguna cases   (leak)
get_common_moe_test_cases(backend='trtllm') -> 90 Laguna cases   (leak)
```

Why this is data poisoning rather than harmless over-collection: only `collector/vllm/collect_moe.py` was taught Laguna's routing, and this row sets none of the `sglang_moe_*` fields, so the leaked cases fall back to softmax/no-bias defaults (`case_generator.py:1065,1069`) — while routing is a launch parameter of the fused kernels the other collectors benchmark (`collector/sglang/collect_moe.py:513,588` maps it to `RoutingMethodType`; the trtllm collector builds `DeepseekV3Gate` vs `RenormalizeMoeRoutingMethod`). `moe_perf` has no model or routing column, the 3072/1024/topk10/256 tuple is unique to Laguna, and merges are append-missing (this PR's own meta: `replaced_rows: 0`) — so the next routine sglang/trtllm campaign would persist wrong-routing rows at Laguna's exact query keys **and block a later correct collection from ever replacing them**. If the cases crashed they'd be recorded and tolerated; succeeding with the wrong identity is the failure mode `case_authoring.md` singles out.

**Fix (4 lines, committable suggestion posted inline on the YAML):** explicit empty `allowed_modes` for sglang/trtllm — the exact pattern `KimiK3ForConditionalGeneration_cases.yaml` documents for the same situation (review 2026-08-04). Alternative if sglang/trtllm coverage is actually wanted: teach those collectors the laguna routing mapping and set the `sglang_moe_*` fields — either close the gate or make the leaked cases correct; the current in-between is the only wrong state.

### 2. Every tp=8 topology is silently dropped: the per-head gate GEMM falls below the GEMM-table floor

The attention gate is modeled as `GEMM(n = n_q_per_gpu, k = hidden)`. At tp=8 that is n=6 (global, 48/8) and n=9 (SWA, 72/8); the smallest collected n in the h200/vllm/0.24.0 GEMM table is 32, and n=6/9 fall outside what interpolation will extrapolate — so every tp=8 candidate errors and is skipped while the sweep completes green:

```
sweep_agg: error at tp=8 pp=1 dp=1 moe_tp=1 moe_ep=8, skipping
sweep_agg: error at tp=8 pp=1 dp=1 moe_tp=8 moe_ep=1, skipping
sweep_disagg/prefill: ... err=GEMM perf data not available for requested shape.
  quant_mode='bfloat16', m=4000, n=6, k=3072
```

**Suggested fix:** model the per-head gate as `ElementWise` instead of a table GEMM — a hidden x n_q matvec per token is memory-bound on reading the hidden vector (norm-like), which removes the degenerate-width query class for every current and future tp. Alternatives: append small-n GEMM rows (n in {6, 9, 12, 18, 24, 48} x k=3072) to the H200 table, or state explicitly that Laguna recommendations exclude tp8 — a defensible call for a 3072-hidden model, but it should be a documented decision rather than a silent skip. (Same failure class as the below-grid gate GEMM found in #1503.)

### 3. LAGUNA is missing from `MOE_WORKSPACE_FAMILIES` — block-scale dispatch workspace unmodeled

`base_backend.py` adds a MoE block-scale dispatch workspace (`num_tokens * width * attn_dp * num_experts * topk / moe_ep / 128 * 4` bytes) for exactly the block-FP8-served families (GEMMA4MIX, DEEPSEEK, DEEPSEEKV32, DEEPSEEKV4, KIMIK25). Laguna is served fp8_block on vLLM through the same flashinfer_cutlass/triton expert kernels but is not in the list, so its activation estimate omits the term. With 256 experts x topk 10 that is roughly **2.0 GB at an 8k-token budget with ep=1** (3072 width; ~250 MB at ep=8) — unmodeled memory that inflates the predicted KV budget and max concurrency on TP-only topologies, which is exactly where this model deploys. No test asserts memory for the family, so nothing fails.

**Suggested fix:** verify against a real vLLM 0.24.0 Laguna serving memory profile (the compressed-tensors path may differ from the DeepGEMM-style one), then either add `"LAGUNA"` to `MOE_WORKSPACE_FAMILIES` plus the `_moe_workspace_width` override (`_hidden_size`, i.e. 3072, matching the GEMMA4MIX branch — not `num_heads*head_size` = 6144), or record the measured justification for excluding it.

<details>
<summary>Non-blocking notes (no action required to merge, listed for completeness)</summary>

- The new laguna routing mappings and the `CompressedTensorsConfig` branch in `collect_moe.py` carry no serving citations (file:line @0.24.0), unlike the sibling deepseek_v4 and Fp8Config comments; `test_vllm_laguna_runtime_config_matches_serving` asserts the mapping against itself, so a citation is the only serving-truth anchor.
- `gating=bool(config.get("gating", False))` in `utils.py`: the checkpoint value is the string `"per-head"`, so any future string (`"elementwise"`, even `"none"`) silently parses as per-head gating; every other Laguna field in this parser raises on unknown values.
- `_validate_fp8_block_quantized_moe_config` reads `quantization_config.weight_block_size`, which compressed-tensors checkpoints do not define (Laguna declares it at `config_groups.group_0.weights.block_structure`) — correct today only via the hardcoded [128,128] default; it is also a third verbatim copy of the helper in `moe.py`/`hybrid_moe.py`, worth extracting.
- The absolute pins in `test_getter_deduplication.py` (1926/52002) and `test_model_cases.py` (5025) were replaced with Laguna-scoped asserts; restoring updated absolute pins alongside them is what would have caught finding 1 on the MoE side.
- The SWA attention entry lacks `sglang_runtime_window_size: 511` (Gemma4 maps 1024 to 1023, GPT-OSS 128 to 127; the generator fallback would pass 512 verbatim).
- `LagunaModel.get_kvcache_elements_per_token` is byte-identical to the base GQA formula and unreachable for Laguna (its only production caller is the base `get_kvcache_bytes_per_sequence`, which Laguna overrides) — dead code.
- The moe `collection_meta.yaml` rewrite dropped the dated Kimi-K3 wave-3 append note (pipeline/job ids, EP-coverage rationale); please restore it or relocate it to the Kimi-K3 ledger under `docs/perf_database/`.
- Triage aid: CodeRabbit's "missing `pytest.mark.unit`" comment is a false positive — module-level `pytestmark = pytest.mark.unit` (line 15) already covers the new test.

</details>

collector/vllm/collect_moe.py
```diff
@@ -399,14 +401,19 @@ def gemma4_routing_function(hidden_states, gating_output, topk, renormalize):
...
-        quant_config = Fp8Config(
-            is_checkpoint_fp8_serialized=True,
-            activation_scheme="dynamic",
-            weight_block_size=[128, 128],
-        )
+        model_quantization = _load_model_moe_config(model_name).get("quantization_config")
+        if isinstance(model_quantization, dict) and model_quantization.get("quant_method") == "compressed-tensors":
+            # Construct the same artifact-owned quantization method serving
+            # selects instead of approximating it with generic Fp8Config.
+            quant_config = CompressedTensorsConfig.from_config(model_quantization)
```
jasonqinzhou, line 428: The current mandatory `Check collector data` job still fails after the recollection: the table emits `vllm_compressedtensorsw8a8fp8moe_flashinfer_cutlass_flashinferexperts` and `vllm_compressedtensorsw8a8fp8moe_triton_tritonexperts`, but neither label is registered, so both resolve to `backend=unverified` and the generated backend facts drift from the checked-in registry. Please add both labels to `collector/kernel_source_backends.yaml` with their verified backend identities, regenerate `collector/op_backend_facts.yaml`, and commit the source and generated artifacts with the parquet data. Otherwise this head cannot pass the required Collector-data gate.

collector/cases/models/LagunaForCausalLM_cases.yaml
```diff
@@ -0,0 +1,50 @@
...
+  attention:
+    - id: laguna_s2_full_attention
```
jasonqinzhou, line 39: The new `max_tp_exclusive` constraint narrows only the MoE cases; these attention profiles remain framework-neutral under `include_base: true`, so they still expand the SGLang 0.5.14 structural population. The current full-unit job and the exact local test both fail `test_sglang_sm90_full_structural_population_is_stable` with 154 cases instead of 151. Please either scope the attention profiles to vLLM if that is the intended collection surface, or explicitly validate Laguna for SGLang and update the structural fixture with that justification. The PR should restore the required unit gate either way.

aic-core/src/aiconfigurator_core/systems/data/h200_sxm/attention/vllm/0.24.0/collection_meta.yaml
```diff
@@ -1,13 +1,34 @@
...
+  path: poolside/Laguna-S-2.1-FP8
+  revision: fc72eb9025540b2225a681bb81950cc1225ff286
+  config_sha256: 6337348c4988936be54276c637e7a2c1c6b25ec44114a199240944a3e0f7212e
+collection:
+  scope: append_missing_laguna_keys
+  run_root: artifacts/laguna_s2_repro/collection_vllm_0_24_minimal_broad_moe_20260810
+  collected_at: '2026-08-10'
 tables:
   context_attention_perf:
+    collector_hash: sha256:fa0ee72fb4fdce4686b2325299579efa42bf09129e78a6990d845df434fb011a
```
jasonqinzhou: This `collector_hash` is stale: `sha256:fa0ee72f...` is exactly the `collector.vllm.collect_attn` closure hash at the previous reviewed head `ccf63d720f83`, while the current head computes `sha256:2f0fff50...`. The new `max_tp_exclusive` field lives in the shared `__model_cases__` closure, which the attention collector intentionally hashes even though this particular constraint is MoE-specific. Please re-finalize/recollect both attention tables from the final current-head closure and commit sidecars carrying the current hash; a previous-head provenance attestation should not be published as current data evidence.

jasonqinzhou (changes requested):

## What this PR does

This PR adds first-class Poolside Laguna-S 2.1 FP8 support for H200 and vLLM 0.24.0: packaged configuration and model-family parsing, a dedicated hybrid full/SWA plus dense/MoE operation graph, KV-cache sizing, Collector cases and performance data, compatibility exports, and tests.

The follow-up correctly adds Laguna's sigmoid and FP32-bias routing contract, restores the established full-module MoE timing path, recollects all 972 Laguna MoE measurements, rejects unsupported TP=16 block-FP8 cases, and cleans up the public export and formatting issues.

## Why this score

6.0/10, REQUEST_CHANGES. The incremental commit fixes the prior routing-contract and Ruff findings, and the focused Laguna and import tests remain strong. This head still fails the required Collector-data and full-unit gates because the compressed-tensors kernel labels are unregistered and the framework-neutral Laguna attention profiles expand SGLang's stable population. The new attention sidecar also attests the previous head's Collector closure hash rather than the current head. Existing CodeRabbit feedback and required copyright, title, and DCO checks also remain unresolved.

DanBlanaru:
Thank you guys! Will address your comments

tests/unit/collector/test_getter_deduplication.py
```diff
@@ -552,3 +551,27 @@ def test_vllm_standard_topk_does_not_require_group_fields(monkeypatch):
     assert runtime_config["use_grouped_topk"] is False
     assert runtime_config["num_expert_group"] is None
     assert runtime_config["topk_group"] is None
+
+
+def test_vllm_laguna_runtime_config_matches_serving(monkeypatch):
```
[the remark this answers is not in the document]
DanBlanaru: No change needed because the module already applies pytestmark = pytest.mark.unit.

collector/cases/models/LagunaForCausalLM_cases.yaml
DanBlanaru: Fixed. Laguna now declares empty MoE quantization allowlists for SGLang and TRT-LLM, so its geometry is expanded only by the vLLM collector.

aic-core/src/aiconfigurator_core/sdk/models/laguna.py
DanBlanaru: Agreed with your recommendation. I replaced the skinny gate GEMM with an ElementWise memory model that reads the hidden vector and writes the per-head gate values; the model-scoped GEMM machinery/data were removed and the GEMM parquet remains unchanged.

collector/vllm/collect_moe.py
DanBlanaru: Fixed. The observed labels now map to flashinfer_cutlass and triton_fused_moe, the affected H200/vLLM 0.24 backend fact was updated, and the collector-data gate passes.

collector/cases/models/LagunaForCausalLM_cases.yaml
DanBlanaru: Fixed. Both Laguna attention profiles are scoped to frameworks: [vllm], and the SGLang structural population remains unchanged.

aic-core/src/aiconfigurator_core/systems/data/h200_sxm/attention/vllm/0.24.0/collection_meta.yaml
DanBlanaru: Fixed. The attention sidecars now use hashes from the final collector closure and contain the final row counts and parquet hashes. I also removed the workstation-local run_root fields.

jasonqinzhou (approved):
**APPROVE · 9.3/10 · high confidence**

## What this PR does

This PR adds first-class Poolside Laguna-S 2.1 FP8 and Laguna-XS.2 FP8 support for H200 with vLLM 0.24.0: bundled model parsing, hybrid full/SWA dense-plus-MoE operation graphs, checkpoint-specific quantization and topology validation, KV and activation-memory accounting, Collector plans and measured data, compatibility exports, and tests.

The Laguna-XS addition is especially thoughtful about deployment boundaries: it separates block-FP8 FFN weights from BF16 attention, rejects invalid TP8 shapes before database lookup, preserves every existing performance row while extending the measured attention surface, and keeps the colliding XS MoE collection lane fail-closed with direct tests.

## Why this score

9.3/10, APPROVE. The three previously reviewed Laguna-S commits are patch-identical after the rebase, and the only new PR patch adds Laguna-XS with coherent model, data, Collector, and validation coverage. All targeted checks pass, the exact-head native SILICON sweep completes with 1,152 aggregated and 95 disaggregated results while rejecting only intentionally invalid TP8/MoE-TP8 topologies, the branch merges cleanly, and all 26 GitHub checks are successful. No prior finding reopens and no new actionable finding is recommended. The main residual risk is accuracy rather than functional correctness: XS MoE estimates deliberately reuse an existing consumer-key-identical fp8_block surface instead of new routing-specific measurements.

PR #1516, opened 2026-08-11
feat: support meta-models/Muse-Glimmer-30B (dense hybrid SWA) with windowed attention data

Arsene12358:
## What / Why

[Muse Glimmer 30B](https://huggingface.co/meta-models/Muse-Glimmer-30B) is Meta Superintelligence Labs' first open-weight release (2026-08-10): a dense 52-layer causal LM with hybrid attention — 39 `sliding_attention` layers (window 2048) interleaved 3:1 with 13 `full_attention` layers — uniform GQA at 32 heads / 2 KV heads / head_dim 128, hidden 6656, SwiGLU inter 19968, vocab 202K, 128K context, plus an unpriced vision tower. This PR makes it a first-class AIC model: a dedicated `MUSEGLIMMER` family, and freshly collected window-2048 attention silicon so its SWA layers are priced from measured kernels instead of the window-0 fallback (whose reconstruction the repo's own validation notes put at ~25–77% error).

Design per the approved spec (Linear project "Model: Meta Muse Glimmer", design doc mirrored there): a standalone model class copy-adapted from gemma4's dense path — deliberately NOT a subclass and NOT shared code; structural similarity to `gemma4.py` is the intended shape, chosen so a gemma4 refactor can never silently change Muse pricing.

## Modeling

- `MuseGlimmerModel` (`aic-core/.../sdk/models/muse_glimmer.py`, registered `MUSEGLIMMER` via auto-discovery): two layer-type recipes driven by `layer_types` — SWA ops at `window_size=2048`, global at `window_size=0` — uniform geometry from standard model_info fields; dense (no MoE/router/MTP); window-capped KV accounting (SWA layers cap at 2048, global grow) with exact binary-search capacity inverse; sglang-only prefill CP.
- **Per review (jasonqinzhou):** the recipe now models the Muse attention gate (per-layer `gate_proj` GEMM `hidden → num_heads·head_dim` + sigmoid-multiply before o_proj, all 52 layers both phases — +1,417,674,752 params flowing into weight accounting, kimi_k3 gate-multiply convention), the TP collectives (context/generation `embedding_ar` + `ar_1`/`ar_2` per layer, mirroring the llama.py dense reference including CP `seq_split` semantics), and QK RMS-norm (`use_qk_norm=True` on both attention ops). Regression tests pin the gate shapes and +1.418B weight constant, the six AR names/scales at tp=8, `_seq_split` under CP, and the QK-norm flags.
- One deliberate deviation from the gemma4 CP pattern: a single uniform `context_cp_all_gather` weighted by all 52 layers instead of per-layer-type gathers — arithmetically identical here because KV bytes/token are uniform across layer types (39x + 13x = 52x); documented in-code and pinned by test.
- Parse branch (`utils.py`) validates `layer_types` length/values and requires a positive `sliding_window`; `MuseGlimmerConfig` carries only `layer_types` + `sliding_window_size`.
- Backend activation coefficients land at the dense LLAMA tier (trtllm `{1: 11, 2: 6.5, 4: 5, 8: 5}`, sglang `{1: 14, 2: 8.5, 4: 6.5, 8: 6.5}`; vllm inherits trtllm's table). Not added to any MoE family/workspace list.
- Deliberately unmodeled (shape/timing-neutral): NoPE on global layers (`layer_rope_theta` 0), logit softcapping, the `qk_scale_factor` scalar (the QK RMS-norm itself IS modeled), `output_multiplier`, and the vision tower (text-serving convention shared with Kimi-K2.5/K3, Llama-4, Qwen3.5).
- Bundled `meta-models--Muse-Glimmer-30B_config.json` is byte-faithful to the HF download; `DefaultHFModels` registration enables offline resolution.

## Collected data

Collector cases (`collector/cases/models/MuseGlimmerForConditionalGeneration_cases.yaml`): one common attention profile for trtllm/vllm (`window_sizes: [0, 2048]`, TP 1/2/4/8) plus sglang-specific SWA/global profiles (`sglang_runtime_window_size: 2047` per the window−1 convention). Collection ran on the aiconfigurator-ci pipeline (h200/b200/gb200/gb300) and via JET (b300), 15/15 jobs, every artifact verified by content (row counts, dtype combos, zero bad latencies, windowed-vs-global physics sanity at ISL 8192: windowed context runs 0.43–0.60× full attention on sglang/trtllm).

Shipped payload: **29,333 window-2048 rows across 24 attention parquets** (h200_sxm, b200_sxm, gb200, gb300 × 3 backends × context/generation). Pre-existing rows are preserved as an unmodified leading prefix in every file; duplicate-key window-0 anchors from staging were dropped against the loader key (all columns except latency, device excluded), so the shipped deltas are exactly the new windowed slices. Per-backend context dtype coverage matches each backend's existing windowed rows (trtllm additionally carries bf16-fmha/fp8-kv). Two sub-0.1% row variances (b200 sglang gen 1332, gb300 sglang gen 1333 vs 1334) are isolated case retries, documented in the task evidence.

**b300_sxm windowed slices are deferred (per review).** The b300 collection ran on GB300 silicon (the B300-SXM cluster is blocked) and anchor-to-anchor divergence vs native B300-SXM rows reaches 11–15% at p90 — too large to publish as native B300 silicon. The GB300-collected rows were dropped from this PR (b300 parquets are byte-identical to main); b300 Muse SWA layers use the documented window-0 empirical fallback, and the affected support-matrix rows honestly report `HYBRID_PASS,empirical`. The validated staging data is retained for a follow-up: re-collection when B300-SXM returns, or an explicit reviewed cross-system transfer mechanism.

## Fidelity

Component level: at (b=1, s=8192, n=32, n_kv=2, hd=128) on h200 sglang, windowed attention resolves from silicon at 0.4488 ms vs 0.9802 ms for full attention — 2.18× — with `source=silicon` provenance (b200 trtllm: 1.79×). End-to-end (h200 sglang, ISL 8192 / OSL 1024, top-3 sweep): landing the silicon data corrected same-config TTFT estimates up ~0.4% (the window-0 fallback was ~4.8% optimistic at the operating point), and the review-driven modeling additions (attention gate, TP all-reduces, QK-norm — all previously unmodeled) corrected rank-1 TTFT a further +233.67 ms (+17.5%). These are corrections toward physical completeness, not regressions; the full mechanism analysis is in the session fidelity note, digested here.

## Support matrix & validation

- Support matrix: 134 Muse rows across the 5 target system CSVs (agg+disagg, latest pinned versions), zero non-Muse churn; the 12 b300 rows report `HYBRID_PASS,empirical` per the deferral above.
- Sanity charts (`tools/sanity_check/create_charts.py`, base→head): 75/75 checks pass, no regressions for other models (pre-existing rows are provably untouched).
- Suite: ruff check/format clean repo-wide; touched suites green (test_utils 109/109, CP 18/18, import contract 62/62, collector 616); full-suite failures (98) were all verified pre-existing by replaying the exact node IDs on clean main@8fc57cf6 (97 fail + 1 error there; the one remainder is a 120 s-timeout load flake that passes solo in 10 s).
- Every commit DCO-signed; collector changes respect the repo's collector layer-permission rules; case counts and dedup behavior evidenced per `collector-v2` population policy.

## Process

Single PR by design (per the tracking project): code wave → 15 collection jobs (h200 pilot first) → data wave, each task gated by an independent review, a final whole-branch review, and a re-reviewed fix wave addressing all review findings.

Resolves AIC-1693
Resolves AIC-1694
Resolves AIC-1695


aic-core/src/aiconfigurator_core/sdk/models/muse_glimmer.py
```diff
@@ -0,0 +1,252 @@
...
+                count,
+                self._num_heads // tp,
+                d["n_kv_per_gpu"],
+                kvcache_q,
+                window_size=window_size,
+                head_size=self._head_size,
+            )
+        return [
+            ops.ElementWise(f"{phase}_{kind}_attn_norm", count, 2 * h, 2 * h, 0.8),
+            ops.GEMM(f"{phase}_{kind}_qkv_gemm", count, d["qkv_out"], h, gemm_q),
```
jasonqinzhou, line 156: **[P1] Account for the Muse attention gate.** Upstream `MuseGlimmerTextAttention` defines a separate `gate_proj` from `hidden_size` to `num_attention_heads * head_dim`, then computes `attn_output * sigmoid(gate_proj(hidden_states))` before `o_proj` ([reference implementation](https://github.com/huggingface/transformers/blob/main/src/transformers/models/muse_glimmer/modeling_muse_glimmer.py#L322-L370)). This recipe jumps from QKV/attention directly to the output projection, so AIC omits 52 gate GEMMs and the gate multiply in both phases. Because model weights are summed from ops, it also drops `52 * 6656 * 4096 = 1,417,674,752` parameters (about 2.64 GiB at BF16/TP1), not just latency. Please add the per-TP gate projection and elementwise gate, or explicitly fold the gate width into a fused projection while preserving weight and latency accounting, and add an op/weight regression.

aic-core/src/aiconfigurator_core/sdk/models/muse_glimmer.py
```diff
@@ -0,0 +1,252 @@
...
+            attention,
+            ops.GEMM(f"{phase}_{kind}_proj_gemm", count, h, d["proj_in"], gemm_q, low_precision_input=True),
+            ops.ElementWise(f"{phase}_{kind}_ffn_norm", count, 2 * h, 2 * h, 0.8),
+        ] + self._shared_mlp_ops(f"{phase}_{kind}", count, h, d["inter_per_tp"])
```
jasonqinzhou, line 165: **[P1] Model the row-parallel reductions for TP > 1.** `_resolve_dims` shards both the attention projection input and the MLP intermediate across TP ranks, but this block emits neither the post-`proj_gemm` nor the post-`mlp_down_gemm` `CustomAllReduce` needed to combine those shards. A constructed TP8 model has zero TP collectives in both phases; by comparison, the established dense-model recipes charge two reductions per transformer layer. TP2/4/8 predictions therefore omit 104 reductions per pass and can select the wrong configuration. Please add the attention and FFN reductions with the correct layer-type/phase scale factors, plus a TP>1 regression that asserts their counts.

aic-core/src/aiconfigurator_core/sdk/models/muse_glimmer.py
```diff
@@ -0,0 +1,252 @@
...
+        kvcache_q = self.config.kvcache_quant_mode
+        if phase == "context":
+            attention = ops.ContextAttention(
+                "context_attention",
+                count,
+                self._num_heads // tp,
+                d["n_kv_per_gpu"],
+                kvcache_q,
+                self.config.fmha_quant_mode,
+                window_size=window_size,
```
jasonqinzhou, line 139: **[P2] Enable Q/K normalization on both attention ops.** Muse Glimmer RMS-normalizes Q and K on every layer before attention, and AIC's `ContextAttention` / `GenerationAttention` constructors have `use_qk_norm` specifically to charge that work. Both constructors here take the default `False` (the built model reports `[False, False]` in each phase), so the new family systematically omits Q/K-normalization cost. Please pass `use_qk_norm=True` for context and generation and pin that invariant in the builder tests.

aic-core/src/aiconfigurator_core/systems/support_matrix/b300_sxm.csv
```diff
@@ -739,6 +739,30 @@ meta-llama/Meta-Llama-3.1-8B,LlamaForCausalLM,b300_sxm,vllm,0.22.0,agg,PASS,,uv
...
+meta-models/Muse-Glimmer-30B,MuseGlimmerForConditionalGeneration,b300_sxm,trtllm,1.3.0rc15,agg,HYBRID_PASS,,uv run aiconfigurator cli default --model-path meta-models/Muse-Glimmer-30B --total-gpus 32 --system b300_sxm --backend trtllm --backend-version 1.3.0rc15 --database-mode HYBRID --isl 256 --osl 256 --prefix 128 --ttft 2000.0 --tpot 50.0 --top-n 1 --no-color,empirical
+meta-models/Muse-Glimmer-30B,MuseGlimmerForConditionalGeneration,b300_sxm,trtllm,1.3.0rc15,disagg,HYBRID_PASS,,uv run aiconfigurator cli default --model-path meta-models/Muse-Glimmer-30B --total-gpus 32 --system b300_sxm --backend trtllm --backend-version 1.3.0rc15 --database-mode HYBRID --isl 256 --osl 256 --prefix 128 --ttft 2000.0 --tpot 50.0 --top-n 1 --no-color,empirical
+meta-models/Muse-Glimmer-30B,MuseGlimmerForConditionalGeneration,b300_sxm,trtllm,1.3.0rc20,agg,PASS,,uv run aiconfigurator cli default --model-path meta-models/Muse-Glimmer-30B --total-gpus 32 --system b300_sxm --backend trtllm --backend-version 1.3.0rc20 --database-mode SILICON --isl 256 --osl 256 --prefix 128 --ttft 2000.0 --tpot 50.0 --top-n 1 --no-color,silicon
```
jasonqinzhou: **[P1] Do not publish GB300 measurements as B300-SXM silicon.** All six new `b300_sxm` window-2048 slices contain `device == "NVIDIA GB300"` (7,334 rows), while the neighboring TRT-LLM/vLLM B300 slices use `NVIDIA B300 SXM6 AC`. The attention loaders index only on dtype/shape/window and never inspect `device`, so these `PASS` / `SILICON` rows silently price B300 with GB300 measurements. The reported cross-platform comparison reaches 11-15% at p90 and 40-52% at the maximum, which is too large to present as native B300 silicon. Please defer/drop the B300 windowed slices and mark the affected matrix entries hybrid/unsupported until real B300-SXM collection is available, or introduce an explicit reviewed cross-system transfer mechanism that preserves provenance in query results.

jasonqinzhou (changes requested):

## What this PR does

This PR adds first-class Muse Glimmer 30B support: model-family mapping and offline config parsing, a dense hybrid sliding-window/global-attention model with KV-cache and context-parallel modeling, collector cases, 36,667 attention rows, support-matrix entries for five systems, and focused import/model/collector tests.

The bundled config is byte-identical to upstream, and the layer/window validation, duplicate-key checks, CP/import coverage, and backend-specific collector cases show careful end-to-end enablement work.

## Why this score

The breadth and verification are strong, but 5.8/10 and REQUEST_CHANGES are warranted because the prediction model omits the 1.42B-parameter attention gate and all tensor-parallel reductions, disables Q/K-normalization accounting, and labels GB300 measurements as B300-SXM silicon. The P1 issues materially undercount weights/latency or attach the wrong silicon provenance, so the current predictions are not safe to publish.

aic-core/src/aiconfigurator_core/sdk/models/muse_glimmer.py
Arsene12358: Fixed in d33b18cc. `_attention_block` now emits a per-layer gate GEMM (`{phase}_{kind}_gate_gemm`, n = num_heads·head_dim/tp, k = hidden) plus the sigmoid-multiply ElementWise before o_proj, both phases, following the kimi_k3 output-gate convention (kimi_k3.py:247-248). Weights flow through GEMM op accounting: the regression test pins the +52×6656×4096 = 1,417,674,752 param constant and the gate shapes at tp=1 and tp=8 (tests/unit/sdk/test_utils.py).

aic-core/src/aiconfigurator_core/sdk/models/muse_glimmer.py
Arsene12358: Fixed in d33b18cc. Muse now emits the six collectives name-for-name and scale-for-scale with the llama.py:202-213 dense reference: `context_embedding_ar` (×1) + `context_ar_1`/`context_ar_2` (×52) and the generation twins (×mtp scale). CP consistency: the ARs are constructed without seq_split and picked up by the existing CP loop's `_seq_split = cp` assignment — equivalent to llama's construction-time `seq_split=cp` since CustomAllReduce consumes the value once at query time (communication.py:262). Regression tests pin all six names/scales at tp=8 and `_seq_split==2` at cp=2. One adjacent observation while verifying: gemma4.py emits no CustomAllReduce on its dense path either — likely the same pre-existing gap in that family; can file a follow-up issue if you agree.

aic-core/src/aiconfigurator_core/sdk/models/muse_glimmer.py
Arsene12358: Fixed in d33b18cc — `use_qk_norm=True` on both ContextAttention and GenerationAttention, pinned as `[True, True]` per phase in the builder tests. One footnote from verifying the plumbing: GenerationAttention stores `_use_qk_norm` but nothing on the decode query path consumes it today (the only reader is ContextAttention.query, attention.py:569), so the decode-side flag is currently latency-inert — passing it anyway matches llama/moe and future-proofs the recipe if decode-side norm pricing lands.

aic-core/src/aiconfigurator_core/systems/support_matrix/b300_sxm.csv
Arsene12358: Complied via defer/drop in be9fd8a3 + de538891: the GB300-collected window-2048 rows are removed — all six b300_sxm attention parquets are byte-identical to main again — and the 12 affected matrix rows now report `HYBRID_PASS,empirical` (window-0 fallback, no silicon claim). The validated GB300 staging data is retained outside the repo for either re-collection when B300-SXM returns or an explicit reviewed cross-system transfer mechanism, whichever the team prefers. One pre-existing item surfaced by this check: main's b300_sxm sglang windowed rows ({128, 1024, 8192}) also carry device "NVIDIA GB300" — same provenance question, predates this PR; happy to file a follow-up issue.

Arsene12358:
All four review findings addressed in three commits (d33b18cc model+tests, be9fd8a3 b300 data revert, de538891 matrix refresh), with per-finding evidence in the inline threads. Net modeling effect of the corrections (attention gate + TP all-reduces + QK-norm): rank-1 TTFT on the h200 sglang reference sweep moves +233.67 ms (+17.5%) — the undercount the review identified, now priced. b300 windowed slices are deferred per the provenance finding (parquets byte-identical to main; 12 matrix rows honestly report HYBRID_PASS,empirical). PR body updated to match (29,333 window-2048 rows across 24 parquets on the four native-silicon systems). Suites: test_utils 109/109, CP 18/18, support-matrix 104/104, ruff clean.

aic-core/src/aiconfigurator_core/sdk/models/muse_glimmer.py
```diff
@@ -0,0 +1,282 @@
...
+        else:
+            attention = ops.GenerationAttention(
+                "generation_attention",
+                count,
+                self._num_heads // tp,
+                d["n_kv_per_gpu"],
+                kvcache_q,
+                window_size=window_size,
+                head_size=self._head_size,
+                use_qk_norm=True,
```
jasonqinzhou, line 152: **[P2] Make decode Q/K normalization contribute latency.** This argument only stores `_use_qk_norm`; unlike `ContextAttention.query()`, `GenerationAttention.query()` never reads it, and `engine._generation_attention()` omits it from the Rust op spec. A direct probe returns identical latency and an identical serialized payload with the flag set to `False` or `True`, so the new test pins a dormant attribute while this recipe documents Q/K norm as modeled in both phases. Please either account for decode Q/K RMS-norm in both the Python and Rust generation paths with parity coverage, or emit explicit generation-phase ops here and assert that enabling the norm changes query latency.

jasonqinzhou (approved):
**COMMENT · 8.7/10 · high confidence**

## What this PR does

This PR adds first-class Muse Glimmer 30B support across offline model-family/config parsing, dense hybrid sliding-window/global-attention modeling, KV-cache and context-parallel accounting, collector cases, measured attention data, support-matrix entries for five systems, and focused tests.

The response to the first review is disciplined: the gate and TP collectives now follow established in-repo recipes with shape/count regressions, and the GB300-derived B300 slices were removed rather than obscuring their provenance.

## Why this score

The current head earns 8.7/10 and COMMENT: all three prior P1 findings are fixed, and prefill Q/K-normalization is now charged, leaving one non-blocking P2. The new GenerationAttention flag is not consumed by the Python query path or serialized to the Rust engine, so decode Q/K-normalization still contributes zero latency despite the new documentation and attribute-level test claiming both phases are modeled.

PR #1518, opened 2026-08-11
feat: add aic-estimate config adapters to aic-core sdk

simone-chen:
## Summary

```
InferenceX config ─┐
Dynamo recipe ─────┼─> aic-estimate-request/1.0.0 ─> cli_estimate kwargs
Custom input ──────┘
```

- Add strict `aic-estimate-request/1.0.0` models, packaged JSON Schema, ordered adaptation reports, and lowering to existing `cli_estimate()` arguments.
- Adapt InferenceX records, standard DynamoGraphDeployments, concrete Dynamo CI SGLang recipes, and literal Helm benchmark workload points without executing source commands.
- Add the repo-local `adapt-server-config` skill with schema validation and explicit estimate execution.
- Keep gap-analysis infrastructure, skills, tools, datasets, reports, and web assets out of both wheels and the Rust crate.

## Validation

- `60 passed`: adapter unit tests, real agg/disagg estimate integration, skill tests, package layering, and release verifier tests.
- Ruff lint and format checks passed.
- Clean upper/core wheels passed payload ownership verification.
- Isolated full-wheel install passed runtime and package-layer verification.
- Rust `cargo package --list` remained core-only.
- Current benchmark corpus audit:
  - 14 concrete recipes completed a real estimate.
  - 1 adapted recipe reached AIC and surfaced an OOM.
  - 2 recipes produced actionable mapping rejections for source topology/concurrency inconsistencies.
  - 16 rendered Helm recipe groups produced 20 canonical requests; unsupported or non-literal points remained explicit rejections.

## Scope

This is Phase 1 only. It does not migrate the dashboard, support-matrix UX, gap data, or gap regression CI, and it does not change `aic-auto-gap-analysis`.

## Dynamo CI API example

Pass a concrete, rendered `dynamo-ci` benchmark recipe to `DynamoRecipeSource`. Adaptation returns one ordered outcome per benchmark concurrency and never runs an estimate itself.

```python
from pathlib import Path

from aiconfigurator.cli.api import cli_estimate
from aiconfigurator.sdk.config_adapter import (
    AdapterOverrides,
    DynamoRecipeSource,
    adapt_config,
    to_cli_estimate_kwargs,
)

recipe_path = Path("benchmarks/<benchmark-name>/recipe.yaml")
report = adapt_config(
    DynamoRecipeSource(
        deployment=recipe_path,
        source_reference=str(recipe_path),
    ),
    AdapterOverrides(
        # Optional when the recipe already provides these values.
        backend_version="0.5.5",
        database_mode="SILICON",
    ),
)

for outcome in report.outcomes:
    if outcome.status == "rejected":
        for diagnostic in outcome.diagnostics:
            print(outcome.point_id, diagnostic.code, diagnostic.message)
        continue

    request = outcome.request
    assert request is not None

    # Canonical aic-estimate-request/1.0.0 payload.
    print(request.model_dump_json(indent=2))

    # Run only when estimation is explicitly requested.
    result = cli_estimate(**to_cli_estimate_kwargs(request))
    print(outcome.point_id, result.ttft, result.tpot)
```

For a recipe with `benchmark.concurrencies: "8x16x32"`, the report contains `concurrency-8`, `concurrency-16`, and `concurrency-32` in source order. Helm, Jinja, and Slurm templates must be rendered first; the adapter never executes recipe commands.

## New Features
- Added a versioned configuration adapter for InferenceX and Dynamo/Dynamo CI serving configurations.
- Converts supported configurations into validated estimate requests for aggregated and disaggregated deployments.
- Supports workload points, topology, hardware, backend, runtime, quantization, speculative decoding, and explicit overrides.
- Converts adapted requests into CLI estimation inputs and reports structured diagnostics for rejected configurations.
- Added a versioned JSON Schema for estimate-request validation.

## Documentation
- Added guidance covering supported formats, schema, limitations, deterministic behavior, workflows, and package usage.


src/aiconfigurator/sdk/config_adapter/dynamo.py
```diff
@@ -0,0 +1,888 @@
...
+        overrides.nextn
+        if overrides.nextn is not None
+        else (_as_int(source_nextn, "speculative depth") if source_nextn else 0)
+    )
+
+    free_fraction = overrides.free_gpu_memory_fraction
+    if free_fraction is None:
+        raw_fraction = _runtime_value(
+            services,
+            ("free-gpu-memory-fraction", "gpu-memory-utilization"),
```
jasonqinzhou: **[P1] Preserve the standard SGLang runtime flags here.** SGLang deployments use `--mem-fraction-static` and `--context-length` (the repo's own `backend_config_mapping.yaml` and the new `dynamo_ci.py` adapter use those names), but these two `_runtime_value` calls only recognize the vLLM/TRT-LLM aliases. I reproduced a valid SGLang DGD containing `--mem-fraction-static 0.73 --context-length 4096` being marked adapted with both `runtime.free_gpu_memory_fraction` and `runtime.max_seq_len` set to `None`; lowering then omits both kwargs. That makes the estimate use a different KV-cache budget/context limit from the deployment. Please add the SGLang aliases to both lookups and cover a standard DGD using the actual SGLang flags.

src/aiconfigurator/sdk/config_adapter/dynamo.py
```diff
@@ -0,0 +1,888 @@
...
+    if overrides.workload_points is not None:
+        return [
+            _Point(point.point_id or f"point-{index}", point.isl, point.osl, point.concurrency, point.prefix or 0)
+            for index, point in enumerate(overrides.workload_points)
+        ]
+    explicit: list[_Point] = []
+    for document in documents:
+        points = document.get("points")
+        if isinstance(points, list):
+            explicit.extend(_point_from_mapping(point, len(explicit)) for point in points if isinstance(point, Mapping))
```
jasonqinzhou: **[P2] Preserve a rejection for every explicit point instead of filtering malformed entries out.** This `isinstance` guard silently drops non-object items from `points`; with `[valid, "malformed", valid2]` I get only two adapted outcomes. That contradicts the documented ordered one-outcome-per-point/fail-closed contract and can hide a bad benchmark point from callers. Please either emit a rejected placeholder for each invalid entry or reject the explicit points container, and add a mixed valid/invalid regression test.

src/aiconfigurator/sdk/config_adapter/dynamo.py
```diff
@@ -0,0 +1,888 @@
...
+    def replace(match: re.Match[str]) -> str:
+        name = match.group("braced") or match.group("plain")
+        if name not in env:
+            raise ValueError(f"command depends on unresolved environment variable {name!r}")
+        value = env[name]
+        if "$(" in value or "`" in value:
+            raise ValueError(f"environment variable {name!r} is shell-derived, not literal")
+        return value
+
+    return _ENV_PATTERN.sub(replace, command)
```
jasonqinzhou: **[P2] Reject shell parameter forms that remain after literal expansion.** `_ENV_PATTERN` only recognizes `$NAME` and `${NAME}`, so expressions such as `${MODEL_PATH:-Qwen/Qwen3-32B}` (and positional forms such as `$1`) pass through unchanged. I reproduced the former being accepted as the canonical model path even though the adapter promises to reject arbitrary shell-derived values. Please detect remaining shell parameter expressions after substitution, or validate extracted values before constructing the request, so these inputs become structured rejections rather than adapted requests.

src/aiconfigurator/sdk/config_adapter/dynamo.py
```diff
@@ -0,0 +1,888 @@
...
+                assumptions=assumptions,
+            ),
+        )
+        systems = SystemSettingsV1(prefill=system, decode=overrides.decode_system_name or system)
+
+    speculative_values = [
+        _flag(service, "num-speculative-tokens", "speculative-num-steps", "speculative-token-num")
+        for service in services
+    ]
+    active_speculation = any(value not in (None, False, "0", 0) for value in speculative_values) or any(
```
jasonqinzhou: **[P2] Do not activate speculation solely because a speculative flag key exists.** The first predicate correctly treats `--num-speculative-tokens 0` as disabled, but the following key-name scan makes `active_speculation` true anyway. A valid DGD with that explicit zero is therefore rejected for missing `nextn_accepted`. Please use the same disabled-value handling as `dynamo_ci._speculation` (while still recognizing algorithm/model enable flags), and add a zero-depth regression test.

README.md
```diff
@@ -191,6 +192,29 @@ agg, disagg = cli_support(model_path="Qwen/Qwen3-32B-FP8", system="h200_sxm")
...
+```python
+from aiconfigurator.sdk.config_adapter import (
+    AdapterOverrides,
+    DynamoRecipeSource,
+    adapt_config,
+    to_cli_estimate_kwargs,
+)
+
+report = adapt_config(
+    DynamoRecipeSource("deploy.yaml", "perf.yaml"),
```
jasonqinzhou: **[P2] Make this example use `Path` objects, as the detailed guide now does.** `DynamoRecipeSource` deliberately treats a plain `str` as YAML text, not as a filesystem path, so these two values parse as scalar YAML and the example returns a rejected report instead of loading either file. Please import `Path` and wrap both filenames (or pass actual YAML text) so the primary README example exercises the supported API.

jasonqinzhou (changes requested):

## What this PR does

This PR adds a versioned upper-SDK adapter that converts InferenceX records, DynamoGraphDeployments, and concrete Dynamo CI recipes into validated AIC estimate requests, plus a repo-local agent workflow and wheel-layer safeguards.

The ordered outcome model, strict cross-field validation, non-executing source parser, and explicit upper/core packaging checks show thoughtful work on both safety and release boundaries.

## Why this score

The 5.8 score and REQUEST_CHANGES recommendation reflect one blocking correctness gap and four narrower contract/documentation issues. The primary problem is that standard SGLang DGD flags for memory fraction and context length are silently omitted, so the generated estimate can describe a different runtime budget than the source deployment; malformed point entries and unresolved shell expressions can also bypass the advertised fail-closed reporting behavior.

src/aiconfigurator/sdk/config_adapter/dynamo.py
jasonqinzhou: Still reproducible at 07ade14e. The placeholder preserves an outcome only when workload overrides are absent. With AdapterOverrides(isl=512, osl=64, concurrency=4), [valid, "malformed", valid2] yields three adapted outcomes because the global overrides fill the placeholder fields. Please retain an explicit invalid-entry marker (or reject the explicit points container) so a malformed source point cannot become adapted through overrides, and add that regression case.

jasonqinzhou: Fixed in 780504dd3. Malformed explicit points now carry an explicit mapping error into request construction, so global workload overrides cannot convert them into adapted requests. The regression test supplies isl, osl, and concurrency overrides and still verifies the malformed middle entry is rejected in source order. Local validation: 89 relevant tests passed, plus Ruff lint and format checks.

jasonqinzhou (approved):
Approved on exact head 780504dd38c80b12a2f32c0751e80d865c53181c after all review threads were addressed. The final fail-closed workload-point fix passed 89 relevant local tests and Ruff lint/format checks. CI is still running and must be fully green before merge.

Issue #1357, opened 2026-07-14
Python engine-step path: freeze and retirement plan (post-#1355 full parity)


tianhaox:
Tracking issue for freezing and retiring the Python engine-step (op-graph walk) path, now that #1355 lands full Python↔Rust parity for the SILICON / HYBRID / EMPIRICAL database modes (engine-step 303, compile-engine + perf gate, `parity_tests/` 361 green; typed errors + empirical provenance tier cross the FFI; capability/selection audits closed).

## Already landed in #1355 (defenses)

- [x] Full numeric + selection-rule parity, guarded by `rust/aiconfigurator-core/parity_tests/`
- [x] Typed errors (`PerfDataNotAvailableError` / `EmpiricalNotImplementedError`) and empirical provenance tiers across the FFI (support-matrix `HYBRID_PASS` labelling works on rust-routed runs)
- [x] Opspec coverage tripwire (`tests/unit/sdk/test_opspec_coverage.py`) — new `Operation` subclasses must convert or carry a justified exemption
- [x] Freeze discipline doc (`.claude/rules/rust-core/parity.md`) + `/rust/` CODEOWNERS routing (same owners as `/src/`)

## Phase 1 — freeze (small PR)

- [ ] Flip `engine_step_backend` default to `rust` for SILICON/HYBRID/EMPIRICAL; Python becomes explicit opt-out (one release of escape hatch)
- [ ] Upgrade silent Python-step fallbacks to warn-once + counters (SOL modes, AFD, VL encoder, `RustEngineUnsupportedError` rescue) — freeze-window telemetry for "who still uses the Python step"
- [ ] **Repo setting (manual):** make the `rust-engine-step-parity` CI job a required check on `main` — this is the actual enforcement of the freeze
- [ ] Decide SOL/SOL_FULL: port the SOL dispatch branches to Rust (small — per-op `get_sol` already exists Rust-side) and close the routing gate's last delegation, or declare them permanently Python-side diagnostics

## Phase 2 — retirement prerequisites (parallel PRs)

- [ ] **Energy/power across the FFI**: thread the energy data column through the interp core, every loader, every op, and the FFI returns (rust-routed reports currently show `power=0.0W`)
[...]

PR #1521, opened 2026-08-12
feat(sdk): retire the Python engine-step path — the compiled engine is the only step executor (Phase 2 PR-3)

tianhaox:
> [!NOTE]
> **Rebuilt from `main`** (`fa8ebbb3`, post-#1508/#1442). An earlier draft of this PR was stacked on #1508 and scoped before #1461 (FPM-on-rust) landed; rather than rebase through the inverted dispatch order, the branch was rebuilt commit by commit against current main — and #1461 let this PR delete MORE than the draft did (the Python FPM walk goes too).

## What

Phase 2 PR-3 of the dedup plan (`aic-core/rust/aiconfigurator-core/docs/python-dedup-plan.md`), following #1454 (PR-1, the default flip), #1496 (PR-2, gap closure + the golden anchor), and #1508 (PR-2.5). **The compiled Rust engine becomes the only engine-step executor — op-level and FPM models alike**; the Python step path, the Python FPM walk, the routing escape hatch, the live-Python golden capture harness, and the python-vs-rust comparison tooling are deleted. Eleven layered commits: ten Python-side, plus one focused Rust commit (a perf-table sharing memo the retirement itself exposed the need for — §6, values proven unchanged).

**Scope note, up front:** the plan's original PR-3 ("delete `operations/*.py` `query()` + `perf_database.query_*` wholesale") is still not fully deliverable — the survey (recorded in the plan's new disposition section) found the per-call query stack load-bearing for the AFD comm ops (which query the nccl/p2p/mem_op tables EMPIRICALLY), the sanity notebook (10 `query_*` methods incl. SOL_FULL raw tuples), and internal SOL/empirical couplings inside the op bodies. So the per-call stack stays, and its retirement is the recorded sequel (PR-5 in the ladder below; #1357 Phase 3). What DID become deliverable since the plan was written: #1461 ported FPM to the compiled engine including the `fpm_sol.rs` roofline, which removed the last consumer of the Python phase runners — this PR deletes them entirely.

### 1. `fix(sdk)`: close the #1461 leftover FPM guards
#1461 moved only `run_mixed` to rust-first; `_run_static_breakdown` and `_get_genonly_step_latency` kept #1384's `forward_model != "fpm"` short-circuits, so **FPM static/decode always ran the Python walk even under an explicit `rust` request — and two of the four FPM parity tests silently compared python against python.** The guards are removed (rust-first everywhere); verified answer-preserving to full precision on the synthetic parity fixture (all 10 static points + genonly, incl. both error-symmetry cases) before the change.

### 2. `feat(sdk)`: retire the Python engine-step path
- `base_backend` loses the non-FPM Python branches: `run_mixed`'s three-pass composition (+ the `op_filter` plumbing #1508 added for it), the encoder `op.query()` fallback loop, and every `RustEngineUnsupportedError` "parity by delegation" arm — an inexpressible op graph is a **hard error** (the opspec tripwire keeps that unreachable for shipped ops); a non-`PerfDatabase` database on any step surface raises a `TypeError` explaining the disk-identity requirement. The agg cache key drops its now-constant backend slot.
- `"python"` becomes a **warn-once deprecation no-op** with true as-if-unset semantics (config or env; a config-level `"python"` no longer shadows the env, and the non-`PerfDatabase` delegation still applies — retained one release cycle). Unknown values now **raise** instead of silently taking the Python path (`"auto"` used to). The gate keeps only the non-`PerfDatabase` delegation (sole consumer: AFD's per-call loop).
- One SDK-facing contract note: FPM mixed steps now always report through the uniform `StepEstimate` breakdown — the decode marginal rides the literal `"generation_attention"` key instead of `"fpm_forward_decode"`; totals unchanged to full precision.

### 3. `feat(sdk)`: delete the Python FPM walk
The phase runners (`_run_context_phase`/`_run_generation_phase`), `_get_fpm_mix_step_latency`, and `fpm_forward.py`'s entire query machinery (query/query_totals/query_pass_baseline, parquet+sidecar loader/validators, perf_interp configs, `_oplevel_sol_fn`) go — the Rust core owns FPM end to end (`perf_database/fpm_forward.rs`, `operators/fpm_forward.rs`, `operators/fpm_sol.rs`). `FPMForwardOp` keeps exactly the construction surface `_to_opspec` serializes and the memory model sums (identity, weights, sol_ops). The FPM parity class becomes **live-Rust vs frozen references** captured from the Python walk immediately before deletion (inline `_FPM_*_FROZEN` — this class's dataset is generated per-run, so it sits outside the golden files). `test_fpm_forward.py` drops the query/loader classes (Rust twins live in `fpm_forward.rs`'s test module) and keeps construction + routing tests, now exercising the compiled engine.

### 4. `chore(parity)`: freeze the goldens
`regenerate_goldens.py` (+ guard tests) deleted — no Python reference left to capture; the Gate-2 fixtures are the permanent regression oracle. **`pin_goldens.py`** handles the post-freeze lifecycle: append-only pinning for cases added after the freeze (values from the live rust engine, provenance-marked in `post_freeze_pins`), and `--refresh`/`--refresh-all` for deliberate modeling changes where **the golden diff is the review artifact**. Same guards as the retired script (clean tree, thread caps, byte-reproducible, all-payloads-before-any-write). The relative Rust-vs-Python perf gate (+ its CI step) retires with its denominator — running its "python" arm today would time rust against itself; `benchmark_engine_step.py` becomes a rust-only timing tool.

### 5. `chore(tools)` / `docs`
`scan_rust_parity.py`, `--compare-engine-step-backends`, and the PR-e2e live-python pass retire (HYBRID_PASS provenance labelling untouched); `collect_static.py`'s baseline pin flips to `"rust"`. The dedup plan gains the PR-3 disposition + sequel ladder; `.claude/rules/rust-core/parity.md` gets a **proposal-only** golden-diff rewrite (`docs/parity-rule-golden-diff-proposal.md` — rule files are human-owned) that also keeps the dual-implementation rule for the still-dual per-call surface.

### 6. `perf(rust-core)`: share parsed perf tables across engines with the same DB identity
Flipping the Prediction Regression Gate's pin to `"rust"` exposed a cost the retired Python path had been hiding: every compiled engine owned a private `PerfDatabase`, so each engine's **first query lazily re-parsed the same parquet files** (~0.5 s per engine on data-rich systems). Sweeps compile one engine per model x parallelism x quant identity — the gate builds ~650 per combo — so its collect step went from ~4 CPU-minutes (python, one shared DB) to **102 CPU-minutes** and a 30-minute CI timeout. The fix is a process-wide `Weak` memo of `Arc<PerfTables>` keyed by the full load identity: only the immutable-after-load tables are shared, while every engine keeps fresh mode/policy fields, memo caches, and its own provenance accumulator (no cross-engine state coupling; 4 new cargo tests pin exactly that). Gate collect drops back to **3 m 08 s CPU** — faster than the python side it replaced. This also removes the same per-config tax from every multi-config CLI/pareto sweep on `main` (present since the #1454 default flip, just never load-bearing in CI before).

### 7. `fix(core-sdk)`: public-compat restorations (review round 4)
Review-driven (see thread): the exported `Mamba2` composite **stays** (deprecated, opspec-EXEMPT, its `query_gemm`/`query_mem_op` substrate survives until the per-call retirement); `FPMForwardOp` keeps its legacy constructor layout `(phase, model_config, model_path, sol_fn=None, weight_bytes=0.0, sol_ops=None)` — positional callers keep their meaning, a passed `sol_fn` callback raises a targeted migration error (it cannot cross the compiled boundary), omitted `sol_ops` still raises. New **baseline** tests pin the operations export surface and the constructor signature literally (the facade-vs-facade contract tests can't catch a name vanishing from both sides at once). The unknown-backend `ValueError` names the re-resolved value actually rejected, and the CLI guide names `--engine-step-backend`'s five real modes.

## Evidence

| Check | Result |
|---|---|
| **Prediction Regression Gate, full grid, python-old vs rust-new** | **134,028 rows across 28 combos — zero differences** (identical statuses AND values; run locally with `report.py` against a merge-base python-pinned snapshot). The strongest whole-surface parity statement so far: every bundled model x parallelism x quant x shape the gate samples answers identically on the compiled engine |
| Gate collect cost (the CI-timeout regression, §6) | before the memo: old(python) 3 m 39 s CPU vs new(rust) **102 m 26 s** CPU (30-min job timeout on 4-core runners); after: **3 m 08 s** CPU — faster than the python side it replaces |
| engine-step + compile-engine parity (live Rust vs frozen goldens/references) | **447 passed** on the rebased tree — goldens byte-untouched (incl. main's new wideep keys); the FPM class asserts against its frozen `_FPM_*_FROZEN` references |
| Unit + cross-package suites | **3,378 passed / 4 skipped, 0 failures** on the final rebased tree. Four torch-importing files excluded locally: their collection breaks identically on pristine `main` under CPU-only torch (TORCH_LIBRARY registration; CI's container is unaffected) |
| cargo test (default + --no-default-features) | **463 passed** each (421 pre-existing + main's large-EP additions + 4 new shared-tables tests + review-round additions) |
| FPM reroute answer-preservation (pre-change probe) | 11/11 points match the Python walk to full precision |
| `pin_goldens.py` (append-only, clean tree) | "nothing to pin" — matrix↔fixture population match, zero writes; `--refresh-all --refresh <bogus>` aborts before any compute |
| ruff check / format | clean |
| Behavior deltas | explicit-python requests behave as if unset (compiled engine on real databases; the synthetic-database delegation still applies); synthetic-DB steps raise `TypeError`; unknown backend values raise `ValueError` naming the re-resolved value; FPM mixed per-op keys use the uniform breakdown contract; support-matrix tasks and replay commands pin `--engine-step-backend rust`; `FPMForwardOp(sol_fn=...)` raises a migration error |
| LoC | 46 files, +1,694 / −5,092 (net **−3,398**) |

## Notes for reviewers
- **One Rust commit, zero numeric changes**: the shared-tables memo (§6) touches loading, not math — proven by the gate comparison above (132,192 rows, python-old vs rust-new, zero differences) and by the byte-untouched goldens. Everything else is Python-side; the only numeric surface touched is routing (and the reroute in commit 1 was probed answer-preserving before landing).
- `runtime.rs` doc-comments still cite the (now retired) Python functions as their historical spec; left as-is deliberately (the golden-diff rule proposal describes the new source-of-truth story) — §6 deliberately confines itself to the loading layer.
- Known pre-existing local-env failures (NOT this PR): 7 `test_moe_dispatch.py` tests fail identically on pristine `main` under CPU-only torch (Sm100 detection); CI's container is unaffected.
- The freeze-window telemetry (#1357 Phase 1) served its purpose: the remaining `python_step_fallback_counts()` reasons are the two permanent AFD-side delegations.

## What comes after this PR (the remaining ladder)

- **PR-4 — sanity-notebook re-oracle** (independent of this PR; can run in parallel): a small Rust FFI addition exporting the per-op `(sol_time, sol_math, sol_mem)` tuple (the engine already computes both components internally), then `tools/sanity_check/validate_database.ipynb` + `create_charts.py` re-point from `PerfDatabase.query_*` to the compiled engine's per-op evaluation. This is the prerequisite that unblocks the per-call deletion — the notebook is the `query_*` facade's biggest live consumer.
- **PR-5 — per-call query-stack retirement** (needs this PR + PR-4): rewire `Operation.query()` as a thin delegation to the op-list evaluation FFI (per-call `database_mode` support incl. SOL/SOL_FULL), then delete the duplicated Python latency math — the 8 op families' `_query_*_table` bodies (~2.9 kLoC of SOL/silicon/empirical), the 30 `PerfDatabase.query_*` facade methods, Python `perf_interp/`, and `util_empirical`'s math (provenance constants stay). The AFD synthetic comm ops either gain OpSpec variants or remain the last per-call island. #1357's family-by-family cadence with a parity run per family.
- **PR-6 — deprecation removal** (time-locked): drop the `"python"` `engine_step_backend` value, the routing gate, and the CLI choice after the one-release bake this PR starts.

Post-review hardening (CodeRabbit round + an adversarial self-review pass): the deprecation shim's as-if-unset semantics fix above, upfront `--refresh` validation and a before-any-write `pytest.skip` guard in `pin_goldens.py` (plus per-case handle-cache reuse), the support-matrix rust pin moved inside `_create_task`, the dead `--aic-rust-core-autobuild` option removed, and the parity helpers' vestigial engine-selector parameter dropped (passing `"python"` would have been a silent rust-vs-rust self-comparison).

Refs #1357 (Phase 3 execution, revised scope). Builds on #1454, #1496, #1508, #1461.


aic-core/rust/aiconfigurator-core/docs/parity-rule-golden-diff-proposal.md
```diff
@@ -0,0 +1,118 @@
...
+  diff. Per-op query math is still dual-implemented (FPM/AFD/tooling) until
+  the #1357 Phase-3 retirement.
+paths:
+  - "rust/aiconfigurator-core/**"
+  - "src/aiconfigurator/sdk/operations/**"
+  - "src/aiconfigurator/sdk/perf_database.py"
+  - "src/aiconfigurator/sdk/perf_interp/**"
+  - "src/aiconfigurator/sdk/engine.py"
+  - "src/aiconfigurator/sdk/rust_engine_step.py"
+  - "tests/unit/sdk/test_opspec_coverage.py"
```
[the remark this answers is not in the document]
tianhaox: Fixed in 81c9e450: every path in the proposed rule now uses the repository-root-relative REAL source locations (aic-core/rust/aiconfigurator-core/** and aic-core/src/aiconfigurator_core/sdk/** — the old entries named the post-restructure compatibility shims), and the proposal's motivation list gained a bullet calling the stale paths out explicitly.

aic-core/rust/aiconfigurator-core/docs/python-dedup-plan.md
```diff
@@ -5,7 +5,81 @@ SPDX-License-Identifier: Apache-2.0
 
 # Phase 2 Execution Plan — Rust-default flip and Python latency-path removal
 
-**Status (2026-08-06): PR-1 (#1454) MERGED 2026-08-05; PR-2 in flight.**
+**Status (2026-08-12): PR-1 (#1454) and PR-2 (#1496) MERGED; PR-2.5 (#1508)
+in flight; PR-3 delivered as the ENGINE-STEP-path retirement — see "PR-3
+disposition" below.** The original PR-3 scope ("delete the per-call query
+stack wholesale") was revised after FPM (#1384, merged 2026-08-10, after
+this plan's 2026-08-06 revision) landed a live consumer of that stack.
```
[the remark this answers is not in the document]
tianhaox: Fixed in 81c9e450: the status line now reads "PR-3 (#1521 — the PR carrying this text) delivers the ENGINE-STEP-path retirement and is in flight until it merges" — true both before and after merge.

aic-core/rust/aiconfigurator-core/parity_tests/pin_goldens.py
```diff
@@ -0,0 +1,407 @@
...
+    """Tracked-file modifications other than the goldens themselves (this
+    script's output) and untracked files."""
+    dirty: list[str] = []
+    for line in porcelain.splitlines():
+        if not line.strip() or line.startswith("??"):
+            continue
+        paths = line[3:].split(" -> ")
+        if all(path.strip().strip('"').startswith(_GOLDEN_REL_PREFIX) for path in paths):
+            continue
+        dirty.append(line[3:].strip())
```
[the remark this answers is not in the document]
tianhaox: Fixed in 81c9e450: untracked files now count as dirty. Agreed the retired capture script's contract was wrong to inherit here — unlike `git describe`, a pin's VALUE can depend on untracked inputs (an untracked parquet under systems/data changes what the live engine loads), so the docstring now states that rationale.

docs/cli_user_guide.md
```diff
@@ -16,8 +16,8 @@ These flags are shared across modes (a few are sweep-only, as noted):
...
+- `--engine-step-backend`: Engine-step latency backend. The compiled Rust engine is the only step executor; `rust` is the only live value. `python` is DEPRECATED and now a no-op — it warns once and runs on the compiled engine anyway (accepted for one release cycle, then removed). Accepted by all modes but inert in `generate`, which performs no latency estimation. (`default`, `exp`, `generate`, `estimate`)
```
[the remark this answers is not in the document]
tianhaox: Fixed in 81c9e450: the flag description now ends with "any other value raises an error".

tools/support_matrix/support_matrix.py
```diff
@@ -922,45 +687,20 @@ def _attempt(mode: str, db_mode: str | None) -> tuple[str, str | None, str, bool
...
                         mode=mode,
                         model=model,
                         system=system,
                         backend=backend,
                         version=version,
                         constraints=constraints,
-                        engine_step_backend="python" if compare_engine_step_backends else None,
+                        engine_step_backend=None,
                         database_mode=db_mode,
                     )
```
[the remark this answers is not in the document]
tianhaox: Fixed in 81c9e450: SupportMatrix now passes engine_step_backend="rust" for the executed task (the config value outranks the ambient env var, so an unknown ambient value can no longer flip PASS/FAIL host-dependently), and _support_matrix_row_command emits --engine-step-backend rust so the replay command reproduces the executed path. The replay-expectation test was updated to match.

aic-core/src/aiconfigurator_core/sdk/operations/dsv4.py
```diff
@@ -1158,38 +1158,50 @@ def _csa_topk_top_last(cls, database: PerfDatabase, isl: int, step: int, native_
...
+        # in one process — ``databases_cache`` itself keys on the flag).
         key = (
             database.systems_root,
             database.system,
             database.backend,
             database.version,
             database.enable_shared_layer,
+            database.strict_provenance,
             native_heads,
         )
```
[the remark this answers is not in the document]
tianhaox: Skipped — false positive: PerfDatabase.enable_shared_layer is a @property (perf_database.py:2548-2549; verified isinstance(PerfDatabase.__dict__['enable_shared_layer'], property) is True), so the dsv4 cache key stores the resolved bool, not a bound method, and cache reuse across equivalent instances works. The same property-access pattern is the established convention across the op families' _cache_key helpers (e.g. mamba.py).

aic-core/src/aiconfigurator_core/sdk/operations/__init__.py
```diff
@@ -38,7 +38,7 @@
 from aiconfigurator_core.sdk.operations.embedding import Embedding
 from aiconfigurator_core.sdk.operations.fpm_forward import FPMForwardOp
 from aiconfigurator_core.sdk.operations.gemm import GEMM
-from aiconfigurator_core.sdk.operations.mamba import GDNKernel, KDAKernel, Mamba2, Mamba2Kernel
+from aiconfigurator_core.sdk.operations.mamba import GDNKernel, KDAKernel, Mamba2Kernel
```
jasonqinzhou: [P1] **Preserve the exported `Mamba2` compatibility surface.**

Although the composite is unused by in-repo model builders, it was part of the canonical `operations.__all__` surface and is re-exported through the user-facing `aiconfigurator.sdk.operations` compatibility package. On this head, `from aiconfigurator.sdk.operations import Mamba2` now raises `ImportError`. The cross-package contract tests compare the two current facades, so they stay green even when a previously exported name disappears.

Please keep this public symbol for a deprecation window (or provide a compatibility stub with a clear migration) and add a baseline import-contract test. An internal call-site search does not establish that downstream SDK users are absent.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
```diff
@@ -154,449 +83,25 @@ def _norm_identity(value) -> str:
...
-
     def __init__(
         self,
         phase: str,
         model_config,
         model_path: str,
-        sol_fn: Callable[..., float] | None = None,
         weight_bytes: float = 0.0,
-        sol_ops: list | None = None,
+        *,
```
jasonqinzhou: [P1] **Preserve the exported `FPMForwardOp` constructor contract.**

`FPMForwardOp` is in the canonical `operations.__all__` surface and is re-exported through `aiconfigurator.sdk.operations`, but `main` exposes `(phase, model_config, model_path, sol_fn=None, weight_bytes=0.0, sol_ops=None)` while this head exposes `(phase, model_config, model_path, weight_bytes=0.0, *, sol_ops)`. Existing `sol_fn=` callers now get an unexpected-keyword `TypeError`, a fourth positional callback is reinterpreted as `weight_bytes`, and callers that supplied `sol_ops` positionally are rejected. The current import-contract tests compare the two rewritten facades and do not pin this baseline signature.

Please retain the legacy parameter layout for a deprecation window, or provide an explicit compatibility/migration path without shifting positional meanings, and add a baseline signature test. If arbitrary `sol_fn` callbacks cannot cross the compiled boundary, fail that legacy case with a targeted migration error rather than silently rewriting the public call shape.

aic-core/src/aiconfigurator_core/sdk/rust_engine_step.py
```diff
@@ -280,39 +284,60 @@ def _normalize_tuning_iterations(iterations: dict[str, Any] | list[Any]) -> list
...
     """
     backend = getattr(runtime_config, "engine_step_backend", None) or os.environ.get(ENGINE_STEP_BACKEND_ENV)
     requested = str(backend).lower() if backend else None
+    if requested == "python":
+        _warn_python_backend_deprecated(requested)
+        # The no-op contract: ignore the retired value entirely and
+        # re-resolve the remaining signal (a config-level "python" no longer
+        # shadows the env), then proceed as if it were never set.
+        env_value = os.environ.get(ENGINE_STEP_BACKEND_ENV)
+        requested = str(env_value).lower() if env_value else None
```
jasonqinzhou, line 333: [P2] **Report the re-resolved environment backend in the error.**

When the config says deprecated `python` and the environment contains an invalid value such as `auto`, this branch correctly re-resolves `requested` to `auto`, but the later `ValueError` still interpolates the stale `backend` value. I reproduced a warning that `python` is accepted as a no-op followed by `unknown engine_step_backend 'python'`, while the actual rejected value is the environment's `auto`.

Please carry the effective value/source through the re-resolution (or format the error from `requested`) and add the config-`python` plus invalid-env case so the diagnostic identifies the value users must fix.

docs/cli_user_guide.md
```diff
@@ -16,8 +16,8 @@ These flags are shared across modes (a few are sweep-only, as noted):
...
+- `--engine-step-backend`: Engine-step latency backend. The compiled Rust engine is the only step executor; `rust` is the only live value. `python` is DEPRECATED and now a no-op — it warns once and runs on the compiled engine anyway (accepted for one release cycle, then removed); any other value raises an error. Accepted by all modes but inert in `generate`, which performs no latency estimation. (`default`, `exp`, `generate`, `estimate`)
```
jasonqinzhou: [P2] **Document the actual mode set for `--engine-step-backend`.**

The parser exposes this option on `default`, `recommend`, `exp`, `generate`, and `estimate`, but not `support`. The current sentence both says “all modes” and omits `recommend` from its parenthetical list. Please name those five modes explicitly and retain the note that `generate` accepts the option but does no latency estimation.

jasonqinzhou (changes requested):

## What this PR does

This PR retires the Python engine-step and Python FPM execution paths, makes the compiled Rust engine the sole step executor, freezes the former Python outputs as regression goldens, and replaces comparison tooling with a Rust-side golden-pinning workflow. The latest two commits fix deprecated-backend routing and harden the pinning, parity, and support-matrix paths.

The incremental update directly fixes the synthetic-database routing bug and the refresh-option failure from the prior review, adds focused regression coverage, and simplifies the parity helpers so callers cannot accidentally create a vacuous Rust-vs-Rust comparison.

## Why this score

Score 5.8/10; REQUEST_CHANGES with high confidence. The latest delta resolves two prior findings and passes focused tests, but the current head still removes the exported Mamba2 symbol and now rewrites the exported FPMForwardOp constructor without a compatibility window, leaving two confirmed public SDK regressions. The CLI mode documentation remains inaccurate, and the re-resolved environment-backend error reports the stale config value instead of the invalid environment value.

aic-core/src/aiconfigurator_core/sdk/operations/__init__.py
tianhaox: Restored — and you were right that an internal call-site search doesn't establish absence of downstream users. `Mamba2` is back exactly as on `main` (its `query_gemm`/`query_mem_op` substrate survives until the per-call stack retirement, so it stays fully functional), marked DEPRECATED in its docstring with the disposition pointer, with its opspec-coverage `EXEMPT` entry back. Added `test_operations_baseline_exports_survive` to the cross-package import contract: a literal frozen baseline of the `operations` export surface, so a name vanishing from both facades at once now fails a test instead of slipping through the facade-vs-facade comparison.

aic-core/src/aiconfigurator_core/sdk/operations/fpm_forward.py
tianhaox: Restored the legacy layout `(phase, model_config, model_path, sol_fn=None, weight_bytes=0.0, sol_ops=None)` — positional `weight_bytes`/`sol_ops` callers keep their meaning. Since an injected Python roofline callback genuinely cannot cross the compiled boundary, a non-None `sol_fn` now raises the targeted migration `TypeError` you suggested (pointing at `sol_ops`), and omitting `sol_ops` still raises `ValueError` (main's "exactly one" contract minus the retired half). Baseline tests added: `test_fpm_forward_op_keeps_legacy_constructor_layout` pins the literal signature in the import contract, and `test_fpm_legacy_positional_layout_preserved` + `test_fpm_sol_fn_raises_targeted_migration_error` pin the behavior.

aic-core/src/aiconfigurator_core/sdk/rust_engine_step.py
tianhaox: Fixed — the `ValueError` now interpolates the re-resolved `requested` value (your exact scenario: config `"python"` + env `"auto"` now reports `'auto'`). Regression test: `test_invalid_env_behind_deprecated_python_config_is_named_in_the_error`.

docs/cli_user_guide.md
tianhaox: Fixed — the sentence now names the five modes that actually accept the option (`default`, `recommend`, `exp`, `generate`, `estimate`), notes `support` doesn't, and keeps the `generate`-is-inert note.

aic-core/rust/aiconfigurator-core/docs/parity-rule-golden-diff-proposal.md
```diff
@@ -0,0 +1,124 @@
...
+   moved under `aic-core/rust/`). The proposal below uses the
+   repository-root-relative real locations.
+
+One thing the rule must KEEP: the per-call Python query stack
+(`operations/*.py` `_query_*_table` bodies, `perf_database.query_*`,
+`perf_interp/`) is still live — the AFD comm ops and the sanity notebook
+consume it — so per-op query math remains DUAL-implemented until the #1357
+Phase-3 sequel deletes it family-by-family. The dual rule therefore
+survives for the per-op surface, while the step-level rule becomes
+golden-diff.
```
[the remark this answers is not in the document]
tianhaox: Added the third consumer: the retained-boundary sentences now name the op-internal SOL/empirical couplings alongside the AFD comm ops and the sanity notebook.

aic-core/rust/aiconfigurator-core/docs/parity-rule-golden-diff-proposal.md
```diff
@@ -0,0 +1,124 @@
...
+
+`src/aiconfigurator/sdk/operations/**` `_query_*_table` bodies,
+`perf_database.query_*`, and `perf_interp/**` remain live consumers'
+substrate: the AFD comm ops (empirical nccl/p2p/mem_op queries plus the
+`_sum_latency` fallback loop) and
+`tools/sanity_check/validate_database.ipynb`. Until the #1357 Phase-3
+retirement lands, a latency-affecting change to that per-call math MUST
+still be mirrored in the corresponding Rust operator (and vice versa), with
+an anchor per Rule 1 — otherwise AFD/tooling silently diverge from the
+engine step.
```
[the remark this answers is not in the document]
tianhaox: Fixed — Rule 2 now names `aic-core/src/aiconfigurator_core/sdk/operations/**` (the real location), matching the frontmatter.

aic-core/rust/aiconfigurator-core/parity_tests/pin_goldens.py
```diff
@@ -0,0 +1,410 @@
...
+    references["wideep_sglang::mixed_step"] = float(sglang_handle.mixed_step_latency(1024, 2, 1024, 4, 0))
+    references["wideep_sglang::decode_step"] = float(sglang_handle.decode_step_latency(2, 1024, 4))
+
+    _m, _b, _d, trtllm_spec = compile_parity._build_wideep_trtllm()
+    trtllm_handle = compile_parity._handle_from_spec_json(trtllm_spec)
+    wt_ctx, wt_gen, _ = trtllm_handle.run_static(batch_size=1, isl=1024, osl=4, prefix=0, stride=1)
+    references["wideep_trtllm::static_ctx"] = float(wt_ctx)
+    references["wideep_trtllm::static_gen"] = float(wt_gen)
+
+    return references
```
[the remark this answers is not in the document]
tianhaox: This one is stale against the current head: the `matched_refresh` accumulator it describes no longer exists. Since ae7ce02, `--refresh` names are validated against the full key universe up front in `main()` — before any live-engine compute, and unconditionally (with or without `--refresh-all`), so a typo aborts in milliseconds and nothing can fail after computation. When both flags are given, `--refresh-all` recomputes every record (a strict superset of any `--refresh` list, see `_plan`) and the run completes normally. Verified empirically: `pin_goldens.py --refresh-all --refresh bogus_key` aborts in milliseconds with `--refresh names keys outside every test matrix: ['bogus_key']` and zero writes.

tianhaox:
Both P1s and both P2s are addressed in `fix(core-sdk): restore the public compat surface and review fixes (PR review round 4)` — details on each inline thread. Two additional things landed in the same push:

1. **A real regression your review window didn't see yet**: flipping the Prediction Regression Gate's pin to `"rust"` exposed that every compiled engine re-parses its perf tables privately (~0.5 s per engine identity on data-rich systems, ~650 identities per gate combo) — the gate's collect step went 4 → 102 CPU-minutes and hit the 30-minute CI timeout. Fixed in `perf(rust-core)` with a process-wide `Weak` memo sharing only the immutable-after-load `PerfTables` across engines with the same DB identity (fresh mode/policy/memo-caches/provenance per engine; 4 cargo tests pin the isolation). Collect is now 3 m 08 s CPU — faster than the python side it replaces — and the full-grid comparison (132,192 rows, python-old vs rust-new) reports **zero differences**, which doubles as the strongest whole-surface parity statement this migration has produced.
2. Rebased onto latest `main` (the large-EP native path #1464/#1448 and the new wideep golden surfaces merged in; the retired capture-workflow helpers they carried are dropped with the rest of the capture harness — their golden values are already frozen in `compile_engine.json` and `pin_goldens.py` re-pins them from the live rust engine).

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -1717,7 +1190,7 @@ def run_agg(
             runtime_config.num_images_per_request,
         )
         cache_key = (
-            self._make_agg_cache_key(isl, osl, b, ctx_tokens, engine_step_backend_key, agg_extra),
+            self._make_agg_cache_key(isl, osl, b, ctx_tokens, agg_extra),
```
jasonqinzhou, line 1193: [P2] Validate backend before returning cached aggregate results

The cache lookup returns before `should_use_rust_engine_step` or `_require_rust_engine_step` runs. Since this PR removed the backend selector from the cache key, a previously cached Rust result can satisfy a call with an invalid backend value such as `auto`, skipping the required `ValueError`; the same path also bypasses synthetic-database rejection and the deprecated-`python` warning. Validate the runtime/database before this lookup and add a cache-hit regression test.

jasonqinzhou: Fixed in `97bebde0f37a4944e23b8580e02635e9d1706a4e`: `run_agg()` now enforces the compiled-engine request/database contract before consulting the aggregate cache. Added cache-hit regressions for both an unknown `auto` selector and a synthetic database. Local validation: Ruff format/check clean; 30/30 focused tests and 112/112 backend unit tests passed.

jasonqinzhou (approved):
Approved at exact head `97bebde0f37a4944e23b8580e02635e9d1706a4e` after re-review. The aggregate cache-hit validation gap is fixed with regression coverage; all 14 review threads are resolved, all 19 checks are green, and the head merges cleanly with current `main` (`d2af06a6b602866c559f493a34429654991e97a8`).

PR #1528, opened 2026-08-13
feat(models): model Gemma 4 vision encoder

jasonqinzhou:
## Summary

- parse `gemma4_vision` alongside the existing Gemma 4 hybrid text/MoE configuration for `google/gemma-4-26B-A4B`
- model the Gemma-specific patch embedding, learned 2D positions, full-2D-RoPE ViT, gated MLP, 3x3 average pooling, standardization, replicated language adapter, and encoder communication
- account for fixed visual soft-token budgets, encoder latency/memory/energy, visual context tokens, and encoder contribution to TTFT
- make support-matrix and tier-2 coverage image-bearing so the encoder path is exercised automatically

This intentionally does not reuse Qwen3-VL pixel-shuffle, partial-RoPE, or two-layer projector assumptions.

## Validation

- `ruff check` on all changed Python files
- `307 passed, 5 warnings` across parser, encoder/runtime, Gemma model, context-parallel, backend, and support-matrix unit/regression tests
- image-bearing support-matrix check on B200 / TRT-LLM `1.3.0rc20`: aggregated PASS, disaggregated PASS, Python/Rust engine-step comparison enabled, silicon provenance
- tier-2 `gemma4_vision_encoder`: OK (`TTFT 97.720922 ms`, `TPOT 3.704695 ms`)
- direct TP4/EP4 estimate for one 672x960 image: encoder latency `8.693332 ms`, encoder memory `1.115152 GiB`, context latency `6.513895 ms`, TTFT `15.207228 ms`

## Known data limitations

The packaged performance database does not contain nonzero encoder energy samples for the queried operations, so the real silicon-backed run reports zero energy/power even though accounting and nonzero-energy propagation are covered by unit tests. Pooling, normalization, 2D RoPE, and standardization use the existing analytical/empirical elementwise model; no end-to-end Gemma 4 vision-tower silicon trace or CPU preprocessing latency is included.

Linear: AIC-1739

## Summary by CodeRabbit

- **New Features**
  - Added Gemma 4 multimodal vision support, including image encoding, pooling, projection, and text integration.
  - Added bidirectional visual attention, configurable image-token budgets, aspect-ratio-aware processing, and multi-image workloads.
  - Added image-aware workload constraints and replay commands.

- **Bug Fixes**
  - Improved vision latency, memory, energy, power, and time-to-first-token estimates.
  - Improved visual-token handling for disaggregation and batch sizing.
  - Improved handling of model metadata failures during support-matrix execution.

- **Tests**
  - Expanded coverage for vision architecture, configuration, parallelism, image workloads, and regression scenarios.


aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -229,14 +229,34 @@ def _encoder_pre_merge_per_visual(
...
+
         Returns ``(tokens_post_merge_per_image, pre_merge_per_image)``.
         Returns ``(0, 0)`` when neither is set (text-only path).
         """
         has_image_dims = runtime_config.image_height > 0 and runtime_config.image_width > 0
+        if isinstance(enc_cfg, common.Gemma4VisionEncoderConfig):
+            if not has_image_dims and runtime_config.num_image_tokens <= 0:
+                return 0, 0
+            tokens_per_image = runtime_config.num_image_tokens or enc_cfg.soft_tokens_per_image
+            if tokens_per_image not in enc_cfg.supported_soft_token_budgets:
```
jasonqinzhou: [P2] Derive Gemma soft-token counts from image aspect ratio

For Gemma 4 images whose resized pooled grid does not exactly fill the configured maximum, this ignores the supplied dimensions and conflates `max_soft_tokens` with the actual output count. `Gemma4ImageProcessor` derives `num_soft_tokens_per_image` from the aspect-ratio-preserving resized grid; for example, a 448x448 image with a 280-token maximum becomes 768x768 and produces 256 soft tokens, while this returns 280. Consequently projector work, injected context length, memory, and TTFT are overestimated. Track the maximum patch budget separately from the actual pooled-token count.

aic-core/src/aiconfigurator_core/sdk/models/gemma4.py
```diff
@@ -112,6 +113,16 @@ def set_gemma4_config(self, cfg: common.Gemma4MixConfig) -> None:
         self._gemma4_config = cfg
         self._build_context_ops()
         self._build_generation_ops()
+        self.encoder_ops = []
+        self.encoder_config = cfg.vision_config
+        if cfg.vision_config is not None:
```
jasonqinzhou, line 153: [P2] Model bidirectional attention for Gemma visual blocks

The bundled checkpoint specifies `use_bidirectional_attention="vision"`, but that flag is discarded and `_build_context_ops()` still uses ordinary causal `ContextAttention` for every sliding-attention layer. Gemma 4 applies `(causal OR same-vision-block) AND sliding_window`, so visual requests perform additional bidirectional attention work that the TTFT estimate currently omits. Carry this contract into the sliding-attention model.

jasonqinzhou (commented):
Automated Codex PR review found two actionable multimodal modeling issues. I’ll address both in a follow-up commit and resolve the threads after verification.

jasonqinzhou (commented):
Automated Codex follow-up review found two additional accuracy issues after tracing Gemma 4 through padded tower execution and legacy disaggregated sizing.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -229,14 +229,34 @@ def _encoder_pre_merge_per_visual(
...
+                raise ValueError(
+                    f"Gemma 4 num_image_tokens must be one of {enc_cfg.supported_soft_token_budgets}, "
+                    f"got {tokens_per_image}"
+                )
+            pre_merge_per_image = tokens_per_image * enc_cfg.pooling_kernel_size**2
```
jasonqinzhou: [P2] Keep the padded patch count for encoder work

When dimensions round below the soft-token budget (for example 448x448), Gemma4ImageProcessor still pads pixel_values to max_soft_tokens * pool² patches, and vLLM executes patch embedding and the vision transformer on that padded tensor before stripping invalid outputs after pooling. Preserve the smaller valid soft-token count for language context, but use the padded maximum for pre-pool encoder latency and activation memory.

aic-core/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -798,13 +860,15 @@ def _parse_hf_config_json(config: dict) -> dict:
             global_head_dim=global_hd,
             sliding_window_size=config.get("sliding_window", 0),
             attention_k_eq_v=bool(config.get("attention_k_eq_v", False)),
+            vision_config=gemma4_vision_config,
```
jasonqinzhou, line 864: [P2] Unwrap the Gemma vision config for legacy disaggregated sizing

DisaggInferenceSession passes extra_params to the visual-token helper. Gemma now nests the encoder under Gemma4MixConfig, so the helper sees a non-VisionEncoderConfig and returns zero. Unwrap vision_config before calculating effective ISL so prefill batch candidates respect image soft tokens.

jasonqinzhou (commented):
Automated Codex final review found one additional encoder-TP memory-accounting issue.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -1502,10 +1526,18 @@ def _get_encoder_component_memory(self, model: BaseModel, num_tokens: int, embed
         enc_cfg = getattr(model, "encoder_config", None)
         activations = 0.0
         if isinstance(enc_cfg, common.VisionEncoderConfig) and num_tokens > 0:
-            # ~3x hidden_size per patch covers QKV, attention output, and FFN intermediates (bfloat16)
-            activations = 2 * num_tokens * enc_cfg.hidden_size * 3
-            # Projected embeddings (all projector instances concatenated along hidden)
-            activations += 2 * embed_tokens * enc_cfg.out_hidden_size * enc_cfg.projector_n_instances
+            if isinstance(enc_cfg, common.Gemma4VisionEncoderConfig):
+                # Peak gated-MLP workspace: input plus gate/up intermediates.
```
jasonqinzhou: [P2] Apply encoder TP sharding to activation memory

When Gemma 4 uses `enable_encoder_dp=False` with `tp_size > 1`, the vision graph shards QKV and gated-MLP intermediates per rank, but this formula charges every rank the full `intermediate_size` and QKV width. Because this value feeds the OOM gate, encoder-TP runs can be rejected or forced onto extra GPUs despite fitting; compute the peak from TP-local widths in this mode.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
jasonqinzhou: Fixed in 00c0ed9a. Gemma now mirrors the upstream aspect-ratio-preserving resize math, keeps the maximum patch budget separate, and derives valid soft-token counts (for example 448x448 -> 256).

aic-core/src/aiconfigurator_core/sdk/models/vit_ops.py
```diff
@@ -221,3 +221,168 @@ def build_encoder_ops(enc_cfg: common.VisionEncoderConfig, tp_size: int, enable_
...
+        result.append(
+            ops.NCCL(
+                "encoder_dp_all_gather",
+                1,
+                "all_gather",
+                num_elements_per_token=enc_cfg.out_hidden_size * tp_size,
+                num_gpus=tp_size,
+                comm_quant_mode=common.CommQuantMode.half,
+            )
+        )
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 8ffa0eff. Encoder-DP all-gather now uses the per-rank projected width for both Gemma 4 and Qwen, since the backend already supplies rank-local image/token count. Added regressions for both builders; full unit and image-bearing Gemma support-matrix tests pass.

tools/support_matrix/support_matrix.py
```diff
@@ -198,26 +213,66 @@ def _support_matrix_row_command(
...
+@cache
+def _get_matrix_visual_workload(model_path: str) -> tuple[int, int, int] | None:
+    """Return one representative image workload when AIC models an encoder.
+
+    Matrix rows used to exercise multimodal architectures as text-only, which
+    could report PASS without querying any encoder op.  Derive this workload
+    from the parsed model configuration so both agg and disagg checks cover the
+    encoder automatically whenever ``encoder_ops`` are expected.
+    """
+    model_info = _get_model_info(model_path)
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 8ffa0eff. Model constraint and architecture resolution failures are now wrapped as fatal `ModelMetadataResolutionError` values, propagated out of parallel execution, and also fail closed during sequential retry. Added direct and parallel-worker regression coverage.

aic-core/src/aiconfigurator_core/sdk/models/gemma4.py
jasonqinzhou: Fixed in 00c0ed9a. The parser retains `use_bidirectional_attention=vision`; the model adds the strict upper-triangle work only on the 25 sliding layers, using independent per-image blocks and including it once in TTFT and energy.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
jasonqinzhou: Fixed in 00c0ed9a. The valid language context remains 256/532 soft tokens while tower, pooling, latency, and activation memory use the padded 2520/5040 patch inputs. Regression cases cover both budgets.

aic-core/src/aiconfigurator_core/sdk/utils.py
jasonqinzhou: Fixed in 00c0ed9a. Legacy disaggregated sizing unwraps `Gemma4MixConfig.vision_config`; the regression verifies text ISL 256 plus 280 image tokens under a 1024-token cap yields prefill batch list `[1]`.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
jasonqinzhou: Fixed in 00c0ed9a and strengthened in ef379d4b. Encoder-TP activation peaks now use TP-local QKV and gated-MLP widths while retaining the replicated residual width. The regression runs above the 32 MiB floor and checks the exact value.

tests/unit/sdk/backends/test_encoder.py
```diff
@@ -550,3 +551,162 @@ def test_vl_model_with_images_has_encoder_memory(self, model_config):
...
+        unsharded_bytes = (
+            2
+            * 2520
+            * max(
+                3 * enc_cfg.hidden_size,
+                enc_cfg.hidden_size + 2 * enc_cfg.intermediate_size,
+            )
+        )
+        unsharded_bytes += 2 * 280 * (2 * enc_cfg.hidden_size + enc_cfg.out_hidden_size)
+        assert memory["activations"] < unsharded_bytes / (1 << 30)
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in ef379d4b. The regression now uses the 5,040-patch budget, explicitly asserts the expected activation allocation exceeds 32 MiB, and then verifies the exact TP-local memory value.

tools/support_matrix/support_matrix.py
```diff
@@ -1069,6 +1152,11 @@ def _run_parallel_combinations(
                         retry_combos.add(futures[remaining])
                     pbar.update(len(unprocessed_futures))
                     break
+                except ModelMetadataResolutionError:
+                    for remaining in futures:
+                        if remaining is not future:
+                            remaining.cancel()
+                    raise
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a94d563e. The executor is no longer managed by a waiting context on this fatal path: pending futures are cancelled, `shutdown(wait=False, cancel_futures=True)` is called, and the metadata error is immediately re-raised. The regression now submits a second pending future and asserts both cancellation and the non-waiting shutdown arguments; all 120 support-matrix unit tests pass.

tools/support_matrix/support_matrix.py
```diff
@@ -1069,6 +1154,15 @@ def _run_parallel_combinations(
...
                     pbar.update(len(unprocessed_futures))
                     break
+                except ModelMetadataResolutionError:
+                    for remaining in futures:
+                        if remaining is not future:
+                            remaining.cancel()
+                    # A context-manager exit would wait for already-running
+                    # workers, defeating the fatal metadata fail-fast contract.
+                    shutdown_without_wait = True
+                    executor.shutdown(wait=False, cancel_futures=True)
```
jasonqinzhou: **[P2] Terminate running workers on fatal metadata errors**

When a metadata failure arrives after another combination has started, `cancel_futures=True` only cancels queued work; `shutdown(wait=False)` leaves running worker processes alive, and Python still keeps the interpreter alive until those futures finish. A full matrix invocation can therefore continue expensive tests for minutes instead of failing fast and keep consuming resources. Terminate active workers before re-raising.

jasonqinzhou (commented):
Final Codex review found one additional actionable process-lifecycle issue.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in a219d080. Fatal metadata failures now terminate active process-pool workers before re-raising: Python 3.14+ uses `terminate_workers()`, while older supported versions explicitly terminate and join each worker. Pending futures remain cancelled. The updated lifecycle regression verifies active-worker termination and all 120 support-matrix unit tests pass.

jasonqinzhou (commented):
The final Codex review found one additional parity-probe integration issue.

tools/support_matrix/support_matrix.py
```diff
@@ -198,26 +251,66 @@ def _support_matrix_row_command(
...
-            return constraints
+            selected = constraints
+            break
+    visual_workload = _get_matrix_visual_workload(model_path)
+    if visual_workload is not None:
+        selected = replace(
+            selected,
+            image_height=visual_workload[0],
+            image_width=visual_workload[1],
+            num_images=visual_workload[2],
```
jasonqinzhou, line 305: **[P2] Pass visual constraints to the parity probe**

When `scan_rust_parity.py --scan-mode probe_only` handles a multimodal PASS row, `_get_test_constraints()` now returns these image fields, but `_run_probe()` still forwards only the text constraints to `cli_estimate`. The probe therefore exercises the text-only path and can report Python/Rust parity without checking encoder or Gemma visual-context work. Pass `image_height`, `image_width`, and `num_images` through that consumer.

jasonqinzhou: Fixed in ada8ee5b. The parity probe now forwards `image_height`, `image_width`, and `num_images` to both Python and Rust `cli_estimate` calls, and records them in `probe_shape`. Added direct argument and shape regressions; all 122 support-matrix unit tests pass.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -461,6 +461,9 @@ def _run_probe(
         isl=constraints.isl,
         osl=constraints.osl,
         prefix=constraints.prefix,
+        image_height=constraints.image_height,
+        image_width=constraints.image_width,
+        num_images=constraints.num_images,
```
jasonqinzhou, line 509: **[P1] Require evidence that the encoder actually ran**

When either estimator silently skips encoder modeling—for example after an `encoder_ops` regression—it can still return matching text-only TTFT/TPOT values, causing this visual probe to be recorded as `PASS`. For positive image constraints, validate the returned raw encoder latency/memory evidence and fail the probe when that evidence is missing.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -492,7 +495,9 @@ def probe_entry(entry: Entry) -> ProbeRecord:
     started = time.monotonic()
     constraints = _get_test_constraints(entry.model)
     shape_str = (
-        f"isl={constraints.isl},osl={constraints.osl},prefix={constraints.prefix},total_gpus={constraints.total_gpus}"
+        f"isl={constraints.isl},osl={constraints.osl},prefix={constraints.prefix},total_gpus={constraints.total_gpus},"
+        f"image_height={constraints.image_height},image_width={constraints.image_width},"
+        f"num_images={constraints.num_images}"
```
jasonqinzhou: **[P1] Invalidate text-only checkpoints for visual probes**

When resuming a pre-change database with `--continue-across-commits`, `pending_entries_for_probe` treats any completed row for the unchanged `Entry.key` as done, regardless of `probe_shape`. Therefore existing text-only results skip the new visual probe entirely. Purge or reschedule completed results whose stored shape does not match the current image-aware probe shape.

jasonqinzhou (commented):
The scoped Codex review found two fail-closed requirements in the new visual parity probe.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in 98fc66af. On startup, resumable scans derive the canonical visual workload per model, delete completed probe rows whose stored shape does not exactly match it, and therefore reschedule those entries. Tests cover retiring a legacy text-only result and preserving a current visual result.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in 98fc66af. Positive image workloads now require raw `encoder_latency` and `encoder_memory` values greater than zero from each engine. Missing evidence produces `ENCODER_EVIDENCE_ERROR`, is bucketed as a regression, and is surfaced in reports. The regression verifies matching text-only results cannot pass.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -481,6 +522,20 @@ def _run_probe(
...
         else:
             return None, None, f"unsupported mode {entry.mode!r}"
+        if _has_visual_workload(constraints):
+            raw = getattr(result, "raw", None)
+            if not isinstance(raw, dict):
+                raise RuntimeError("ENCODER_NOT_EXERCISED: parity probe returned no raw encoder evidence")
+            missing = []
+            for field in ("encoder_latency", "encoder_memory"):
+                try:
+                    value = float(raw.get(field, 0.0) or 0.0)
```
jasonqinzhou, line 538: **[P1] Read disaggregated encoder memory from its actual field**

For every visual `disagg` entry, `cli_estimate` exposes encoder memory as `raw["(e)memory"]`, not `raw["encoder_memory"]`. This loop therefore marks healthy Python and Rust estimates as `ENCODER_EVIDENCE_ERROR`, counts them as regressions, and makes the probe scan fail for all visual disaggregated rows. Select the evidence field according to `entry.mode`.

jasonqinzhou (commented):
The final scoped review found one mode-specific encoder-evidence schema issue.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in cdf54d10. The evidence gate now reads `encoder_memory` for aggregate results and `(e)memory` for disaggregated results. Added a disaggregated schema regression and verified a live B200/TRT-LLM Gemma probe: Python and Rust both return TTFT 35.849 ms and TPOT 3.227 ms with positive evidence. All 126 support-matrix tests pass.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -481,6 +522,21 @@ def _run_probe(
             )
         else:
             return None, None, f"unsupported mode {entry.mode!r}"
+        if _has_visual_workload(constraints):
+            raw = getattr(result, "raw", None)
+            if not isinstance(raw, dict):
+                raise RuntimeError("ENCODER_NOT_EXERCISED: parity probe returned no raw encoder evidence")
+            memory_field = "(e)memory" if entry.mode == "disagg" else "encoder_memory"
```
jasonqinzhou, line 534: **[P1] Invalidate cached disagg evidence failures**

When `scan.sqlite` is resumed from the parent commit using `--continue-across-commits`, already-probed visual disaggregated rows have the current `_probe_shape` but an `ENCODER_EVIDENCE_ERROR` caused by the old `encoder_memory` lookup. `_retire_stale_visual_probe_results` retains these rows and `pending_entries_for_probe` treats them as complete, so this corrected `(e)memory` check never runs. Version the visual probe/evidence shape so these cached results are requeued.

jasonqinzhou (commented):
The final scoped review found one resumable-cache invalidation requirement.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in d425ad53. Visual probe shapes now include `encoder_evidence_schema=2`. Resume startup deletes any visual result with a prior schema (including the false disaggregated evidence failures from schema 1), which requeues those entries while preserving current schema-2 results. Added both invalidation and preservation regressions; all 127 support-matrix tests pass.

aic-core/src/aiconfigurator_core/sdk/models/vit_ops.py
```diff
@@ -215,7 +215,172 @@ def build_encoder_ops(enc_cfg: common.VisionEncoderConfig, tp_size: int, enable_
                 "encoder_dp_all_gather",
                 1,
                 "all_gather",
-                num_elements_per_token=enc_cfg.out_hidden_size * enc_cfg.projector_n_instances * tp_size,
+                num_elements_per_token=enc_cfg.out_hidden_size * enc_cfg.projector_n_instances,
```
tianhaox, line 218: **[Blocking] Restore the `* tp_size` factor here and in `build_gemma4_vision_encoder_ops` (line 383) - the per-rank payload undercounts the all-gather by ~`tp_size`x.**

The 8ffa0eff change assumed `message_size` is the per-rank send payload because the backend queries with the rank-local token count. That only looks at the query side; what `message_size` *means* to the NCCL table and SOL in this codebase is the **total gathered buffer**:

- `NCCL._query_nccl_table.get_sol` uses `sol_time = mem * message_size * (n-1)/n / p2p_bw` - the nccl-tests busbw form, which is only correct when `message_size` is the total receive buffer. For a per-rank payload `P` the wire time is `P*(n-1)/bw`; the formula returns `P*(n-1)/(n*bw)`, an `n`x underestimate.
- The packaged table is collected by `collector/network/collect_nccl.py`, which runs nccl-tests `all_gather_perf -b <size> -e <size>` and stores `size // bytes_per_element`. In nccl-tests, the all-gather size parameter is the total receive buffer (`sendcount = count/nranks`), so the collected `message_bytes` axis is total-size, matching the SOL.
- Existing in-repo usage agrees: `base.py`'s `context_cp_all_gather` queries with full-sequence tokens x the full per-token KV width (total gathered volume), not a per-rank share.

With the rank-local `x = images_local * tokens_per_image` that `_run_encoder_phase` passes, the original `out_hidden_size * projector_n_instances * tp_size` reconstructed exactly the total gathered size. The current code queries the collected curve at a `tp_size`x smaller message and inherits the SOL underestimate - and it silently changes the merged Qwen3-VL (#1393) predictions inside a Gemma PR. The magnitude is small today (the gather is us-scale vs the ~ms encoder), but it breaks the library-wide message-size convention for any future larger-payload gather.

Please revert both builders to `* tp_size` (or equivalently query with the global post-merge token count) and update `test_qwen_encoder_dp_all_gather_uses_per_rank_payload` / `test_gemma4_encoder_dp_replicates_compute_and_gathers_soft_tokens`, which currently pin the per-rank semantics.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -516,7 +637,15 @@ def _run_static_breakdown(
 
         # FPM models compile to no Rust op variant yet; force the Python phases
         # (which handle the single whole-model op naturally).
-        if model.forward_model != "fpm" and should_use_rust_engine_step(runtime_config, database):
+        # Gemma 4's visual-block overlay has a runtime-derived per-image shape,
+        # which the compiled context walk cannot express yet. Keep its prefill
+        # on the Python orchestrator; encoder and decode ops can still use Rust.
+        has_visual_context_work = mode != "static_gen" and self._has_visual_context_work(model, runtime_config)
```
tianhaox, line 643: **[Non-blocking] Emit `note_python_step_fallback` for this delegation (here and in `run_mixed`, line 1297).**

Every other Python delegation route emits a fallback note; this gate is silent, and it also silently overrides an explicit `engine_step_backend="rust"` request (which normally forces Rust per `should_use_rust_engine_step`). A `note_python_step_fallback("visual_context_ops", ...)` keeps the post-#1454 telemetry complete.

One consequence worth stating in the PR description: with an image workload, the parity scan's "rust" probe runs the Python prefill on both sides, so Gemma image rows carry no Rust-parity signal for the context portion. That's fine given the compiled walk can't express the per-image overlay yet, but it should be a documented consequence rather than an accident.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -429,8 +496,62 @@ def _run_context_phase(
             else:
                 context_source_dict[op._name] = "mixed"
 
+        visual_latency, visual_energy, visual_source = self._run_visual_context_phase(
```
tianhaox, line 499: **[Non-blocking / structure] Consider calling `_run_visual_context_phase` explicitly at the consumers instead of embedding it here.**

This embedded call ignores `op_filter`. The reason `run_mixed`'s pass 1/2 don't double-count the overlay is only that their synthetic pass configs happen to carry no image fields - a fragile invariant for a shared primitive (any future caller that passes an image-bearing config plus an `op_filter` double-counts silently). Invoking the visual phase explicitly in `_run_static_breakdown` (the run_mixed call already exists) is the same line count with no hidden coupling.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -229,14 +236,70 @@ def _encoder_pre_merge_per_visual(
...
+        the patch tensor back to the maximum before running the vision tower, so
+        the returned pre-pool count is the padded maximum while the post-pool
+        count remains the number of valid soft tokens. The SDK-only
+        ``num_image_tokens`` field selects a supported alternate maximum.
+
         Returns ``(tokens_post_merge_per_image, pre_merge_per_image)``.
         Returns ``(0, 0)`` when neither is set (text-only path).
         """
         has_image_dims = runtime_config.image_height > 0 and runtime_config.image_width > 0
+        if isinstance(enc_cfg, common.Gemma4VisionEncoderConfig):
```
tianhaox, line 256: **[Nit]** This ~45-line Gemma-specific arithmetic could live on `Gemma4VisionEncoderConfig` (e.g. a `resolve_token_counts(height, width, budget)` method), keeping the shared backend to a two-line isinstance dispatch. Also, the final `if tokens_per_image <= 0 or resized_patches <= 0` guard is unreachable - after the zero-fixups both sides are >= `side_multiple`, so the pooled count is >= 1.

aic-core/src/aiconfigurator_core/sdk/utils.py
```diff
@@ -790,6 +791,67 @@ def _parse_hf_config_json(config: dict) -> dict:
...
+                )
+            if vision_cfg.get("model_type") not in (None, "gemma4_vision"):
+                raise ValueError(
+                    f"Gemma 4 vision_config.model_type must be 'gemma4_vision', got {vision_cfg.get('model_type')!r}"
+                )
+
+            pooling_kernel_size = int(vision_cfg["pooling_kernel_size"])
+            if pooling_kernel_size <= 0:
+                raise ValueError("Gemma 4 vision pooling_kernel_size must be positive")
+            soft_tokens_per_image = int(
```
tianhaox, line 810: **[Non-blocking] Validate `soft_tokens_per_image` against `supported_soft_token_budgets` at parse time.**

The budget list is a hardcoded dataclass default and the checkpoint value is only checked deep inside `_encoder_pre_merge_per_visual` at query time - a Gemma 4 variant shipping a different budget would fail mid-sweep rather than at model load. Parse-time validation (and/or reading the budget list from the checkpoint when present) fails fast with a clearer message.

aic-core/src/aiconfigurator_core/sdk/models/gemma4.py
```diff
@@ -112,6 +147,36 @@ def set_gemma4_config(self, cfg: common.Gemma4MixConfig) -> None:
...
+            )
+            if cfg.use_bidirectional_vision_attention:
+                # The ordinary SWA ContextAttention already accounts for the
+                # causal half of each visual block. Gemma's blockwise overlay
+                # additionally unmasks the upper triangle for every SWA layer.
+                # BaseBackend supplies the per-image pooled length; the op
+                # scales Gemma's collected causal curve to the added pairs.
+                d = self._resolve_dims(self.config.tp_size)
+                self.visual_context_ops.append(
+                    _Gemma4VisualBlockAttention(
```
tianhaox, line 169: **[Non-blocking] Worth a comment on the CP convention divergence.**

This op models CP as `swa_layers / cp_size` (balanced average), while the language `ContextAttention` under CP uses the zigzag `_cp_size` representative-rank path (`_build_context_ops` wiring). Fine as a first-order choice for per-image blocks, but a one-line comment noting the deliberate divergence will keep a later CP audit from flagging it as a missed `_cp_size` wire-up.

tools/support_matrix/support_matrix.py
```diff
@@ -86,6 +87,44 @@
...
+    # ProcessPoolExecutor.terminate_workers() is only available on Python 3.14+.
+    # Terminate the processes explicitly on older supported Python versions,
+    # then join them so they cannot outlive the fatal error.
+    processes = getattr(executor, "_processes", None) or {}
+    for process in processes.values():
+        process.terminate()
+    executor.shutdown(wait=True, cancel_futures=True)
+
+
+class ModelMetadataResolutionError(RuntimeError):
```
tianhaox, line 108: **[Non-blocking / policy] Fatal metadata errors now abort the entire matrix run - worth an explicit ack from the matrix owners.**

Fail-closed is the right direction for vision coverage (a metadata failure must not silently degrade an image-bearing row to text-only PASS). But the tradeoff is that one bad or newly-added model now kills the whole nightly job instead of producing a FAIL row, and the mechanism is generic infra riding a model PR.

Minor simplification while you're here: `_resolve_test_constraints`/`_resolve_architecture` are twin wrappers that could share one helper, and the manual `executor_shutdown` flag + `finally` could stay a plain `with executor:` if the fatal path terminates workers before re-raising.

tianhaox (changes requested):
Reviewed the modeling core, the runtime integration, and the tooling changes; also independently verified the pieces most likely to be wrong, so they don't need re-litigating:

- the aspect-ratio/soft-token arithmetic against the three pinned cases (280/2520 @ 672x960, 256/2520 padded @ 448x448, 532/5040 @ budget 560) is internally consistent, and the floor-rounding cannot exceed `max_patches`;
- the visual-block overlay is applied exactly once on every path (run_static -> `_run_static_breakdown`; run_agg -> `_get_mix_step_latency` -> run_mixed; the synthetic pass configs carry no image fields);
- `_Gemma4VisualBlockAttention`'s strict-upper-triangle scaling off the collected causal SWA curve is right to first order (the sliding window only matters at the 1120-token budget, ~0.7%);
- encoder TP/DP activation widths match the op graph; the `inference_session` unwrap covers the only `extra_params` -> encoder call site; the agg cache-key extension closes a real collision surface.

**One blocking item**: the encoder-DP all-gather payload change made mid-PR in response to the CodeRabbit thread (see the `vit_ops.py` comment). The original `* tp_size` was correct under the collected table's total-size convention, and the change also silently alters merged Qwen3-VL (#1393) behavior.

Non-blocking notes are inline. Two without a good anchor:

- The daily support matrix and parity scan become image-bearing for existing Qwen3-VL rows too (their workloads change; prior probe rows are retired via `encoder_evidence_schema=2`). Intended, but please state it explicitly in the PR description.
- Data-side follow-ups worth filing as a tracking issue if Gemma 4 multimodal is a priority target: the packaged DB has no nonzero encoder energy samples (as you noted), the encoder shapes (head_dim-72 varlen attention, K=768 patch-embed GEMM) ride grid interpolation without dedicated collection points, and there is no end-to-end vision-tower trace to calibrate the ~8.7 ms encoder estimate.

Test coverage is the strongest of the recent VL PRs - architecture pins, the resize contract, DP/TP sharding, fail-closed evidence checks, and resume retirement are all covered.

aic-core/src/aiconfigurator_core/sdk/models/vit_ops.py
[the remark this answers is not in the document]
tianhaox: Coming back to this thread after full review: I believe the original `* tp_size` was correct and 8ffa0eff should be reverted - see the blocking comment in review 4928515032 (vit_ops.py line 218) for the full evidence chain.

Short version: the premise "NCCL's all-gather message size is the per-rank send payload" doesn't match this codebase's convention. The packaged table is collected via nccl-tests `all_gather_perf` (whose size parameter is the *total* receive buffer, `sendcount = count/nranks`), and `get_sol` uses the busbw form `size*(n-1)/n / bw`, which is likewise only correct for total-size semantics. With the rank-local token count the backend passes, `* tp_size` reconstructed exactly that total; the per-rank payload queries the curve at a `tp_size`x smaller message and underestimates the collective by ~`tp_size`x.

jasonqinzhou:
Closing this AIConfigurator PR because the repository is being refrozen for feature development; ongoing feature work, review, and landing are moving to AISimulate.

The one-to-one continuation is ai-dynamo/aisimulate#86: https://github.com/ai-dynamo/aisimulate/pull/86

Please continue review and development there. This AIC PR is retained as source history.

PR #1529, opened 2026-08-13
feat: exercise multimodal encoders in support matrix

jasonqinzhou:
## Summary

- Detect vision encoders independently from whether AIC implements them.
- Automatically exercise one canonical nonzero image workload (`1 x 1024 x 1024`) for implemented encoders in both agg and disagg, without a redundant text-only run.
- Require strictly positive encoder latency and memory evidence before a multimodal row can pass.
- Persist and validate image workload/replay evidence in the support-matrix CSV schema.
- Classify declared-but-unimplemented encoders as explicit `FAIL / ENCODER_UNSUPPORTED` results while leaving text-only checkpoints unchanged.

## Why

Support-matrix runs previously used a zero-image workload, so a successful text-backbone estimate could masquerade as multimodal encoder coverage. AIC-1738 makes encoder execution and evidence part of the PASS contract.

## Acceptance

- Dense: `Qwen/Qwen3-VL-8B-Instruct`
- MoE: `Qwen/Qwen3-VL-30B-A3B-Instruct`
- Both acceptance cases cover agg and disagg with the compiled engine-step executor and positive encoder latency/memory evidence.
- Checkpoints that declare vision but lack an AIC encoder implementation return `ENCODER_UNSUPPORTED` rather than PASS.

## Main integration

- Merged current `origin/main` (`a9e012ac1`) into the branch and resolved all conflicts.
- Adapted the feature to main's compiled-only engine-step path; obsolete Python/Rust support-matrix parity code was removed with main.
- Preserved main's Docker test assets while retaining the support-matrix documentation copy needed by the UI contract test.

## Validation

- `.venv/bin/pytest -q -p no:timeout tests/unit/support_matrix tests/unit/sdk/backends/test_encoder.py tests/unit/generator/test_module_bridge_image_workload.py` - 213 passed.
- `.venv/bin/pytest -m unit -p no:timeout --ignore=tests/unit/sdk/database/test_moe_dispatch.py -q` - 3,629 passed, 10 skipped, 712 deselected.
- `.venv/bin/pytest -q -p no:timeout tests/e2e/support_matrix/test_pr_support_matrix.py -k 'Qwen3-VL-8B or Qwen3-VL-30B' -vv` - 2 passed, 4 deselected.
- `.venv/bin/ruff check .` - passed.
- `.venv/bin/ruff format --check .` - 588 files already formatted.
- `git diff --check origin/main` - passed.

## Review

The built-in Codex PR reviewer audited the full current diff, traced the support-matrix schema and encoder paths, independently reran focused and full validation, and reported no actionable correctness defects on `e12feed57`.

## Notes

- Historical checked-in support-matrix CSVs are intentionally not regenerated in this feature PR. The next scheduled full generation will emit the 13-column image-evidence schema and explicit unsupported classifications.
- This PR is non-draft and ready for maintainer review.

Linear: AIC-1738


jasonqinzhou (commented):
Codex review found four actionable gaps. I am addressing all four in a follow-up commit and will re-run the review before resolving these threads.

tools/support_matrix/support_matrix.py
```diff
@@ -885,6 +1007,14 @@ def run_single_test(
             for mode in modes_to_test
         }
 
+        if encoder_coverage.checkpoint_declares_encoder and not encoder_coverage.aic_encoder_implemented:
```
jasonqinzhou, line 1164: [P1] The first scheduled regeneration converts existing declared-but-unimplemented encoder rows to FAIL, while the comparator currently blocks those transitions. Add an explicit recognized encoder-coverage migration path so the workflow can publish the intended reclassification without weakening unrelated regression checks.

tools/support_matrix/support_matrix.py
```diff
@@ -112,12 +115,81 @@ class TestConstraints:
...
+# daily matrix bounded. This matches the existing tier-2 multimodal regression case.
+SUPPORT_MATRIX_IMAGE_WORKLOAD = ImageWorkload(image_height=1024, image_width=1024, num_images=1)
+
+
+def _get_encoder_coverage(model: str) -> EncoderCoverage:
+    """Return checkpoint declaration and AIC implementation state separately."""
+    model_info = _get_model_info(model)
+    raw_config = model_info.get("raw_config") or {}
+    vision_config = raw_config.get("vision_config")
+    checkpoint_declares_encoder = isinstance(vision_config, dict) and bool(vision_config)
```
jasonqinzhou: [P1] Bundled Llama-4 and Step-3.7 configs are trimmed and omit vision_config even though their architectures are registered as multimodal. Include known multimodal architecture evidence so these checkpoints cannot retain a text-only PASS.

tools/support_matrix/support_matrix.py
```diff
@@ -799,6 +916,7 @@ def _run_mode(
         version: str,
         constraints: TestConstraints,
         engine_step_backend: str | None,
+        image_workload: ImageWorkload | None = None,
```
jasonqinzhou, line 1070: [P2] The full Rust parity scanner calls _run_mode directly and currently discards the CSV image metadata, so Qwen3-VL parity can still run text-only. Carry the image workload through parity Entry/probe/pareto execution and assert encoder evidence there.

tools/support_matrix/support_matrix.py
```diff
@@ -1352,15 +1519,33 @@ def save_results_to_csv(self, results: list[tuple[str, ...]], output_file: str)
...
+                    mode,
+                    status,
+                    err_msg,
+                    command,
+                    source,
+                    image_height,
+                    image_width,
+                    num_images,
+                ) = row
+            elif len(row) == 10:
```
jasonqinzhou, line 1747: [P2] Upgrading a legacy 10-column encoder PASS currently writes a current 13-column row with blank image metadata, which sanity validation accepts. Reject legacy multimodal PASS upgrades without explicit image evidence.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -272,6 +284,28 @@ def load_entries(baseline_dir: Path, *, status_filter: tuple[str, ...] = ("PASS"
...
+                    try:
+                        image_height, image_width, num_images = (int(value) for value in image_values)
+                    except ValueError as exc:
+                        raise ValueError(f"Invalid image workload metadata in {csv_path}: {row}") from exc
+                    if min(image_height, image_width, num_images) <= 0:
+                        raise ValueError(f"Non-positive image workload metadata in {csv_path}: {row}")
+                    image_workload = ImageWorkload(image_height, image_width, num_images)
+                else:
+                    coverage = _get_encoder_coverage(row["HuggingFaceID"].strip())
+                    if coverage.checkpoint_declares_encoder and not coverage.aic_encoder_implemented:
```
jasonqinzhou, line 306: [P1] Migrate the baseline before rejecting legacy PASS rows

The checked-in 10-column baseline still has hundreds of PASS rows for encoders that this change now classifies as unimplemented. Raising here makes the default parity scan abort before seeding any work. Handle these legacy rows compatibly (without certifying their text backbones) or migrate the baseline in this change.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -37,6 +37,8 @@
 from datetime import datetime, timezone
 from pathlib import Path
 
+import pandas as pd
```
jasonqinzhou: [P1] Import pandas only after setting worker thread caps

Importing pandas here may initialize NumPy/native runtimes before `_WORKER_THREAD_CAPS` is applied. Forked workers then inherit uncapped pools, defeating the scanner's documented slowdown/OOM safeguard. Move the import below the environment setup or make it lazy.

tools/support_matrix/support_matrix.py
```diff
@@ -885,6 +1013,14 @@ def run_single_test(
             for mode in modes_to_test
         }
 
+        if encoder_coverage.checkpoint_declares_encoder and not encoder_coverage.aic_encoder_implemented:
+            reason = _encoder_unsupported_error(model, encoder_coverage)
+            statuses = dict.fromkeys(modes_to_test, STATUS_FAIL)
+            error_messages = dict.fromkeys(modes_to_test, reason)
+            if include_commands:
```
jasonqinzhou, line 1178: [P2] Make encoder-rejection replay commands reproduce the failure

For an unimplemented encoder, the persisted command is text-only and succeeds even though the row records `ENCODER_UNSUPPORTED`. Persist a replay path that exercises the same support-matrix preflight so the classification can be reproduced and investigated.

jasonqinzhou (commented):
Second Codex review pass found three additional actionable integration defects after the first fixes. I will address all three in the next commit.

jasonqinzhou (commented):
Codex review pass 3 found three actionable edge cases after focused and full validation.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -481,6 +535,16 @@ def _run_probe(
             )
         else:
             return None, None, f"unsupported mode {entry.mode!r}"
+        if entry.image_workload is not None:
+            raw = getattr(result, "raw", None)
+            if not isinstance(raw, dict):
+                raise RuntimeError("ENCODER_NOT_EXERCISED: parity probe returned no raw encoder evidence")
+            _assert_encoder_evidence(
```
jasonqinzhou, line 581: [P1] Fail probes when both engines omit encoder evidence. For a multimodal row, both Python and Rust can raise `ENCODER_NOT_EXERCISED`; the generic both-error parity classification currently turns that into `BOTH_ERROR_PASS`, so a probe-only scan reports green while neither engine exercised the encoder. Treat missing encoder evidence as a scan failure even when both implementations fail identically.

tools/support_matrix/compare_support_matrix.py
```diff
@@ -338,7 +418,18 @@ def find_blocking_status_transitions(changed_rows: list[tuple]) -> list[str]:
...
     """
     errors = []
+    encoder_migration_keys = {
+        tuple(row[:6])
+        for row in new_data_rows or []
+        if len(row) > 7 and row[6] == STATUS_FAIL and row[7].strip().startswith("ENCODER_UNSUPPORTED:")
+    }
     for huggingface_id, architecture, system, backend, version, mode, old_status, new_status in changed_rows:
+        key = (huggingface_id, architecture, system, backend, version, mode)
+        if new_status == STATUS_FAIL and key in encoder_migration_keys:
```
jasonqinzhou, line 489: [P1] Limit the `ENCODER_UNSUPPORTED` migration waiver to legacy rows without persisted image metadata. Once an image-backed PASS exists, a later encoder regression must remain a blocking PASS-to-FAIL transition instead of being mistaken for the one-time migration.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -178,10 +183,17 @@ class Entry:
     version: str
     mode: str
     baseline_status: str
+    image_workload: ImageWorkload | None = None
 
     @property
     def key(self) -> str:
-        return f"{self.model}|{self.system}|{self.backend}|{self.version}|{self.mode}"
+        workload = self.image_workload
```
jasonqinzhou, line 195: [P2] Preserve the existing parity key for text-only entries. Adding `|text` makes `--continue-across-commits` duplicate every prior text row and rerun completed work; only image-bearing entries need a workload suffix.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -593,7 +668,15 @@ def pareto_entry(entry: Entry) -> ParetoRecord:
             version=entry.version,
             constraints=constraints,
             engine_step_backend="python",
+            image_workload=entry.image_workload,
         )
+        if entry.image_workload is not None and python_df is not None and not python_df.empty:
```
jasonqinzhou, line 716: [P1] Fail default Pareto scans when encoder evidence is missing. The assertion is caught as a Python error and returned as `PARETO_STATUS_ERROR`, while `_run_pool` only exits nonzero for DRIFT/REGRESSION. Classify `ENCODER_NOT_EXERCISED` as a regression or otherwise make it fail the scan.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -178,10 +184,18 @@ class Entry:
...
     @property
     def key(self) -> str:
-        return f"{self.model}|{self.system}|{self.backend}|{self.version}|{self.mode}"
+        workload = self.image_workload
+        base_key = f"{self.model}|{self.system}|{self.backend}|{self.version}|{self.mode}"
+        if workload is None:
+            # Keep text-only keys stable so --continue-across-commits can reuse
+            # results written before image workloads became part of the key.
+            return base_key
+        workload_key = f"image={workload.image_height}x{workload.image_width}x{workload.num_images}"
```
jasonqinzhou, line 200: [P2] Retire the superseded base key when an image-specific key is seeded. With `--continue-across-commits`, the insert-only SQLite seeding leaves the old text-only PASS beside the new image run, double-counting and continuing to present backbone-only coverage as certified.

tools/support_matrix/compare_support_matrix.py
```diff
@@ -338,7 +420,35 @@ def find_blocking_status_transitions(changed_rows: list[tuple]) -> list[str]:
...
+            return True
+        try:
+            return not (int(row[10]) > 0 and int(row[11]) > 0 and int(row[12]) > 0)
+        except (TypeError, ValueError):
+            return True
+
+    encoder_migration_keys = set()
+    for row in new_data_rows or []:
+        key = tuple(row[:6])
+        if (
```
jasonqinzhou, line 479: [P2] Verify `_get_encoder_coverage` reports a declared but unimplemented encoder before granting the legacy migration waiver. A text-only PASS can currently be relabeled with an `ENCODER_UNSUPPORTED:` prefix and bypass the regression gate.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -1037,7 +1128,9 @@ def cmd_report(args: argparse.Namespace) -> int:
                        COUNT(p.entry_key) AS probed,
                        SUM(CASE WHEN p.status IN ('PASS','BOTH_ERROR_PASS') THEN 1 ELSE 0 END) AS passed,
                        SUM(CASE WHEN p.status = 'DRIFT' THEN 1 ELSE 0 END) AS drifted,
-                       SUM(CASE WHEN p.status = 'RUST_ERROR_ONLY' THEN 1 ELSE 0 END) AS regressed
+                       SUM(
```
jasonqinzhou, line 1182: [P2] Include `ENCODER_EVIDENCE_ERROR` in the detailed regression query, not only the aggregate count, and display the available Python/Rust evidence error so the report is actionable.

jasonqinzhou (commented):
Codex review pass 4 found four remaining parity/validation integration gaps.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -265,13 +282,40 @@ def _git_head_sha() -> str:
...
+                        # The published baseline predates AIC-1738 and still has
+                        # backbone-only PASS rows for unsupported encoders. Do
+                        # not certify those rows, but let the rest of a default
+                        # parity scan proceed until the daily matrix migration
+                        # replaces them with ENCODER_UNSUPPORTED failures.
+                        model = row["HuggingFaceID"].strip()
+                        skipped_unsupported[model] = skipped_unsupported.get(model, 0) + 1
+                        continue
+                    # Legacy matrices predate explicit image columns. Infer the
+                    # canonical workload for supported encoders so a resumed
```
jasonqinzhou, line 316: **[P2] Purge skipped encoder rows from resumable parity scans**

When `scan.sqlite` was seeded before this change and resumed with `--continue-across-commits`, this `continue` omits the now-unsupported checkpoint from the entries passed to `seed_entries` but never deletes its existing SQLite entry or probe/Pareto results. Consequently, `cmd_report` continues to count the stale text-backbone PASS even though the warning says it was not certified. Purge these retired keys similarly to supported image-backed replacements.

jasonqinzhou (commented):
Fifth Codex review pass found one resumability gap.

docs/support-matrix/model-roster.md
```diff
@@ -22,6 +22,24 @@ Before adding a model, confirm that its runtime path is viable on at least one
...
+
+An encoder-supported PASS means that the agg or disagg run used the canonical
+image workload and produced strictly positive encoder latency and encoder
+memory for every result row. `ImageHeight`, `ImageWidth`, and `NumImages` in the
+generated CSV, plus the matching replay-command arguments, record that workload.
+
+If a checkpoint declares `vision_config` but AIC cannot normalize and build its
+encoder operations, the row fails with an `ENCODER_UNSUPPORTED` reason. It must
+not inherit PASS from a successful text-backbone-only estimate. Text-only
+checkpoints keep their existing workload and leave the image metadata empty.
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a8d53ff3. The docs now distinguish `ENCODER_UNSUPPORTED` normalization failures from `ENCODER_NOT_EXERCISED` evidence failures.

tools/support_matrix/compare_support_matrix.py
```diff
@@ -210,6 +212,48 @@ def check_csv_sanity(header: list[str], data_rows: list[list[str]]) -> list[str]
...
+                                actual_values = []
+                                for part_index, part in enumerate(command_parts):
+                                    if part == flag and part_index + 1 < len(command_parts):
+                                        actual_values.append(command_parts[part_index + 1])
+                                    elif part.startswith(f"{flag}="):
+                                        actual_values.append(part.partition("=")[2])
+                                if actual_values != [expected_value]:
+                                    errors.append(
+                                        f"Row {i}: replay command must include exactly one {flag} {expected_value}; "
+                                        f"found {actual_values or 'none'}"
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a8d53ff3. Shared option parsing now records bare flags, and a regression test covers a valid image-height pair followed by a trailing bare flag.

tools/support_matrix/support_matrix.py
```diff
@@ -862,6 +981,8 @@ def run_single_test(
             if unsupported_modes:
                 raise ValueError(f"Unsupported support-matrix mode(s): {sorted(unsupported_modes)}")
         constraints = _get_test_constraints(model)
+        encoder_coverage = _get_encoder_coverage(model)
+        image_workload = encoder_coverage.workload
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a8d53ff3. Model metadata failures now use a fatal `ModelMetadataError`, are validated in the parent, and propagate through parallel and sequential retry paths. Added tests for both paths.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in e027dc5e. The comparator permits only the one-time legacy unsupported-encoder migration, with later hardening in 10cf6687/e7ab7267 requiring a real declared/unimplemented encoder and a legacy row without image evidence.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in e027dc5e. Encoder declaration detection includes architectures registered in `MULTIMODAL_TEXT_CONFIG_KEY`, covering trimmed Llama 4 and Step-3.7 configs.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in e027dc5e. Parity entries carry the canonical image workload through probe and Pareto execution and assert positive encoder evidence.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in e027dc5e. Legacy multimodal PASS rows cannot be upgraded to the current schema without explicit image workload evidence.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in 36405a36. Default parity scans skip legacy unsupported-encoder PASS rows without certifying their text backbones; f228e5c6/59fdaea3 also purge stale resumed results and fail closed on metadata errors.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in 36405a36. Worker thread caps are applied before importing pandas/native runtimes.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in 36405a36. Unsupported-encoder rows persist an exact support-matrix preflight replay command with expected FAIL/error-prefix assertions.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in 10cf6687. `ENCODER_NOT_EXERCISED` is a regression even when both engines fail identically, so probe-only scans cannot report green.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in 10cf6687. Text-only entries retain their historical base key; only image-bearing entries gain a workload suffix.

tools/support_matrix/compare_support_matrix.py
jasonqinzhou: Fixed in 10cf6687. The migration waiver is limited to legacy rows without persisted image metadata; image-backed PASS regressions remain blocking.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in e7ab7267. Missing encoder evidence is classified as a Pareto regression and makes the scan fail.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in e7ab7267. Seeding an image-backed entry deletes the superseded base entry and its probe/Pareto results.

tools/support_matrix/compare_support_matrix.py
jasonqinzhou: Fixed in e7ab7267. The waiver now verifies `_get_encoder_coverage` reports a declared but unimplemented encoder.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in e7ab7267. Encoder evidence errors are included in detailed regression output with Python/Rust error text.

tools/support_matrix/compare_support_matrix.py
```diff
@@ -190,26 +201,105 @@ def check_csv_sanity(header: list[str], data_rows: list[list[str]]) -> list[str]
...
+                            )
+                    if "--no-save" not in command_parts:
+                        errors.append(f"Row {i}: encoder-unsupported replay command must include --no-save")
+                else:
+                    database_modes = [value.upper() for value in _option_values("--database-mode") if value]
+                    if database_modes != [expected_database_mode]:
+                        errors.append(
+                            f"Row {i}: replay command must include exactly one effective "
+                            f"--database-mode {expected_database_mode}; found {database_modes or 'none'}"
+                        )
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in a8d53ff3. `_option_values` now records bare flags and is reused for image flags; the duplicate/trailing-bare regression test is included.

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in f228e5c6. Resumed scans purge skipped encoder-unsupported entries plus probe/Pareto results; 59fdaea3 makes coverage resolution fail closed.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -309,11 +362,42 @@ def init_db(db_path: Path) -> None:
...
+        retired_keys = []
+        for entry_key, model in conn.execute("SELECT entry_key, model FROM entries"):
+            if model not in coverage_by_model:
+                try:
+                    coverage_by_model[model] = _get_encoder_coverage(model)
+                except Exception:
+                    # Historical scan databases can outlive bundled configs.
+                    # Leave an unresolvable entry alone rather than preventing
+                    # the rest of the resumable scan from starting.
+                    coverage_by_model[model] = None
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 59fdaea3. Coverage lookup exceptions now abort the scan and roll back the retirement transaction; a regression test verifies the existing database is not reported or partially mutated.

tools/support_matrix/compare_support_matrix.py
```diff
@@ -190,26 +201,100 @@ def check_csv_sanity(header: list[str], data_rows: list[list[str]]) -> list[str]
...
+                    ):
+                        errors.append(
+                            f"Row {i}: replay command includes image arguments but image workload metadata is empty"
+                        )
+
+                    try:
+                        encoder_coverage = _get_encoder_coverage(row[0])
+                    except Exception:
+                        encoder_coverage = None
+                    if status in {STATUS_PASS, STATUS_HYBRID_PASS} and encoder_coverage is not None:
```
jasonqinzhou: [P2] Enforce the canonical workload in CSV validation

For an encoder-supported PASS row, this only requires that all three image fields are populated, so values such as `448,448,1` pass validation when the command matches, despite generation defining the required workload as `1024,1024,1`; likewise, a text-only model can carry arbitrary image metadata. This lets the comparison gate accept rows that violate the new coverage contract. Compare the persisted values with `encoder_coverage.workload.csv_values()` and require empty values when the workload is `None`.

jasonqinzhou (commented):
Automated final review found one remaining support-matrix validation gap.

tests/unit/support_matrix/test_scan_rust_parity_multimodal.py
```diff
@@ -0,0 +1,366 @@
...
+    )
+
+    with pytest.raises(RuntimeError, match="metadata unavailable"):
+        seed_entries(db_path, [])
+
+    # The failed retirement transaction leaves the database unchanged for an
+    # operator to diagnose; it is never reported by this aborted scan.
+    with _connect(db_path) as conn:
+        assert conn.execute("SELECT entry_key FROM entries").fetchall() == [(legacy_entry.key,)]
+
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 15fd22c6. The rollback test now seeds two entries plus probe/Pareto rows, performs the unsupported-encoder retirement for the first entry, fails metadata resolution for the second, and verifies that both entries and every result row are restored by transaction rollback.

tools/support_matrix/compare_support_matrix.py
jasonqinzhou: Fixed in 400dd34f. CSV validation now requires the exact canonical workload for implemented encoders and rejects image metadata for text-only or otherwise workload-free models. Two regression tests cover noncanonical encoder metadata and text-only image metadata.

tools/support_matrix/support_matrix.py
```diff
@@ -1053,6 +1264,18 @@ def _run_parallel_combinations(
...
+                        if remaining is not future:
+                            remaining.cancel()
+                    logger.exception(
+                        "Fatal model metadata failure for %s/%s/%s/%s; aborting matrix generation",
+                        model,
+                        system,
+                        backend,
+                        version,
+                    )
+                    raise
```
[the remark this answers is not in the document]
jasonqinzhou: Fixed in 15fd22c6. Fatal `ModelMetadataError` handling now cancels pending futures, terminates active executor-owned worker processes, and shuts down the pool without relying on `ProcessPoolExecutor.__exit__`; normal completion still shuts down gracefully. The worker retry test asserts termination and fatal shutdown.

jasonqinzhou (commented):
Final independent review found two remaining multimodal validation/persistence gaps.

tools/support_matrix/compare_support_matrix.py
```diff
@@ -190,26 +201,116 @@ def check_csv_sanity(header: list[str], data_rows: list[list[str]]) -> list[str]
...
-                        else:
-                            database_modes.append(command_parts[part_index + 1].upper())
-                    elif part.startswith("--database-mode="):
-                        value = part.partition("=")[2]
-                        if value:
-                            database_modes.append(value.upper())
-                        else:
-                            missing_database_mode_value = True
-                if missing_database_mode_value or database_modes != [expected_database_mode]:
+                encoder_unsupported_replay = err_msg.startswith("ENCODER_UNSUPPORTED:") and (
```
jasonqinzhou: [P2] Require the encoder-preflight replay contract

When an `ENCODER_UNSUPPORTED:` row contains an ordinary `aiconfigurator cli default` command, this expression evaluates false and the row falls through to generic SILICON validation; the path also never requires `Status` to be `FAIL`. Consequently, an unreproducible command or even a `HW_INCOMPATIBLE`/`FRAMEWORK_INCOMPATIBLE` row with this error passes sanity checks. Derive the special case from the error prefix and require both `FAIL` and the generator assertion command.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -309,11 +362,34 @@ def init_db(db_path: Path) -> None:
...
+                conn.execute("DELETE FROM probe_results WHERE entry_key = ?", (entry_key,))
+                conn.execute("DELETE FROM pareto_results WHERE entry_key = ?", (entry_key,))
+                conn.execute("DELETE FROM entries WHERE entry_key = ?", (entry_key,))
+
         for entry in entries:
+            if entry.image_workload is not None:
+                # Pre-image scans used the base key for multimodal rows. Once
+                # encoder coverage is required, remove that stale text-only
+                # certification and its results before inserting the image key.
+                conn.execute("DELETE FROM probe_results WHERE entry_key = ?", (entry.base_key,))
```
jasonqinzhou: [P2] Retire superseded image-workload parity keys

When `--continue-across-commits` reuses a database after an existing row's image workload changes, this only deletes the text-only base key. An old key such as `...|image=448x448x1` and its probe/Pareto results therefore remain alongside the new `...|image=1024x1024x1` entry, causing reports and CSV exports to continue counting obsolete certification. Delete sibling workload keys sharing `entry.base_key`, except the current key, before insertion.

jasonqinzhou: Fixed in 2a06e336. Resumed parity seeding now retires the text-only base key and every sibling `|image=...` key except the current workload, including associated probe/Pareto rows. A regression test seeds and removes a stale 448x448x1 certification.

tools/support_matrix/compare_support_matrix.py
jasonqinzhou: Fixed in 2a06e336. `ENCODER_UNSUPPORTED` now unconditionally selects the preflight contract, requires `FAIL`, requires the generator replay path, and validates every expected assertion flag. Regression cases cover an ordinary CLI command and a mismatched incompatibility status.

tools/support_matrix/support_matrix.py
```diff
@@ -186,6 +274,74 @@ def _support_matrix_row_command(
...
+            "ENCODER_NOT_EXERCISED: "
+            f"{mode} {engine_step_backend} results for the canonical "
+            f"{workload.num_images}x{workload.image_height}x{workload.image_width} image workload "
+            f"are missing encoder evidence column(s): {', '.join(missing_columns)}."
+        )
+
+    invalid_columns = []
+    for column in required_columns:
+        values = pd.to_numeric(pareto_df[column], errors="coerce")
+        if values.isna().any() or (values <= 0).any():
```
jasonqinzhou, line 335: [P2] Compare encoder metrics before accepting parity

Both engines are checked independently for positive evidence, but `_compare_pareto_dfs` can fall back to a frontier envelope that excludes encoder fields. A large Python/Rust mismatch in encoder latency or memory can therefore still PASS. Compare encoder evidence directly before allowing relaxed frontier fallback.

tools/support_matrix/compare_support_matrix.py
```diff
@@ -140,6 +150,7 @@ def check_csv_sanity(header: list[str], data_rows: list[list[str]]) -> list[str]
         return errors
 
     seen_keys: dict[tuple[str, ...], int] = {}
+    header_has_source = header in (SUPPORT_MATRIX_HEADER, SUPPORT_MATRIX_HEADER_WITH_SOURCE)
```
jasonqinzhou, line 164: [P2] Reject legacy headers for the new matrix

The legacy 10-column header remains readable, but if it is accepted as `--new`, image workload validation is skipped. That permits a backbone-only Qwen3-VL PASS—or an unimplemented encoder reported PASS—to masquerade as current coverage. Keep legacy schema support for `--old`, but require the 13-column evidence schema for the new matrix.

jasonqinzhou (commented):
Final automated review found two multimodal evidence gaps to address.

tools/support_matrix/compare_support_matrix.py
jasonqinzhou: Fixed in eac6948d. New-matrix sanity validation now requires the current 13-column image-evidence schema by default; legacy headers remain explicitly readable only for legacy input inspection. A regression test rejects the prior Command/Source header as newly generated output.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in eac6948d. Python/Rust parity now compares every shared encoder evidence column with strict tolerances before any relaxed Pareto-frontier fallback. Dense Qwen3-VL agg/disagg still passes both engines, and regression tests cover aggregate and disaggregated memory drift.

jasonqinzhou (commented):
The clean-head review found remaining strict-parity and legacy-row gaps.

tools/support_matrix/support_matrix.py
```diff
@@ -556,6 +753,10 @@ def _compare_pareto_dfs(
     if python_df.empty and rust_df.empty:
         return None
 
+    encoder_mismatch = _compare_encoder_evidence(python_df, rust_df, rtol=rtol, atol=atol)
```
jasonqinzhou: [P1] Keep encoder tolerances strict during envelope fallback

`scan_rust_parity.pareto_entry` retries this comparator with `rtol=1.0, atol=1.0` to classify relaxed envelope passes. Those values now loosen encoder evidence too, allowing Python encoder latency 1 and Rust latency 2 to become `ENVELOPE_PASS`. Give encoder evidence dedicated strict tolerances or preserve the first encoder mismatch.

tools/support_matrix/support_matrix.py
```diff
@@ -1390,7 +1737,21 @@ def _row_values(row: tuple[str, ...]) -> tuple[str, str, str, str, str, str, str
             else:
                 raise ValueError(f"Invalid support-matrix result row length: {len(row)}")
 
-            if legacy_row:
+            if len(row) != 13 and status in {STATUS_PASS, STATUS_HYBRID_PASS}:
```
jasonqinzhou: [P2] Reject legacy multimodal failures without image evidence

The legacy-row guard only covers PASS/HYBRID_PASS. An 8/9/10-column non-PASS row for an implemented encoder is upgraded with blank image fields, then immediately fails current CSV sanity validation. Reject all legacy multimodal rows before writing (or define a validator-compatible migration).

jasonqinzhou:
Automated review follow-up — [P2] Preserve the unsupported-encoder replay command in the support-matrix UI.

Every `ENCODER_UNSUPPORTED` row stores a generator assertion command, but `docs/support-matrix/index.html::rowCommand` currently rejects commands containing `tools/support_matrix/generate_support_matrix.py` and synthesizes a text-only CLI command. The Reproduce Command UI can therefore pass the language backbone instead of replaying the recorded failure. Return the persisted preflight command.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in cf9249ba. All legacy 8/9/10-column rows for declared multimodal checkpoints are rejected before serialization, regardless of status. The parameterized regression test covers both PASS and FAIL inputs.

jasonqinzhou:
Fixed the UI replay finding in cf9249ba. `rowCommand` now returns persisted support-matrix generator commands verbatim, so `ENCODER_UNSUPPORTED` cells replay the asserted failure instead of synthesizing a text-only CLI command. Added a focused docs regression check.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in cf9249ba. `_compare_pareto_dfs` now accepts dedicated encoder tolerances, and both parity-scan passes pin them to the strict 1%/1e-3 thresholds even when ordinary Pareto metrics use the relaxed envelope classification. A regression test proves 1→2 encoder latency is `DRIFT`, never `ENVELOPE_PASS`.

tools/support_matrix/support_matrix.py
```diff
@@ -538,6 +695,46 @@ def _compare_frontier_envelope(
...
+    python_df: pd.DataFrame,
+    rust_df: pd.DataFrame,
+    *,
+    rtol: float,
+    atol: float,
+) -> str | None:
+    """Compare encoder metrics before any relaxed Pareto-envelope fallback."""
+    mismatches: list[str] = []
+    for column in sorted(_ENCODER_EVIDENCE_COLUMNS & set(python_df.columns) & set(rust_df.columns)):
+        python_values = sorted(float(value) for value in pd.to_numeric(python_df[column], errors="coerce").dropna())
```
jasonqinzhou: [P2] Preserve encoder metrics' configuration association

Independently sorting encoder values lets Rust attach the right set of metrics to the wrong TP/configuration rows. The later relaxed frontier comparison ignores encoder metrics and can still PASS. Compare evidence keyed by configuration columns when configurations match; use only an explicit encoder envelope when configurations differ.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -532,6 +640,10 @@ def _classify_probe(
     ttft_drift: float | None,
     tpot_drift: float | None,
 ) -> str:
+    if any("ENCODER_NOT_EXERCISED:" in error for error in (python_err, rust_err) if error):
```
jasonqinzhou, line 649: [P2] Fail image probes on encoder execution errors

For an image entry, matching Python/Rust encoder execution errors that lack the `ENCODER_NOT_EXERCISED` marker fall through to `BOTH_ERROR_PASS`. Pass image-workload context into classification and make any probe error non-passing when encoder execution is required.

tools/support_matrix/scan_rust_parity.py
```diff
@@ -618,7 +746,9 @@ def pareto_entry(entry: Entry) -> ParetoRecord:
 
     # Regression: baseline says PASS but Python or Rust failed this run.
     if not python_ok:
-        outcome = PARETO_STATUS_ERROR
+        outcome = (
+            PARETO_STATUS_REGRESSION if python_err and "ENCODER_NOT_EXERCISED:" in python_err else PARETO_STATUS_ERROR
```
jasonqinzhou: [P2] Make image Pareto execution failures block the scan

A normal Python encoder error or empty image Pareto frame maps to `ERROR`, but `_run_pool` exits nonzero only for DRIFT/REGRESSION. A baseline-PASS image entry can therefore fail execution while the scan exits 0. Classify any unsuccessful image Python result as a regression.

jasonqinzhou (commented):
The clean-head review found three multimodal parity false-success paths.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in d327d72f. Encoder evidence is now aligned by non-approximate configuration columns whenever Python and Rust select the same configurations; only differing configuration sets fall back to an explicit min/max encoder envelope. Added a regression test that swaps encoder metrics between TP rows. Verified by 172 support-matrix tests and the full unit suite (3,333 passed, 10 skipped, 751 deselected).

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in d327d72f. Image-workload probes now require successful execution from both engines, so any Python or Rust error classifies as `ENCODER_EVIDENCE_ERROR` and blocks the scan. Added a matching non-marker encoder execution-error regression test. Verified by 172 support-matrix tests and the full unit suite (3,333 passed, 10 skipped, 751 deselected).

tools/support_matrix/scan_rust_parity.py
jasonqinzhou: Fixed in d327d72f. Any unsuccessful Python Pareto result for an image-workload entry now classifies as `REGRESSION`, including ordinary execution errors and empty results, so `_run_pool` exits nonzero. Added a focused regression test. Verified by 172 support-matrix tests and the full unit suite (3,333 passed, 10 skipped, 751 deselected).

tools/support_matrix/support_matrix.py
```diff
@@ -556,6 +782,15 @@ def _compare_pareto_dfs(
...
+        python_df,
+        rust_df,
+        rtol=rtol if encoder_rtol is None else encoder_rtol,
+        atol=atol if encoder_atol is None else encoder_atol,
+    )
+    if encoder_mismatch:
+        return encoder_mismatch
+
     approximate_columns = [col for col in python_columns if col in _APPROXIMATE_ENGINE_STEP_COLUMNS]
     identity_columns = [col for col in python_columns if col not in _APPROXIMATE_ENGINE_STEP_COLUMNS]
```
jasonqinzhou: [P1] Exclude encoder metrics from exact configuration identity

When multimodal Python/Rust encoder values differ non-exactly but remain within `encoder_rtol`, the dedicated encoder check succeeds, yet those columns remain in `identity_columns` below. The exact frame comparison then treats the same configuration as different and switches to the relaxed frontier comparison. For example, a 0.5% encoder delta plus a 4% TPOT delta can return no mismatch in a 1% strict / 5% frontier scan even though TPOT violates the strict tolerance. Exclude `_ENCODER_EVIDENCE_COLUMNS` from the subsequent identity comparison and cover this combination with a regression test.

jasonqinzhou (commented):
Built-in Codex clean-head review found one additional strict-parity issue; applying and validating it now.

tools/support_matrix/support_matrix.py
jasonqinzhou: Fixed in ce7d92d5. Encoder evidence remains independently aligned and compared with its dedicated tolerance, but encoder columns are now excluded from exact deployment-configuration identity. The new regression proves that an allowed 0.5% encoder delta cannot route a 4% TPOT delta around the strict 1% check. Verification: 23 focused parity tests passed, all 173 support-matrix tests passed, and the full unit suite passed with 3,334 passed, 10 skipped, and 751 deselected.

tools/support_matrix/compare_support_matrix.py
```diff
@@ -190,25 +212,122 @@ def check_csv_sanity(header: list[str], data_rows: list[list[str]]) -> list[str]
...
+                        actual_values = _option_values(flag)
+                        if actual_values != [expected_value]:
+                            errors.append(
+                                f"Row {i}: encoder-unsupported replay command must include exactly one "
+                                f"{flag} {expected_value}; found {actual_values or 'none'}"
+                            )
+                    if "--no-save" not in command_parts:
+                        errors.append(f"Row {i}: encoder-unsupported replay command must include --no-save")
+                else:
+                    database_modes = [value.upper() for value in _option_values("--database-mode") if value]
```
jasonqinzhou: [P2] Reject empty duplicate database-mode flags

When a replay command contains a valid `--database-mode SILICON` plus a trailing `--database-mode` or `--database-mode=`, `_option_values` returns an empty occurrence but this comprehension discards it, so `check_csv_sanity` accepts a command that fails argument parsing when executed. Preserve empty occurrences in the cardinality check and cover both malformed duplicate forms.

jasonqinzhou (commented):
Built-in Codex clean-head review found one replay-validator edge case; applying and validating it now.

tools/support_matrix/compare_support_matrix.py
jasonqinzhou: Fixed in `f35d9339`. The database-mode cardinality check now preserves empty occurrences, so both a valid `--database-mode SILICON` followed by a bare `--database-mode` and by `--database-mode=` are rejected. Added both regression forms to `tests/unit/support_matrix/test_compare_support_matrix.py`; the focused comparator suite passes (`39 passed`).

jasonqinzhou:
Closing this AIConfigurator PR because the repository is being refrozen for feature development; ongoing feature work, review, and landing are moving to AISimulate.

The one-to-one continuation is ai-dynamo/aisimulate#85: https://github.com/ai-dynamo/aisimulate/pull/85

Please continue review and development there. This AIC PR is retained as source history.

PR #1533, opened 2026-08-13
fix(collector): align Blackwell GEMM and GDN collection with serving

Arsene12358:
## What this PR does

This PR fixes two collector/data mismatches found while auditing SGLang v0.5.14 decode and addresses the subsequent review findings around serving selection, B300 data quality, and provenance.

**AIC-1744 (Closes AIC-1744): remove load-time FP8-block weight packing from timed GEMM calls.** The SGLang collector previously passed raw FP32 weight scales into `gemm_nt_f8f8bf16` on every replay, charging DeepGEMM's one-time UE8M0 preparation to every inference step. The collector now prepares weights and scales once during case setup while retaining per-token activation quantization in the measured call. A behavioral regression test pins that boundary.

**AIC-1745 (Closes AIC-1745): make GDN collection and estimation follow SGLang's serving predicate exactly.** FlashInfer decode is eligible only for SGLang on compute-capability major 10 with BF16 SSM state; FP32 state, Hopper, and SM120 use FLA. The collector executes only the selected lane, and the Rust estimator uses the same predicate, prefers the matching physical lane, and falls back to dtype-aware SOL when required FlashInfer data is absent rather than consuming non-serving FP32 FLA silicon. The recurrence/state byte model uses two-byte traffic for BF16/FP16 state and four-byte traffic for FP32 state.

## Exact data delta

- The final diff changes exactly four parquet files, all GEMM. It changes no packaged GDN parquet.
- `b200_sxm`, `gb200`, and `gb300` retain their exact 130,536-row key sets; all and only the 29,526 `fp8_block` latencies per system are replaced with setup-prepacked measurements. No other dtype changes on those systems.
- `b300_sxm` is a full replacement of the borrowed 130,536-row `NVIDIA GB300` table with the first true `NVIDIA B300 SXM6 AC` table: 130,906 unique rows, preserving every prior logical shape and adding 222 BF16 plus 148 FP8 shapes. Its 29,526 FP8-block values are serving-true measurements.
- The final packaged SGLang v0.5.14 plan contains 74 GDN cases, all with FP32 state, so it schedules zero FlashInfer invocations. The eight retained GDN tables contain 8,820 rows each and zero `flashinfer_gated_delta_rule_decode` rows; they are byte-identical to `main`.
- After `main`'s queryable-version pruning, the backend catalog remains a 1,746-key registry. This PR changes exactly nine surviving SGLang generation facts from `causal_conv1d`/FLA to FLA/`unverified`; it adds or removes no fact keys and changes no kernel-source sets.

## Review findings addressed

- The GDN predicate is covered at the SM100 boundary, SM103, SM120, Hopper, BF16, FP32, eligible-data-present, and eligible-data-absent cases. The repo-owned FP32 recipes are also checked end to end to ensure no non-serving FlashInfer collection is planned.
- The current GDN parquet identity has no state-dtype column, so repository collection rejects non-FP32 cases until a separately approved schema migration can represent them. Untyped packaged state-sensitive rows are treated as FP32; non-FP32 queries use silicon only for an exact SM-major-10 BF16 FlashInfer lane and otherwise fall back to dtype-aware SOL. State-independent causal-convolution rows remain safely reusable across state dtypes.
- The reported native-B300 BF16 stall at `(m,n,k)=(96,3584,256)` was re-collected from `0.068639` ms to `0.002486` ms. Five other isolated spikes were healed in the same review pass (five BF16, one NVFP4 total). Each flagged key and both adjacent `m` neighbors were collected in two independent runs; the per-key minimum for the six flagged keys was landed. The greater-than-3x two-neighbor scan now reports zero isolated spikes.
- `GdnOp.mamba_ssm_dtype` changes positional bincode layout, so `ENGINE_SPEC_SCHEMA_VERSION` is bumped from 13 to 14. A regression reconstructs a real pre-field v13 GDN payload and verifies rejection as `UnsupportedSchemaVersion` before op decoding.
- All four GEMM sidecars use metadata schema v2 with exact collector refs, collector hashes, case-plan hashes, runtime image digests, shipped-row attribution, and separate source-campaign counts/status. The B300 healing events record 1 shipped row from a 3-case campaign and 5 shipped rows from a 15-case campaign; their case-plan hashes are explicitly identified as reconstructed from the exact logged filters because those scoped jobs did not export sidecars.
- Collector publication is transactionally recoverable across the complete parquet batch, sidecar, retained staging files, and checkpoint ledgers. The original merge-versus-replacement decision and exact row contribution are made durable before the first target mutation, so an exception or hard exit cannot expose a silent partial batch or turn an incompatible replacement into a merge with stale history on retry.
- The earlier measured-impact and FlashInfer-row claims were removed: they came from an integration state that assumed BF16 Qwen state, while the shipped Qwen configurations use FP32 and remain on FLA.

## Validation

- Rust core: 513 passed, 2 ignored.
- Engine-step and compile-engine parity: 336 passed against a source-built release extension.
- Regenerated golden populations remain unchanged at 243 engine-step records, 49 compile references, and 9 per-op cases. An independent audit attributed every material value shift to the intended GEMM replacements; Qwen GDN surfaces changed only at floating-point ULP scale, and a second refresh was byte-identical.
- Collector unit suite after the `main` integration and crash-recovery fix: 1,381 passed, 7 skipped; the final crash-safety/schema/GDN/data gate passes all 411 tests.
- SDK unit suite: 2,135 passed. The CI-equivalent full unit suite with strict provenance enabled passes 4,451 tests with 11 skips. Collector data rules R1-R7 pass, and the committed backend catalog matches all 1,746 fact slices.
- Rust formatting, Ruff, `git diff --check`, DCO across all 43 branch commits, exact parquet key/value census, metadata closure hashes, and full-branch Standards/Spec reviews pass.

## Disclosures

- GB200 FP8-block values are the deterministic per-key minimum of two independent full campaigns. Attribution exhausts all 29,526 keys: 14,824 from the first run (including 65 ties assigned first) and 14,702 strict improvements from the redo. The final median new/old latency ratio is 0.6497; 692 keys remain above 1.3x their old value and six `m=32` shapes remain above 1.5x both immediate neighbors because the elevated measurements reproduced across runs rather than disappearing as one-sided noise.
- The true-B300 healing case-plan hashes were reconstructed only for the two exact scoped review jobs; collector refs, collector hashes, runtime identity, logged filters, source counts, and shipped counts are preserved explicitly.
- SGLang 0.5.12 continues to inherit approved data through the existing reuse mappings.
- The compiled `forward_model="fpm"` whole-model path remains unchanged and has no lane-specific empirical selection.
- The vLLM and TensorRT-LLM GEMM collectors were audited for the same packing defect and already perform their one-time weight transformations outside the timed call.

## Summary by CodeRabbit

* **New Features**
  * Added dtype-aware GDN performance modeling, memory sizing, and kernel routing.
  * Added FlashInfer decode support for compatible BF16 workloads and GPU configurations.
  * Added richer multi-event collection provenance and metadata validation.
  * Improved collection recovery, resumability, atomic publication, and performance-file integrity.

* **Bug Fixes**
  * Prevented unsupported or incomplete kernel data from silently falling back to incorrect measurements.
  * Fixed FP8 block GEMM setup to pre-pack weights and scales once.

* **Documentation**
  * Updated engine schema and collection metadata documentation for the new formats.


aic-core/src/aiconfigurator_core/systems/data/b200_sxm/gemm/sglang/0.5.14/collection_meta.yaml
```diff
@@ -8,4 +8,12 @@ runtime:
...
 tables:
   gemm_perf:
+    # 2026-08-13: REPLACED all 29526 fp8_block rows (AIC-1744) — the prior
+    # rows billed load-time UE8M0 weight-scale packing inside every timed
+    # call (~9 us on a ~7 us gemm); the collector now pre-packs at setup
+    # exactly as serving's process_weights_after_loading does, and the
+    # timed op is act-quant + gemm only (so these rows now legitimately
+    # include the per-token-group quant serving pays each step).
+    # Example m32 @ n5120xk4096: 20.6 -> 11.6 us. Same key grid;
+    # values-only replacement, deliberate append-only exception.
```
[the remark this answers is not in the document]
Arsene12358: This is a deliberately coordinated cross-module change, declared in the PR body's "Disclosures for reviewers" section (same pattern as #1435/#1503/#1519): the collector fixes, the SDK/Rust lane preference, and the re-collected data are one atomic correctness story — landing the code without the data (or vice versa) would leave predictions in a worse mixed state. The data changes were human-directed (AIC-1744/1745/1759 track the decisions), each parquet edit is documented in its collection_meta.yaml with provenance, and the landing invariants (key-multiset preservation for replacements, cross-disjointness for appends) are asserted in the landing scripts and re-verified in the final whole-branch review.

aic-core/src/aiconfigurator_core/systems/data/gb200/gemm/sglang/0.5.14/collection_meta.yaml
```diff
@@ -8,4 +8,18 @@ runtime:
...
+    # timed op is act-quant + gemm only (so these rows now legitimately
+    # include the per-token-group quant serving pays each step).
+    # This gb200 run is noisier than its gb300/b200_sxm siblings: 19 (n,k)
+    # shapes show m=32 rows >1.5x their m-neighbors (12 of them >15us),
+    # ~360 small-m interior spikes, and 1203 keys (4.1%) sit >1.3x above
+    # the replaced rows (vs 4 on gb300, 142 on b200_sxm). Median new/old
+    # is 0.673, the same overall win as the sibling systems, so the
+    # corrected rows still strictly improve on the packing-biased ones;
+    # a gb200 gemm re-collection is tracked as a follow-up. Same key grid;
+    # values-only replacement, deliberate append-only exception.
```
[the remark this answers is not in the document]
Arsene12358: Done — this exact re-collection happened before the PR left draft-era state: a second independent gb200 run (auto-collector pipeline 62701570) confirmed the noise is environmental and one-sided (runs agree at median 1.000; ~6% of keys spike >1.3x in each run, mostly at different keys). The landed rows are now the per-key MIN of the two independent runs: m=32 spike shapes drop 24 -> 7, and the ~6 surviving shapes spike identically in both runs — reproduced platform behavior, not noise, kept as measured. The meta comment documents the full profile; further re-runs would not change the picture (a third run + median-of-3 is possible if reviewers want the last few shapes smoothed, tracked under AIC-1759 which is now closed with this analysis).

collector/sglang/collect_gdn.py
```diff
@@ -329,6 +335,41 @@ def run_gdn_scan():
...
+        return None
+    try:
+        from flashinfer.gdn_decode import gated_delta_rule_decode_pretranspose
+    except (ImportError, RuntimeError) as e:
+        print(
+            f"  SM{sm_version}: FlashInfer bf16 GDN decode lane unavailable "
+            f"(flashinfer.gdn_decode import failed: {e}); skipping "
+            "flashinfer_gated_delta_rule_decode rows for this case."
+        )
+        return None
```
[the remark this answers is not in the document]
Arsene12358: Fixed in f63178f: on SM100+ the flashinfer import failure is now routed through the collector's classified-failure mechanism (raised after the flashinfer-independent lanes have logged their rows, so the always-collectible data isn't lost), matching failure_handling.md's fail-closed doctrine. The silent None-skip remains only for sm<100 where the lane is genuinely not applicable, with the logged reason. Seven new unit tests pin both branches.

Arsene12358:
Rebased onto origin/main (899034f5). Three conflicts resolved: `mamba.py` took main's side wholesale — the pyo3 op-unification (#1552/#1555/#1566) deleted `GDNKernel._query_gdn_table` and its `get_sol` closure, both now banned def shapes under the rust-core parity rules, so this branch's Python diff had nothing left to apply; its two behaviors (SM100+ lane preference, per-lane state bytes) already live in Rust as the sole implementation. `perf_database/mod.rs` and `state_space.rs` merged main's new fallible `SourceResolver` table construction with this branch's `sm_version` threading, keeping main's mechanism and adding `sm_version: Option<u32>` to the signature. `engine_step.json` took main's side, then re-pinned this branch's own 7 records fresh against the rebased tree — the same records the original pin moved, still shifted by this branch's b200/b300 `gemm_perf` replacement.

Two new commits sit on top of the 15 rebased ones: the goldens re-pin, and a from-scratch rewrite of `test_gdn_flashinfer_lane.py` — the old file's synthetic table-injection technique is invisible to the compiled engine post-unification, so every test now runs the sanctioned engine-query path against real on-disk tables. The lane behavior was re-proven post-rebase: mutation-tested (inverting the SM100 gate or the 2-byte state constant fails the suites), plus a 17-check fallback probe covering lane-absent systems (h200 SM90, pre-lane versions) and the 32 uncovered geometries on gb300 — the fla lane prices with exact 4-byte SOL semantics or its own silicon rows, and the 5 covered geometries resolve the flashinfer rows at measured ratios of 0.455–0.548, clustering on the 0.50 state-bytes prediction.

No schema bump; all 8 perf parquet files are byte-identical to their pre-rebase state (SHA-256 verified across five refs — main contributed zero bytes to them). Known coverage ceiling, disclosed in the collection_meta and worth a follow-up: the flashinfer lane ships rows for the 5 Qwen3.5-397B TP-shard geometries only; the other 32 GDN decode geometries on SM100+ still price via the fla lane until their lanes are collected.

Gates at f21fbe5d: cargo 505+3, unit 3904, parity 447, cross-package 88, collector checks R1–R7 + facts registry (3,319 slices) green, ruff clean.

aic-core/rust/aiconfigurator-core/src/perf_database/state_space.rs
```diff
@@ -139,6 +149,7 @@ impl StateSpaceTable {
             gdn_sources,
             kda_sources,
             vllm_024_gdn_aliases: backend == "vllm" && version == "0.24.0",
+            sglang_sm100_gdn_flashinfer_lane: backend == "sglang" && sm_version.unwrap_or(0) >= 100,
```
jasonqinzhou: **[P1] Gate the Rust FlashInfer preference on the actual serving contract.** After the pyo3 unification this boolean is the sole runtime gate, and it enables the alias for every SGLang system with `sm_version >= 100`. In pinned SGLang v0.5.14, `_handle_linear_attn_backend` auto-selects FlashInfer only when `is_sm100_supported()` (capability major 10, not SM120) and `mamba_ssm_dtype == "bfloat16"`. Every bundled Qwen3.5/Qwen3.6 config that declares this field pins `float32`, and `Qwen35Model` still budgets a 4-byte state, so the default AIC query should remain on FLA. At the packaged B300 397B TP4 shape, current-head routing resolves the FLA query to 0.001918/0.006195/0.016733 ms at batch 1/32/128 instead of the FLA rows' 0.004145/0.012736/0.048911 ms. Please thread state dtype or the explicit decode-backend choice into Rust construction and mirror the exact major-10 + BF16 predicate, keeping FLA precedence otherwise, with regressions for FP32 on SM100, BF16 on SM100, and SM120.

collector/sglang/collect_gdn.py
```diff
@@ -329,6 +335,55 @@ def run_gdn_scan():
...
+      this message as a classified failure (layer_permissions.md: "execute
+      it, or raise") once it has attempted the other decode lanes for the
+      point, so the row is recorded as failed instead of silently omitted.
+
+    The lazy import mirrors sglang's own guard for this kernel
+    (gdn_flashinfer.py:34-56 @0.5.14, _get_flashinfer_gdn_kernels ->
+    gated_delta_rule_decode_pretranspose).
+    """
+    sm_version = get_sm_version()
+    if sm_version < 100:
```
jasonqinzhou: **[P1] Match the collector's required lane to SGLang's exact serving predicate.** This resolver treats every SM >= 100 as requiring FlashInfer and describes BF16 as the model default. In pinned v0.5.14, the auto-select branch is `is_sm100_supported() && mamba_ssm_dtype == "bfloat16"`, and that helper restricts capability majors to 10. The repository's Qwen3.5/Qwen3.6 configs pin `float32`, while `rtx_pro_6000_server` is SM120; at this head `_resolve_flashinfer_gdn_decode()` returns a classified required-lane error on SM120 even though serving would not select that lane. Please carry state dtype or the explicit decode-backend choice into the collection case and mirror the exact major-10 + BF16 gate, keep FLA-only cases from requiring FlashInfer, and add FP32-SM100 and SM120 regressions.

aic-core/src/aiconfigurator_core/systems/data/b300_sxm/gemm/sglang/0.5.14/collection_meta.yaml
```diff
@@ -8,4 +8,12 @@ runtime:
   version: 0.5.14
 tables:
   gemm_perf:
+    # 2026-08-14 (AIC-1759): FULL-TABLE replacement with the first true
```
jasonqinzhou, line 17: **[P1] Re-collect the isolated B300 GEMM stall before marking this replacement complete.** The BF16 row at `(m=96, n=3584, k=256)` is still 0.068639 ms, while its immediate `m=81/97` neighbors are 0.002564/0.002522 ms and the replaced main-branch row is 0.002483 ms—an isolated roughly 27x stall. The current table has 130,906 rows with no duplicate keys or non-positive latencies, but this point can still distort the measured curve and interpolation. Please repeat the affected collection (or combine independent clean runs with a documented rule), replace or substantiate the one-sided stall, and rerun the targeted sanity scan before publishing the table as complete.

aic-core/src/aiconfigurator_core/systems/data/b300_sxm/linear_attention/sglang/0.5.14/collection_meta.yaml
```diff
@@ -9,6 +9,11 @@ runtime:
   image_digest: sha256:5027e95bf6ec536856b1b52a91d1f35ff5c564ab83e8a94758a169ff09bb8df3
 tables:
   gdn_perf:
+    # 2026-08-14 (AIC-1759): appended 54 flashinfer_gated_delta_rule_decode
```
jasonqinzhou: **[P2] Refresh the structured provenance for these new collections.** This comment says the 54 rows came from the 2026-08-14 manual JET run, but the machine-readable entry still attests the 2026-08-10 collector ref/hash/case plan and `rows: 8820`; the table now has 8,874 rows. The sibling full B300 GEMM replacement likewise remains `provenance: legacy` with only `status`. Please regenerate or merge the structured entries with the actual collector ref, hashes, date, row counts, and derived status for all four appended GDN tables and the new B300 GEMM collection.

jasonqinzhou (changes requested):

## What this PR does

This PR corrects Blackwell SGLang FP8-block GEMM measurements by moving weight-scale packing out of the timed loop, replaces the affected tables including a first native-B300 collection, and adds a FlashInfer BF16-state GDN decode lane across collection, Rust routing, and packaged data. The current head rebases that work across the pyo3 operation unification, rewrites the GDN lane tests through the compiled engine, and narrowly re-pins the parity records shifted by the B200/B300 GEMM replacements.

The rebase adaptation is careful and well evidenced: the obsolete Python query path was not resurrected, the GDN tests now exercise the compiled engine against real packaged tables, and the parity re-pins remain limited to the seven surfaces affected by the corrected GEMM donor chain.

## Why this score

5.5/10 — REQUEST_CHANGES. The pyo3 rebase adaptation, targeted local checks, and all 26 exact-head CI checks are clean, but the three previous correctness blockers still apply: the sole Rust runtime gate and the collector both select or require FlashInfer more broadly than SGLang's major-10-plus-BF16 serving predicate, while the packaged Qwen3.5/Qwen3.6 configs use FP32 state; and the native-B300 GEMM table still contains an isolated roughly 27x BF16 latency stall. Structured provenance for the appended and replaced data also remains stale.

aic-core/rust/aiconfigurator-core/src/perf_database/state_space.rs
Arsene12358: Fixed in 0397baec. The state dtype is now threaded from model construction into the rust gate (`GdnOp.mamba_ssm_dtype`, pyo3 kwarg with a conservative `float32` default so pre-fix serialized ops deserialize to FLA behavior), and the gate mirrors the exact serving predicate: capability major 10 only (`(100..110).contains(&sm_version)` — sm 120 excluded) AND `mamba_ssm_dtype == "bfloat16"`, FLA precedence otherwise. `qwen35.py` resolves the dtype from the HF config exactly as serving's `mamba_utils` elif chain does (text_config declares → wins outright, invalid declared value keeps float32); all 8 bundled Qwen3.5/3.6 configs pin float32.

Your cited magnitude reproduces and flips: b300_sxm 397B TP4 shard at batch 1/32/128 now returns the FLA rows **0.004145/0.012736/0.048911 ms** by default (was 0.001918/0.006195/0.016733 from the flashinfer rows); an explicit bf16 state still selects flashinfer on major-10. Regressions added for FP32-SM103, BF16-SM103, SM120 (both dtypes), SM90, and lane-absent degrade, plus a new whole-run parity case on the 0.5.14 b200 tables (both lanes present there; append-only golden pin) so the routing is pinned end-to-end.

collector/sglang/collect_gdn.py
Arsene12358: Fixed in 0397baec, mirroring the same predicate collector-side: `_resolve_flashinfer_gdn_decode` now takes the case's `mamba_ssm_dtype` and gates on `100 <= sm < 110`; SM120 is a logged not-applicable skip and can never produce a required-lane error; fp32/fp16-state cases never REQUIRE flashinfer (a missing kernel is a logged sibling-lane skip — the lane is still collected as sibling coverage on major-10 when importable, preserving the shipped tables' reproducibility); the classified failure remains exactly for bf16-state cases on major-10 with the kernel missing. The dtype rides the collection case (`GdnCommonTestCase.mamba_ssm_dtype`, in the dedup key — a proven no-op today: 74 cases before and after, all float32) and is now declared explicitly on every Qwen3.5/3.6 gdn row in both cases yamls; the incorrect "defaults to bfloat16" comment is corrected. Regressions added: SM120 both dtypes, FP32-SM103 with kernel unavailable (no error), bf16 boundary/classified-error, kernel-available for both dtypes.

aic-core/src/aiconfigurator_core/systems/data/b300_sxm/gemm/sglang/0.5.14/collection_meta.yaml
Arsene12358: Re-collected and landed in 0a205cf5. The flagged bf16 (m=96, n=3584, k=256) row re-measures at **0.002486 ms** (was 0.068639 — confirming a 27x measurement artifact; m=81/97 neighbors reproduced their stored ~0.0025 values within noise). Method: scoped `--case-filter` re-collection of the stall plus both m-neighbors on 1x B300 SXM6 AC at aws-pdx in the mirrored v0.5.14 amd64 image, two independent runs, per-key MIN landed (values-only swap; row count 130,906, keys and schema unchanged).

We also went one step further than the finding: an isolated >3x m-neighbor scan over the full table surfaced five more artifacts of the same species — bf16 (4353,32,2048) 0.0803→0.00558, (14,32,10240) 0.0525→0.00495, (96,768,32) 0.00661→0.00182, (193,12288,1024) 0.0819→0.00860, and nvfp4 (128,2048,6144) 0.0651→0.00796 — all re-collected the same way (spike + both neighbors, two runs, per-key MIN) and landed. The targeted sanity scan now reports **0 isolated >3x m-neighbor spikes table-wide**, so the table is published as complete on a clean scan rather than clean-except-unexamined.

aic-core/src/aiconfigurator_core/systems/data/b300_sxm/linear_attention/sglang/0.5.14/collection_meta.yaml
Arsene12358: Regenerated in 0397baec with honest limits. All four appended GDN metas (b200/b300/gb200/gb300 linear_attention sglang 0.5.14) now carry a merged structured attestation: top-level `rows: 8874` (verified against the actual parquets — each has exactly 54 flashinfer rows appended) plus a per-event `collections:` list that preserves the fully-attested 2026-08-10 base run (original ref/hashes/8820 rows intact) and dates the 2026-08-13/14 54-row appends (b300 via the manual JET run, mirrored `sglang:v0.5.14-cu130-amd64` image). The b300 gemm meta moves `provenance: legacy` → `local` (the kda/mla_bmm precedent for manual out-of-band collections) with `collected_at: 2026-08-14`, verified `rows: 130906`, and derived status.

One honest limitation, stated in-file rather than papered over: `collector_ref`/`collector_hash`/`case_plan_hash` for the re-collection events are not reconstructable — the branch states that actually ran were destroyed by the two post-review force-push rebases (2026-08-15 and 2026-08-20), and the run checkpoints weren't retained. Those fields are left absent with an explanatory comment; nothing was fabricated. `tools/perf_database/check_collector_data.py` passes all 7 rules on the new structure.

Arsene12358:
Review wave landed — all four open threads addressed and resolved (details in each thread). The branch is also rebased onto current main (the CONFLICTING state is cleared; 3 static goldens re-pinned for main's #1478 prefill-imbalance pricing, no overlap with this PR's content).

- **0397baec** `fix(review)`: the GDN FlashInfer lane is gated on sglang v0.5.14's exact serving predicate — capability major 10 only AND `mamba_ssm_dtype == "bfloat16"`, threaded from model construction (all bundled Qwen3.5/3.6 configs pin float32, so their defaults return to the FLA rows — your cited B300 397B TP4 numbers flip back exactly); same predicate collector-side with SM120 never erroring and fp32-state never required; structured provenance regenerated with verified row counts (non-reconstructable run hashes left absent with in-file comments, not fabricated).
- **0a205cf5** `data(gemm)`: the flagged 27x bf16 stall re-measured (0.0686 → 0.00249 ms) plus five more isolated spikes our own >3x scan surfaced, all re-collected on B300 with two independent runs / per-key MIN, values-only swap; the targeted sanity scan now reports **0** isolated spikes across the 130,906-row table.

Suites at head: cargo 520, parity 451, sdk 2040, collector 1087, collector-data checker 7/7, ruff clean. @jasonqinzhou ready for your re-review.

Arsene12358:
@jasonqinzhou The requested changes are ready for another pass.

- The collector and Rust estimator now use SGLang's exact FlashInfer GDN predicate: compute-capability major 10 plus BF16 SSM state. FP32 selects FLA; non-FP32 collection fails closed while the dtype-less GDN schema remains in place; and a required FlashInfer data miss falls back to dtype-aware SOL rather than non-serving FP32 FLA silicon. State-independent causal-convolution rows remain reusable across dtypes.
- The repo-owned plan has 74 GDN cases, all FP32, so it schedules zero FlashInfer invocations. No GDN parquet changes remain in the PR; all retained packaged GDN tables are byte-identical to `main` and contain zero FlashInfer rows.
- The native-B300 outlier at `(96,3584,256)` was re-collected from `0.068639` to `0.002486` ms, along with five other isolated spikes. Two independent neighbor-inclusive runs support the six landed minima, and the full-table greater-than-3x two-neighbor scan is now zero.
- The four GEMM sidecars now carry exact schema-v2 histories: collector refs/hashes, case-plan hashes, runtime image digests, source-campaign counts/status, and shipped-row attribution. Historical closure hashes and row attribution were independently recomputed.
- The positional GDN field requires EngineSpec schema 14; a reconstructed v13 payload is rejected at the version gate before op decoding.
- Finalization now makes the original merge/replacement decision durable before parquet publication and recovers the complete parquet/sidecar/staging/checkpoint transaction after exceptions or hard exits, including incompatible-schema replacement retries. Output-root and checkpoint operations remain bound to their locked directory descriptors, deterministic private slots stay bounded across repeated crashes, and foreign or nonregular replacements fail closed without blocking or deletion.
- The branch is synced with `main`'s queryable-version pruning. The final data diff is four GEMM parquets only; the PR body no longer claims packaged FlashInfer rows or the earlier unverified GDN impact.
- The strict-provenance CI failure is fixed without weakening schema-v2 validation: three synthetic SDK sidecars now identify themselves honestly as compatibility-schema fixtures. CodeRabbit's three follow-up findings are also addressed—the dtype-resolution comment matches serving precedence, the private-claim test guards the descriptor-relative rename actually exercised, and rollback leak assertions match the real artifact name.

Local verification includes 513 Rust core passes (2 ignored), 4 public-API Rust passes, 336 parity passes, 2,135 SDK passes, 1,381 collector passes (7 skipped), 4,451 CI-equivalent strict-provenance unit passes (11 skipped), a 411-test focused crash-safety/schema/GDN/data gate, deterministic golden regeneration, exact parquet and metadata audits, formatting, Ruff, DCO across all 43 branch commits, and clean final whole-branch Standards and Spec reviews.

Please re-review when convenient.

Arsene12358:
@coderabbitai review

aic-core/src/aiconfigurator_core/sdk/models/qwen35.py
```diff
@@ -173,6 +184,21 @@ def __init__(
...
+        root — serving's branches are an ``elif`` chain, so root is never
+        consulted once text_config declares the field); an invalid declared
+        value keeps the ``float32`` default, as serving does."""
+        text_config = raw_config.get("text_config")
+        for source in (text_config if isinstance(text_config, dict) else {}, raw_config):
+            if "mamba_ssm_dtype" in source:
+                dtype = source["mamba_ssm_dtype"]
+                return dtype if dtype in _MAMBA_SSM_DTYPE_BYTES else "float32"
+        return "float32"
+
```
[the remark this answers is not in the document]
Arsene12358: Fixed in f5766047. The caller comment now documents `text_config` precedence and the `float32` fallback for unsupported declared values, matching the existing resolver and pinned SGLang behavior.

Arsene12358:
@coderabbitai review

jasonqinzhou (approved):
**APPROVE · 9.2/10 · low confidence**

## What this PR does

This PR aligns Blackwell GEMM collection and GDN estimation with SGLang's serving behavior, corrects and attests the affected performance data, and adds crash-safe transactional publication for parquet, provenance, and checkpoint artifacts. The incremental head also versions multi-event provenance and makes GDN state dtype explicit without publishing ambiguously keyed non-FP32 rows.

The schema-v2 event history, locked-root attestations, deterministic recovery journal, and adversarial crash/replacement tests show unusually careful treatment of both auditability and failure safety.

## Why this score

The 9.2 quality score reflects the completed Codex review, which found both prior P2 findings fixed and no new actionable defect at the current head. The canonical recommendation is nevertheless BLOCKED because the mandatory independent Claude Fable review timed out after 1,800 seconds and produced no model-verified structured result; under the packet contract, an approval cannot be issued without both reviewers completed.

PR #1534, opened 2026-08-13
fix(collector): preserve power data across B200 collection paths

kaim-eng:
## Summary

- require explicit CUDA device indices for NVML power sampling
- keep power columns stable when an individual measurement is unavailable or a retained power-off CSV is resumed with power enabled
- propagate measured power through SGLang MLA, DSV4 sparse, GLM-5 sparse, MoE, and all-reduce collection paths
- preserve structural zero-work rows with zero work-attributable power and a real device power limit
- preserve communication energy through Rust MoE dispatch so Python and Rust power predictions remain aligned
- fail closed when serial chunks have incomplete power, while preserving their valid latency
- keep FP8-static GEMM energy unavailable when its base GEMM row has no power instead of subtracting powered overheads into a negative value

This PR intentionally contains **code, documentation, and tests only**. The B200 SGLang 0.5.14 and vLLM 0.24.0 parquet updates are in separate data-only PRs #1535 and #1536.

## Producer / consumer contract

This is an explicit coordinated data-contract change:

- On a power-enabled run, log_perf appends optional power and power_limit metrics. Final parquet fields are nullable doubles; an empty/null value means that the performance row is valid but no power sample is available.
- The non-power identity schema and measured latency remain unchanged. A retained power-off CSV is upgraded atomically under the writer lock before a power-enabled append.
- Collectors that cannot associate a sampled region with one emitted row opt out explicitly with include_power_columns=False; moe_a2a remains power-free.
- Consumer handling was audited and updated in attention.py, communication.py, dsa.py, dsv4.py, gemm.py, mamba.py, mla.py, and moe.py. Missing/null power loads as zero energy rather than failing database load.
- Public single-point power estimates apply the existing 90% latency-weighted coverage gate and report power unavailable below it.
- Data PRs #1535 and #1536 **must not merge before this PR** because their partial tables rely on nullable-power loader support.

## Review follow-up

- sampling is now stopped from power_monitoring_only even when a kernel or synchronize call raises
- the SGLang framework MoE benchmark resolves the string CUDA device before entering power sampling
- mixed present/missing chunk samples warn and return no aggregate power instead of dropping the latency row
- missing power / power_limit values produce actionable validation errors
- nullable parquet loading, failed-body sampler cleanup, mixed chunks, and missing-base FP8 energy have regression coverage
- power_limit is treated as a measurement by the parquet diff tool
- the dead moe_a2a power helper was removed

The proposed stale-lock heartbeat was not added: the schema upgrade is already serialized by the writer lock, campaign staging files are bounded, and introducing a lease protocol is not justified by an observed failure. The CodeRabbit request to check a falsey log_perf return is also not applicable: success returns True and failures raise PerfLogError.

## Validation

- focused WSL/Linux collector, loader, operation, and diff-tool suite: **186 passed**
- additional WSL/Linux fail-closed, schema-contract, nullable-loader, GEMM, coverage-gate, and tooling suite: **114 passed**
- Ruff check and format: passed on all changed Python files
- Rust MoE dispatch tests: **6 passed**
- Rust engine-step bridge tests: **32 passed**
- B200 vLLM MiniMax-M2.5 disaggregated support-matrix comparison: all **40** Pareto rows have identical Python/Rust power values
- prior independent final diff review: no actionable findings


collector/helper.py
```diff
@@ -113,8 +113,16 @@ class PowerMonitor:
     def __init__(self, device_id: int):
         """
         Args:
-            device_id: CUDA device index to monitor
+            device_id: Non-negative CUDA device index to monitor (must not be None;
```
tianhaox, line 116: [P1] This contract calls `device_id` a CUDA ordinal, but `_init_handle()` passes it to `nvmlDeviceGetHandleByIndex()`. Those index spaces are not equivalent, especially here because `launch_mla_subprocess` sets `CUDA_VISIBLE_DEVICES=<physical N>` and then passes `cuda:0`: CUDA ordinal 0 maps to the selected device, while NVML index 0 still denotes NVML device 0. Workers for N != 0 can therefore silently sample the wrong board. NVIDIA also documents that NVML index ordering need not correlate with CUDA ordering: https://docs.nvidia.com/deploy/pdf/NVML_API_Reference_Guide.pdf

Please resolve the selected CUDA device once to a stable UUID or PCI bus ID and obtain the NVML handle from that identity; avoid using one integer for both namespaces. A regression should cover a `CUDA_VISIBLE_DEVICES` remap.

aic-core/src/aiconfigurator_core/sdk/operations/gemm.py
```diff
@@ -971,7 +974,7 @@ def load_gemm_data(gemm_file):
         latency = float(latency)
 
         # NEW: Read power with backward compatibility
-        power = float(row.get("power", 0.0))
+        power = float(row.get("power") or 0.0)
```
tianhaox, line 977: [P1] `or 0.0` makes unavailable power a numeric 0 W anchor *before* `perf_interp`. During the intended staged rollout a table can contain both measured rows and legacy/null rows. Interpolating between, for example, 100 W and a null row converted to 0 W yields a positive but downward-biased power result; because its energy is then positive, the coverage gate treats it as available.

Gradual energy coverage is fine, but missing must remain missing through power interpolation. Please interpolate only from measured-power anchors, or return unavailable when there is insufficient measured support, and convert to the public zero/uncovered representation only at the boundary. The same normalization occurs in the other loaders, so this needs a shared contract-level fix rather than only changing GEMM.

collector/helper.py
```diff
@@ -741,58 +829,94 @@ def log_perf(
...
+                )
+            if include_power_columns is False and existing_power:
+                raise ValueError(
+                    f"Existing CSV contains power columns that this table explicitly omits: {perf_filename}"
+                )
+
+            # A retained power-disabled CSV may be resumed with power enabled.
+            # Upgrade its header and pad existing rows before appending so the
+            # staging file remains rectangular and finalizes cleanly.
+            if requested_power_cols and not existing_power:
```
tianhaox, line 874: [P1] This migrates a retained CSV, but it does not cover the normal finalized-resume case: an existing legacy parquet without `power` columns followed by a new power-enabled CSV. `_merge_perf_rows()` sees different column sets and returns `new_table` via its `overwriting instead of merging` path, silently dropping the historical rows. In a minimal repro, existing rows `s1/s2` plus a new `s3` row finalized to a parquet containing only `s3`.

Please normalize the legacy table by adding nullable `power`/`power_limit` columns before merging, or use a fixed nullable metric schema for staging and parquet. Add a regression that finalizes a power-off run, then resumes/finalizes with power enabled and asserts all old and new rows remain.

tianhaox (changes requested):
The incremental rollout model makes sense: energy does not need to land for every op at once, nullable power is a valid compatibility contract, and collecting data before every consumer is wired is a reasonable staged rollout. I am not asking this PR to complete energy support for all ops.

The three inline findings are narrower correctness issues: sampling the wrong GPU, turning unavailable power into a numeric interpolation anchor, and losing legacy rows while adding the new nullable columns. Those can silently corrupt the data during the rollout and should be fixed before merge.

Scope clarification: `DispatchFlavor::TrtllmAlltoall` still extracts only `.latency_ms` from its communication queries and reconstructs a zero-energy `PerformanceResult`. If that flavor is intentionally outside this rollout, please state/test that it remains unavailable and narrow the end-to-end propagation claim; if it is in scope, carry the full `PerformanceResult` through the branch.

The PR also currently conflicts with `main`, whose loader/data-plane shape has moved substantially, so these contracts should be preserved in the rebased implementation rather than mechanically restoring the stale Python loader paths.

PR #1535, opened 2026-08-13
chore(data): add matched B200 SGLang 0.5.14 power

kaim-eng:
> [!IMPORTANT]
> ## Dependency and zero-sentinel update
>
> This pure-data PR depends on #1590. Merge #1590 first, then rebase/update this PR onto the resulting `main` before merging it.
>
> Latest signed-off data head: `6bc4f243cb5c7bf20a4e8e27d90819a41f186d93`.
>
> The latest commit normalizes every unavailable `power` / `power_limit` cell in the changed power-enabled tables to typed `0.0`. Row coverage, measured values, logical keys, and physical row order are unchanged.

## Summary

Pure-data PR constructed from the then-current `main` data snapshot (`095f58a51c4ca8e61b66ec108d86f223f8d559ce`). This revision rebuilds the B200 SGLang 0.5.14 power overlay from the third-pass campaign pinned to the same backend image used for the on-stock perf-only collection.

- updates exactly **24** B200 SGLang 0.5.14 parquet files in the aggregate PR diff
- retains the complete stock row set and physical row order
- preserves the schema/type/value of every original field
- appends non-null `float64` `power` / `power_limit` only through exact logical-key matches
- retains unmatched stock rows as typed `0.0` / `0.0` sentinel pairs and excludes collection-only keys

No code, documentation, sidecar, or other artifact is included in this PR.

## Exact-image provenance

- on-stock perf-only snapshot: `cb43a2f412983fee709c087882ea3438ab5cb8f5`
- collector code used for power collection: #1534 head `0c6d722ecbbbce39c477adceae2687eb7245d22c`
- image: `lmsysorg/sglang:v0.5.14-cu130@sha256:5027e95bf6ec536856b1b52a91d1f35ff5c564ab83e8a94758a169ff09bb8df3`
- local image (`.sqsh`) SHA-256: `b3ecf5c8fcd521dcca67ceaa4c786d0b73b6adabdc5e08885c092b7c64c4551c`
- SGLang source revision in the image: `49e384ce9d304648e9959666ecb8ce8cd98d0deb`

Broad collection passes used `umb-b200-*` 8x-B200 allocations and parallelized independent single-GPU cases across the available GPUs. Final exact-key retries used isolated 1-GPU allocations where the operation did not require multiple GPUs; the custom-allreduce collection retained its required multi-GPU topology.

The only approved key normalization is the one-to-one HCA model alias `sgl-project/DeepSeek-V4-Pro-FP8` -> `deepseek-ai/DeepSeek-V4-Pro` for the two HCA tables. No performance field is relabeled or changed.

## Coverage for the packaged stock tables

- stock rows retained across the 24 changed tables: **624,463 / 624,463**
- rows with complete measured power pairs: **575,908**
- rows remaining perf-only: **48,555**
- raw row coverage: **92.2245%**
- partial `power` / `power_limit` pairs: **0**

Twenty-one files are byte-identical to the prior PR head. Three tables changed on `main` after the original collection snapshot, so their power was rejoined by the complete logical key while preserving every current-main row, field, latency value, type, and physical row order:

| Table | Stock rows | Exact-key powered rows | Unmeasured `0.0` pairs |
| --- | ---: | ---: | ---: |
| `moe_perf.parquet` | 151,470 | 145,314 | 6,156 |
| `dsa_context_module_perf.parquet` | 95,203 | 55,255 | 39,948 |
| `dsa_generation_module_perf.parquet` | 6,048 | 3,600 | 2,448 |

No power was inferred for new or changed current-main keys that were absent from the collection snapshot.

## Overlay policy

The output is constructed from the stock tables, then power is joined by the complete logical key. Later isolated retries take precedence only for the same exact key. This makes the stock support surface authoritative: collected-only configurations are not inserted, and stock latency or configuration values cannot be overwritten.

## Power-quality checks

- every measured power pair is finite
- every measured pair has `power > 0` and `power_limit > 0`; unavailable pairs are exactly `0.0` / `0.0`
- **905** measurements exceed the recorded 1000 W limit briefly; the maximum is **1.027278x** the limit (**+2.7278%**), within the approved 5% transient allowance
- duplicate stock logical keys have consistent power pairs

## Rebase and validation

- overlay construction base: then-current `main` data snapshot `095f58a51c4ca8e61b66ec108d86f223f8d559ce`; rebase onto the post-#1590 `main` is required before merge
- current signed-off head: `6bc4f243cb5c7bf20a4e8e27d90819a41f186d93`
- history: two signed-off data-only commits
- diff: exactly **24** B200 SGLang 0.5.14 parquet files; no code, documentation, sidecar, or legacy perf artifact
- independent Arrow audit: all **624,463** stock rows, original columns, values, types, latency, and physical order preserved
- exact-key transfer: **575,908** complete measured power pairs, **48,555** typed `0.0` sentinel pairs, **0** null or partial pairs
- strict repository parquet diff: **PASS**; +0 / -0 / ~575,908 rows, 0 legacy perf files

The PR is structurally independent of #1534. Targeted support-matrix validation with #1590 produced 118 `PASS`, 2 `HYBRID_PASS`, and 22 unrelated `FAIL` rows, with zero null-access errors.


tianhaox:
Independent Arrow audit of the 24 files at c8be1754c against the merge-base (2ed278a91, the #1590 merge): every non-power column is value- and order-identical to stock, row counts unchanged, `power`/`power_limit` are typed float64 with 0 nulls, max power/limit ratio 1.0273, no duplicate-key power inconsistencies. Two things need fixing before this can merge.

**1. 7,966 partial pairs (`power=0.0`, `power_limit=1000.0`) in the two topk tables.**

| table | rows | `power>0 & limit>0` | `power==0 & limit==1000` | `0.0/0.0` |
|---|---:|---:|---:|---:|
| `dsv4_csa_topk_calib_perf` | 10,364 | 4,842 | 5,522 | 0 |
| `glm5_topk_module_perf` | 6,664 | 4,220 | 2,444 | 0 |

They are exactly the stock rows with `latency == 0.0` (5,522 + 2,444). The PR contract says unavailable pairs are `0.0 / 0.0`, and the body reports "partial pairs: 0" and 575,908 measured rows; the actual measured count is 567,942 (90.95%), the 7,966 partials were counted as measured. Runtime-safe (both loaders read only `power`, so these rows already resolve as uncovered), but please normalize them to `0.0 / 0.0` and refresh the numbers. Side note: #1590's `parquet_diff` gate only checks type/null/finite/non-negative, so it cannot catch this shape; a pair-consistency check would be a cheap follow-up.

**2. The rebase onto current `main` is not mechanical for `gemm_perf`.**

#1533 (b31d899fb) re-collected the SGLang 0.5.14 fp8_block DeepGEMM rows after moving the UE8M0 weight-scale pack out of the timed loop. Against this PR's stock: 29,526 rows have new latency on `main`, 26,071 faster, median |drift| 32.9%, 18,442 rows above 25%. The power here was collected at 0c6d722 with the old timed loop, so it describes a different kernel mix than the latency it would be joined to. Please apply the same 25% per-row drift gate used in #1536/#1584 (those rows fall to `0.0 / 0.0`), or re-collect GEMM power with the post-#1533 collector.

`moe_perf` is the easy one: `main` added 3,159 nvfp4 rows (#1519), zero latency changes on the 151,470 surviving keys, so a plain key re-overlay gives 145,314 / 154,629 powered and the new rows get sentinels.

PR #1537, opened 2026-08-13
fix(sdk): gpt-oss hybrid-SWA KV memory + decode-share MTP activation scaling

BenHamm:
Fixes two memory-accounting defects that make `recommend`/sweep over-prune deployable configurations, found while validating AIC against recent Dynamo recipes. Full investigation, replication evidence, and validation data: [AIC-1743](https://linear.app/nvidia/issue/AIC-1743/aic-vs-dynamo-recipes-50-configs-2-predictable-gpt-oss-swa-kv-bug-agg).

## 1. gpt-oss KV memory ignores the hybrid SWA layout (~2x overcharge, false OOM)

`GptOssForCausalLM` maps to the MoE family, whose timing path models the alternating banded/global layout (`window_size = 128`, half the layers) while the memory path previously billed all 36 layers at full context.

At ISL 65536 / fp8 KV / TP1 that is 2.2637 GiB/seq modeled vs 1.1341 GiB/seq actual (**1.996x**). Consequently, the shipped Dynamo gpt-oss-120b B200 agg recipe (64 concurrent/GPU, measured 2896 tok/s/GPU on the agentic trace) was rejected as OOM: modeled KV 144.9 GiB vs 72.6 GiB corrected against 180 GiB capacity.

Fix: `MOEModel` now overrides `get_kvcache_bytes_per_sequence` and `get_kvcache_max_tokens` for `GptOssForCausalLM`, using the same hybrid SWA/global layout as the timing path. After the unified-MoE refactor, this one model class covers both regular and large-EP/DeepEP graph regimes. Tests cover short, boundary, long, capacity-inverse, BF16-storage, and TP16 KV-head-partition cases.

## 2. Agg MTP activation scaling multiplies the whole footprint by (nextn+1)

Agg steps mix context and decode tokens (`num_tokens = ctx_tokens + num_gen_requests`), but only decode requests verify `nextn+1` tokens. Scaling the entire activation footprint models `(nextn+1) * (context + decode)` instead of `context + (nextn+1) * decode`. At the GLM-5.2-NVFP4 B200/SGLang best point that turns 25.876 GiB of activations into 103.505 GiB (exactly 4x on a 99.99%-context step), collapsing the spec-on recommendation to concurrency 8 / 58 tok/s/GPU where the shipped recipe sustains concurrency 64 / 176 tok/s/GPU with EAGLE on.

Fix: `mtp_scaled_tokens` identifies the decode share in mixed agg steps. Prefill-only steps pass zero, decode-only steps keep the full multiplier, and a single-request agg run with `OSL > 1` retains the larger of its prefill and later decode-only memory footprints. TensorRT-LLM's `max_num_tokens` budget path deliberately retains legacy behavior pending separate analysis, documented inline.

## Validation

- Production-path regression: `run_agg` at batch size 1 proves `OSL=1` keeps the prefill footprint while `OSL>1` selects the 4x decode activation peak.
- Focused final-head tests: **7 passed**.
- `tests/unit/sdk/backends` + `tests/unit/sdk/models`: **554 passed**.
- Full local unit selection: **3965 passed, 10 skipped** (torch-only test excluded per the repository development setup).
- Ruff lint, Ruff format, and `git diff --check`: clean.
- gpt-oss repro (64k/400/90% agentic point, bs=64, TP1, B200): flips from `RuntimeError: OOM` to fitting at 143.4 GiB total, with corrected KV at 72.580 GiB.
- GLM-5.2-NVFP4 B200/SGLang 16-GPU sweep (ISL 65536 / OSL 400 / prefix 58982, `--nextn 3 --nextn-accepted 1.69`): best agg **57.98 → 317.22 tok/s/GPU** (concurrency 8 → 112); disagg headline unchanged at 342.72.

Known residual, documented in AIC-1743: after removing the pruning error, the agg estimate overshoots the measured 176.42 tok/s/GPU by about 80%, so at least one opposite-sign timing mismatch remains. This PR fixes memory accounting; it does not claim end-to-end estimator accuracy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory estimates for mixed prefill and decode workloads involving speculative token generation.
  * Corrected activation scaling across prefill-only, decode-only, and combined execution steps.
  * Improved GPT-OSS-120B KV-cache capacity estimates, including sliding-window behavior for long sequences.

* **Tests**
  * Added coverage validating activation-memory scaling and GPT-OSS hybrid KV-cache calculations.


Harrilee (approved):

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
```diff
@@ -1803,8 +1812,14 @@ def run_agg(
...
             # in order to indicate what's happening
             num_tokens = num_gen_requests + ctx_tokens
+            # Only the decode requests' tokens verify nextn+1 under speculative
+            # decoding; the context share is processed once (see the MTP
+            # correction in _get_memory_usage).
+            mtp_scaled_tokens = int(num_gen_requests)
         else:
+            # b == 1: context-only step, no decode share to scale.
             num_tokens = ctx_tokens
+            mtp_scaled_tokens = 0
```
jasonqinzhou, line 1823: [P1] Please keep the decode-only activation peak in the `b == 1` memory verdict. This branch has a context-only first step, but when `osl > 1` the code above also creates `num_genonly_steps`, and those later steps require the full `(nextn + 1)` activation scaling. Passing `mtp_scaled_tokens=0` sizes only the prefill step. I revalidated the gap on this head with gpt-oss and `nextn=3`: for `ctx_tokens=256`, this path reports 0.068 GiB of activations while the decode-only step reports 0.273 GiB, so a near-capacity single-request candidate can be accepted as fitting. Please size this case as the maximum of the prefill-only and decode-only footprints (with the decode side conditional on `osl > 1`) and add a `run_agg` regression test for `b=1`.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.9/10 · high confidence**

## What this PR does

This PR corrects two memory-accounting paths: it models gpt-oss's hybrid sliding-window/global KV cache in the unified MoE implementation, and applies MTP activation scaling only to the decode-token share of mixed aggregated steps while deliberately retaining TensorRT-LLM's existing budget behavior.

The latest commit directly addresses the exact sliding-window boundary gap with clear assertions at 128 and 129 tokens, and those focused tests integrate cleanly with the existing model test suite.

## Why this score

6.9/10, REQUEST_CHANGES. The new boundary-only delta is correct, all 544 backend/model tests pass, and lint, formatting, and diff hygiene are clean. However, the batch-size-1 aggregated path still records only its context-step activation footprint even when OSL > 1 schedules later decode-only iterations. A current-head production-path probe again reproduces the 4x MTP activation undercount, so a near-capacity candidate can be admitted despite not fitting at its decode peak.

Arsene12358 (approved):
**Approve — validated end-to-end on head `fc08d6ec`, no regressions across the large-EP rebase.**

Independent verification run against this branch (full evidence chain on [AIC-1743](https://linear.app/nvidia/issue/AIC-1743/aic-vs-dynamo-recipes-50-configs-2-predictable-gpt-oss-swa-kv-bug-agg) / our review artifacts):

- Unit suite `-n 2`: **4101 passed, 50 skipped, 0 failed**.
- gpt-oss false-OOM reproducer (bs=64/TP1/fp8-KV/64k/400/90%): fits with **kvcache exactly 72.580 GiB** (total 143.367/180).
- GLM-5.2-NVFP4 agg spec-decode sweep (16 GPU, EAGLE 3/1.69): **331.133 tok/s/GPU @ concurrency 128** — the fixed value to three decimals; the broken value was ~58.
- Pinned exact-replay harness (41-config Dynamo recipe census, gpt-oss subset + control): verdicts and ratios **byte-identical** to the pre-refactor validated tree — zero semantic drift from the rebase across #1442/#1448/#1464/#1470/#1508.
- Negative control: reverse-applying only this PR's `moe.py` override makes the relocated tests fail with the exact 1.996× over-billing signature (2430664704 vs 1217691648 bytes), then pass on restore — the moved tests kept their teeth.
- Commit provenance: the four base commits match the series attached to AIC-1743 (patch-id drift is purely the rebase); the compat commit's only production change is a ruff reflow with identical operands, and its test relocation correctly adapts to main deleting `SGLangEPMOEModel` (DeepEP now routes through `MOEModel`, so the override still covers that path).

Four additive suggestions, none blocking — **we'll carry all of them in a slim follow-up PR** once this merges, so nothing needs to change here:
1. The deleted `test_mtp_activation_decode_share.py` took the TRT-LLM budget-path no-op regression pin with it (the deliberately-retained multiplier is now unpinned; the underlying budget question is tracked in AIC-1755).
2. There's no call-site reversion guard for `run_agg`'s `mtp_scaled_tokens` threading — the new tests exercise `_get_memory_usage` semantics directly, so a refactor that stops passing the kwarg would keep the whole suite green.
3. Plain-`static` mode and AFD `get_partition_memory_usage` still apply the full `(nextn+1)` multiplier to context-dominated activations (live via `cli estimate`; AFD prefill activations drop 176→44 GiB with the extension at gpt-oss/64k/bs4/nextn3).
4. nextn-independence / KV-dtype-linearity / backend-path pins (guards for the bf16-vs-fp8 misdiagnosis class recorded on AIC-1746).

Coordination note: our #1539 duplicated this series (applied verbatim from the AIC-1743 attachments before this PR appeared) — closing it in favor of this PR; #1540/#1541 are file-disjoint and unaffected.

aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
jasonqinzhou: Addressed in 1b73ec6a and retained on final head f84230b3. For b=1, run_agg now compares the context-only footprint with a decode-only footprint when OSL > 1 and keeps the larger total; the decode call retains the full nextn+1 activation multiplier. The production-path regression covers OSL=1 and OSL=5 and pins the expected 0.068359375 vs 0.2734375 GiB activation footprints. Final validation: 3965 local unit tests passed and all 19 GitHub checks are green.

jasonqinzhou (approved):
Approved on exact head f84230b3 after the b=1 decode-peak fix, BF16/TP16 KV coverage, current-main merge, 3965 passing local unit tests, zero unresolved review threads, and all 19 GitHub checks passing.

PR #1543, opened 2026-08-13
feat(adapter): support Dynamo nightly estimate reporting

simone-chen:
## Summary

- harden the Dynamo adapter against resolved DSV4 and Kimi nightly bundles
- preserve ordered literal concurrency sweeps and shell-comment handling
- detect TRT-LLM speculative settings from engine configuration
- keep uneven concurrency, role-specific memory fractions, and missing acceptance rates fail-closed
- publish checksum-verified platform-wheel manifests and completion markers

## Why

Dynamo-CI needs reproducible AIC estimates after direct K8s nightlies. The adapter must emit estimates only when the resolved recipe maps faithfully and return stable diagnostics otherwise.

This adds the AIC-side contract required by the Dynamo-CI reporting MR. It improves continuous coverage for AIC-1743, but does not itself fix SWA KV accounting, MTP activation, prefix reuse, database coverage, or PASS semantics.

## Validation

- `uv run ruff check` on changed Python and tests
- `uv run pytest tests/unit/sdk/config_adapter tests/unit/tools/test_finalize_wheels_in_artifactory.py tests/integration/test_config_adapter_estimate.py -q`
- 73 tests passed
- workflow YAML and `git diff --check` passed

## Summary by CodeRabbit

- **New Features**
  - Platform wheel releases now include immutable manifests with SHA-256 and size metadata, plus completion markers for verification before installation.
  - Dynamo performance configuration supports literal concurrency sweeps, improved comment handling, and speculative decoding detection from mounted engine configurations.
- **Bug Fixes**
  - Improved validation for unsupported concurrency, unresolved engine configurations, and conflicting settings.
  - Strengthened wheel publication checks to ensure complete, verifiable releases.
- **Documentation**
  - Clarified installation, wheel verification, concurrency expansion, and speculative decoding guidance.
  - Documented that copied pull-request workflows validate wheels without publishing artifacts.

### DIFF
--- a/.github/workflows/build-platform-wheels.yml
+++ b/.github/workflows/build-platform-wheels.yml
@@ -8,11 +8,10 @@ on:
     branches:
       - main
       - "release/*"
-  workflow_call:
 
 concurrency:
   group: platform-wheels-${{ github.ref }}
-  cancel-in-progress: ${{ startsWith(github.ref, 'refs/heads/pull-request/') }}
+  cancel-in-progress: false
 
 permissions:
   contents: read
@@ -75,21 +74,6 @@ jobs:
           platform: ${{ matrix.platform }}
           wheel_base: ${{ matrix.wheel_base }}
 
-      - name: Report disabled copied-PR Artifactory staging
-        if: >-
-          matrix.upload_upper &&
-          startsWith(github.ref, 'refs/heads/pull-request/') &&
-          vars.PLATFORM_WHEEL_STAGING_ENABLED != 'true'
-        shell: bash
-        run: |
-          echo "::notice title=Copied-PR Artifactory staging disabled::Set PLATFORM_WHEEL_STAGING_ENABLED=true when staging is ready."
-          {
-            echo "### Copied pull-request Artifactory staging"
-            echo
-            echo '- Status: skipped because `PLATFORM_WHEEL_STAGING_ENABLED` is not `true`.'
-            echo '- Follow-up: [#1432](https://github.com/ai-dynamo/aiconfigurator/issues/1432).'
-          } >> "${GITHUB_STEP_SUMMARY}"
-
       - name: Report disabled post-merge Artifactory staging
         if: >-
           matrix.upload_upper &&
@@ -104,31 +88,6 @@ jobs:
             echo '- Status: skipped because `PLATFORM_WHEEL_STAGING_ENABLED` is not `true`.'
           } >> "${GITHUB_STEP_SUMMARY}"
 
-      - name: Resolve copied pull-request staging path
-        id: pr-staging
-        if: startsWith(github.ref, 'refs/heads/pull-request/')
-        shell: bash
-        run: |
-          pr_number="${GITHUB_REF_NAME#pull-request/}"
-          if [[ ! "${pr_number}" =~ ^[0-9]+$ ]]; then
-            echo "::error::Expected a copy-pr-bot branch named pull-request/<number>, got ${GITHUB_REF_NAME}"
-            exit 1
-          fi
-          echo "subpath=PR/${pr_number}/${GITHUB_SHA}/${GITHUB_RUN_ID}/${GITHUB_RUN_ATTEMPT}" >> "${GITHUB_OUTPUT}"
-
-      - name: Stage copied pull-request wheels in Artifactory
-        if: >-
-          vars.PLATFORM_WHEEL_STAGING_ENABLED == 'true' &&
-          startsWith(github.ref, 'refs/heads/pull-request/')
-        env:
-          ARTIFACTORY_URL: ${{ secrets.ARTIFACTORY_URL }}
-          ARTIFACTORY_TOKEN: ${{ secrets.ARTIFACTORY_PR_TOKEN }}
-          ARTIFACTORY_PYPI_REPO_NAME: ${{ secrets.ARTIFACTORY_PYPI_REPO_NAME }}
-          ARTIFACTORY_SUBPATH: ${{ steps.pr-staging.outputs.subpath }}
-          UPLOAD_UPPER: ${{ matrix.upload_upper }}
-        shell: bash
-        run: bash tools/stage_wheels_in_artifactory.sh
-
       - name: Stage post-merge wheels in Artifactory
         if: >-
           vars.PLATFORM_WHEEL_STAGING_ENABLED == 'true' &&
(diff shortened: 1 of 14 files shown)


.github/workflows/build-platform-wheels.yml
```diff
@@ -78,15 +78,14 @@ jobs:
...
+          startsWith(github.ref, 'refs/heads/pull-request/')
         shell: bash
         run: |
-          echo "::notice title=Copied-PR Artifactory staging disabled::Set PLATFORM_WHEEL_STAGING_ENABLED=true when staging is ready."
+          echo "::notice title=Copied-PR Artifactory staging disabled::Copied PRs validate wheels but never receive Artifactory credentials or publish artifacts."
           {
             echo "### Copied pull-request Artifactory staging"
             echo
-            echo '- Status: skipped because `PLATFORM_WHEEL_STAGING_ENABLED` is not `true`.'
+            echo '- Status: disabled. Copied PRs validate wheels but do not publish artifacts.'
```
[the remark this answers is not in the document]
simone-chen: Fixed in ffb41b30: copied-PR pushes now call the reusable validate-platform-wheels workflow, whose jobs run only on GitHub-hosted runners. The privileged build-platform-wheels workflow is now post-merge only, with copied-PR handling removed. I did not change the repository ruleset from this code PR; the existing ruleset already requires code-owner review, while making the separate codeowners status check required is an administrator policy change.

src/aiconfigurator/sdk/config_adapter/dynamo.py
```diff
@@ -731,19 +767,40 @@ def _request_for_point(
...
 
     speculative_flag_names = {"num-speculative-tokens", "speculative-num-steps", "speculative-token-num"}
-    speculative_values = [
-        _flag(service, "num-speculative-tokens", "speculative-num-steps", "speculative-token-num")
-        for service in services
-    ]
+    speculative_values: list[Any] = []
+    speculative_configs: list[Mapping[str, Any]] = []
+    for service in services:
+        config = service.engine_config.get("speculative_config")
```
jasonqinzhou, line 813: **[P1] Validate that the engine ConfigMap is actually mounted.** `service.engine_config` is currently populated from any ConfigMap-backed volume declared in `extraPodSpec.volumes`; `_mounted_config` does not correlate that volume with `mainContainer.volumeMounts` or the mount path used by `--extra-engine-args`. I reproduced this by deleting `volumeMounts` from the new Kimi fixture: adaptation still returned `adapted` with `model.nextn == 3`, even though `/config/agg.yaml` would not exist in the worker. Since this block now uses that config to change the estimate, please make config discovery verify the matching volume mount/path and add a regression case for a declared-but-unmounted ConfigMap. This needs coordinated parser and test changes, so a one-click suggestion is not safe.

src/aiconfigurator/sdk/config_adapter/dynamo.py
```diff
@@ -731,19 +767,40 @@ def _request_for_point(
...
+        speculative_config = config if isinstance(config, Mapping) else {}
+        if speculative_config:
+            speculative_configs.append(speculative_config)
+        speculative_values.append(
```
jasonqinzhou, line 815: **[P1] Reject conflicting speculative depths across services.** The loop coalesces command and ConfigMap values only within each service, but the resulting `speculative_values` list is later reduced with `next(...)`. A prefill ConfigMap with `num_nextn_predict_layers: 2` and a decode ConfigMap with `num_nextn_predict_layers: 3` therefore returns an `adapted` request with `nextn == 2` instead of failing closed. Please coalesce the active per-service depths after this loop (the Dynamo-CI adapter already rejects differing role depths) and cover the conflict explicitly.

src/aiconfigurator/sdk/config_adapter/dynamo.py
```diff
@@ -731,19 +767,40 @@ def _request_for_point(
...
         value
         for service in services
         for key, value in service.flags.items()
         if "specul" in key and key not in speculative_flag_names
-    ]
+    ] + speculative_configs
     speculation_disabled = {"", "0", "false", "none", "disabled"}
 
     def speculation_enabled(value: Any) -> bool:
+        if isinstance(value, Mapping):
```
jasonqinzhou: **[P2] Keep zero-depth speculative configs disabled.** Returning `bool(value)` for every non-empty ConfigMap means `speculative_config: {num_nextn_predict_layers: 0}` is considered active. I reproduced this as a rejection requiring `nextn_accepted`, while the existing command-flag path correctly treats depth `0` as disabled. Determine activation from the recognized algorithm/depth fields using the same disabled sentinels, and add zero-depth ConfigMap coverage.

jasonqinzhou (changes requested):

## What this PR does

This PR extends the Dynamo adapter for resolved nightly DSV4 and Kimi configurations, including literal concurrency sweeps, shell comments, and TRT-LLM speculative settings. It also isolates copied-PR wheel validation and adds checksummed Artifactory manifests for post-merge platform wheels.

The ordered sweep fixtures, explicit fail-closed diagnostics, and separation of copied-PR validation onto GitHub-hosted runners show thoughtful attention to reproducibility and runner safety.

## Why this score

The score is 5.5/10 with REQUEST_CHANGES because the new mounted-ConfigMap speculation path still has two silent fidelity failures: it can use a ConfigMap that is not mounted, and it can select the first of conflicting prefill/decode depths. It also treats a valid zero-depth ConfigMap as active. The focused suite and CI are green, but these adversarial cases violate the adapter's fail-closed contract and are not covered by the new tests.

src/aiconfigurator/sdk/config_adapter/dynamo.py
simone-chen: Fixed in f857cb1b. `_mounted_config` now correlates each ConfigMap volume with `mainContainer.volumeMounts` and resolves the exact `--extra-engine-args` path, including projected `items` and `subPath`. An unresolved or unmounted path is rejected. The regression test removes `volumeMounts` from the resolved Kimi fixture and verifies the fail-closed diagnostic.

src/aiconfigurator/sdk/config_adapter/dynamo.py
simone-chen: Fixed in f857cb1b. Active per-service depths are normalized to positive integers and compared after service parsing; differing values now reject with a stable `worker services use different speculative depths` diagnostic. Added explicit prefill=2/decode=3 ConfigMap coverage.

src/aiconfigurator/sdk/config_adapter/dynamo.py
simone-chen: Fixed in f857cb1b. ConfigMap activation now examines recognized depth/algorithm fields instead of treating any non-empty mapping as active. The same disabled sentinels used for command flags apply, and a ConfigMap depth of zero is covered as adapted with `nextn == 0` and no acceptance override.

jasonqinzhou (approved):
Re-reviewed exact head 5e0586ee. The mounted ConfigMap, conflicting-depth, zero-depth, and override-only acceptance paths are fail-closed and covered. The focused local suite passes (77 tests), all GitHub checks are green, all review threads are resolved, and the changes remain limited to the Dynamo adapter, tests/docs, and wheel workflow plumbing. Approved.

PR #1544, opened 2026-08-14
fix(sdk): decode-share MTP scaling for static/AFD paths + reversion guards (follow-up to #1537)

Arsene12358:
## What

Drops the MTP `(nextn+1)` draft multiplier from the two remaining prefill-sized activation footprints, and pins the whole decode-share contract with call-site guard tests. Linear: AIC-1746.

**This is a rework of the original branch onto current main.** The engine migration (PR-5/PR-6/PR-7) landed while this PR was in flight and absorbed most of the original production change; the branch was rebuilt from scratch on main so the diff now carries only what main still lacks. The prior approval predates the rework and should be considered stale.

## Production change (2 sites, both declare `mtp_scaled_tokens=0`)

1. **`run_static`'s plain `"static"` mode** (`base_backend.py`): the `static_ctx` branch already declares `mtp_scaled_tokens=0` on main, but the `"static"` else-branch — which sizes activations from the same derived context token count `((isl - prefix) * batch_size)` — did not, so with `nextn > 0` it still applied the full `(nextn+1)` multiplier to a footprint that contains no draft tokens.
2. **`get_partition_memory_usage` with `num_tokens=0`** — the AFD prefill-phase sentinel (`InferenceSession._estimate_{a,f}_memory_dict` passes `num_tokens=0` for `phase == "prefill"`, `num_tokens=batch_size` for decode). On the measured shape (gpt-oss-120b, ISL 65,536, batch 4, TP1, nextn 3) the multiplier turned a 44.0 GiB context footprint into 176.0 GiB — a 4x over-count that shrinks the A/F worker KV budget and over-prunes AFD batch sizes.

Decode-shaped callers are untouched and keep the full multiplier (every token they process is verified): `static_gen`, AFD decode (`num_tokens=batch_size`), and `run_agg`'s decode share (already on main). TRT-LLM's agg budget path also intentionally keeps the legacy multiplier (`max_num_tokens` is a per-iteration budget; whether the correction applies there needs its own analysis — tracked in AIC-1755; the only non-test edit outside `base_backend.py` is that comment line).

## Tests (re-sited onto the engine-era surfaces)

The compiled engine is now the only static/agg engine-step executor, so the guards stub the latency side exactly the way the neighboring tests do (`estimate_static_latency_breakdown_with_rust` bridge / `run_mixed`), while memory sizing — the surface under test — remains a Python `_get_memory_usage` call:

- `test_run_static_declares_mtp_decode_share_per_mode`: `static`/`static_ctx` → `mtp_scaled_tokens=0`, `static_gen` → full multiplier retained.
- `test_run_agg_declares_mtp_decode_share_at_call_site`: the `b > 1` mixed step passes the `num_gen_requests` decode share (`b == 1` is already pinned on main by `test_run_agg_b1_uses_scheduled_activation_peak`).
- `test_trtllm_budget_path_ignores_the_decode_share`: pins the intentional drop on the TRT-LLM budget path.
- `TestAFDPartitionActivationScaling`: the AFD `num_tokens` contract end-to-end through `get_partition_memory_usage` (prefill 44.0 GiB pin + nextn-invariance; decode keeps the 4x).
- `TestGptOssHybridKVCache` additions: the window-capped KV curve is nextn-independent, stays window-aware through `_get_memory_usage` at the recipe operating point (batch 48, ISL 65,536 → 54.4 GiB, not the ~2x linear value that false-OOMs the shipped 8xB200 agg recipe), and scales linearly with KV-cache dtype.

Red/green verified: on unpatched main, `test_prefill_partition_does_not_scale` fails with exactly the 176.0-vs-44.0 over-count and `test_run_static_declares_mtp_decode_share_per_mode[static-0]` fails with `None != 0`; with the fix, both pass.

## What the original PR carried that is now on main (dropped from this diff)

The `static_ctx` branch fix, `run_agg`'s decode-share call site and `_memory_usage_kwargs_for_agg` helper, the TRT-LLM override, `TestMTPActivationMemoryScaling`, and `test_run_agg_b1_uses_scheduled_activation_peak`.

## Verification

- Full unit suite: 3722 passed, 10 skipped (`-n 2`).
- `tests/cross_package` + Rust parity suite: green (memory sizing is outside the engine-step latency goldens; no re-pin needed).
- `ruff check` / `ruff format --check`: clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved memory estimates for workloads using speculative decoding.
  - Prefill calculations now avoid applying decode-only token scaling.
  - Decode calculations continue to account for speculative tokens correctly.
  - Corrected window-aware KV-cache sizing, including BF16 and FP8 precision differences.
  - Improved multimodal token calculations by aligning image dimensions with processing strides.
  - Improved aggregate result validation and cache consistency.

- **Tests**
  - Added coverage for activation scaling, aggregation memory paths, speculative decoding, and KV-cache limits.

### DIFF
--- a/aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
+++ b/aic-core/src/aiconfigurator_core/sdk/backends/base_backend.py
@@ -664,6 +664,12 @@ class BaseBackend:
                 **memory_extra,
             )
         else:
+            # "static": activations default to the context token count
+            # ((isl - prefix) * batch_size), i.e. the prefill peak — the decode
+            # steps of a static batch only ever process
+            # batch_size * beam_width * (nextn + 1) tokens, far fewer. So this
+            # footprint has no decode share either (mtp_scaled_tokens=0), the
+            # same rule the static_ctx branch above already applies.
             memory = self._get_memory_usage(
                 model,
                 database,
@@ -673,6 +679,7 @@ class BaseBackend:
                 osl,
                 prefix=prefix,
                 encoder_memory=encoder_memory,
+                mtp_scaled_tokens=0,
                 **memory_extra,
             )
 
@@ -897,8 +904,16 @@ class BaseBackend:
             "num_tokens": num_tokens,
             "prefix": prefix,
         }
-        if "max_seq_len" in inspect.signature(self._get_memory_usage).parameters:
+        memory_usage_params = inspect.signature(self._get_memory_usage).parameters
+        if "max_seq_len" in memory_usage_params:
             kwargs["max_seq_len"] = max_seq_len
+        # ``num_tokens == 0`` (AFD's ``phase == "prefill"`` callers) makes
+        # ``_get_memory_usage`` derive the count from ``(isl - prefix) * batch_size``
+        # — a prefill-only footprint, so it carries no decode share that
+        # verifies nextn+1 draft tokens. Decode callers pass ``num_tokens > 0``
+        # (one token per request) and keep the full ``(nextn+1)`` multiplier.
+        if num_tokens == 0 and "mtp_scaled_tokens" in memory_usage_params:
+            kwargs["mtp_scaled_tokens"] = 0
 
         memory = self._get_memory_usage(
             model,
--- a/aic-core/src/aiconfigurator_core/sdk/backends/trtllm_backend.py
+++ b/aic-core/src/aiconfigurator_core/sdk/backends/trtllm_backend.py
@@ -132,6 +132,7 @@ class TRTLLMBackend(BaseBackend):
         # whether the decode-share correction applies to a budget-based footprint
         # needs its own analysis (compare the AIC-1110 suppression on the
         # KV-capacity path, which treats the budget as already draft-inclusive).
+        # Tracked in AIC-1755.
         return {
             "num_tokens": agg_extra["max_num_tokens"],
             "max_seq_len": agg_extra["max_seq_len"],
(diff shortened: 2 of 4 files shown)


Harrilee (approved):

Arsene12358:
Converting to draft: this PR's green checks are stale (they predate #1552/#1555). Verified locally against current main (merge-ref equivalent): 5 of its tests fail with `TypeError: the compiled engine is the only aggregate engine-step executor` — the run_agg call-site guard tests monkeypatch the python aggregate step that the migration retired behind the engine. The PR needs the same engine-era rework #1540 got (re-audit which of its production changes survive on the new architecture, re-site the guards) before it is mergeable. Will follow up.

Arsene12358:
Reworked onto current main and un-drafted. The engine migration (#1552/#1555/#1566) absorbed most of the original branch (the static_ctx fix, run_agg's decode-share call site, the TRT-LLM override, and two of the test groups), so the branch was rebuilt from scratch: the diff now carries only the two production sites main still lacks (the plain `static` mode and the AFD `num_tokens==0` partition path) plus the guard tests re-sited onto the engine-era surfaces (latency stubbed at the engine bridge / `run_mixed`, matching the neighboring tests; the Python `_get_memory_usage` call remains the surface under test). Red/green verified against unpatched main: the AFD prefill pin fails with exactly the 176.0-vs-44.0 GiB over-count and the `static` mode guard with `None != 0`. Full unit (3722), cross-package + Rust parity (535) suites green locally; memory sizing is outside the engine-step goldens, so no re-pin.

@Harrilee your earlier approval predates this rework — please re-review when you get a chance.

Arsene12358:
Proceeding to merge on the standing approval: the rework's CI is fully green, CodeRabbit's pass over the rebuilt branch left no actionable threads, and the residual production change was red/green-verified against unpatched main (the 176.0-vs-44.0 GiB AFD pin and the static-mode share guard). @Harrilee @tianhaox happy to address anything post-merge.

Issue #1462, opened 2026-08-02
follow-up: vLLM Kimi-K3 W4A8 MegaMoE (deep_gemm) data lane — unblocked on post-merge vllm main


tianhaox:
## Context

Kimi-K3's checkpoint (compressed-tensors mxfp4-pack + SiTU) serves two vLLM MoE paths:

1. **Single-node / TP**: MarlinExperts weight-only **W4A16** (CUTLASS W4A4's activation whitelist has no SITU) — collected and shipped in #1435 on all eight systems, kernel lane `w4a16_mxfp4`.
2. **Recommended DEP scale-out** (`--moe-backend deep_gemm_mega_moe`): DeepGEMM MegaMoE, fp8 activations × fp4 weights = **W4A8** semantics. This path was **unrunnable** on the pinned preview image (`vllm/vllm-openai@sha256:e90e2603`, 0.1.dev19262 — no `deep_gemm` package), so #1435 carries no vLLM W4A8 rows; W4A8 scope is covered by the sglang trtllm-gen lane only.

## What changed upstream (2026-08-02 research)

- K3 merged to vllm **main** Jul 29–30 (vllm-project/vllm#50089 + vllm-project/vllm#50000); no tagged release contains it yet (v0.26.1rc0 was cut pre-merge; tracking issue vllm-project/vllm#50001).
- **The blocker is gone on main**: DeepGEMM is vendored into the official image (`vllm.third_party.deep_gemm`, vllm-project/vllm#41516) and the K3 merge pinned a SITU-capable vllm-project/DeepGEMM fork (`fp8_fp4_mega_moe` kernels; `ep_gather` fix vllm-project/vllm#50458). `deep_gemm_mega_moe` should run on `nightly-*` images (Jul 31 onward).
- CUTLASS W4A4 SITU whitelist is still not extended → the single-node W4A16 lane label stays serving-true; nothing to re-collect there.
- New on main: FlashInfer trtllm-gen mxfp4×mxfp8 SITU experts exist behind a FlashInfer capability check, but whether the compressed-tensors scheme can route into them is undetermined.

## Follow-up work (the actual scope — large, hence this issue)

- [ ] Pick and pin a post-merge image (nightly or the first release containing K3). A pin bump triggers the standing **full kda dispatch re-audit** (merged-main routing may differ from the preview branch: FlashKDA/fused_kda_decode availability per SM, backend selector changes).
[...]

PR #1553, opened 2026-08-15
feat(megamoe): model + collect Kimi-K3 MegaMoE on vLLM (GB300, 256 rows)

Ma1oneZhang:
Hi from Novita AI

## What

Opens the vLLM lane of the fused DeepGEMM MegaMoE module and ships the first measured Kimi-K3 dataset for it (gb300, vllm 0.27.0, 256 rows). Ref #1462 (not closed by this PR).

## Modeling (SDK)

- kimi_k3/deepseek_v4: `moe_backend=megamoe` now allowed on vLLM; K3 builds the fused `DeepSeekV4MegaMoEModule` in latent space (3584/3072, topk 16, 896 experts) instead of the decomposed MoEDispatch+MoE pair. `pre_dispatch` is per-backend (`vllm` vs `sglang_jit`); gate/top-k/latent projections/shared experts stay outside the measured boundary.
- Quant remap is moe_backend-aware: vLLM's plain K3 lane keeps `w4a16` (Marlin), vLLM+MegaMoE gets `w4a8_mxfp4_mxfp8` on Blackwell.
- CLI: `--moe-backend` on default/exp/estimate; SOL sidecar degrades to measured-only columns when a measured-only lane is in the graph; `--backend auto` enumerates vLLM only for models with measured vLLM MegaMoE rows where they resolve (today: Kimi-K3 on gb300).

## Rust engine

`Dsv4MegaMoeTable` routes its single primary through `resolve_op_sources`, honoring the family-first layout (`<system>/megamoe/<backend>/<version>`). Previously the compiled engine hard-pointed at the legacy dir and every vLLM MegaMoE query came back empty while Python answered the same points. Rust unit tests + a run_agg E2E regression pin the resolution.

## Collector (serving truth @ vllm 0.27.0, verified in-container on GB300/SM103)

- `collector/vllm/collect_k3_megamoe.py`: thin wrapper on the shared DSv4 MegaMoE harness forcing `--model-config kimi_k3 --pre-dispatch vllm --framework VLLM`.
- vLLM's own triton `prepare_megamoe_inputs` is the timed pre-dispatch (NOT sglang's deep_gemm JIT copy); DeepGEMM reached via serving's vendored `vllm.third_party.deep_gemm`; SiTU passed as `activation="situ"` with `activation_clamp=None`.
- Harness memory rewrite (the EP=32×32768 cgroup OOM kills): `balanced_logits` one-hot cube → scatter_add_ (bit-identical at these strides); `build_routing_plan` full-perm+slice copy → same RNG stream, per-rank gather only. Peak host RSS 177GiB → 11.1GiB/worker. Legacy-oracle equivalence test included.

## Data

New `megamoe` family in the op catalog; the 3 pre-existing MegaMoE tables move to `<system>/megamoe/`, this commit ships the 4th (gb300 vllm 0.27.0): 256 rows, EP 4/8/16/32 × (24 ctx + 40 gen) × 4 routing distributions, `w4a8_mxfp4_mxfp8/fp8_fp4`, `validate_perf` PASS per EP and merged. Manifest megamoe family pinned to vllm 0.27.0 (amd64+arm64 digests).

## Verification

- 3990 python unit tests + 461 rust tests; ruff clean; `backend_facts --check` + `check_collector_data` R1-R6 green.
- E2E: `cli default moonshotai/Kimi-K3 gb300 vllm 16gpu nextn7/3 megamoe` → top1 agg tp1dp16ep16 bs=7 = 233.01 tokens/s/gpu; per-op sources show the fused module silicon-sourced throughout.
- Compiled-engine path hard-asserted in `test_kimi_k3_megamoe_vllm_agg_answers_via_compiled_engine`.

## Known boundaries

- Table EP axis tops at 32 — queries beyond loud-fail (`PerfDataNotAvailableError`) and are excluded from sweeps.
- K3 + AFD remains unsupported upstream (AFD partitioner can't classify `kda_*`; same as plain K3, not new).
- `--moe-backend deepep_moe` remains deprecated/inert per the large-EP rework (#1442/#1448).

## CodeRabbit review round

Fixed in-place (amended): DSv4 stays SGLang-only at both the model and Task gate (no vLLM DSv4 rows exist); Kimi-K3 identity is architecture-keyed, so local/offline checkpoint mounts qualify equally with the hub id (quota remap + Task allowlist + auto-sweep gate all share it); MegaMoE tail mask hoisted out of the timed region for both pre-dispatch lanes (**note: shipped rows predate this hoist and carry a small pessimistic tail-mask overhead at small decode shapes — refresh on re-collection**); vllm lane hard-requires --version (no "unknown" rows); `--framework` choices pinned; the k3 wrapper validates forced options across both `--flag v`/`--flag=v` forms incl. duplicates/missing values; the harness import fallback narrows to `ModuleNotFoundError.name == "collector"`; disagg tasks gate vllm+megamoe on decode-side MegaMoE rows; Task backend validation now checks every role.

Intentionally not changed: AFD pinned executors with `tp_a>1` already fail loudly at model construction ("K3 MegaMoE requires moe_tp_size=1"); K3+AFD remains unsupported upstream, so no silent-wrong-data window exists there.

## Summary by CodeRabbit

* **New Features**
  * Added MegaMoE support for Kimi-K3 with both SGLang and vLLM.
  * Added configurable MoE backend selection to estimation workflows and CLI options.
  * Added vLLM MegaMoE performance coverage, validation, collection, and metadata.
  * Added backend-specific handling for activation, routing, dispatch, and quantization.
* **Bug Fixes**
  * Reduced memory usage when calculating balanced expert counts.
  * Improved local Kimi-K3 checkpoint detection and routing-plan memory efficiency.
* **Documentation**
  * Expanded MegaMoE backend availability and model coverage guidance.


Ma1oneZhang:
CodeRabbit review round — fixes landed in amended HEAD (45345f1):

Fixed:
- DSv4 admission reverted to SGLang-only at both the model and Task gate (no measured vLLM DSv4 MegaMoE rows exist; construction-then-query-fail was worse than failing fast)
- Kimi-K3 identity is now architecture-keyed (`_is_kimi_k3_checkpoint`), so local/offline checkpoint mounts get the remap, Task allowlist, and auto-sweep gate the same way the hub id does; helpers-level regression covers a real on-disk config
- MegaMoE tail mask hoisted out of the timed region for both pre-dispatch lanes; shipped rows predate the hoist and carry a small pessimistic overhead at small decode shapes — will refresh on re-collection
- vllm lane now hard-requires `--version` (no `unknown` provenance)
- `--framework` constrained to {sglang, vllm} in finalize
- k3 wrapper: forced options validated across both `--flag v`/`--flag=v` forms, duplicates and missing values raise; import fallback narrowed to `ModuleNotFoundError.name == 'collector'`
- disagg vllm+megamoe tasks now require decode-side MegaMoE rows
- Task mega validation checks every disagg role, not only the first

Not changed (with reasons):
- AFD pinned executor + megamoe with `tp_a>1` already rejects loudly at model construction ("requires moe_tp_size=1"); K3+AFD is unsupported upstream of this PR, so no silent window exists — proper AFD EP-only A-worker construction is follow-up.

Ma1oneZhang:
On the two provenance findings:

**1. collector_hash ghost state — confirmed, and not recoverable as-is.** I verified f32c9ebe against every committed state of this branch (all pre-rebase/amend iterations through to tip): none reproduces it. The 2026-08-13 collection ran from working-tree bytes during the EP32-OOM fix + vllm-lane split churn; the exact bytes never landed in git, so I cannot attribute the ghost to a specific file. Rather than papering over it:
- the shipped closure now records `collector_hash: sha256:8995d04c…` recomputed over the **final shipped closure bytes** (content-addressed, rebase-stable);
- `collection_meta.yaml` carries an honesty note: the known behavior-affecting delta between collection-time and shipped closures is the tail-mask hoist out of the timed region (pessimistic direction on shipped rows, mostly visible at small decode shapes); remaining deltas are arg-gating/docs only;
- rows and hash will be refreshed together on the next re-collection.

**2. Wrapper in hash closure — fixed in 064aedf.** `collector/vllm/collect_k3_megamoe.py` is now in the dsv4_megamoe closure (its argv forcing bends row identity, agreed — the drift was invisible before). One precision note: `validate_perf.py` was actually already listed in the closure (line 56 of hash_closures.yaml); the newly-added member is only the vllm wrapper.

aic-core/src/aiconfigurator_core/systems/data/gb300/megamoe/vllm/0.27.0/collection_meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    # the vLLM lane wrapper collector/vllm/collect_k3_megamoe.py). Honesty note:
+    # the 2026-08-13 collection ran from mid-development working-tree bytes and
+    # its recorded hash (sha256:f32c9ebe...) is not reproducible from any git
+    # state; the behavior-affecting deltas vs the collected closure are (a) the
+    # tail-mask hoist out of the timed region (rows carry a small PESSIMISTIC
+    # overhead, especially at small decode shapes — safe direction) and (b)
+    # non-timing changes only (lane arg gating, doc/comments). Both are covered
+    # in this PR's review thread. Refresh the rows+hash together on the next
+    # re-collection.
+    collector_hash: sha256:8995d04c90def9bf9e5cd511706b1bc5afa5afa4a226275f2ae5fa9fabd227fc
```
jasonqinzhou: [P1] This `collector_hash` still identifies the closure as shipped, not the working-tree bytes that produced these 256 rows. Lines 33-40 explicitly say the collection-time hash is unreproducible and that the shipped closure moves the tail-mask work out of the timed region; recomputing the current closure yields the declared `sha256:8995d04c...`, confirming the mismatch while the table remains `status: complete`. Please recollect with the current closure and update the rows, ref, and hash together, or use an explicitly approved provenance state that records the actual collection bytes before treating this table as measured support.

src/aiconfigurator/cli/main.py
```diff
@@ -1584,7 +1628,20 @@ def build_default_tasks(
 
     backends_to_sweep = [b.value for b in common.BackendName] if backend == "auto" else [backend]
     if backend == "auto" and moe_backend == "megamoe":
+        # SGLang carries measured MegaMoE data on every supported Blackwell
```
jasonqinzhou: [P1] The auto path still seeds SGLang unconditionally. On this head, the default Kimi-K3/GB300 build returns `agg_sglang@0.5.16` alongside `agg_vllm@0.27.0`, but executing the generated SGLang configuration fails with `PerfDataNotAvailableError` because no Kimi MegaMoE table resolves for GB300/SGLang 0.5.16. Please apply a model/system/version-aware data probe to SGLang too (or omit Kimi/GB300 SGLang until matching rows ship), and add a regression that runs the generated task far enough to exercise the measured query rather than asserting construction only.

aic-core/rust/aiconfigurator-core/src/perf_database/mod.rs
```diff
@@ -473,8 +473,7 @@ impl PerfDatabase {
                 .next()
                 .map(|PerfSource(path, _)| path)
                 .unwrap_or_else(|| data_root.join("dsv4_megamoe_module_perf.parquet")),
-            ),
-            mhc: MhcTable::with_sources(data_root.clone(), perf_db_sources),
+            ),            mhc: MhcTable::with_sources(data_root.clone(), perf_db_sources),
```
jasonqinzhou: [P2] The rebase joined `mhc` onto the closing line of the preceding field, and a scoped rustfmt check proposes splitting it. Please keep this changed hunk formatted:

```suggestion
            ),
            mhc: MhcTable::with_sources(data_root.clone(), perf_db_sources),
```

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.2/10 · high confidence**

## What this PR does

This PR adds Kimi-K3 fused MegaMoE modeling and 256 measured GB300/vLLM 0.27.0 rows, moves MegaMoE data into family-specific paths, threads backend selection through CLI and task construction, and adds collector, provenance, documentation, and focused model coverage. The updated head rebases that work onto the compiled Rust single-oracle data plane.

The rebase preserves the feature across a substantial engine migration: the new end-to-end test now proves the Kimi-K3 vLLM MegaMoE path reaches measured silicon rows through the compiled engine, while the model, task, collector, and data-integrity suites remain well targeted and the exact-head CI rollup is fully green.

## Why this score

6.2/10, REQUEST_CHANGES. The rebased implementation compiles, the affected Python and data checks pass, and the vLLM compiled-engine path is working. Both prior blocking correctness findings remain unchanged at this head: the published GB300 table is still marked complete even though its collection-time harness cannot be reproduced and differs in a timing-affecting detail, and the default Kimi-K3/GB300 auto sweep still creates a SGLang experiment that deterministically fails because matching MegaMoE rows are unavailable. The rebase also introduced one non-blocking rustfmt issue in its only PR-local Rust diff line.

src/aiconfigurator/cli/main.py
Ma1oneZhang: Fixed on abbe8b6. MegaMoE availability now probes the exact model shape, backend pre-dispatch lane, moe_tp=1, moe_ep>1, and both context plus generation phases instead of treating a family file as model support. The generated Kimi-K3 GB300 auto task is vLLM-only and a constrained task.run reaches measured silicon rows.

aic-core/rust/aiconfigurator-core/src/perf_database/mod.rs
Ma1oneZhang: The rebased head keeps this Rust hunk formatted correctly. Rebase validation and the scoped Rust checks pass; no joined mhc field remains.

aic-core/src/aiconfigurator_core/systems/data/gb300/megamoe/vllm/0.27.0/collection_meta.yaml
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in 11923e6 by full recollection, not by relabeling the old table. All 256 rows now come from committed collector ref abbe8b6 with collector hash sha256:72800f900313fe9f50f4885390f974a62644027daa89f539593c1c6125d59781 and case-plan hash sha256:4ae27f93d317ac0f79562c627818dcf214221edf2508fbddc0197f803d4a3531. EP4, EP8, EP16, and EP32 each exited zero and passed validation; the merged table has 256 unique loader keys and status complete.

Ma1oneZhang: Recollected and replaced in 11923e6. The new parquet contains no rows from the unreproducible August 13 working tree: every row was measured from the committed abbe8b6 closure, whose declared hash recomputes exactly. The sidecar records that ref, hash, unchanged 832-case plan hash, 2026-08-21 date, 256 rows, and complete status. EP32 additionally passed the 32-rank IB/MNNVL/GDR transport smoke and full 64-row validator.

src/aiconfigurator/cli/main.py
```diff
@@ -1479,6 +1495,137 @@ def _resolve_version_for_matching(system_name: str, backend_name: str, backend_v
...
+    explicit pin wins), then ``current``, then ``next``. Returns the version
+    the lane must run at, or None when no measured rows resolve anywhere.
+
+    Lanes whose rows would arrive only through reuse donors are deliberately
+    not chased: the auto sweep should enumerate only lanes holding their own
+    measured MegaMoE rows.
+    """
+    candidates: list[str] = []
+    if backend_version is not None:
+        candidates.append(backend_version)
```
jasonqinzhou: [P1] Resolve query-slot aliases before probing MegaMoE rows

When `backend_version` is the advertised `current` or `next` alias, this appends the alias unchanged and `_megamoe_rows_resolve_at()` passes it to `resolve_op_data_path()`, which looks for a literal `<backend>/<version>/` directory. On GB300/vLLM, `get_version_slots()` maps `next` to `0.27.0`, but `_megamoe_lane_version("vllm", "gb300", "next", "moonshotai/Kimi-K3")` returns `None`; `build_default_tasks(..., backend="auto", backend_version="next", moe_backend="megamoe")` then drops the valid vLLM lane and exits with `No backends available`. Please resolve the explicit alias with `resolve_query_version` (and return/use the literal) before the row probe, and add a regression for `--backend-version next` on the auto MegaMoE lane.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.8/10 · high confidence**

## What this PR does

This PR adds fused Kimi-K3 MegaMoE modeling and measured GB300/vLLM 0.27.0 coverage, with collector provenance, compiled-engine support, CLI task generation, and documentation. Since the previous review, it has been rebased onto queryable version slots and now resolves default MegaMoE lanes into the slot carrying measured fused-module rows while bypassing separate DeepEP A2A requirements for that fused boundary.

The rebase follow-up correctly recognized both integration boundaries: the default no-pin Kimi-K3 auto task now reaches the vLLM next-slot data, and the focused measured execution test proves that the fused module answers from silicon rows.

## Why this score

6.8/10 quality assessment, BLOCKED packet. The required Claude Fable lane could not be verified: `claude-fable-5` was requested, but only Claude Sonnet 4.6 and Claude Haiku 4.5 were observed, so the skill forbids a non-blocked canonical recommendation. Independently, Codex confirmed one current-head P1: explicit query-slot aliases such as `--backend-version next` are passed directly to a filesystem probe, causing the valid GB300/vLLM MegaMoE lane to be dropped. The previous provenance finding remains fixed; 128 focused tests, all 26 exact-head checks, and all 18 review threads are otherwise clean.

src/aiconfigurator/cli/main.py
jasonqinzhou: Fixed in 30f0393d. Explicit query-slot aliases now resolve to their literal backend version before the exact MegaMoE row probe and before task construction. The new regression covers backend_version="next" on the auto Kimi-K3/GB300 lane and asserts that the valid vLLM tasks use literal 0.27.0. Validation on the exact b4d2b31b tree: 4,519 Python unit tests passed (12 skipped), 508 Rust tests passed (2 ignored), focused MegaMoE/version coverage passed, and Ruff/format/diff checks are clean.

Issue #1416, opened 2026-07-26
[Collector] Multi-node custom allreduce collection unsupported; TRT-LLM MNNVL path never exercised (GB200/GB300)


tianhaox:
## Summary

TODO: the collector currently cannot produce multi-node custom-allreduce data, and even if launched multi-node, the TRT-LLM MNNVL allreduce path would never be exercised. On GB200/GB300 (NVL, 4 GPUs per node) this means the `custom_allreduce_perf` table has no data for the TP sizes (8/16/32/...) that actually span nodes in serving, and any naive multi-node run would silently measure the wrong kernel path.

## Findings (verified against TRT-LLM 1.3.0rc20 installed sources)

**1. Default collection is single-node only.**
`collector/network/collect_comm.sh` derives `gpu_count_list` from the local node's GPU count (`nvidia-smi ... | wc -l`). On a GB200/GB300 node (4 GPUs) it collects only `tp=2,4`. There is no multi-node path in the default driver.

**2. Slurm custom-allreduce collectors are also single-node only.**
`collector/network/slurm/` has `slurm_custom_ar_2gpu.sh` / `slurm_custom_ar_4gpu.sh`, both `#SBATCH -N 1`; the README section is literally titled "Custom allreduce collection in one node". By contrast, the NCCL sweep has scripts up to 16 nodes / 64 GPUs. So there is currently no sanctioned way to collect custom allreduce for TP sizes that span nodes.

**3. Even a multi-node launch would not measure the serving kernel (MNNVL fidelity gap).**
Both workers construct the module without a dtype:

- `collector/network/collect_all_reduce.py` (`AllReduce(mapping=mapping)`)
- `collector/network/slurm/collect_allreduce.py` (`AllReduce(mapping=mapping)`)

In TRT-LLM, MNNVL enablement is decided at **construction time**, not call time (`tensorrt_llm/_torch/distributed/ops.py`):

```python
self.mnnvl_allreduce = MNNVLAllReduce(self.mapping, dtype) if dtype else None
```

[...]

Issue #1260, opened 2026-06-29
Define strict policy for communication rank-count empirical extrapolation


tianhaox:
## Context

PR #1238 calibrates NCCL and CustomAllReduce EMPIRICAL estimates as `query SOL / measured boundary utilization`.

When the requested communication group is larger than the collected grid, the implementation currently uses the largest available rank-count slice for utilization and reconstructs latency with the requested rank count SOL:

- CustomAllReduce: node-capped TP slice, for example TP8 utilization for a TP16 query.
- NCCL: largest collected GPU-count slice, for example 8-GPU utilization for a 16-GPU query.

This preserves the existing SILICON boundary-correction behavior and keeps configurations beyond the collected rank-count grid runnable. The result is currently reported with `xshape` provenance even when `transfer_policy=off`.

## Design gap

Strict empirical semantics should be able to say that an unmeasured rank count has no calibration data. Under that interpretation, `transfer_policy=off` should not silently use a neighboring rank-count slice.

Enforcing that immediately in PR #1238 would remove support for many otherwise valid multi-node/model configurations, so the current compatibility behavior is intentionally retained for now.

## Follow-up

Decide and implement an explicit policy for communication rank-count overflow without conflating it with model/kernel shape transfer. Options to evaluate include:

1. Treat rank-count boundary utilization as a separately controlled extrapolation policy.
2. Make strict mode fail accurately when the exact rank-count calibration is absent, while retaining compatibility extrapolation in the default mode.
3. Expand collection coverage enough that strict behavior does not regress supported configurations.

## Acceptance criteria

- The behavior of `transfer_policy=off` for NCCL and CustomAllReduce rank-count overflow is explicit and documented.
- Strict mode can distinguish exact measured calibration from boundary rank-count extrapolation.
- Provenance accurately reports the selected behavior.
- Support-matrix impact is measured before changing the default.
- Regression tests cover exact counts and overflow counts for both communication operations.

Related review: #1238.

PR #1554, opened 2026-08-15
fix(collector): enable multi-node custom allreduce on GB300 (#1416)

Ma1oneZhang:
Hi from Novita AI

## What

Refs #1416 (not closed by this PR).

Enables multi-node custom allreduce collection on GB300 and ships the GB300 TP8/TP16 silicon that was previously node-capped (NVL72 = 4 GPU/tray, so TP8 required 2 trays).

## Changes

- `collector/network/collect_all_reduce.py` + trtllm slurm stack: pass `dtype` into TRT-LLM AllReduce so MNNVL is constructed on multi-node TP (without it the collector silently records the IB fallback path — wrong data).
- New `slurm_custom_ar_{8,16}gpu.sh` 2/4-node launchers; torchrun static rendezvous under hostNetwork (LWS c10d rendezvous gaierror documented).
- Prefer a measured TP8/TP16 slice over the node-capped extrapolation in the SDK/rust comm paths.

## Data (GB300, custom allreduce eager+graph)

- vllm 0.27.0: 184 rows across TP 2/4/8/16, 127MiB–1GiB payloads (fresh collection; dev542-era data removed — comm family has no reuse.yaml and no shared-layer fallback).
- sglang 0.5.14 and trtllm 1.3.0rc20 TP8/TP16 slices refreshed with the same dtype fix.

## Verification

- Path fingerprint confirmed on-node (`Using ['CUSTOM','SYMM_MEM','PYNCCL']` + "Registering N cuda graph addresses").
- 783 collector unit tests + rust 470 lib tests green; `backend_facts --check` green.
- dev542 vs 0.27.0 cross-check on shared shapes: TP8/16 differ 3.2%, graph payload 1.35% — version noise, not drift.

## Known boundaries

- TP32 remains extrapolated from the TP4-capped util policy (#1260 policy still open).
- Communication dtype measured this wave is bfloat16 only (fp8/int8 comm lanes not collected).
- (Cluster IMEX/DRA concern resolved at collection time: trtllm multi-node TP8/16 rows are MNNVL-verified A/B, see the table's collection_meta comment.)

## Summary by CodeRabbit

* **New Features**
  * Added GB300 support for vLLM 0.27.0 custom all-reduce performance across TP2, TP4, TP8, and TP16.
  * Added multi-node collection workflows for 8- and 16-GPU configurations.
  * Enabled dtype-aware all-reduce behavior and MNNVL kernel-source reporting.

* **Bug Fixes**
  * Improved performance estimation by prioritizing measured tensor-parallel data and scaling only fallback estimates.
  * Improved Slurm rank detection and GPU task binding.

* **Documentation**
  * Clarified benchmark coverage, message sizes, data provenance, kernel selection, and multi-node collection requirements.


Ma1oneZhang:
CodeRabbit review round:
- slurm launchers: fixed — trailing `--mca` args were landing on `collect_allreduce.py`'s argv (which doesn't parse them); MCA settings now live in `OMPI_MCA_*` env exports (b873352).
- `kernel_source="TRTLLM"` labelling: intentionally not changed — the dtype axis is a provenance column consumers key on; the MNNVL-vs-fallback distinction is attested collection-side (A/B evidence in `collection_meta.yaml`, 128B 4.42us vs 27.27us on identical hardware). Recording it in-row is a data-contract change that needs producer+consumer coordination; tracked separately.

collector/network/slurm/slurm_custom_ar_16gpu.sh
```diff
@@ -0,0 +1,37 @@
...
+export OMPI_MCA_btl=tcp,self
+export OMPI_MCA_btl_tcp_if_include=bond0
+# Cross-node MNNVL allreduce (same recipe as the NCCL multi-node sweep)
+export NCCL_NVLS_ENABLE=1
+export NCCL_MNNVL_ENABLE=1
+export NCCL_CUMEM_ENABLE=1
+export NCCL_CUMEM_HOST_ENABLE=0
+export TRTLLM_DEEPSEEK_EAGER_FUSION_DISABLED=0
+
+# collect_allreduce.py reads SLURM_NTASKS / SLURM_NTASKS_PER_NODE / SLURM_LOCALID / RANK.
```
jasonqinzhou: [P1] Slurm's `srun` task environment exports the global process rank as `SLURM_PROCID` and the node-local task ID as `SLURM_LOCALID`; it does not define a per-task `RANK` (see the official output-environment contract: https://slurm.schedmd.com/srun.html#SECTION_OUTPUT-ENVIRONMENT-VARIABLES). The worker reads `os.environ["RANK"]` at import time, while both launchers only use `--export=ALL` and never synthesize it, so a standard launch exits with `KeyError: 'RANK'` before CUDA or MPI setup. Please read `SLURM_PROCID` in the worker (or set `RANK=$SLURM_PROCID` inside every task in both launchers) and add a static or subprocess test for this launcher/worker contract.

aic-core/rust/aiconfigurator-core/src/perf_database/communication.rs
```diff
@@ -798,6 +827,43 @@ mod tests {
...
+                (tp4.clone(), latency_curve(points.clone())),
+                (
+                    tp8,
+                    latency_curve(BTreeMap::from([(1024, 7.0), (4096, 9.0)])),
+                ),
+            ]),
+            BTreeMap::new(),
+            BTreeMap::new(),
+        );
+        assert_eq!(with_tp8.measured_tp_slice(CommQuantMode::Half, 8, 4), 8);
```
jasonqinzhou: [P2] This test's comment says it proves that a measured cross-node curve is not bandwidth-scaled, but the assertions call only `measured_tp_slice`; a regression later in `query_custom_allreduce_scaled` would still pass. The deleted Python test previously asserted the returned TP8/TP16 latency. Please query both synthetic tables through `query_custom_allreduce_scaled` and assert that the measured TP8 result is the raw TP8 value while the no-TP8 case uses the scaled TP4 fallback, so the policy remains covered at the actual query boundary.

aic-core/src/aiconfigurator_core/systems/data/gb300/comm/trtllm/1.3.0rc20/collection_meta.yaml
```diff
@@ -7,5 +7,14 @@ runtime:
   framework: trtllm
   version: 1.3.0rc20
 tables:
+  # TP2/TP4 (single tray, original wave) + TP8/TP16 multi-node added
+  # 2026-08-12 from a 4-node x 4-GPU GB300 NVL fleet via mpirun.
+  # The multi-node rows required the issue #1416 collector fix: the old
+  # code path passed a dtype that disqualified the TRT-LLM MNNVL
```
jasonqinzhou: [P2] This provenance sentence states the opposite of the collector defect fixed by the PR. The old constructor omitted `dtype`; the new `dtype=torch_dtype` argument makes MNNVL eligible. Please record that accurately so the sidecar remains trustworthy.

```suggestion
  # code path did not pass a dtype, which disqualified the TRT-LLM MNNVL
```

aic-core/src/aiconfigurator_core/systems/data/gb300/comm/trtllm/1.3.0rc20/collection_meta.yaml
```diff
@@ -7,5 +7,14 @@ runtime:
...
+  # allreduce, so a naive multi-node run silently measured the fallback.
+  # Measured A/B on identical hardware/payload: 128B 4.42us (MNNVL) vs
```
jasonqinzhou: [P2] The `message_size` table key is an element count, not a byte count: the collectors create BF16 tensors containing `size` elements, and the Rust query path also consumes element counts. The recorded 128..536,870,912 range therefore represents 256 B..1 GiB of physical BF16 payload, and this 128-cell A/B point is 256 B rather than 128 B. Please correct the byte labels here, on line 18, and in the SGLang sidecar's corresponding range statement.

jasonqinzhou (dismissed):
**REQUEST_CHANGES · 5.4/10 · high confidence**

## What this PR does

PR #1554 enables measured multi-node custom-allreduce coverage on GB300. It makes TensorRT-LLM MNNVL eligible by passing dtype, adds TP8/TP16 Slurm launchers and GB300 parquet slices for SGLang, TensorRT-LLM, and vLLM, and updates the Rust query oracle to prefer an exact measured TP slice before applying node-capped extrapolation.

The follow-up correctly removes a fixture wired to retired Python loaders, while the focused collector dtype tests and Rust selector test keep the single-oracle migration and the intended measured-slice policy explicit.

## Why this score

Score 5.4/10, REQUEST_CHANGES with high confidence. The new commit fixes the prior red unit job by deleting the obsolete Python-loader fixture, so that P1 is dropped and exact-head unit CI now passes. The launchers still depend on a per-task RANK variable that standard srun does not export, however, and the existing CodeRabbit-covered MNNVL fallback/provenance risk remains valid. The replacement Rust test checks only the slice selector rather than the scaled query result, and the two sidecar accuracy issues remain. Four exact-head compatibility/public-API checks also fail on DSA table-view baseline mismatches outside this PR's diff; those are not attributed to this PR, but they still leave the head merge-blocked.

collector/network/slurm/slurm_custom_ar_8gpu.sh
```diff
@@ -0,0 +1,38 @@
...
+export NCCL_MNNVL_ENABLE=1
+export NCCL_CUMEM_ENABLE=1
+export NCCL_CUMEM_HOST_ENABLE=0
+export TRTLLM_DEEPSEEK_EAGER_FUSION_DISABLED=0
+
+# collect_allreduce.py reads SLURM_NTASKS / SLURM_NTASKS_PER_NODE / SLURM_LOCALID;
+# the per-task rank comes from RANK if set, else SLURM_PROCID (what srun exports).
+srun -l \
+    --ntasks 8 --ntasks-per-node 4 \
+    --container-image=/path/to/trtllm_aarch64_release_v1.0.0rc2.sqsh \
```
tianhaox: Great to finally have sanctioned multi-node launchers here! Two small things I noticed (both apply to `slurm_custom_ar_16gpu.sh` as well):

1. The placeholder image is `trtllm_aarch64_release_v1.0.0rc2.sqsh`, while the shipped TP8/16 data and the dtype-gate reference in the comments are both against 1.3.0rc20 — and the `MNNVLAllReduce(mapping, dtype) if dtype else None` construction gate is what I verified on 1.3.0rc20 in #1416 (I haven't checked how 1.0.0rc2 behaves there). Since the PR description mentions the actual collection went through torchrun / mpirun rather than these sbatch scripts, could you update the placeholder to an image matching the shipped data, and maybe add a short note in the README on whether these launchers have been exercised on a real Slurm cluster as-is? Just so the next person collecting GB200/GB300 data knows what to expect.

2. A couple of leftovers that look specific to your cluster: the `/path/to/:/kimi` mount target, and `TRTLLM_DEEPSEEK_EAGER_FUSION_DISABLED=0` (which shouldn't matter for a pure allreduce sweep). Probably worth cleaning these up to generic placeholders so the template doesn't carry them forward.

collector/op_backend_facts.yaml
```diff
@@ -452,6 +452,7 @@ ops:
       - {framework: "vllm", version: "0.24.0", system: "h100_sxm", op_name: "all_reduce", allreduce_dtype: "bfloat16", backends: ["custom_allreduce"], kernel_sources: ["vLLM_custom_eager", "vLLM_custom_graph"]}
       - {framework: "vllm", version: "0.24.0", system: "h200_sxm", op_name: "all_reduce", allreduce_dtype: "bfloat16", backends: ["custom_allreduce"], kernel_sources: ["vLLM_custom_eager", "vLLM_custom_graph"]}
       - {framework: "vllm", version: "0.24.0", system: "rtx_pro_6000_server", op_name: "all_reduce", allreduce_dtype: "bfloat16", backends: ["custom_allreduce"], kernel_sources: ["vLLM_custom_eager", "vLLM_custom_graph"]}
+      - {framework: "vllm", version: "0.27.0", system: "gb300", op_name: "all_reduce", allreduce_dtype: "bfloat16", backends: ["custom_allreduce"], kernel_sources: ["vLLM_custom_eager", "vLLM_custom_graph"]}
```
tianhaox, line 458: The TP8/TP16 rows are really valuable, and recording what vLLM serving actually does cross-node is exactly what the table should capture. One labeling question though: since vLLM's custom allreduce doesn't span nodes, the TP8/16 rows presumably ran the SYMM_MEM or PyNCCL lane — the latencies suggest this too (vllm graph TP16 small-message is ~46us vs ~5us for the trtllm MNNVL rows), yet the parquet still labels them `vLLM_custom_eager` / `vLLM_custom_graph`.

Per the "kernel_source records ground truth" collector rule (point 3 in #1416), would it be possible to confirm which backend actually ran at TP8/16 and reflect that in `kernel_source`? Same question for the sglang slice. Totally fine to defer to the #1416 point-3 data-contract discussion if you'd rather not expand the axis values in this PR — a note in the collection_meta would already help.

aic-core/rust/aiconfigurator-core/src/perf_database/communication.rs
```diff
@@ -147,6 +147,25 @@ impl CommunicationTable {
         interp_message_size(curve, message_size)
     }
 
+    /// Resolve which measured TP slice backs a query for `tp_size`.
+    ///
+    /// Returns `tp_size` when the table has rows for that exact rank count,
+    /// otherwise the node-capped slice. Mirrors Python's `measured_tp_slice`.
```
tianhaox: Tiny doc nit: this comment (and the similar one in `operators/communication.rs`) says it mirrors Python's `measured_tp_slice` in `operations/communication.py`, but on main that file is just the engine delegation shell — the query logic lives entirely in the Rust engine, and there's no `measured_tp_slice` on the Python side. That's actually good news (no Python/Rust divergence to worry about — I checked), but the reference might send future readers hunting for a function that doesn't exist. Maybe reword to say the Rust engine is the single implementation?

tianhaox (commented):
Thanks a lot for picking this up — this goes well beyond the minimal fix I sketched in #1416, and it's great to see it come with fresh multi-node silicon. I pulled the three parquet files and checked them: TP2/4/8/16 × 23 sizes are complete with no duplicate keys and monotonic curves, and the trtllm TP8 small-message latencies (~4.3–4.4us) line up with the MNNVL A/B numbers in the collection_meta, so the dtype fix clearly did its job. The `measured_tp_slice` change on the Rust side also looks right to me — the fallback branch stays bit-for-bit compatible with the old node-capped behavior, and the four-quadrant test coverage is much appreciated.

I left a few small inline comments/questions below — mostly cleanups and one labeling question. None of them are blockers from my side.

collector/network/collect_all_reduce.py
```diff
@@ -169,7 +169,9 @@ def benchmark_trtllm_allreduce(
 
         op_list = []
         for i in range(repeat_n):
-            allreduce = trtllm_mods["AllReduce"](mapping=mapping).cuda()
+            # dtype enables MNNVL for multi-node TP (issue #1416):
+            # _torch/distributed/ops.py @v1.3.0rc20 builds `MNNVLAllReduce(mapping, dtype) if dtype else None`.
+            allreduce = trtllm_mods["AllReduce"](mapping=mapping, dtype=torch_dtype).cuda()
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in 7f901ee. The collector now requires an active TensorRT-LLM MNNVL object, reads the pinned rc20 one-shot threshold, and records TRTLLM_MNNVL_oneshot or TRTLLM_MNNVL_twoshot. If MNNVL is inactive, regular AUTO remains unobservable and collection fails loudly instead of writing an ambiguous label. The GB300 TP8 and TP16 rows were relabeled from the exact rc20 threshold; all latency values remain byte-for-byte unchanged.

collector/network/slurm/slurm_custom_ar_16gpu.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# Multi-node custom allreduce sweep: TP=16 across 4 nodes (GB200/GB300: 4 GPUs/node,
+# so TP>4 always spans nodes). Env template mirrors slurm_nccl_test_4node16gpu.sh;
+# adjust NCCL_IB_HCA / NCCL_SOCKET_IFNAME / container-image to the cluster.
+#SBATCH -N 4
+#SBATCH --gpus 16
+#SBATCH --ntasks-per-node=4
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in 7f901ee. Both launchers request four GPUs per node plus one GPU per task with gpu-bind single:1. The worker verifies exactly one visible CUDA device and selects logical device 0, while SLURM_LOCALID remains topology metadata. Static launcher-worker contract tests cover both scripts.

collector/network/slurm/slurm_custom_ar_16gpu.sh
Ma1oneZhang: Addressed in e4f608c and tightened in 7f901ee. The worker resolves int(os.environ.get RANK or os.environ subscript SLURM_PROCID), so standard srun launches use the exported global process rank and missing rank state fails loudly. The AST test now pins the exact operand order and fallback expression.

aic-core/rust/aiconfigurator-core/src/perf_database/communication.rs
Ma1oneZhang: Addressed in 59fd243. The Rust test now queries both synthetic tables through query_custom_allreduce_scaled: measured TP8 returns the raw TP8 latency, while a table without TP8 uses the scaled TP4 fallback. It also covers TP16 fallback. Rebase validation ran 23 communication tests successfully.

aic-core/src/aiconfigurator_core/systems/data/gb300/comm/trtllm/1.3.0rc20/collection_meta.yaml
Ma1oneZhang: Addressed in 3d24a81. The sidecar now states that the old constructor omitted dtype, which disqualified MNNVL, while the fixed constructor makes the multi-node MNNVL path eligible.

aic-core/src/aiconfigurator_core/systems/data/gb300/comm/trtllm/1.3.0rc20/collection_meta.yaml
Ma1oneZhang: Addressed in 3d24a81. The sidecars now describe message_size as a BF16 element count, the 128 cell as 256 B, and the full range as 256 B through 1 GiB.

tests/unit/collector/test_slurm_custom_ar_env_contract.py
```diff
@@ -0,0 +1,79 @@
...
+
+def test_worker_rank_errors_when_neither_var_is_set():
+    assigns = _module_assignments(_WORKER, "rank")
+    assert assigns
+    # A subscript (os.environ["SLURM_PROCID"]) raises KeyError instead of
+    # silently collapsing every task to rank 0 (which would make every rank
+    # the perf-row writer).
+    assert any(isinstance(child, ast.Subscript) for child in ast.walk(assigns[0].value)), (
+        "the `rank` resolution must raise (not default to 0) when neither RANK nor SLURM_PROCID is present"
+    )
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in 7f901ee. The rank contract test now requires the exact int(RANK get or SLURM_PROCID subscript) AST, including RANK first and a raising subscript fallback. It no longer accepts arbitrary environment references.

collector/network/slurm/slurm_custom_ar_8gpu.sh
Ma1oneZhang: Fixed in 7f901ee. Both templates now use the rc20 image placeholder, a generic aiconfigurator mount, no Kimi-specific mount, and no DeepSeek fusion variable. The README explicitly says the Slurm templates have not been exercised as-is and that the published rows came from torchrun or mpirun.

collector/op_backend_facts.yaml
Ma1oneZhang: Documented in 7f901ee without inventing historical labels. The vLLM and SGLang sidecars now cite the exact cross-node dispatch and rejection paths, state that the old kernel_source values encode launch mode rather than the per-call backend, and require instrumented recollection before relabeling. TRT-LLM TP8 and TP16 could be relabeled because the recorded A/B plus rc20 active-MNNVL terminal path and exact size threshold make that provenance recoverable.

aic-core/rust/aiconfigurator-core/src/perf_database/communication.rs
Ma1oneZhang: Fixed in 7f901ee. Both Rust comments now identify the Rust engine as the single implementation and describe Python or SDK as delegates, with no reference to a nonexistent Python measured_tp_slice helper.

collector/network/slurm/README.md
```diff
@@ -35,14 +35,20 @@ cat log_nccl/1node2gpu.out | grep 0:\ \  > 1u2g
 python3 cvt_log_to_perf_txt.py
 ```
 
-# 2.Custom allreduce collection in one node
+# 2.Custom allreduce collection
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in c951898. The section heading is now `# 2. Custom allreduce collection`.

collector/network/slurm/README.md
```diff
@@ -35,14 +35,20 @@ cat log_nccl/1node2gpu.out | grep 0:\ \  > 1u2g
...
 
-## 2.1 Replace the "/path/to" in the sctipt with actual path
+Single-node scripts cover TP=2/4. On GB200/GB300 (4 GPUs per node) serving-relevant TP sizes span nodes; use the multi-node scripts so the TRT-LLM MNNVL path is exercised (issue #1416). `collect_allreduce.py` must construct `AllReduce(..., dtype=...)` — without dtype the collector silently measures the fallback kernel.
+
+The TP8/TP16 launchers are reviewed templates and have not been exercised as-is on a Slurm cluster; the published GB300 rows were collected through `torchrun`/`mpirun`. Before use, replace the rc20 image and repository-mount placeholders, and verify the site supports `--gpus-per-task=1 --gpu-bind=single:1`; the worker intentionally fails unless each task sees exactly one CUDA device.
+
+## 2.1 Replace the "/path/to" in the script with actual path
 
 ## 2.2 Run the collector
 ```
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in c951898. The sbatch command fence now declares the bash language.

Ma1oneZhang:
> Thanks a lot for picking this up — this goes well beyond the minimal fix I sketched in #1416, and it's great to see it come with fresh multi-node silicon. I pulled the three parquet files and checked them: TP2/4/8/16 × 23 sizes are complete with no duplicate keys and monotonic curves, and the trtllm TP8 small-message latencies (~4.3–4.4us) line up with the MNNVL A/B numbers in the collection_meta, so the dtype fix clearly did its job. The `measured_tp_slice` change on the Rust side also looks right to me — the fallback branch stays bit-for-bit compatible with the old node-capped behavior, and the four-quadrant test coverage is much appreciated.
> 
> I left a few small inline comments/questions below — mostly cleanups and one labeling question. None of them are blockers from my side.

Good catch! Thanks for pointing this out. Fixed in the latest commit. Thanks for your comments! @tianhaox

tianhaox:
Thanks for the update — the MNNVL provenance work is a real improvement. Second-round review: there are 4 blocking correctness issues that need fixing, plus a few smaller cleanups introduced in this PR that I'd suggest addressing while you're here.

## Blocking

**1. `collector/network/collect_all_reduce.py:120` — the fail-closed MNNVL check breaks TRT-LLM allreduce collection on every non-MNNVL system.**
`_trtllm_mnnvl_kernel_source` raises whenever `mnnvl_allreduce is None`, and it's called unconditionally. On h100_sxm / h200_sxm / b200_sxm / b300_sxm / a100_sxm / l40s / gb200 — and on GB300 TP2/TP4, where this PR's own `collection_meta` says the rows stay plain `"TRTLLM"` — the first message size raises `RuntimeError` and zero rows are collected. `op_backend_facts.yaml` still declares `["TRTLLM"]` for all those systems, and the new test asserts TP<8 rows carry `"TRTLLM"`, a label this code path can no longer produce. Fix: return `"TRTLLM"` when `mnnvl_allreduce is None`; only raise when MNNVL is active but the variant can't be determined.

**2. `collector/network/slurm/collect_allreduce.py:58` — the new "exactly one visible GPU per task" requirement breaks the still-documented 2/4-GPU launchers.**
`slurm_custom_ar_2gpu.sh` / `4gpu.sh` use `#SBATCH --gpus 2` with no `--gpus-per-task` / `--gpu-bind`, so each task sees both GPUs and every rank dies with `RuntimeError('...found 2')` before any measurement — where the old `torch.cuda.set_device(local_rank)` worked. The README edited in this PR still documents that path (section 2.2), and the new env-contract test only covers the 8/16-GPU launchers, so CI won't catch it. Either update the 2/4-GPU scripts to the same one-GPU-per-task contract or relax the check.

**3. `aic-core/rust/aiconfigurator-core/src/perf_database/communication.rs:173` — `measured_tp_slice` only honors an exact TP match, producing non-monotonic estimates for unmeasured TP.**
Example with the shipped GB300 trtllm data (half, message_size 134217728): TP16 is measured at 1.0394 ms, but a TP32 query falls through to `min(32, 4) = 4` and scales the TP4 slice → 0.8748 ms, i.e. the model claims doubling ranks from 16 to 32 makes the allreduce 16% *faster*. Any user-supplied TP>16 sweep would systematically favor bigger TP on fabricated numbers. `query_nccl_scaled` already implements the right policy (cap to the largest measured fan-out and scale from there) — mirroring it here makes the exact-hit case fall out for free.

**4. `collector/network/collect_all_reduce.py:1008` — the trtllm `--use-slurm` branch still reads `os.environ["RANK"]` with no `SLURM_PROCID` fallback.**
An `srun` launch (the file's own docstring documents this; `--backend` defaults to trtllm) hits `KeyError('RANK')` before any benchmarking, since srun exports `SLURM_PROCID` and not `RANK` — the exact contract this PR establishes and tests for the sibling worker. The vllm (line 1023) and sglang (line 1047) branches already have the fallback; only the trtllm branch — the path this PR's dtype/MNNVL fix targets — is missing it.

## Suggested cleanups (non-blocking, all introduced in this PR)

**5. `tests/unit/collector/test_slurm_custom_ar_env_contract.py` — the AST/source-string pinning is stricter than the contract it protects.**
Pinning the exact expression shape means the equivalent `os.environ.get('RANK', os.environ['SLURM_PROCID'])` would fail the suite; the placeholder image filename (`trtllm_aarch64_release_v1.3.0rc20.sqsh`) is a template the README says users must edit, so every routine image bump breaks a test; and the `/kimi` / `TRTLLM_DEEPSEEK_EAGER_FUSION_DISABLED` absence checks encode one past copy-paste incident permanently. Suggest: extract a small `resolve_rank(env)` helper and test its behavior under three fake environs, keep the semantic launcher assertions (`--gpus-per-task=1`, `--gpu-bind=single:1`, no synthesized `RANK`), drop the rest.

**6. `tests/unit/collector/test_allreduce_mnnvl_dtype.py` — pins shipped-data contents and re-derives a TRT-LLM private constant.**
`len(rows) == 92` and the hardcoded `64*1024*8*2` threshold mean any legitimate re-collection (extra message size, TP32, or a TRT-LLM version where `_MNNVL_ONE_SHOT_THRESHOLD_BYTES` changed) breaks the test even when the new parquet is correct. Asserting just the schema invariant (`num_gpus < 8 → "TRTLLM"`, `>= 8 → startswith("TRTLLM_MNNVL_")`) keeps the protection without the lockstep. Also, `_load_pure_function` duplicates the `_load_function` helper already in `test_vllm_collect_attn.py` / `test_vllm_collect_mla_module.py` — worth hoisting into the collector tests' conftest.

**7. `collector/network/collect_all_reduce.py:207` — the cross-repeat `kernel_sources` consistency check is dead code.**
Every input to `_trtllm_mnnvl_kernel_source` is loop-invariant (tensor created once per size, fixed world_size, AllReduce rebuilt from identical args), so the set can never hold more than one element and the `len != 1` raise is unreachable. Compute `kernel_source` once before the loop and drop the set.

**8. `collector/network/slurm/slurm_custom_ar_8gpu.sh` / `16gpu.sh` — copy-paste twins differing only in `-N`/`--ntasks`/log names.**
Any future env-block fix has to land in both and will drift (the stale `slurm_custom_ar_2gpu.sh` next to them shows how that goes). `submit_moe_a2a.sh` / `submit_trtllm_alltoall.sh` in the same directory already demonstrate the parameterized pattern; one `submit_custom_ar.sh` could replace both and subsume the 2/4-GPU scripts too. Fine as a follow-up if you'd rather not do it here.

**9. Rust test in `communication.rs` — the new test re-derives the production scale formula.**
The `beyond_node_scale` closure re-implements the same math as the code under test, so it can only catch bugs in *when* the formula is applied, not in the formula itself. Asserting fixed numeric constants would make it a real check. (The custom-AR/NCCL duplication in production code predates this PR — not asking to fix that here.)

Ma1oneZhang:
Addressed the August 23 review in signed-off commit `152b1c0`.

Blocking fixes:
1. Non-MNNVL TRT-LLM now records the established `TRTLLM` family label. Variant-threshold validation only runs when `mnnvl_allreduce` is active.
2. TP2 and TP4 Slurm launchers now use the same one-GPU-per-task contract as TP8/TP16: `--gpus-per-task=1 --gpu-bind=single:1`, with the rc20 generic worker paths.
3. Custom-allreduce TP overflow now caps at the maximum measured slice for the quant mode. A TP32 request backed by TP16 scales from TP16 rather than falling back to TP4. The Rust regression uses fixed expected values, including `12.0 -> 12.4` for the TP16-to-TP32 case.
4. Every `--use-slurm` backend now resolves rank through the same fail-closed `RANK` / `SLURM_PROCID` helper; no rank-0 default remains.

Scoped cleanups included:
- replaced exact-expression rank tests with behavior tests for RANK-only, SLURM_PROCID-only, precedence, and missing-env failure;
- reduced shipped-data checks to the stable kernel-source schema invariant;
- removed the loop-invariant `kernel_sources` set and classify once while preserving per-object initialization order;
- removed the stale image/path and historical-copy-paste assertions.

I left the cross-file AST loader hoist and the four-launcher consolidation for a follow-up: both are non-blocking refactors, and launcher consolidation would change the documented `sbatch` interface.

Verification:
- collector unit suite: `1011 passed, 6 skipped, 55 deselected`
- Rust crate: `502 passed, 1 ignored`
- focused review regressions: `11 passed`
- targeted Ruff lint/format, rustfmt, `bash -n`, Python compile, and `git diff --check`: passed

@tianhaox

tianhaox (approved):

tianhaox:
Re-reviewed at 72b5195ef (`e01ca8825` "harden multinode allreduce fallbacks"). All my remaining concerns are resolved — no blockers left from my side.

**Verified fixed:**

- `_trtllm_mnnvl_kernel_source` now returns the established `TRTLLM` family label when MNNVL is inactive and only raises on genuine API drift (missing threshold constant). This unbreaks trtllm custom-AR collection on the 7 non-MNNVL systems that `collect_comm.sh` targets by default, and removes the contradiction with the sidecar / `op_backend_facts.yaml`, which both accept `TRTLLM` for regular AUTO. Probing only the first `AllReduce` instance per size is fine — construction is deterministic per process/mapping/dtype.
- The shared `_resolve_rank` (RANK → SLURM_PROCID → KeyError) wired into all three backends' `--use-slurm` branches plus the worker also fixes the pre-existing silent rank-0 collapse in the vllm/sglang branches. Replacing the AST structure assertions with a parametrized behavior test of the loaded function is a strict improvement.
- TP2/TP4 launcher cleanup + extending the contract tests to all four templates goes beyond what was asked — appreciated.
- The `measured_tp_slice` change (exact-hit-else-node-cap → cap-at-largest-measured-slice) is a real fix, not just plumbing: previously a TP32 query on GB300 would ignore the measured TP16 slice and extrapolate from TP4. I checked the two behavioral edges: (1) the "max measured < per_node" silent-extrapolation case is unreachable in today's data tree (only three old rtx_pro_6000_server version dirs hit it, and sibling-version source merging donates the TP8 rows); (2) an unmeasured intermediate cross-node TP (e.g. TP12 on GB300) now errors loudly instead of scaling from TP4 — the right failure mode for this repo.

Verified locally at 72b5195ef: 11/11 new collector unit tests, `cargo test --all-features custom_allreduce` all green; 26/26 CI checks pass.

**Non-blocking follow-ups** (fine to land as-is):

1. `slurm/collect_allreduce.py` still has no active-MNNVL check and writes the `ALLREDUCE_AUTO` text prefix, so the sanctioned Slurm path can't yet produce the provenance the mpirun path now records. Since the README declares these templates unexercised and the text output has no automated path into the DB, this can wait for the instrumented recollection already promised in the sidecars.
2. `perf_data_reuse_manifest.yaml` hasn't been regenerated, so the two new `TRTLLM_MNNVL_*` groups are absent. Runtime impact is nil today (comm has no cross-framework kernel_source group), but the daily manifest job will open a drift PR against main after this merges — regenerating here would save that round-trip.

PR #1557, opened 2026-08-17
feat(sdk): add GLM-5.3 support (BF16 / FP8 / NVFP4)

natoscott:
## Overview

Registers GLM-5.3 (`zai-org/GLM-5.3`, `zai-org/GLM-5.3-FP8`, and
`nvidia/GLM-5.3-NVFP4`) across the SDK, collector plans, and tests. GLM-5.3
uses the same modeled geometry as GLM-5.2, so collection reuses the existing
GLM-5.2 MoE rows instead of creating duplicate physical cases.

## Details

- Adds cached BF16, FP8, and NVFP4 model configs, plus the companion NVFP4 HF
  quantization config.
- Registers all three model paths in `DefaultHFModels`.
- Adds the three paths to the GLM DSA/MoE plan and MLA-module targeting.
- Maps the new MoE identities to the corresponding GLM-5.2 rows through
  `model_aliases`; GLM-5.3 therefore contributes zero new physical MoE cases.
- Keeps GLM-5.2 as the default sparse-collection checkpoint because it is the
  first registered artifact with the same longest-range geometry, while a
  targeted GLM-5.3 request still preserves the requested checkpoint.
- Adds quantization-policy coverage and verifies every GLM-5.3 config matches
  GLM-5.2 geometry, including `len(indexer_types) == num_hidden_layers`.

## Validation

- [x] Merged current `main` (`0dc8a2b438da88e39a2c63ee14f42d5dea65c1dc`)
  and resolved the aggregate MoE case-count conflict against main's 6,606-case
  baseline.
- [x] `ruff check` and `ruff format --check` on the changed tests.
- [x] 191 focused collector tests across model-case planning, SGLang MLA/MoE,
  vLLM MoE, and TRT-LLM MoE.
- [x] Earlier exact-head E2E recommendations: GLM-5.3 FP8 and NVFP4 on B200
  matched their GLM-5.2 baselines; GLM-5.3 BF16 succeeded on B300.
- [ ] Fresh hosted checks for the merged head.


collector/sglang/glm5_dsa_sparse_modules.py
```diff
@@ -67,8 +67,9 @@ def _selected_glm5_models():
     are checkpoint-quantization-independent and use only model geometry. On
     SM90, select the registered BF16 GLM-5 identity to match the supported
     full-module plan without advertising an NVFP4 artifact. On SM100/103,
-    prefer GLM-5.2-NVFP4 because its longer range covers GLM-5. Return no cases
-    when a targeted model filter selects an unsupported or non-GLM checkpoint."""
+    prefer GLM-5.3-NVFP4 (or GLM-5.2-NVFP4) because its longer range covers
+    GLM-5. Return no cases when a targeted model filter selects an unsupported
+    or non-GLM checkpoint."""
```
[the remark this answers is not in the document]
natoscott: Fixed.

aic-core/src/aiconfigurator_core/model_configs/nvidia--GLM-5.3-NVFP4_config.json
```diff
@@ -0,0 +1,492 @@
...
+        "shared",
+        "full",
+        "shared",
+        "shared",
+        "shared",
+        "full",
+        "shared",
+        "shared",
+        "shared"
+    ],
```
jasonqinzhou, line 106: **P1 — Restore a complete per-layer indexer schedule**

This remains reproducible on the updated head: `indexer_types` has 74 entries while `num_hidden_layers` is 78. Transformers constructs every GLM decoder layer with `config.indexer_types[layer_idx]`, so a full load reaches `IndexError` for layers 74–77; the collector's default two-layer dummy override masks this. The repository's GLM-5.2 NVFP4 config has the complete 78-entry schedule, with the following four entries at the tail. Please restore them and add a config invariant test such as `len(indexer_types) == num_hidden_layers`.

```suggestion
        "full",
        "shared",
        "shared",
        "shared"
    ],
```

jasonqinzhou (changes requested):

## What this PR does

This PR registers GLM-5.3 BF16, FP8, and NVFP4 configs throughout the SDK and collector, extends the MoE and DSA collection plans, and makes the Blackwell sparse-kernel selector prefer the new NVFP4 artifact.

The integration is carried coherently across bundled configs, model registration, quantization policy, sparse selection, and tests, and the follow-up commit promptly repaired both WideEP population contracts while clarifying the artifact-selection rationale.

## Why this score

6.5/10, REQUEST_CHANGES: the head update fixes the prior WideEP failures and CodeRabbit docstring concern, the required unit job is now green, and targeted checks pass. One blocking correctness issue remains: the bundled GLM-5.3 NVFP4 config still has only 74 indexer entries for 78 decoder layers, so a full model construction can fail on the final four layers.

aic-core/src/aiconfigurator_core/model_configs/nvidia--GLM-5.3-NVFP4_config.json
natoscott: Updated, thanks @jasonqinzhou

natoscott: The failing test test_supported_quant_modes_structure is unrelated to the GLM-5.3 changes (which only modify model configs, collector case YAML, and test count assertions).

- Failure: assert len(supported[op]) > 0 — one of the ops returned an empty list for supported quant modes
- Test passes locally on the exact same commit (d2281474) that failed in CI
- This test uses the comprehensive_perf_db fixture which mocks all data loaders, though none of these are touched by this PR

Possible causes: flaky test or pytest-xdist race condition? The test infrastructure was recently refactored in commit fa8ebbb3 (MoE schema unification).

tianhaox (changes requested):
Requesting changes — one structural ask: **GLM-5.3 should ride GLM-5.2's collected data via aliasing, not new collection cases.** Everything else in the PR is solid, and I verified the tests locally.

## Why reuse is the right model

GLM-5.3 is post-training-only on the GLM-5.2 base (Z.ai's own statement); the config delta is literally zero — same IndexShare layout (21 full / 57 shared indexers, `index_topk_freq=4`), same 1M range, same MoE dims, same quant recipes. Contrast with 5.1→5.2, which was a real change (IndexShare indexer-KV reuse introduced, `index_topk_freq` 1→4, range 202752→1M) and therefore genuinely needed fresh collection — including the new skip-indexer op. 5.2→5.3 changes nothing about the deployment form, so the collection plan should express reuse, not re-measurement. This is also the criterion to apply to future 5.x additions: new collection only when the deployment form actually changes.

## Blocking

1. **`GlmMoeDsaForCausalLM_cases.yaml` — replace the three new moe rows with `model_aliases`** on the corresponding GLM-5.2 rows (`zai-org/GLM-5.2` ← `zai-org/GLM-5.3`, `-FP8` ← `-FP8`, `-NVFP4` ← `-NVFP4`). This is the established pattern: `deepseek-ai/DeepSeek-V3`'s moe row aliases DeepSeek-R1 + DeepSeek-V3.2, and `nvidia/DeepSeek-V3.1-NVFP4` aliases `DeepSeek-V3.2-NVFP4`. As written, the PR adds +351 moe / +24 vLLM moe_ep / +8 SGLang moe_ep collection cases that would re-measure identical geometry at identical quant on every system. With aliases, the case-count pins revert to their current values (62/186, 110, 5340) — no test-count churn at all.

2. **`base_ops/moe.yaml` — drop the `nvidia/GLM-5.3-NVFP4` addition** to the NVFP4 `allowed_model_paths`. That list gates which *physical* checkpoints may expand nvfp4 moe cases (`case_generator.py` `moe_model_allows_quantization`); under aliasing, no case ever carries `model_path=nvidia/GLM-5.3-NVFP4`, so the entry gates nothing and muddies the list's "physical load targets" semantics. Precedent: `nvidia/DeepSeek-V3.2-NVFP4` (an alias) is deliberately not listed. Consequently, the new `moe_model_allows_quantization("sglang", "nvidia/GLM-5.3-NVFP4", "nvfp4")` assertion in `test_model_cases.py` must also go (it returns False for unlisted alias names — same reason the DeepSeek quant-policy test only asserts fp8_block for the R1/V3.2 aliases). The `GLM-5.3-FP8` fp8_block assertions can stay: fp8_block has no whitelist and the alias resolves to the 5.2-FP8 row's `allowed_modes`.

3. **`glm5_dsa_sparse_modules.py` — revert the preference flip to 5.3-NVFP4.** It buys nothing (same 1M range, byte-identical geometry — the docstring says as much), needlessly changes the recorded provenance of future sparse rows, and leaves the two GLM DSA collection paths on different canonical artifacts while the mla_module canonical stays on GLM-5.2. Keep 5.2 as the sparse canonical; revert the selector test change with it.

## Keep as-is

- **The three config JSONs and the `common.py` registration.** The configs are byte-for-byte copies of 5.2's, and that's fine here — the real GLM-5 vs GLM-5.1 FP8 artifacts (also a same-base pair) carry identical 712-entry `modules_to_not_convert` lists (same set, same order), so zai's quant recipe is demonstrably stable across same-base releases, and the modeling only consumes the recipe-determined fields. The configs are also load-bearing for the mla_module additions (the collector reads the bundled config for `max_position_embeddings`). One addition: please also copy `nvidia--GLM-5.3-NVFP4_hf_quant_config.json` — every sibling NVFP4 model ships the companion file.
- **The `mla_module` `model_paths` additions** — canonicalization already dedups them onto GLM-5.2 at zero cost; the updated test comments are accurate (verified: `max()` returns the first maximal element on the YAML-order tie at 1048576).
- **The `@pytest.mark.unit` fix** — load-bearing, thanks for catching it.

## FYI (no action needed)

Registering the three paths in `DefaultHFModels` auto-enrolls them in the daily support-matrix roster (`SupportMatrixHFModels = DefaultHFModels − Retired`). GLM-5.2 and GLM-5.3 will coexist in the matrix — that's intended; the first daily compare after merge will show new GLM-5.3 rows, which is expected.

## Verification

I applied this diff on `origin/main` and ran the four affected test files — 118 passed; all updated count pins are real. After the alias rework, a quick e2e `recommend` run on `zai-org/GLM-5.3` would be a good sanity check that the query path lands on the reused data (DeepSeek-R1/V3.2 already exercise the same mechanism).

aic-core/src/aiconfigurator_core/model_configs/nvidia--GLM-5.3-NVFP4_config.json
jasonqinzhou: Resolved on 81d000a19: the NVFP4 schedule is complete, and the new GLM-5.3 geometry test now asserts len(indexer_types) == num_hidden_layers for BF16, FP8, and NVFP4.

tests/unit/collector/test_model_cases.py
```diff
@@ -400,6 +400,10 @@ def test_moe_model_quantization_policy_is_yaml_backed():
     assert not moe_model_allows_quantization("sglang", "nvidia/GLM-5.2-NVFP4", "bfloat16")
     assert moe_model_allows_quantization("sglang", "zai-org/GLM-5-FP8", "fp8_block")
     assert not moe_model_allows_quantization("sglang", "zai-org/GLM-5-FP8", "nvfp4")
+    assert moe_model_allows_quantization("sglang", "nvidia/GLM-5.3-NVFP4", "nvfp4")
+    assert not moe_model_allows_quantization("sglang", "nvidia/GLM-5.3-NVFP4", "bfloat16")
+    assert moe_model_allows_quantization("sglang", "zai-org/GLM-5.3-FP8", "fp8_block")
+    assert not moe_model_allows_quantization("sglang", "zai-org/GLM-5.3-FP8", "nvfp4")
```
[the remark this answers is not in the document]
jasonqinzhou: Resolved on 81d000a19: added the requested GLM-5.3 BF16 acceptance and incompatible-mode rejection coverage, plus the three-variant geometry invariant.

jasonqinzhou (approved):
Approved at exact head 81d000a191e7493a5180e1c71c6214a41a551c92. The GLM-5.3 aliases preserve GLM-5.2 geometry and collector provenance, the NVFP4 layer schedule is complete and invariant-tested, the merge conflict is resolved against current main, all review threads are resolved, focused local validation passed 191 tests, and the substantive hosted build/test matrix is green.

jasonqinzhou:
@tianhaox Your requested changes are addressed on exact head `81d000a191e7493a5180e1c71c6214a41a551c92`: GLM-5.3 MoE identities now alias the GLM-5.2 physical rows, the base-op whitelist addition is gone, sparse collection retains GLM-5.2 as the canonical provenance, and the companion NVFP4 HF quant config is included. Current `main` is merged, all review threads are resolved, local focused tests pass 191/191, and the substantive hosted checks are green. Re-review requested so the stale change-request state can be cleared.

tianhaox (approved):

PR #1559, opened 2026-08-17
fix(collector): serve-parity conv metadata for vLLM KDA context + gb300 kda recollect at vllm 0.27.0

Ma1oneZhang:
Hi from Novita AI

## Summary

Fix a measurement-parity bug in the vLLM KDA collector **and** recollect
`gb300/kda/vllm/0.27.0` context rows with serve-parity timing.

### Bug

`collector/vllm/collect_kda.py` context timing invoked
`causal_conv1d_fn` **without** `metadata`. The handler's `metadata=None`
branch is a non-serving path that rebuilds token-offset tables with
`np.repeat` + `nums.sum().item()` (D2H sync) inside **every** timed call
(~0.25 ms/conv on GB300 → flat 0.76-0.9 ms floor per `qkv3` iteration).
All rows for `causal_conv1d_fn_qkv3` (context) in pinned-preview table
`kda/vllm/0.1.dev19262` inherit this floor and are shape-flattened (bug
is host-side, channel/token independent).

Serving pre-computes conv metadata once per step in the KDA attention
metadata builder and passes it into every layer's prefill conv call
(vllm 0.27.0 image:
`vllm/models/kimi_k3/nvidia/kda_metadata.py::KimiK3KDAMetadata`
(subclass of `GDNAttentionMetadata`) →
`kda.py::_prefill_conv(..., metadata=m)`).

### Fix

Build `GDNAttentionMetadata` per (batch, seq) via
`compute_causal_conv1d_metadata` and pass `metadata=` — identical to the
GDN precedent in `collector/vllm/collect_gdn.py`. Timing moves to
`repeat_n=10` with CUDA-graph replay (no longer `use_cuda_graph=False`),
so rows report GPU service time, matching serving's deep-queue launches.

### Measured evidence (GB300, vllm/vllm-openai:v0.27.0 image, K3 TP8 shard)

| shape | old path (no metadata) | kernel-only (CUPTI) | fixed path (metadata+graph~rep10) |
|---|---|---|---|
| b1 s1      | 0.838 ms | 0.0037 ms | 0.0029 ms |
| b1 s4096   | 1.157 ms | 0.0132 ms | 0.0115 ms |
| b64 s4096  | 6.120 ms | 2.020 ms  | 1.968 ms |
| b32 s32768 | 21.96 ms | 7.89 ms   | 7.79 ms  |

vs sdk sol bytes model (36*proj*nt @ 8 TB/s): kernel reaches ~90% of SOL
at saturation => current kernel is fine; AIC's table was the artifact.

### Data (committed in this PR)

- [x] `gb300/kda/vllm/0.27.0/`: 1198 rows (context 895, generation 87, verify 216), full base grid recollected with the fix.
- [x] Formerly-guarded `nt*proj >= 2**31` veto deleted: 0.27.0's causal_conv1d is int64 end-to-end (`stride_x_token: tl.int64` verified in-image); the ~40 added context rows pass with SOL-sane timings, and the framework_manifest kda family moves 0.1.dev19262 -> 0.27.0.
- [x] Generation loses the preview-era `fused_kda_decode` rows intentionally: 0.27.0's `is_fused_kda_decode_supported` rejects DS conv-state layout (`VLLM_SSM_CONV_STATE_LAYOUT=DS`, kda.py:135-159) — the collector probes via the same predicate as serving, so decode rows are `causal_conv1d_update` + `fused_recurrent_kda_packed_decode`, matching serving parity.
- Accepted cells (recorded in the failure log, same spectrum as the preview era): `packed_decode` at (bs=1024, nv=96) triton invalid argument; context (nv=96, bs=64, s=32768) OOM at input allocation. SDK interpolation covers both.
- Consumer check: `aiconfigurator cli default --model-path moonshotai/Kimi-K3 --system gb300 --backend vllm --backend-version 0.27.0 --database-mode SILICON` produces agg/disagg pareto rows (ttft 1.1-1.4s @ isl4k/osl1k); at 0.1.dev19262 the same SLA run yields NO agg config — the polluted 0.8ms/conv floor priced prefill out of the SLA.
- Validation gates: ruff, tests/unit/collector (incl. manifest + kda contract pins), tools/perf_database/backend_facts --check, check_collector_data R1-R6, tests/unit/sdk/database (2 pre-existing failures unchanged on clean tree).

## Summary by CodeRabbit

- **New Features**
  - Added GB300 NVL72 performance metadata for vLLM 0.27.0.
  - Added backend coverage and reporting for context, generation, and verification phases.

- **Updates**
  - Refreshed provenance, rerun results, capacity limits, kernel coverage, and layout restrictions.
  - Improved context benchmarking with serving-compatible metadata and repeated measurements.

- **Bug Fixes**
  - Corrected benchmark routing for single-token and longer sequences.
  - Resolved convolution token-offset handling limitations and improved phase-specific error reporting.


collector/framework_manifest.yaml
```diff
@@ -39,13 +39,17 @@ frameworks:
...
+      # Kimi-K3 KDA kernels shipped upstream in 0.27.0
+      # (vllm/models/kimi_k3/* incl. fused_kda_decode + FlashKDA;
+      # verified in-image on GB300, previous preview pin
+      # 0.1.dev19262+gb6bbf29dd retired). Digest below is the public
+      # multi-arch manifest list of v0.27.0; on aarch64 hosts it resolves
+      # to the arm64 entry (sha256:7441a0579b7974fe00eec2c41a5710bcd0dc3da
+      # fc03edcdc5663158dd68286cf — verified serving Kimi-K3 TP8).
       kda:
-        version: "0.1.dev19262"
+        version: "0.27.0"
```
jasonqinzhou: [P1] The 0.27.0 bump also needs a consumer routing change. This PR correctly records that DS layout makes is_fused_kda_decode_supported false and collects only causal_conv1d_update + fused_recurrent_kda_packed_decode, but KimiK3._build_generation_ops still emits only fused_kda_decode for every non-spec vLLM decode and removes generation_kda_onorm. On this head I loaded gb300/vllm/0.27.0 in SILICON mode: the shared view fills 44 fused_kda_decode rows from 0.1.dev19262, so the new-version estimate continues pricing the retired preview kernel (0.5756 ms at bs1); with shared reuse disabled it still falls back to the fused SOL model and still omits the norm. Please keep the family pin on the preview until a coordinated consumer fix lands, or update version-aware Kimi-K3 routing so 0.27.0 uses the collected conv-update + packed-recurrence pair and explicit gated RMSNorm, prevents preview fused rows from filling this version, and pins the behavior with an exact 0.27.0 test.

aic-core/src/aiconfigurator_core/systems/data/gb300/kda/vllm/0.27.0/collection_meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  # everywhere at 0.27.0: the is_fused_kda_decode_supported predicate
+  # rejects DS conv-state layout (VLLM_SSM_CONV_STATE_LAYOUT=DS, matching
+  # serving on this image), so generation runs packed_decode + conv update.
+  kda_perf:
+    collector_ref: 7969588
+    collector_hash: sha256:20bc4a950d53065a484c6d6cb03b6200b293fd6a53124a5c5d2c845216696651
+    case_plan_hash: sha256:6a42ea30c5fe48538d365ca28a5b1f4f0c79a511f808ed5324efd46010d1cd66
+    collected_at: '2026-08-17'
+    rows: 1198
+    status: complete
```
jasonqinzhou: [P1] This must be partial, not complete. The current KDA plan produces 1,200 rows, while the parquet has 1,198: it is missing context nh96/bs64/s32768 flashkda_fwd and generation nh96/bs1024 fused_recurrent_kda_packed_decode. Both benchmark functions record these cell failures and then raise their strict-completeness error, and _write_collector_provenance derives complete only when no unresolved failed cases remain. Please regenerate this sidecar from the actual checkpoint evidence (and correct the note above: the context causal_conv1d_fn_qkv3 row exists; only flashkda_fwd is missing).

tests/unit/collector/vllm/test_collect_kda_contract.py
```diff
@@ -14,15 +14,37 @@
...
+    # that run_kda_context_benchmark passes metadata= structurally.
+    tree = ast.parse(SOURCE_PATH.read_text(encoding="utf-8"), filename=str(SOURCE_PATH))
+    for node in ast.walk(tree):
+        if isinstance(node, ast.FunctionDef) and node.name == "run_kda_context_benchmark":
+            conv_calls = (
+                call
+                for call in ast.walk(node)
+                if isinstance(call, ast.Call) and isinstance(call.func, ast.Name) and call.func.id == "causal_conv1d_fn"
+            )
+            assert conv_calls and all(any(kw.arg == "metadata" for kw in call.keywords) for call in conv_calls)
```
jasonqinzhou: [P2] conv_calls is a generator, so it is always truthy, and all() over the empty iterator is also true. I exercised this exact condition on a run_kda_context_benchmark function with no causal_conv1d_fn call and it still evaluated true. Materialize the calls into a list, assert that the list is non-empty (or assert the expected single call site), and then check that every call supplies metadata=.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 4.8/10 · high confidence**

## What this PR does

This PR fixes GB300 KDA prefill convolution collection by supplying serving-style GDNAttentionMetadata, upgrades the KDA family from the preview build to vLLM 0.27.0, removes the obsolete int32 guard, and commits refreshed backend facts plus a 1,198-row performance table.

The source-level parity work is thoughtful: it pins tagged vLLM source, mirrors the established GDN metadata construction, verifies the int64 stride contract, and ships a duplicate-free dataset whose kernel choices and observed gaps are documented in detail.

## Why this score

4.8/10, REQUEST_CHANGES. The rebase preserves the original seven-file patch and all targeted checks pass, but the 0.27.0 bump leaves non-speculative Kimi-K3 generation modeled with the retired preview fused-decode path: the current SILICON view silently fills 44 fused_kda_decode rows from 0.1.dev19262 and the model omits the now-unfused gated RMSNorm. The prior seq_len=1 serving-dispatch, provenance-status, and vacuous-test findings also remain unchanged; the seq_len=1 finding stays unselected because CodeRabbit already has an exact-head unresolved thread for it.

collector/vllm/collect_kda.py
```diff
@@ -234,6 +230,36 @@ def run_kda_context_benchmark(
...
+                    num_decodes=0,
+                    num_decode_tokens=0,
+                    num_spec_decodes=0,
+                    num_spec_decode_tokens=0,
+                    num_actual_tokens=nt,
+                    nums_dict=nums_dict,
+                    batch_ptr=batch_ptr,
+                    token_chunk_offset_ptr=token_chunk_offset_ptr,
+                )
+
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in cf2c909. The shared context grid still contains seq_len=1, but run_kda_torch now dispatches that cell through the serving decode benchmark and keeps only seq_len>1 in the prefill path. The executable contract test verifies [1,2,4] becomes context-labelled decode plus prefill [2,4], and invalid lengths fail loudly. A GB300 vLLM 0.27.0 smoke emitted exactly causal_conv1d_update and fused_recurrent_kda_packed_decode. The 56 old prefill-labelled rows were removed from the shipped partial table.

collector/framework_manifest.yaml
Ma1oneZhang: Addressed in 6ed6bcf. The KDA family remains pinned to the Kimi-K3 preview runtime until coordinated 0.27.0 consumer routing lands, so automatic collection and estimates do not claim the split-decode table as the active family version. The manifest comment records the exact consumer mismatch and re-pin condition.

aic-core/src/aiconfigurator_core/systems/data/gb300/kda/vllm/0.27.0/collection_meta.yaml
Ma1oneZhang: Addressed in e3c4093 and tightened in cf2c909. The sidecar is status partial, names the two original failed cells correctly, and now reports 1142 rows after removing 56 one-token context rows that were collected through the wrong prefill path. check_collector_data R1-R7 passes.

tests/unit/collector/vllm/test_collect_kda_contract.py
Ma1oneZhang: Addressed in 256bdc6. conv_calls is materialized as a list, the test requires exactly one causal_conv1d_fn call site, and then verifies that site supplies metadata. The focused collector contract suite passes.

collector/kernel_source_backends.yaml
```diff
@@ -203,7 +203,7 @@ mappings:
...
+  - {framework: vllm, kernel_source: causal_conv1d_update, backend: triton, source: "collector/vllm/collect_gdn.py:23,376,401 (gdn_perf at 0.18.0/0.19.0/0.22.0/0.24.0) and collect_kda.py:412,497 (decode) + :581,634 (verify) (kda_perf at 0.27.0; :199-205 records the int64 token-offset verification at v0.27.0) all import vllm/model_executor/layers/mamba/ops/causal_conv1d.py, a pure Triton module at every shipped era — @triton.jit _causal_conv1d_update_kernel :748-749 and causal_conv1d_update :1069 @v0.18.0/v0.19.0/v0.24.0 (:750-751 / :1071@v0.22.0); it imports only vllm.triton_utils, no CUDA causal-conv1d dependency"}
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in cf2c909. Every collect_kda.py citation was refreshed against the rebased file, including int64 evidence, prefill import and invocation ranges, resolver, decode, packed recurrence, and verify ranges. Backend-facts tests and YAML parsing pass.

collector/vllm/collect_kda.py
```diff
@@ -234,6 +240,39 @@ def run_kda_context_benchmark(
...
+                    num_prefill_tokens=nt,
+                    num_decodes=0,
+                    num_decode_tokens=0,
+                    num_spec_decodes=0,
+                    num_spec_decode_tokens=0,
+                    num_actual_tokens=nt,
+                    nums_dict=nums_dict,
+                    batch_ptr=batch_ptr,
+                    token_chunk_offset_ptr=token_chunk_offset_ptr,
+                )
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in e6c2e1e. The collector now cites vLLM v0.27.0 kda_metadata.py lines 261-279 for every num_* count, 403-418 for nums_dict, batch_ptr, and token_chunk_offset_ptr, and 466-485 for object construction plus num_actual_tokens. The contract test now requires the causal_conv1d_fn metadata keyword to be the exact conv_metadata AST name, so metadata=None cannot satisfy it. All shifted backend provenance line references were refreshed again.

collector/vllm/collect_kda.py
```diff
@@ -694,7 +743,26 @@ def run_kda_torch(
         device=device,
     )
     if phase == "context":
-        run_kda_context_benchmark(batch_size_list=batch_size_list, seq_len_list=seq_len_list, **kwargs)
+        if seq_len_list is None:
+            raise ValueError("vLLM KDA context collection requires seq_len_list")
+        if any(seq_len < 1 for seq_len in seq_len_list):
+            raise ValueError(f"vLLM KDA context sequence lengths must be positive: {seq_len_list}")
```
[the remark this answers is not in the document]
Ma1oneZhang: Fixed in e6c2e1e. The context entrypoint now rejects an empty seq_len_list before the positivity check, and the executable extracted-function test verifies that the empty grid raises instead of returning WORKER_RESTART with zero cases.

jasonqinzhou (approved):
**APPROVE · 9.1/10 · high confidence**

## What this PR does

This PR fixes vLLM KDA context timing by constructing and passing serving-style convolution metadata, records a partial GB300 vLLM 0.27.0 dataset, and keeps the active KDA family on the preview runtime until version-aware decode routing lands. It also routes one-token context cells through the serving decode kernels and preserves the related provenance and contract tests.

The author handled the earlier review feedback carefully: the implementation cites every hand-built metadata field to the pinned vLLM serving builder, makes the metadata contract test non-vacuous, rejects empty context grids, and keeps kernel-source anchors synchronized after the rebase.

## Why this score

9.1/10 local quality signal, BLOCKED packet. Codex found the rebased current head approval-quality with no actionable finding: all eight PR patches are identical to the previously reviewed series, focused checks passed, all live checks are green, and all review threads are resolved. The canonical recommendation must nevertheless be BLOCKED because the independent helper requested claude-fable-5 but observed only claude-haiku-4-5-20251001 and claude-sonnet-4-6; the skill forbids treating that output as a verified Fable review.

PR #1556, opened 2026-08-15
fix: add sglang 0.5.14 DSA skip-indexer perf data for h200/h100/b200 (AIC-1747)

Arsene12358:
## What

Ships the sglang 0.5.14 DSA **skip-indexer** perf rows for h200_sxm, h100_sxm, and b200_sxm, registers the two kernel-source labels the collection surfaced, and lifts the SM90 `unverified_sms` marker on the two skip ops now that they are hardware-validated on Hopper. Linear: AIC-1747 (project "Gap: Dynamo Recipe Alignment", umbrella AIC-1743 §5).

GLM-5.2 shares one topk index across `index_topk_freq` layers: only 21 of 78 layers compute the indexer, the other 57 reuse it (the cheaper `skip_indexer` path). Until now the skip tables existed only for Blackwell — Hopper had **zero** skip rows because the registry marked the ops `unverified_sms=(90, …)` and `collect.py` records unverified (op, SM) combos as `unverified_skipped` without running them. On unpatched main the unavoidable skip query kills the GLM-5.2-FP8/h200 model load outright (`PerfDataNotAvailableError` — the AIC-1743 §5 finding).

## Provenance (probe collections)

Collected on branch `probe/aic-1747-dsa-skip-sm90` (temporary; lifts SM90 for the run — B300-probe pattern), pipelines [62700025](https://gitlab-master.nvidia.com/dl/ai-dynamo/aic-auto-collector/-/pipelines/62700025) (h100 + b200) and [62872230](https://gitlab-master.nvidia.com/dl/ai-dynamo/aic-auto-collector/-/pipelines/62872230) (h200, 8h walltime):

| system | context skip rows | generation skip rows | task failures (all classified in the pipeline error logs) |
|---|---|---|---|
| h200_sxm | +35,148 (all four gemm×kv keys) | +2,448 | 12 — guarded illegal-access at the 1M-token-prefix tier, prior rows preserved |
| h100_sxm | +32,384 | +2,448 | 24 — init-time CUDA OOM (80 GB capacity, largest task configs) |
| b200_sxm | +39,948 (adds the fp8_block skip keys) | +2,448 | 9 — same guarded 1M-prefix class as h200 |

[...]

PR #1563, opened 2026-08-18
feat(sdk): pluggable speculation module + verify-on-FPM (zero-calibration)

Ethan-ES:
---

### What

Four commits, re-authored for the current op-shell architecture (ops
self-serialize from Rust; the Python dict builders the original series
targeted are retired):

1. **feat(rust): speculative width channel for generation attention** —
   `GenerationAttentionOp.scale_num_tokens` (sequence-basis batch fold) +
   `verify_query_tokens` (ideal-peak roofline guard). Both are physical
   bounds; no fitted constants. Schema 13→14. py_ops getter/setter surface
   for the materializer.
2. **feat(sdk): pluggable speculation module** — scheme registry (ngram /
   eagle3 / dense-draft / dflash / dspark / draft-model / mtp),
   `SpeculationConfig` on `ModelConfig` (nextn>0 stays mtp sugar),
   compile-time materialization (draft ops prefixed `draft_`, width folds
   via the setter channel), byte-exact draft weight/KV accounting, CLI/task
   `--speculative` block.
3. **feat(sdk): verify-on-FPM** — hybrid shape for `forward_model="fpm"` ×
   speculation: phase lists lead with the whole-model FpmForward (decode op
   carries `verify_width`, mapping the widened verify step to its
   equivalent-AR point: tokens = c×w, total KV unchanged), materialized
   op-level draft ops ride behind. Includes the pass-baseline per-curve KV
   floor fix and the clamp-survivable draft tail. Schema 14→15.
4. **test(parity) + docs** — two append-only parity pins for the hybrid fpm
   config class; measured limitations documented in place.

### Zero-calibration contract

The bracket-blend calibration from the original series (two fitted
parameters) is **dropped, not ported**. Evidence: it fits gsm8k serving at
3.8% rms but blows up +41–45% on long-context pure-decode (out of fitted
domain). Both shipped pricing channels are physical bounds only.

### Validation (Qwen3-8B × H100 × vLLM, real-KV collection, static-batch pure-decode measurement, zero calibration)

| arm | mean\|Δ\| / max (short seq, c1–256) | long-context (isl 1k/3k) |
|---|---|---|
| AR | 1.2% / 2.4% | ±1.7% |
| EAGLE-3 (w4) | 4.7% / 8.8% | ≤3.7% |
| DFlash (w8) | 6.6% / 12.6% | +0.9~+12.0% |
| DSpark (w8) | 6.1% / 10.2% | −1.9~+9.7% |

### Known limitations (documented in code)

- agg mean-field prices every round as a mixed step: below the
  one-full-prefill-per-round workpoint it overprices +12–34% (fix
  direction: workpoint-aware agg caller — follow-up).
- Equivalent-AR row over-prices the wide-verify kernel by ≤14% in the
  mid-token × mid-KV pocket; closing it needs w>1-shaped collection rows
  (data-side follow-up).
- op-level zero-calibration under-predicts deep-c pure decode by 10–18%
  (the hole blend used to patch; structural fix is follow-up).

### Tests

505 Rust lib · 3 public-api · 4034 unit/cross-package · 449 parity (447
existing untouched + 2 new pins). Goldens untouched — answer-preserving,
verified bit-identical on the real Qwen3-8B×H100 collection at 5 spot
points pre- vs post-port.


Ethan-ES:
Force-pushed two changes:

**1. Collector/data changes removed (module boundary).** The original first commit bundled a V4-Flash w4a16 collector fix and H200 `moe_perf` data slices that the earlier validation campaign used. They are not a dependency of this module (no test or validation claim in this PR reads them), and per the repo's module-boundary rules they belong to the collector/data plane, so the history was rewritten to drop them. The data gap they closed (H200 × DeepSeek-V4-Flash × w4a16 MoE rows) is general database coverage; the patch is archived locally and can be offered separately if the data owners want it.

**2. CodeRabbit findings addressed** (commit `fix(sdk): address the CodeRabbit review findings`): identity_hash now includes `draft_model_path`; `--spec-method mtp` maps the CLI token count onto `params['depth']` instead of silently resolving to AR; the `--spec-*` flags are registered for default/recommend (their handlers already consumed them); `serving_mode='afd'` rejects scheme speculation (the AFD paths don't consume the progress profile); V4-style DSpark configs must carry `dspark_target_layer_ids`; dense draft vocab falls back to the target's instead of zeroing weight bytes; plus the input-validation batch (round-model anchors, SGLangRoundMix bounds/guards, fractional-depth rejection, NaN-proof blend params) and the two test hardenings.

One finding kept as-is deliberately: the DFlash draft/target `hidden_size` equality guard stays (mismatched-width drafts are unvalidated modeling territory), but its message no longer claims the injection GEMM requires the match.

Ethan-ES:
The four failing checks on the latest run are inherited from `main`, not introduced by this branch.

All four fail on the same two cases: `tests/cross_package/test_data_plane_baseline.py::test_table_views_match_pre_deletion_python_loaders[b200_sxm/sglang/0.5.14|0.5.16]` — the engine's `_context_dsa_module_skip_data` / `_generation_dsa_module_skip_data` views report ~2× the Python raw loaders' leaves (66950 vs 33488 / 3672 vs 1836).

Reproduced on a pristine checkout of `main` @ 08165d07 (no changes from this PR):

```
FAILED tests/cross_package/test_data_plane_baseline.py::test_table_views_match_pre_deletion_python_loaders[b200_sxm/sglang/0.5.16]
  _context_dsa_module_skip_data: state loaded/loaded, leaves 66950/33488, ordered ..., sorted-match=False
```

The regression window is exactly #1556 (`fix: add sglang 0.5.14 DSA skip-indexer perf data for h200/h100/b200`, merged after this PR's previous green run @ dbeb2822): it grows the b200/h100/h200 sglang DSA module parquets, and the engine view now sees roughly double the leaves the Python loader sees for the skip-data tables — a dedup/keying divergence between the two data planes on the new skip-indexer rows. `main`'s push CI does not run the cross_package jobs (they are PR-only), so the breakage landed silently.

This PR's previous run (same content plus the since-removed collector/data changes) was fully green on base dbeb2822.

Arsene12358:
Migrated to AISimulate

PR #1564, opened 2026-08-18
fix(data): drop superseded NVFP4 b200 skip rows (AIC-1747)

Arsene12358:
## What

Drops superseded `nvidia/GLM-5.2-NVFP4` skip-indexer rows from the B200 SGLang 0.5.14 context and generation parquet tables. Linear: AIC-1747.

## Root cause

#1556's B200 append kept the pre-existing NVFP4 skip rows alongside fresh `zai-org/GLM-5.2*` probe rows at shared loader coordinates. `DsaKey` has no model column and both loaders resolve duplicates last-row-wins, so the zai rows already won in the query plane. The raw enumeration plane introduced by #1555 deliberately retained both rows, leaving charts and support-matrix enumeration with stale duplicate measurements.

## Fix

Remove the superseded NVFP4 rows:

- context table: 115,180 → 95,203 rows
- generation table: 7,272 → 6,048 rows

The query plane remains value-identical because every removed loader coordinate is still supplied by the winning zai measurement. The raw enumeration plane no longer carries duplicate measurements of the same kernels.

## Update after main advanced

#1566 landed after this PR opened and intentionally retired `tests/cross_package/data_plane_baseline.json` and its replay tests. Conflict-resolution commit `dc785f38` merged current `main` and kept that retirement intact. The final PR diff against current `main` is now exactly the two parquet files above; the obsolete baseline refresh is no longer part of this PR.

## Verification

On the original data-fix head `26c18e4f`:

- all 26 hosted checks passed
- `tests/cross_package` with `AIC_DATA_PLANE_BASELINE_FULL=1`: 232/232
- GLM-5.2 B200 engine-step parity: 4/4 against unchanged goldens
- surviving Python DSA parsers and engine views agreed exactly

On the merged tree `dc785f38`:

- both parquet blobs are byte-for-byte identical to the approved data-fix head
- duplicate loader coordinates: 0 in both tables when model and measurement fields are excluded
- `backend_facts.py --check`: 3,143 fact slices, unchanged
- post-push hosted checks are rerunning on the merged tree


jasonqinzhou (approved):
**APPROVE · 9.6/10 · high confidence**

## What this PR does

This PR removes superseded NVFP4 GLM-5.2 skip-indexer rows from the B200 SGLang 0.5.14 context and generation parquet tables, then refreshes only the four affected 0.5.14/0.5.16 DSA skip-table baseline snapshots.

The author explained the #1556/#1555 semantic merge collision precisely and supplied concrete row-diff, loader-parity, seven-pin replay, backend-fact, and GLM-5.2 parity evidence.

## Why this score

The 9.6/10 APPROVE recommendation follows from a narrow, query-neutral cleanup with comprehensive evidence: every removed expanded loader coordinate remains with the same effective value, duplicate coordinates fall to zero, the four refreshed snapshots match the surviving Python loaders and Rust engine views, and all 26 hosted checks pass. No actionable issue was found.

jasonqinzhou (approved):
Exact-head re-approval after resolving the conflict with #1566. The merged tree keeps #1566's retired baseline deleted and leaves exactly the two previously reviewed parquet deduplication changes. Local backend-facts and duplicate-coordinate validation passed, and all hosted checks are green on this SHA.

PR #1565, opened 2026-08-18
ci: stop requiring sudo on the self-hosted macOS wheel runner

Arsene12358:
## What (fixes the other standing main-CI failure)

`Build wheels (macosx_arm64)` has failed on **every main push since #1549** (first red commit `f32b51bf`, green parent `8fe13461`). The job never reaches the wheel build: `actions/setup-python` doesn't find the newly-requested 3.13 in the self-hosted runner's toolcache, downloads it, and its macOS install script dies at `sudo installer` — the runner (`prod-aiconfigurator-default-macos-arm64-v1`) has no passwordless sudo (`sudo: a password is required`).

**Why pre-merge CI never saw it:** the PR-side `validate-platform-wheels.yml` runs the same composite action on GitHub-hosted `macos-14`, which has passwordless sudo and cached interpreters. The failure is exclusive to the self-hosted runner that only the post-merge push workflow uses.

## Fix

Gate `setup-python` to the docker path (whose host interpreter drives the Linux smoke test) and let the native-macos path provision its interpreter through **uv's managed Python** — a user-space `python-build-standalone` download, cached under `~/.local/share/uv` on the persistent runner, with `UV_PYTHON_PREFERENCE=only-managed` for hermeticity (never whatever python the host happens to carry). The macOS path already consumed the interpreter exclusively via `uv venv --python 3.13`, so nothing else changes.

Verified locally on arm64 macOS: `UV_PYTHON_PREFERENCE=only-managed uv venv --python 3.13` provisions 3.13.3/arm64 with no sudo.

## Notes

- The build workflow triggers only on main/release pushes, so the definitive green lands on this PR's merge commit; the PR-side validate workflow exercises the changed composite on GitHub-hosted macOS pre-merge.
- Infra-side alternative (pre-seeding 3.13 into the runner's toolcache or granting passwordless sudo for the installer) would work but needs repeating on every interpreter bump; this change decouples the two permanently.

## Summary by CodeRabbit

- **Chores**
  - Improved build configuration for more consistent Python environment selection across supported platforms.
  - Enhanced packaging workflows to provide more reliable wheel builds.
  - Improved cross-platform build reliability by using managed Python environments where needed and simplifying native macOS builds.


Arsene12358:
Note on the red checks: the four cross-package failures (Python 3.11–3.13 compatibility, aic-core public API contract) are inherited from main via the merge-ref — main has been red in those jobs since 08165d07 (see #1564, which fixes them and is fully green). Nothing in this PR's diff causes them; a re-run after #1564 merges clears them. Suggested merge order: #1564 first.

jasonqinzhou (approved):
**APPROVE · 9.1/10 · high confidence**

## What this PR does

This PR removes actions/setup-python from the native-macOS platform-wheel path and makes the existing uv step create the Python 3.13 venv from a uv-managed interpreter, while Docker/Linux builds retain setup-python for their host-side smoke tests.

The change is tightly scoped to the failing branch of the composite action, preserves the existing wheel build and installed-package verification flow, and documents the post-#1549 self-hosted runner failure with concrete local arm64 validation.

## Why this score

9.1/10, APPROVE. The native-macOS gate is correct, and exact-head CI proves setup-python is skipped before uv provisions managed arm64 CPython 3.13 and completes wheel build plus installed-layer verification; both Linux wheel jobs also pass. No additional review comments are recommended because CodeRabbit already posted the only minor wording issue. Exact-head Build and Test remains red on inherited b200_sxm/sglang data-plane baseline mismatches also failing current main, so this approval is not merge-readiness clearance.

PR #1569, opened 2026-08-19
ci: prune build-test — diff-scope DB sanity check, parallelize suites, cache Rust builds

tianhaox:
## What

Cuts the build-test critical path roughly in half for the common (data-free) PR. Job-level profiling of a recent green run (32211226539) showed the e2e job at 13m41s, ~5.5 min of which was `test_validate_database` re-validating **all 23 system×backend perf databases on every PR** — duplicating the diff-scoped Create Sanity Check Charts workflow (both execute `validate_database.ipynb`).

### Changes

- **Diff-scope the e2e DB sanity check.** `test_validate_database` now reads `AIC_SANITY_TARGETS` (computed by the workflow from the PR diff): changed `systems/` paths select only the affected (system, backend) combos; an always-on `h200_sxm/trtllm` sentinel keeps the loader/query smoke path and `AIC_STRICT_PROVENANCE` check covered on data-free PRs. Changes to the sanity tooling or the DB loader (`perf_database.py`) still run the full matrix; local runs without the env keep today's full-matrix behavior. Shared data with no backend path component (comm/nccl) and `<system>.yaml` spec edits conservatively expand to all backends of that system. The value is passed via step `env:` + `docker -e` (never `${{ }}`-interpolated into the run script) so fork-PR filenames cannot inject into the shell.
- **`validate_database.ipynb` cell 13:** skip the hardcoded GB200 alltoall database load when parameterized for a different combo — previously every per-system CI run paid to warm an unrelated database.
- **e2e pytest `-n 2` → `-n 4`** (hosted runners have 4 vCPUs; unit unchanged — its pytest is already only ~2 min).
- **Parity suites run with `-n 4`** (engine-step + compile-engine). Verified green under xdist locally (303 + 55 cases); every case builds its own engines. The engine-step **perf gate stays serial** — it measures p50 timings.
- **`Swatinem/rust-cache` (pinned v2.9.2)** on the parity and public-API-contract jobs. Caches registry + target dirs keyed on Cargo.lock and rustc version; correctness-neutral (a stale key is a miss, never a wrong artifact).
- **`concurrency` group:** superseded pushes to a PR cancel the in-flight run; push events group by SHA so main/release runs never cancel each other.

### Expected effect (data-free PR)

| Job | Before | After |
|---|---|---|
| Build and Test (e2e) — critical path | ~13m41s | ~6 min |
| Rust/Python engine-step parity | ~9m40s | ~5 min (less with warm rust-cache) |
| aic-core public API contract | ~4m43s | ~2.5 min with warm cache |

### Verification

- Collect-only matrix under six `AIC_SANITY_TARGETS` scenarios (unset → 28 full; `all` → 28; empty → 1 sentinel; single data path → 2; comm/nccl path → system's 3 backends + sentinel; `<system>.yaml` → same).
- Full notebook execution for the sentinel combo passes end-to-end; the cell-13 skip fires for non-GB200 runs.
- Both parity suites green locally under `-n 4`.
- Workflow YAML parse + notebook JSON/compile checks.

One thing to watch on the first CI run: e2e container memory headroom at `-n 4` (CLI subprocesses × 4 workers on a 16 GB runner). If it's tight, `-n 3` retains most of the win.

## Summary by CodeRabbit

* **Build and Validation**
  * Superseded pull-request checks are canceled automatically, while commit-specific push checks run independently.
  * Build caching and increased parallelism improve validation speed.
  * Sanity checks now target affected system and backend combinations, with broader changes triggering the full validation matrix.
  * End-to-end and parity testing support four-way parallel execution.
  * Additional performance-data and kernel manifest changes trigger full-matrix validation.

* **Bug Fixes**
  * Parameterized validation no longer loads unrelated AlltoAll data, while the default visualization remains unchanged.


tests/e2e/tools/test_sanity_check.py
```diff
@@ -16,17 +16,71 @@
...
+        backends = supported.get(system)
+        if not backends:
+            continue
+        backend = next((p for p in rest[2:] if p in backends), None)
+        if backend is not None:
+            combos.add((system, backend))
+        else:
+            # Shared data with no backend path component (e.g. comm/nccl)
+            # feeds every backend of the system.
+            combos.update((system, b) for b in backends)
```
Yiming992: This is a merge blocker because a backend-specific perf file can be consumed by other backends through manifest-gated cross-backend fill. For `gb200/linear_attention/sglang/0.5.14/gdn_perf.parquet`, this selector returns only `gb200/sglang` plus the sentinel, but live provenance shows the exact file is also loaded by latest `gb200/trtllm` and `gb200/vllm` as a `cross_backend` source.

The simplest sound fix is to validate all supported backends for any changed system data path. This still narrows the run to at most the affected system's backends instead of the full 28-combination matrix.

```suggestion
        # Backend-specific rows can feed other backends through manifest-gated
        # cross-backend fill, so validate every backend for this system.
        combos.update((system, b) for b in backends)
```

.github/workflows/build-test.yml
```diff
@@ -150,6 +162,22 @@ jobs:
           else
             echo "run_golden=false" >> "${GITHUB_OUTPUT}"
           fi
+
+          # Scope test_validate_database (e2e sanity check) to the diff:
+          # "all" when the sanity tooling / DB loader changed, otherwise the
+          # changed perf-data paths (empty -> sentinel combo only). Consumed
+          # via AIC_SANITY_TARGETS in tests/e2e/tools/test_sanity_check.py.
+          sanity_all_regex='^(tools/sanity_check/|tests/e2e/tools/test_sanity_check\.py|aic-core/src/aiconfigurator_core/sdk/perf_database\.py)'
```
Yiming992: This full-matrix matcher omits direct dependencies of `validate_database`: the compiled Rust perf-database loader/interpolator, `EngineHandle`/operation loading, and the global source manifest. Emulating the workflow for each proven path collects only the sentinel test, versus 28 with the pre-PR/unset behavior. The prediction regression gate does not replace this because it explicitly uses `shared_layer=False`.

Please include the proven dependency roots in the full trigger. The following is a minimal conservative fix; focused classification tests should pin these cases.

```suggestion
          sanity_all_regex='^('
          sanity_all_regex+='tools/sanity_check/|'
          sanity_all_regex+='tests/e2e/tools/test_sanity_check\.py|'
          sanity_all_regex+='aic-core/src/aiconfigurator_core/sdk/(perf_database\.py|engine\.py|operations/)|'
          sanity_all_regex+='aic-core/rust/aiconfigurator-core/src/perf_database/|'
          sanity_all_regex+='aic-core/src/aiconfigurator_core/systems/op_kernel_source_manifest\.yaml$'
          sanity_all_regex+=')'
```

Yiming992 (changes requested):
Requesting changes because the new diff-scoped sanity selector drops combinations that the changed data or loader can affect. This removes unique real-data, shared-layer, and strict-provenance coverage that existed before this PR.

| Finding | Current behavior | Required behavior |
|---|---|---|
| Cross-backend data dependencies | A backend-specific data path selects only that path's backend. | Select every supported backend for the system, or compute every manifest-resolved consumer. |
| Loader dependency trigger | Only the Python `sdk/perf_database.py` facade forces the full matrix. | Also force the full matrix for the global source manifest, Rust perf-database implementation, and Python engine/operation loader paths. |

Concrete evidence:

- For `gb200/linear_attention/sglang/0.5.14/gdn_perf.parquet`, the selector returns only `gb200/sglang` plus the sentinel, while live provenance shows that exact file is also loaded as a `cross_backend` source by latest `gb200/trtllm` and `gb200/vllm`.
- Emulating the workflow for `op_kernel_source_manifest.yaml`, Rust `perf_database/source_resolution.rs`, Rust `perf_database/perf_interp.rs`, or `sdk/operations/base.py` collects 1 sentinel test; the pre-PR/unset behavior collects 28.

The current green CI run does not cover these routing cases because this PR changes the sanity test and notebook themselves, which forces `AIC_SANITY_TARGETS=all`. Please add focused selector tests for a cross-backend donor, the global manifest, a Rust loader/interpolator path, and the empty/sentinel case alongside the fixes below.

tests/e2e/tools/test_sanity_check.py
tianhaox: Fixed in 2046ca243, suggestion adopted. Verified the claim against the source manifest before applying: `gdn_perf.parquet` entries are `tier: shared` with `frameworks: [sglang, trtllm, vllm]` (op_kernel_source_manifest.yaml:577-600), so the sglang-collected rows do feed trtllm/vllm on the same system — the backend narrowing was unsound. The selector now expands any changed system data path to every supported backend of that system (the `<system>.yaml` branch already did this, so the two branches also collapsed into one rule). Pinned by `test_selector_cross_backend_donor_expands_to_all_system_backends`, which uses your exact donor path.

.github/workflows/build-test.yml
tianhaox: Fixed in 2046ca243 with your trigger set adopted verbatim (sdk/perf_database.py + engine.py + operations/, aic-core/rust/aiconfigurator-core/src/perf_database/, op_kernel_source_manifest.yaml), plus one structural change: the regex now lives in `tools/sanity_check/sanity_full_matrix_triggers.txt` and the workflow `cat`s it. Reason: `.github/` is dockerignored, so classification tests inside the CI container could never pin a regex embedded in workflow YAML — `tools/` ships into the test image, so a single-source file makes the pins actually enforceable in CI. `test_full_matrix_trigger_regex_classification` covers every dependency root you listed (manifest, Rust loader/interpolator path, engine.py, operations/, perf_database.py, the tooling and the selector itself) plus negative cases (a data parquet path and `<system>.yaml` stay diff-scoped), and the empty→sentinel / unset→full / all→full behaviors are pinned separately. Also added while here: campaign-scale data PRs (>200 changed systems paths) fall back to `all` rather than forwarding thousands of paths through an env var (128KB single-string limit on Linux).

tools/sanity_check/sanity_full_matrix_triggers.txt
```diff
@@ -0,0 +1 @@
+^(tools/sanity_check/|tests/e2e/tools/test_sanity_check\.py|aic-core/src/aiconfigurator_core/sdk/(perf_database\.py|engine\.py|operations/)|aic-core/rust/aiconfigurator-core/src/perf_database/|aic-core/src/aiconfigurator_core/systems/op_kernel_source_manifest\.yaml$)
```
jasonqinzhou: **[P1] Include every compiled `EngineReference` dependency in the full-matrix trigger.** The notebook does not stop at `perf_database/`: `engine_reference.py` builds an `EngineHandle` and calls `evaluate_ops_json` / `evaluate_ops_sol_json`, which execute through native `engine/`, `operators/`, `common/`, and `py.rs`; it also imports shared Python SDK types such as `common.py`. With the current pattern, representative changes to `aic-core/rust/aiconfigurator-core/src/engine/runtime.rs`, `src/operators/attention.rs`, `src/py.rs`, `sdk/common.py`, or `sdk/system_spec.py` match neither the `all` trigger nor a systems path, so the workflow passes an empty target set and collects only `h200_sxm/trtllm` instead of the 28 database validations. The current green run cannot expose this because this PR changes the trigger file itself and therefore forces `all`. Please add the proven native engine/operator/binding and shared SDK roots (or conservatively match the whole native `src/` dependency surface) and pin representative paths in `test_full_matrix_trigger_regex_classification`.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.9/10 · high confidence**

## What this PR does

This PR shortens the build-test critical path by diff-scoping the expensive performance-database notebook sweep, increasing e2e and Rust/Python parity parallelism, caching Rust artifacts, cancelling superseded PR runs, and avoiding an unrelated GB200 all-to-all load in parameterized notebook runs. The new head also expands changed perf data to every backend of the affected system, centralizes full-matrix triggers, and adds selector regression tests.

The follow-up carefully verified the cross-backend donor behavior before broadening selection, moved the trigger pattern into a Docker-visible single-source file, and added focused tests for the exact gaps raised on the previous head.

## Why this score

The current head is 6.9/10 with REQUEST_CHANGES. Both findings from the previous packet are fixed: native perf-database/global-manifest changes now force the full matrix, and the selector policy has committed tests. One P1 coverage gap remains, however: the new dependency allowlist excludes direct inputs to the notebook's compiled EngineReference path, including Rust engine/operator/binding code and shared SDK types. Representative changes to those paths still collect only the h200_sxm/trtllm sentinel instead of the 28-database validation matrix, so a system- or backend-specific regression can leave the required e2e check green.

jasonqinzhou (approved):
APPROVE · 8.9/10 · high confidence

No P0, P1, or P2 findings remain at this exact head. The final follow-up closes the compiled EngineReference dependency gap by forcing the full database matrix for the complete native Rust source tree and the shared Python SDK types consumed by the notebook, with representative selector pins. It also tracks the current-main rename to perf_data_reuse_manifest.yaml while retaining the legacy manifest name for maintained branches.

Validation: 6 focused selector tests, strict-provenance h200_sxm/trtllm notebook smoke, Ruff/format/diff checks locally; all 19 hosted checks green on the synthetic merge of main@87caf68d1 and head@8120cf250.

tools/sanity_check/sanity_full_matrix_triggers.txt
```diff
@@ -0,0 +1 @@
+^(tools/sanity_check/|tests/e2e/tools/test_sanity_check\.py|aic-core/src/aiconfigurator_core/sdk/(common\.py|engine\.py|operations/|perf_database\.py|system_spec\.py)|aic-core/rust/aiconfigurator-core/src/|aic-core/src/aiconfigurator_core/systems/(op_kernel_source_manifest|perf_data_reuse_manifest)\.yaml$)
```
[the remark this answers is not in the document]
jasonqinzhou: Verified against both the PR head and current `main`: `aic-core/src/aiconfigurator_core/sdk/engine_reference.py` does not exist. The actual dependency is `tools/sanity_check/engine_reference.py`, which is already covered by the leading `tools/sanity_check/` trigger and by the tooling classification pin. No code change is needed for this comment.

PR #1571, opened 2026-08-19
test(sdk): rebuild the comprehensive singleton after engine-cache eviction (AIC-1777)

Arsene12358:
Fixes a latent, order-dependent unit flake on main (AIC-1777) that any PR changing the collected test count can trip — it currently blocks #1540's CI and passes on main only by xdist chunk-seating luck.

**Minimal repro (fails on clean main, serial, two tests):** `test_comprehensive_router_survives_scoped_stub_patch` followed by `TestSupportedQuantModes::test_supported_quant_modes_structure`.

**Mechanism — corrected per tianhaox's review:** the singleton's fetches never reach the engine (the conftest router answers them), so the original "engine re-fetch over an empty dir" narrative below does not apply to this database. The actual pollution path: an eviction lever fired while a scoped stub fetch is active leaves the shared per-op class caches re-populated from the stub under the singleton's cache key; the next test on the worker that touches the singleton reads stub-shaped tables. The fix records `engine._PROBE_CACHE_GENERATION` explicitly at build (the memo-based predicate from the first revision never executed — this database has no probe-spec memo — and silently degenerated to rebuild-on-every-access; see the review thread for the measured 12-vs-4 build counts).

**Original mechanism notes (kept for the general engine-eviction context, not this singleton's path):** `clear_all_op_caches()` clears the compiled-engine handle LRU and advances `engine._PROBE_CACHE_GENERATION`. The shared `comprehensive_perf_db` singleton's synthetic tables live only in the conftest's module-level router — its tmp data dir holds just the system yaml — so the next engine-backed fetch re-creates a handle over an effectively empty data dir and every table collapses (`supported_quant_mode['gemm']` degrades to `['bfloat16']`, producing the observed `assert 'fp8' in ['bfloat16']` / `assert 0 > 0` failures in whichever tests share the worker afterwards). The conftest comment's claim that routing survives cache clears held for the python fetch path but not for the engine data plane introduced by #1552/#1555/#1566.

**Fix:** the singleton accessor checks the generation recorded in the database's probe-spec memo (the module's own eviction contract) and rebuilds the singleton when a lever has fired since build. No production code touched; polluter-local restores were tried and verified NOT to work (the scoped stub fetch is still active during the polluter's `finally`, so any reload there plants stub tables under the shared class-cache key).

**Verification:** the minimal repro pair passes; polluter + all three victims pass in worst-case (adjacent serial) order; full unit suite green both `-n 2` (3710 passed) and serial (3710 passed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database reliability by detecting when cached engine state is outdated.
  * Automatically rebuilds the database connection when stale state is detected, preventing inconsistent behavior.
  * Preserved temporary database paths for the lifetime of the process, ensuring copied database instances continue to work correctly.
  * Automatically cleans up superseded temporary database directories when the process exits.

### DIFF
--- a/tests/unit/sdk/database/conftest.py
+++ b/tests/unit/sdk/database/conftest.py
@@ -1,7 +1,9 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import atexit
 import os
+import shutil
 import tempfile
 from collections import defaultdict
 from unittest.mock import patch
@@ -401,6 +403,12 @@ def _get_comprehensive_db_singleton() -> PerfDatabase:
     system_spec = cached["system_spec"]
 
     tmp_dir = tempfile.mkdtemp()
+    # Superseded singleton dirs must survive until process exit: deepcopied
+    # databases (mutable_comprehensive_perf_db) keep referencing the old
+    # systems path, and a post-eviction engine re-fetch through such a copy
+    # would turn "empty tables" into a missing-directory error if the dir
+    # were removed eagerly on rebuild. Clean them all up at exit instead.
+    atexit.register(shutil.rmtree, tmp_dir, ignore_errors=True)
     yaml_file = os.path.join(tmp_dir, "test_system.yaml")
     with open(yaml_file, "w") as f:
         yaml.dump(system_spec, f)
--- a/tests/unit/sdk/database/conftest.py
+++ b/tests/unit/sdk/database/conftest.py
@@ -369,11 +369,33 @@ def _build_comprehensive_test_data():
 _comprehensive_db_singleton: PerfDatabase | None = None
 
 
+def _singleton_survived_eviction(db: PerfDatabase) -> bool:
+    """Whether the singleton's engine-backed state is still the build-time one.
+
+    ``clear_all_op_caches()`` (and the other documented eviction levers)
+    advance ``engine._PROBE_CACHE_GENERATION`` and clear the engine-handle
+    LRU. The singleton's synthetic tables exist only in this conftest's
+    module-level router — its tmp data dir holds just the system yaml — so
+    any post-eviction re-fetch through the engine collapses the tables to
+    the near-empty on-disk state and poisons every later test on the worker
+    (order-dependent TestSupportedQuantModes / TestUpdateSupportMatrix
+    failures; see AIC-1777). The probe-spec memo stored on the database
+    carries the generation it was resolved under — a mismatch means a lever
+    fired since the build and the singleton must be rebuilt.
+    """
+    from aiconfigurator_core.sdk import engine as _engine
+
+    memo = db.__dict__.get("_table_view_probe_spec")
+    return memo is not None and memo[0] == _engine._PROBE_CACHE_GENERATION
+
+
 def _get_comprehensive_db_singleton() -> PerfDatabase:
-    """Build and cache a fully-initialized PerfDatabase singleton."""
+    """Build (or rebuild after engine-cache eviction) the shared singleton."""
     global _comprehensive_db_singleton
     if _comprehensive_db_singleton is not None:
-        return _comprehensive_db_singleton
+        if _singleton_survived_eviction(_comprehensive_db_singleton):
+            return _comprehensive_db_singleton
+        _comprehensive_db_singleton = None
 
     cached = _build_comprehensive_test_data()
     system_spec = cached["system_spec"]
--- a/tests/unit/sdk/database/conftest.py
+++ b/tests/unit/sdk/database/conftest.py
@@ -371,33 +371,37 @@ def _build_comprehensive_test_data():
 _comprehensive_db_singleton: PerfDatabase | None = None
 
 
-def _singleton_survived_eviction(db: PerfDatabase) -> bool:
-    """Whether the singleton's engine-backed state is still the build-time one.
-
-    ``clear_all_op_caches()`` (and the other documented eviction levers)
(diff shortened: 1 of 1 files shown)


tianhaox (changes requested):
Verified locally: the flake is real (the minimal repro pair fails on clean main in serial), and with this patch applied the pair passes and `tests/unit/sdk/database/` is green (366 passed). The diagnosis direction (post-eviction pollution of the shared singleton; rebuild as the cure) is right, and the rebuild path itself is sound — the op class-cache keys include `systems_root`, so a fresh `mkdtemp` per rebuild guarantees clean cache entries.

**Blocking: the detection predicate degenerates to "rebuild on every access" — the generation comparison never actually executes.**

`_table_view_probe_spec` is only ever written inside the real `fetch_table_view` (`engine_table_view.py:196`). But this conftest replaces `_etv.fetch_table_view` with `_routed_fetch` at module import, and the router answers every `system == "test_system"` fetch from `_COMPREHENSIVE_OVERRIDES` directly — the real fetch is never reached for this database. So the memo never exists on the singleton: not at build, not after `_warm_lazy_op_caches`, and not after a rebuild either.

Measured on this branch:

- After a full singleton build, `'_table_view_probe_spec' in db.__dict__` → `False`, so `_singleton_survived_eviction` → `False` — and it is still `False` immediately after a rebuild.
- Instrumented build counter, `tests/unit/sdk/database/` serial: the singleton is built **11 times** (intended: 1). Every fixture access logs `memo=None` and takes the rebuild path.

So the mechanism described in the docstring and PR body never fires; the fix works because it silently became "always rebuild", which trivially breaks the pollution chain. The verification runs (repro pass, suite green) can't distinguish the two behaviors. Consequences: the fixture's documented singleton contract ("the same object is returned to every test") is broken, each access leaks a `mkdtemp` dir and accumulates a never-evicted key in every op's class-level `_data_cache`, and — worst — the long `_singleton_survived_eviction` docstring documents a check that never runs, which a future reader will trust.

**Suggested fix (verified locally):** don't infer the generation from the database's memo — this database never fetches through the engine; that's the router's whole point. Record the generation explicitly at build time:

```python
_comprehensive_db_singleton: PerfDatabase | None = None
_comprehensive_db_generation: int | None = None


def _get_comprehensive_db_singleton() -> PerfDatabase:
    global _comprehensive_db_singleton, _comprehensive_db_generation
    from aiconfigurator_core.sdk import engine as _engine

    if (
        _comprehensive_db_singleton is not None
        and _comprehensive_db_generation == _engine._PROBE_CACHE_GENERATION
    ):
        return _comprehensive_db_singleton
    _comprehensive_db_singleton = None
    ...
    # after a successful build:
    _comprehensive_db_generation = _engine._PROBE_CACHE_GENERATION
```

This is the direct implementation of the contract this PR's description states ("rebuilds the singleton when a lever has fired since build"). The polluter (`test_data_loaders.py:421/428`) calls `clear_all_op_caches()` twice, each advancing the generation, so the next access rebuilds. Measured: the repro pair passes with exactly 2 builds (initial + post-eviction rebuild), and the full directory passes 366 with only **4 builds** — rebuilds happen only after a genuine eviction lever, and the steady-state singleton contract is preserved.

**Minor:**

- The mechanism narrative in the PR body ("the next engine-backed fetch re-creates a handle over an effectively empty data dir") doesn't hold for this singleton — its fetches are all intercepted by the router. The actual pollution path is the polluter re-warming the op class caches while the scoped stub fetch is still active (consistent with your own `finally`-section analysis). Worth correcting in the commit message / comment before merge.
- Coupling to private state is fine in test code, but coupling to the memo tuple layout (`memo[0]`) is exactly the shape that broke here; the explicit-generation variant removes that coupling too.

Arsene12358:
Both findings verified and addressed in 7a86aa21.

**Blocking finding — confirmed exactly as measured.** Reproduced your numbers before changing anything: `'_table_view_probe_spec' in singleton.__dict__` is False after a full build (the router answers this database's fetches, so the real `fetch_table_view` never writes the memo), and an instrumented build counter showed **12 builds** across `tests/unit/sdk/database` on the previous revision — rebuild-on-every-access, as you said, with the docstring describing a check that never ran. Implemented your explicit-generation variant verbatim (module-level generation recorded after a successful build; no coupling to the memo layout). Re-measured: the repro pair passes with exactly **2 builds** (initial + post-eviction rebuild), the database directory passes 366 with **4 builds**, full unit suite green at `-n 2` (3710 passed).

**Minor — mechanism narrative corrected** in the conftest comment, the commit message, and the PR body: the pollution path for this singleton is the class caches being re-populated from the scoped stub while it is active (its fetches never reach the engine); the engine-re-fetch story is kept in the body only as general eviction context, explicitly marked as not this singleton's path. Also propagating the same correction to the Linear issue (AIC-1777).

The memo-layout coupling point is well taken — the new version reads only the public-ish generation counter.

tianhaox:
Re-verified at 7a86aa213 — the blocking finding is resolved.

- The explicit `_comprehensive_db_generation` recording replaces the memo-based check entirely, and the new docstring/comment correctly documents why the probe-spec memo can never serve as the eviction signal for this database (the router answers its fetches). 
- Measured (instrumented build counter): minimal repro pair passes with exactly 2 builds (initial + post-eviction rebuild); polluter + TestSupportedQuantModes adjacent serial passes; full `tests/unit/sdk/database/` passes 366 both serial and `-n 2` with only **4 builds** — rebuilds now fire only after genuine eviction levers, and the steady-state same-object singleton contract is preserved.
- The atexit tmp-dir cleanup addresses the leak note, and deferring removal to process exit is right — deepcopied databases (`mutable_comprehensive_perf_db`) keep referencing the superseded systems path.

Nothing further from me.

tianhaox (approved):

PR #1558, opened 2026-08-17
fix(collector): replace borrowed-GB300 b300_sxm tables with true B300 silicon

Arsene12358:
## What this PR does

This PR completes the true-B300 SGLang v0.5.14 data rollout. The remaining `b300_sxm` attention, encoder-attention, quantization, and MHC tables had been copied from GB300; this replaces them with measurements from real B300 hardware without changing loader or estimator code.

Collection used manual JET campaigns on an 8-GPU `NVIDIA B300 SXM6 AC` allocation in the mirrored stock `lmsysorg/sglang:v0.5.14-cu130` amd64 image. The changes are values-only replacements at matched keys plus the documented attention-grid growth.

## Data coverage

| Table | Borrowed rows | B300 rows | Grid versus borrowed | Median latency ratio |
|---|---:|---:|---|---:|
| `context_attention_perf` | 42,262 | 44,167 | strict superset (+1,906 keys, with one documented exception) | 1.029 |
| `generation_attention_perf` | 46,935 | 49,680 | strict superset (+2,745 keys) | 1.000 |
| `encoder_attention_perf` | 7,679 | 7,679 | key-identical | 1.024 |
| `computescale_perf` | 1,628 | 1,628 | key-identical | see disclosure |
| `scale_matrix_perf` | 1,628 | 1,628 | key-identical | 1.027 |
| `mhc_module_perf` | 140 | 140 | key-identical | 1.000 |

All six shipped tables contain only `NVIDIA B300 SXM6 AC` rows and have no duplicate loader keys. The attention supersets use the existing `trtllm_mha`, `flashinfer`, and `triton` lanes, so the fail-closed backend facts registry is unchanged.

## Review follow-ups

- Re-collected the ten loader-selected attention spike shapes twice in JET pipeline 64707151 and landed the per-physical-key minimum. This changes five context rows and seven generation rows: five generation shapes were loader-selected, while two additional physical rows are FlashInfer-lane duplicates of loader keys.
- Added structured healing provenance for both attention tables: exact source ref, collector hash, filtered case-plan hash, collection date, row count/status, two collector invocations, per-key-min merge method, and pipeline/job identifiers.
- Recovered the original collector-produced attention sidecars from their persistent workdirs. The sidecars now carry the emitted 2026-08-17 dates and original case-plan hashes rather than prior recomputations.
[...]

PR #1573, opened 2026-08-19
feat(sdk): Qwen3.8-Max (Qwen/Qwen3.8-2.4T-A95B) full support at sglang 0.5.17

Arsene12358:
## What this PR does

Closes AIC-1762. Full support for **Qwen/Qwen3.8-2.4T-A95B** (open weights 2026-08-12) plus its `-FP8` sibling and the `RadixArk/Qwen3.8-2.4T-A95B-NVFP4` collector artifact: SDK modeling, bundled configs, collector cases, silicon op data on all four Blackwell systems at **sglang 0.5.17** (the day-0 release that first serves this model — v0.5.14 predates it), support-matrix membership, golden parity cases, and a new per-architecture attention-backend defaults mechanism the model needed to price its serving-true attention kernel.

The model is `Qwen3_5MoeForCausalLM` — architecturally the Qwen3.5-397B-A17B with 7 scalars changed (92 layers interval-4 hybrid: 23 full-attention 64h/4kv @ head_dim 256 with output gate + 69 GDN layers 16 K-heads/128 V-heads @128; MoE 512 experts top-10, inter 2048 + shared 2048; vocab 248,320; MTP 1; 256K context). The Qwen3.5 modeling machinery applies as-is: registering the new architecture string and bundling the released configs was sufficient — the modeling-contract test task confirmed zero changes to `qwen35.py` at the new scalars, pinning the GDN TP ladder, the gated-qkv width (34,816 at tp1), and exact KV+SSM state bytes (587,292,672 at tp1 / 73,411,584 at tp8).

## Version story: sparse 0.5.17 + sibling inheritance

Data lands under new sparse `sglang/0.5.17/` version dirs holding only this model's rows (the kda/dsv4_megamoe @0.5.16 precedent); every other op and shape inherits from 0.5.14 at query time via shared-layer sibling resolution — exercised end-to-end by the smoke task (inherited attention resolves silicon at the model's geometry; dataless systems like h200 fail with a clean `PerfDataNotAvailableError`, no nightly crash path). The collector gained **model-scoped runtime pins** in the framework manifest (this model's three ids → the digest-pinned v0.5.17 image; everything else keeps the 0.5.14 default), plus verified `__compat__` bumps for the three collectors this model exercises — including version-branched adaptations to sglang 0.5.17's kernel reorg (the fused-MoE runner-pin mechanism moved from a module global to a runtime-context flag; relocated kernel modules; a server-args access whose replacement was traced to sglang's own bootstrap pattern), each guarded by strict per-version import-surface tests that reproduce the original on-GPU failure. Lanes this model does not use (int4_wo, mxfp4 family) are fail-closed behind version guards rather than silently green-lit at the new version.

## Data (4 Blackwell systems × sglang 0.5.17, all landed with per-table provenance sidecars)

| Table | Rows/system | Coverage |
|---|---|---|
| moe_perf | 9,153 | bf16 3,159 + nvfp4 3,159 (full 117-case × 27-token grids) + fp8_block 2,835 |
| gdn_perf | 1,209 | full v:k=8 TP ladder (16,128)→(1,8), both kernel lanes incl. `flashinfer_gated_delta_rule_decode` |
| gemm_perf | 130,758 | full base grid + this model's below-grid widths (n=1, 16 @ k=8192), 4 dtypes |

Kernel-lane truth was verified on silicon and matches the cited v0.5.17 dispatch exactly: bf16/fp8_block MoE runs `sglang_fused_moe_triton` (this architecture is **not** in sglang's flashinfer_trtllm family-override allowlist — unlike the 397B), nvfp4 runs `sglang_flashinfer_trtllm_moe` (quant-keyed, architecture-independent). Cross-system nvfp4 medians vs gb300: gb200 1.007, b300_sxm 1.036, b200_sxm 1.056 — inside the established device band. The GDN FlashInfer/fla decode ratio reproduces at 0.501–0.522 (median) on all four systems — the state-bytes signature. **Documented boundary:** fp8_block has no moe_tp=32 rows — the checkpoint's 128×128 weight-block quantization cannot shard inter_size 2048 to 64-wide, so serving cannot run that configuration either (SDK: hard `PerfDataNotAvailableError` under SILICON, `empirical` fallback under HYBRID). The fail-closed op-backend facts registry gained the 36 version-keyed slices for the new data.

## Per-architecture attention-backend defaults (new mechanism, owner-designed)

sglang picks attention kernels per model: the 397B's architecture is in the SM100/103 triton-forcing allowlist; this model's is not, so serving resolves `trtllm_mha` — while the SDK's version/SM lane map (verified at 0.5.14 against the 397B) would have priced `triton`. Measured divergence grows from negligible at short sequences to **+51% TTFT at isl=131K**. This PR adds an `architectures:` section to the lane-defaults file (one entry, fully cited against v0.5.17 dispatch) with query-time precedence explicit override > architecture default > global map; the explicit `attention_backend` knob stays first-class; every other model resolves byte-identically to before (equality-tested against the pre-change resolver). Zero Rust changes — the engine already serializes the resolved lane order and Rust replays it; the 463-case parity suites are the enforcement. Post-fix, the model's TTFT error at 131K context is 0 against the forced-lane reference. Known residue tracked in **AIC-1780**: the architecture kwarg is threaded from 2 of ~11 model files (an entry for an unthreaded model would be silently inert), and the global sglang map still needs its own 0.5.17 re-verification for other models.

## Reference points (deterministic estimates, gb300, from the smoke task)

The fp8_block checkpoint needs tp=16 on gb300 (~2.4TB of fp8 weights OOM at tp=8); nvfp4 fits at tp=8. Example: nvfp4 gb300 estimate at bs=16 — TTFT 426 ms, TPOT 18.5 ms, 105.6 tok/s/gpu. Full command lines and a 2-system × 2-quant × 3-concurrency table are in the review trail.

## Cross-module declaration and dependencies

This PR intentionally carries coordinated SDK + collector + data + golden changes (the #1435/#1503/#1519 pattern). It **depends on #1519, #1533, and #1558 merging first** — the branch is currently based on an integration of those three plus main; after they land it will be rebased onto main, at which point the data sidecars' `collector_ref` fields get refreshed to the rebased SHAs and the noted squashes applied (`5a98ecc`→`bef4c39`, `1e9b97b`→`6164174`, `f5b4914`+`cdbfd10`→`6c28c47`/`a51bd04`, `08e3a17`→`a417c25`, `20cee6c`→`4648eb5`) along with a fuller body for the data commit. Opening as **draft** until then.

## Review trail

Every task passed an independent spec+quality review with re-verification (live HF config byte-comparison; blind re-derivation of all modeling arithmetic including exact state-byte totals; v0.5.17 source-trace of every backend citation — twice, by implementer and reviewer independently; mutation-tested regression guards; structural JSON diffs of every golden change), followed by a whole-branch final review on the most capable model whose two Important findings (the attention-lane divergence and the pin-scoping gap) drove the last three commits: the per-architecture defaults mechanism, the golden re-pin to the serving-true lane, and the documented operational invariant for unfiltered collector runs (no active pipeline schedules exist; all runs pass a model filter). Gates at HEAD: collector unit 847, SDK unit 2,385, both parity suites 463/463, both fail-closed data checks green, ruff clean.

## Summary by CodeRabbit

* **New Features**
  * Added attention-backend selection across CLI commands and SDK APIs, including automatic architecture- and hardware-aware defaults.
  * Added attention-lane performance resolution, fallback handling, and lane-density visibility.
  * Added support for Qwen3.8-Max model configurations, including BF16, FP8, and NVFP4 variants.
  * Added FlashInfer GDN decode modeling for supported hardware.
  * Added model-specific runtime selection for collection workflows.

* **Bug Fixes**
  * Improved attention performance lookup accuracy and fallback behavior.
  * Corrected GDN state-memory and latency modeling for FlashInfer decoding.

* **Documentation**
  * Documented the new attention-backend option and updated schema guidance.


Arsene12358:
This branch has been rebuilt onto a fresh integration base (main @ 899034f5 + the rebased heads of #1519, #1533 and #1558) after all three dependencies were themselves rebased; the previous head is superseded and the branch was force-pushed. The rebuild required re-engineering Task 9 (the per-architecture attention-backend default) rather than replaying it: in the post-pyo3-unification world attention ops are constructed by the model layer without a database and carry no `architecture=` constructor kwarg, so the architecture is now read centrally off `model.architecture` in `engine.py::build_engine_spec_json` — exactly where `attention_backend` is already read — and threaded through the single existing resolution choke point. This is a structural simplification, not a workaround: every model family now gets correct per-architecture resolution with no per-model-file threading at all, and no model file is touched by this PR.

Two defects that git's 3-way merge introduced without conflict markers were caught by running the gates rather than trusting the rebase, and are fixed here: a dead test asserting a now-removed `op._architecture` attribute (removed; its intent is covered more strongly by a new test asserting the observable lane on both attention ops, for Qwen3.8-Max and the 397B sibling, against the real shipped gb300/sglang/0.5.17 database), and a stale golden refresh that carried pre-rebuild numeric values onto a base that now also has #1519's density-ranked donor tier (re-pinned via `pin_goldens.py --refresh` for the only two affected records; all six deltas are sub-3%, largest −2.91%).

One further fix landed in this rebuild: the `RadixArk/Qwen3.8-2.4T-A95B-NVFP4` MoE row carried the same `frameworks: [sglang]` quant-gate inversion that #1519's rebase review found on the 397B NVFP4 row — the framework mismatch makes the case generator skip the row entirely instead of narrowing it, inverting the gate to allow-everything on trtllm and losing legitimate cases on vllm. Because this artifact's trtllm/vllm support is not verified at any pinned version (tracked in AIC-1782), the fix follows the Nemotron `allowed_modes: []` precedent rather than #1519's restore-to-`[nvfp4]`, and ships a per-backend-explicit guard test.

Verification of the re-engineering, beyond the full gate suite (cargo 512, unit 4052, parity 463, cross-package 89, collector 882, validate_database 12/12, CLI smoke pricing the serving-true lanes): an 18,144-combination differential sweep of the lane resolver against the base returned zero divergence outside the intended blast radius (exactly 12 combinations: sglang, version ≥ 0.5.17, SM100/103, no override — this model's architecture only), and mutating the central architecture read to `None` fails exactly this branch's own 4 parity records with the other 459 untouched.

Housekeeping: AIC-1780's "central architecture threading" half is structurally moot under this design (there is no per-model threading left to be incomplete) — it will be retargeted to its remaining half (the global sglang lane map still lacks a verified 0.5.17 entry for other models). Note also that the new sglang 0.5.17 gemm grid becomes the "latest sglang" reference for the cross-backend sanity report, re-anchoring 5 pre-existing fp8_block SUSPECT lines from 0.5.14 to 0.5.17 — joined on identical shapes the two grids agree at median ratio 0.991–1.001 on every system and dtype, so this is reference-motion, not new data drift.

Arsene12358:
Pushed `398cb9b4` (fast-forward on `86f07df2`): fixes the CI failure in run 32436997468 — `test_collector_import_surface.py`'s red-proof fetched the pre-fix collector via `git show f5b49140:…`, and that commit is a dangling object after the base-branch rebase, so fresh clones die with exit 128. The pre-fix snippet (collect_moe.py lines 60–86 @f5b49140) is now embedded verbatim in the test with provenance comments, all git/subprocess plumbing removed; the both-direction proof (pre-fix rejected by the 0.5.17 import-surface fake, current module accepted) is unchanged. Collector suite green with `GIT_DIR` pointed at an empty repo (827 passed / 6 skipped). The stacked vLLM branch (AIC-1782) carries the same fix for its own red-proofs, which pinned this branch's head the same way.

PR #1574, opened 2026-08-19
fix(data): recover Lightning and DeepSeek-V4 NVFP4 recipes

Arsene12358:
## What

Recovers the vLLM 0.24.0 NVFP4 MoE coverage needed by AIC-1743 for `nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4` (AIC-1748) and the `nvidia/DeepSeek-V4-{Flash,Pro}-NVFP4` ModelOpt exports (AIC-1749) on B200 SXM and GB200. The branch contains both the collected silicon rows and the vLLM-specific inferred-mode resolution needed for the recipe paths to consume them.

## Data

| Model | Rows per system | Coverage |
| --- | ---: | --- |
| Nemotron-3.5-Lightning | +1,053 | 27 token points × 13 TP/EP layouts × 3 routing distributions |
| DeepSeek-V4-Flash | +1,134 | 27 token points × 14 TP/EP layouts × 3 routing distributions |
| DeepSeek-V4-Pro | +972 | 27 token points × 12 collected TP/EP layouts × 3 routing distributions |
| **Total** | **+3,159** | **+6,318 rows across both systems** |

The merge is append-only: all 6,318 new identity keys are disjoint from the existing tables, with zero removals, collisions, or changed pre-existing values. Every B200 latency is the exact per-key minimum across its independent source runs, including exclusive-node reruns used to remove one-sided scheduler spikes; every GB200 slice is value-identical to its source job. The post-merge isolation sweep found no remaining isolated latency spike above 3× its curve neighbors.

The DeepSeek-V4-Pro source jobs were partial, not zero-failure: three generated TP32/EP1 cases per system—one for each routing distribution—failed FlashInfer's `M % 128 == 0` assertion. Each case spans 27 token points, so 81 TP32 rows per system are intentionally absent. TP32 is waived from the AIC-1743 recipe-recovery scope; hardware-aware case filtering is tracked by [AIC-1883](https://linear.app/nvidia/issue/AIC-1883/make-moe-tensor-parallel-eligibility-hardware-aware-across-collection).

Collection provenance, source jobs, runtime image (`vllm/vllm-openai:v0.24.0-cu129`), collector revision, and the B200 rerun/minimum policy are recorded in each system's `collection_meta.yaml`.

## Mode resolution

For vLLM only, an inferred ModelOpt `w4a16_nvfp4` storage label now resolves to the native `nvfp4` execution lane used by compressed-tensors W4A4 kernels. The Task/default path performs this mapping at the Hugging Face configuration boundary. Direct `cli estimate` paths—aggregate, static, disaggregated, and AFD—now carry backend context into system resolution so Hopper selects `nvfp4_wo` before model operations are constructed, avoiding an incorrect native-FP4 FLOPS lookup.

Explicit quantization modes remain authoritative, Blackwell continues to use native `nvfp4`, and TensorRT-LLM and SGLang behavior is unchanged.

## Verification

- Full supported unit run: 4,040 passed, 10 skipped.
- Focused resolution coverage: 66 NVFP4 model tests, 106 Task v2 tests, 21 CLI API tests, and the direct Hopper pre-construction regression all pass.
- All aggregate, static, disaggregated, and AFD estimate paths reach model construction with `nvfp4_wo` on H100 for the Lightning checkpoint.
- Ruff check, Ruff format verification, and `git diff --check` pass.
- Append-only audit: +6,318/−0/~0 rows, zero identity-key collisions, B200 values equal cross-run minima, GB200 values equal source-job values, and no isolated >3× spikes.


aic-core/src/aiconfigurator_core/systems/data/b200_sxm/moe/vllm/0.24.0/collection_meta.yaml
```diff
@@ -8,6 +8,14 @@ runtime:
   version: 0.24.0
 tables:
   moe_perf:
+    # 2026-08-19: appended 1,053 Nemotron-3.5-Lightning-30B-A3B-NVFP4 nvfp4
```
jasonqinzhou, line 36: Could we re-run or exclude the three B200 cases that the repo's own curve check identifies as isolated spikes in this newly appended slice? The committed values are: `(moe_tp=1, moe_ep=2, power_law_1.2, tokens=192) = 0.3671 ms` versus 0.0869/0.0947 ms at 128/256 tokens and 0.0830 ms for the same GB200 key; `(1, 16, balanced, 384) = 0.6420 ms` versus 0.0726/0.0716 ms at 320/512 and 0.0687 ms on GB200; and `(8, 1, power_law_1.01, 320) = 1.4060 ms` versus 0.0607/0.0617 ms at 256/384 and 0.0623 ms on GB200. That is 3.9x, 8.8x, and 22.8x above the larger adjacent value. These exact points feed SILICON interpolation, so they can also contaminate nearby token counts. Please replace them with clean reruns, or remove them until they can be recollected, before landing the table.

jasonqinzhou (changes requested):

## What this PR does

This PR adds 1,053 Nemotron-3.5-Lightning NVFP4 MoE silicon measurements for each of B200 and GB200 under vLLM 0.24.0, together with collection provenance, so the checkpoint can resolve in SILICON estimates on both systems.

The append-only/key-collision checks, exact pipeline and job provenance, pinned collector and image versions, and two-system end-to-end estimate receipts make this data contribution unusually easy to audit.

## Why this score

The packet scores the PR 6.0/10 and recommends REQUEST_CHANGES because three newly added B200 measurements are isolated latency spikes of 3.9x, 8.8x, and 22.8x versus their adjacent token points, while the same keys are normal on GB200. These committed values are consumed directly by SILICON interpolation and can materially overstate MoE latency. The remaining append, schema, registry, model-plan, and end-to-end checks passed.

aic-core/src/aiconfigurator_core/systems/data/b200_sxm/moe/vllm/0.24.0/collection_meta.yaml
Arsene12358: Fixed in 8729cb61, and your three keys now read 0.0884 / 0.0751 / 0.0656 ms — in-family with their neighbors. What I did, since the finding generalized: (1) ran the same isolation sweep over every slice both data PRs append — it reproduced your 3 and found 14 more in the DSV4 b200 slices (#1577), with gb200 clean everywhere, pointing at computelab shared-node noise; (2) re-collected all three b200 slices on exclusive nodes (harness branch now passes `--exclusive` for computelab timing jobs) — interestingly the exclusive runs showed their own isolated spikes at different keys, so single-run replacement just moves the lottery; (3) merged per-key MINIMUM across the two independent runs — latency noise is one-sided, so the cross-run min removes spikes from either run. Post-merge the isolation sweep is clean on all five system/model slices (569 of 1,053 Lightning values lowered; 558/1,134 Flash and 295/972 Pro in #1577), row counts and key sets unchanged, and the merge rule is documented in the collection_meta.yaml notes with both jobs' ids. E2E receipts re-verified on the rebuilt trees.

Arsene12358:
**Heads-up found while re-verifying after the rebase onto current main: the b28ab8ff wave silently changed which MoE rows this checkpoint queries.** The new `w4a16_nvfp4` alias in `_normalize_mixed_precision_layer_algo` now matches Lightning's ModelOpt expert label (`W4A16_NVFP4`), so on current main the checkpoint resolves `moe=w4a16_nvfp4` and the sweep asks for `MoeKey { quant: "w4a16_nvfp4", distribution: "uniform", ... }` — a key no collector emits — instead of the `nvfp4` rows this PR adds. Three data points suggest `nvfp4` is the intended mode for this checkpoint on vLLM: the merged case plan (#1541) declares `[nvfp4]` for it (pinned by `test_model_cases.py`), the actually-dispatched vLLM 0.24.0 kernel in collection is the FlashInfer w4a4 path (`vllm_compressedtensorsw4a4nvfp4moe_flashinfer_trtllm_trtllmnvfp4expertsmonolithic` — the topk<=10 FlashInfer NVFP4 path, i.e. activations are quantized at run time even though the checkpoint label says W4A16), and the wave's own variant pin table (`test_nvfp4_variants.py`) covers Nemotron-3-Nano-NVFP4 (`moe=nvfp4`) but does not pin Lightning-3.5, so the flip looks like an unintended side effect of the global alias rather than a decision. Nothing in CI catches it because no test exercises Lightning's end-to-end resolution on vLLM. If you agree `nvfp4` is right here, I'll cut a small follow-up scoping the alias (or adding the Lightning pin) so this PR's rows are actually queried; if W4A16-weight-only modeling is the intent for this checkpoint, then the case plan and this collection need to move instead — but there is currently no collector path that emits `w4a16_nvfp4` MoE rows, so that direction needs new collector work. @jasonqinzhou

Arsene12358:
Update on the `w4a16_nvfp4` question above: **the scoping fix is now folded into this PR** (commit 461f4388) rather than waiting, because the evidence made the intent question one-sided — a repo-wide scan shows **zero** `w4a16_nvfp4` rows in any MoE or GEMM table on any system/backend, and no collector has a code path that can produce w4a16 MoE rows, so the alias was routing this checkpoint (on every backend) into a lane that is empty and un-fillable. The fix is deliberately narrow: `_apply_model_quant_defaults` remaps an **inferred** `w4a16_nvfp4` MoE mode to `nvfp4` **on vLLM only** — vLLM's compressed-tensors dispatch quantizes activations at run time and serves the w4a4 FP4-tensor-core kernels regardless of the storage label (the collected `kernel_source` is `vllm_compressedtensorsw4a4nvfp4moe_*`). trtllm/sglang keep the label mode (the weight-only dequant-to-BF16 path this alias was added for, per the Qwen3.6 profile), and an explicit user mode always wins. Per-backend behavior is pinned in `test_nvfp4_variants.py`, and the recipe-point receipt is live again on this branch: 3,325 tok/s/GPU on b200 where main fails with `MoE data missing`. Full unit suite 4,133 passed. One open question stays with you: Lightning on **trtllm** previously rode sibling-inherited nvfp4 rows and now resolves w4a16_nvfp4 (empty lane) — I left trtllm on the label mode since that's your wave's explicit direction, but flagging that its Lightning coverage regressed with the alias and needs either the same treatment or a w4a16 collection story.

aic-core/src/aiconfigurator_core/sdk/models/helpers.py
```diff
@@ -720,6 +721,19 @@ def _apply_model_quant_defaults(
...
+    # quantizes activations at run time and serves the FP4-tensor-core (w4a4)
+    # kernels for NVFP4-weight experts regardless of the ModelOpt storage
+    # label — ``W4A16_NVFP4`` describes the artifact, not the dispatch (the
+    # collected kernel_source is ``vllm_compressedtensorsw4a4nvfp4moe_*``, and
+    # no collector emits w4a16_nvfp4 MoE rows on any backend). Remap the
+    # label-inferred mode to the lane vLLM actually executes; trtllm/sglang
+    # keep the label mode (their weight-only dequant-to-BF16 MoE path is
+    # real, e.g. Qwen3.6). Applies only when the mode was inferred here — an
+    # explicit user mode wins (AIC-1748/AIC-1743).
+    if moe_was_unset and backend_name == "vllm" and model_config.moe_quant_mode == common.MoEQuantMode.w4a16_nvfp4:
```
jasonqinzhou: Thanks for scoping this remap to inferred vLLM values; it fixes the direct `cli estimate` receipt on exact head `461f438822813989535cd110b47984dbc492d78f` (3,325.15 tokens/s/GPU on B200 and 3,413.96 on GB200). The public Task V2/`cli default` path still does not reach this branch, though: `Task._resolve_quant_modes()` materializes the HF-derived `w4a16_nvfp4` mode before `get_model()`, so `moe_was_unset` is false here. At the same ISL 65,536 / OSL 400 / prefix 58,982 SILICON workload, both B200 and GB200 `cli default` runs fail with `Unsupported moe quant mode 'w4a16_nvfp4'`; that CLI also exposes no MoE-mode override. Could you apply the same backend-aware mapping at Task's HF-resolution layer while preserving truly explicit fields, and add a Task/`cli default` regression that reaches these B200 and GB200 rows? As written, the recipe/sweep path still cannot consume the data.

jasonqinzhou (changes requested):
**REQUEST_CHANGES · 6.8/10 · high confidence**

## What this PR does

This PR adds 1,053 de-noised Nemotron-3.5-Lightning NVFP4 MoE silicon measurements for each of B200 and GB200 under vLLM 0.24.0, together with collection provenance, and now attempts to map the checkpoint's W4A16_NVFP4 storage label to vLLM's actual nvfp4 execution lane.

The two-run B200 de-noising and detailed provenance remain unusually auditable, and the new commit thoughtfully scopes the remap to inferred vLLM modes while adding explicit-override and cross-backend tests.

## Why this score

6.8/10, REQUEST_CHANGES. The new remap fixes the previous no-override failure for the direct cli estimate path: exact recipe-point runs now reach the nvfp4 rows on B200 and GB200. One P1 integration gap remains in the public Task V2/cli default path, which resolves the HF mode before this helper, keeps w4a16_nvfp4, and rejects it as unsupported on both systems; because that is the recipe and sweep path and the CLI exposes no MoE-mode override, the PR is not yet safe to merge.

aic-core/src/aiconfigurator_core/sdk/models/helpers.py
Arsene12358: Fixed in daf2b40a — the remap is now a shared helper applied at both sites: get_model's inferred-mode path and `Task._resolve_quant_modes()`'s HF-base layer (adjusting `from_hf` before the explicit-wins check, same pattern as the DSV4/Kimi arch-mode adjustments there, so an explicit field still overrides and validate still fails fast on an explicit `w4a16_nvfp4`). Reproduced your failure first ('Unsupported moe quant mode' on both systems via `cli default`), and the receipt now holds on the recipe path: `cli default` at ISL 65,536 / OSL 400 / prefix 58,982, SILICON, 8 GPUs on b200 sweeps to **4,513 tok/s/GPU (agg)**. One scoping refinement your suite forced: the first cut remapped backend-wide, which broke `test_validate_w4a16_nvfp4_moe_xprofile_reachable_in_hybrid` — Qwen3.6-on-vLLM is deliberately on the weight-only XPROFILE lane — so the remap is scoped to `NemotronHForCausalLM` on vLLM only, where the collected kernel_source (`vllm_compressedtensorsw4a4nvfp4moe_*`) proves the w4a4 dispatch. Task/cli-default regressions added (vllm→nvfp4, trtllm keeps the label mode, explicit w4a16 preserved); full suite 4,136 passed.

Arsene12358:
@jasonqinzhou The outstanding review items on #1574 are addressed and the branch is ready for re-review:

- The previously flagged B200 latency spikes are removed by the committed per-key minimum across independent runs, with a clean post-merge isolation sweep.
- The Task/default vLLM path resolves Lightning's inferred ModelOpt label to the collected `nvfp4` lane (`daf2b40a`).
- Direct aggregate, static, disaggregated, and AFD estimates now resolve that execution lane before the Hopper fallback, so H100 constructs the model with `nvfp4_wo` rather than attempting native-FP4 FLOPS (`d7fcc3f2`).
- The DSV4-Pro provenance now accurately records the three waived TP32/EP1 failures; TP32 is not required for AIC-1743 and hardware-aware filtering is tracked by AIC-1883 (`499b16f9`).
- The title/body now describe the complete stacked payload: +3,159 rows per system (+6,318 total).

Local verification is green: 4,040 unit tests passed (10 skipped), focused NVFP4/Task/CLI regressions passed, Ruff/formatting passed, and the append-only data audit found zero collisions, removals, or changed pre-existing rows. A final independent review found 0 Critical and 0 Important issues. CI is running on the new head.

jasonqinzhou (approved):
**APPROVE · 9.5/10 · medium confidence**

## What this PR does

This PR adds de-noised Nemotron-3.5-Lightning and DeepSeek-V4 NVFP4 MoE silicon rows for B200 and GB200 under vLLM 0.24.0, maps Lightning's inferred W4A16_NVFP4 storage label onto the execution lane each supported path actually uses, and makes the bundled Lightning configuration available offline.

The update is especially thoughtful in three places: it replaces source-inspection coverage with a behavioral Hopper regression, corrects the partial DeepSeek-V4-Pro collection provenance instead of overstating success, and resolves the parquet merge against current main without changing or removing any pre-existing row.

## Why this score

9.5/10 technical score, BLOCKED packet. The independent Codex pass found no actionable current-head issue: the changed paths, row-level merge, formatting, full-PR diff checks, and exact-head hosted gates are sound. The canonical recommendation is nevertheless BLOCKED because the required helper requested claude-fable-5 but observed only claude-haiku-4-5-20251001 and claude-sonnet-4-6. The helper therefore marked the Fable lane unavailable, and its two proposed findings concern main-branch files outside this PR's current-base diff; under the skill contract that output cannot be represented as a Fable review or used to recommend comments.

PR #1578, opened 2026-08-20
fix(cli): resolve cross-node EP through DeepEP data

Harrilee:
## Summary

- share one data-driven MoE communication selection policy between static estimates and `Task.build_model_config`
- prefer exact `(moe_ep_size, node_num)` A2A coverage
- preserve the SGLang DeepEP node-1 substitution established in #1314: when an exact multi-node scale is absent, the canonical legacy `(ep=8, node_num=1)` row for the same model shape may serve the request, with `estimated` provenance and no latency scaling
- keep other frameworks and communication backends exact-scale only
- raise `PerfDataNotAvailableError` when a cross-node tuple has neither exact coverage nor the supported SGLang DeepEP node-1 substitute
- derive the cross-node boundary from the system spec: GB200/GB300 use 4 GPUs per node; B200/B300/H100/H200 use 8

## Why

The static CLI built `ModelConfig` directly and could leave a cross-node MoE request on the fused graph, returning a plausible latency without pricing any A2A operation. The shared resolver now prevents that fused-MoE false positive.

This PR does **not** introduce an NCCL fallback. It also does not change the marked substitution policy merged in #1314:

1. exact multi-node DeepEP measurement is preferred and remains `silicon`;
2. SGLang DeepEP may use the same-shape legacy node-1 measurement when exact scale data is absent;
3. that substitution is `estimated`, optimistic, and is not multiplied by an invented scaling factor;
4. the request fails when neither source is available.

The Python coverage gate and Rust runtime now use the same canonical legacy node-1 coordinate.

## Review follow-ups

- restrict the WideEP MLA `trtllm_mla` compatibility slice to supported `flashinfer` / `fa3` requests; invalid and empty kernel sources remain typed misses
- reuse Task's cached per-role large-EP coverage in the shared resolver instead of probing A2A and expert-compute coverage twice
- refresh the three deliberately changed frozen static parity records through `pin_goldens.py --refresh`, including their Rust/Git provenance

## Validation

- complete Rust engine-step parity: **377 passed**
- targeted Python Cross-node EP coverage tests: **39 passed**
- broader CLI / Task / large-EP tests: **149 passed**; 2 unrelated DSV4 cases require unavailable Hugging Face downloads
- Rust WideEP MLA tests: **7 passed**
- Rust MoE A2A tests: **14 passed**
- Ruff and Rust formatting checks passed
- H200 + SGLang + DeepSeek-R1 + EP32 resolves exact DeepEP coverage
- GB200 + SGLang + DeepSeek-R1 + EP32 succeeds through the #1314 node-1 DeepEP substitution; the final local validation reported TPOT **27.940 ms** and the substituted A2A operation carries `estimated` provenance


src/aiconfigurator/sdk/moe_comm_resolver.py
```diff
@@ -0,0 +1,178 @@
...
+    communication backends remain exact-scale only.
+    """
+    if (moe_ep_size, expected_nodes) in pairs:
+        return True
+    return (
+        framework == "sglang"
+        and comm_backend in _DEEPEP_NODE1_FALLBACK_BACKENDS
+        and expected_nodes > 1
+        and any(node_num == 1 for _ep_size, node_num in pairs)
+    )
```
Arsene12358, line 53: **Blocker — require exact DeepEP coverage for cross-node EP.** The PR says it selects DeepEP when exact A2A coverage exists and raises when an exact cross-node topology is missing, but this branch treats any same-shape node-1 row as coverage for every multi-node request. On the current head, GB200/SGLang/DeepSeek-R1/EP32 needs `(32, 8)`, while the database exposes only legacy `(8, 1)` data; this branch accepts it and the CLI returns TPOT 18.411 ms with estimated provenance instead of raising. It also creates a planner/runtime mismatch: Python accepts any node-1 EP row, while Rust always looks up legacy `(8, 1)`. Per the intended production policy, cross-node EP should require corresponding DeepEP data; neither NCCL AG/RS nor node-1 DeepEP extrapolation is an acceptable substitute. Please require `(moe_ep_size, expected_nodes)` here, remove the node-1 fallback in `operators/moe_a2a.rs`, and update the tests that currently bless the substitution.

aic-core/rust/aiconfigurator-core/parity_tests/goldens/engine_step.json
```diff
@@ -49,11 +49,7 @@
         }
       },
       "static": {
-        "values": {
-          "static_ctx": 115.752,
-          "static_gen": 43.269,
-          "static_total": 159.021
-        }
+        "error": "PerfDataNotAvailableError"
```
Arsene12358, line 52: **Blocker — preserve frozen-golden provenance.** This PR rewrites the frozen Python-era `::static` records for `deepseek-r1-gb200-vllm-014-scan-coverage`, `deepseek-v3-gb200-vllm-014-scan-coverage`, and `kimi-k25-gb300-vllm-019-scan-coverage` from numeric values to `PerfDataNotAvailableError`, but none of those keys was added to `post_freeze_pins`. `parity_tests/README.md` says frozen values must not be edited directly and deliberate modeling changes must use `pin_goldens.py --refresh`, which records the Rust engine and Git head. Please finalize the intended behavior, then refresh all three keys through the documented script or restore the frozen values. Green parity CI cannot validate this provenance because the expected oracle itself was rewritten.

aic-core/rust/aiconfigurator-core/src/perf_database/wideep_mla.rs
```diff
@@ -99,6 +99,39 @@ pub struct GenerationKey {
...
 
+fn resolve_context_key(grids: &WideEpContextMlaGrids, requested: ContextKey) -> ContextKey {
+    if grids.by_keys.contains_key(&requested) {
+        return requested;
+    }
+    let fallback = ContextKey {
+        kernel_source: "trtllm_mla".to_string(),
+        ..requested.clone()
+    };
+    if grids.by_keys.contains_key(&fallback) {
```
Arsene12358, line 113: **Important — restrict this fallback to supported attention backends.** `resolve_context_key` and `resolve_generation_key` currently retry `trtllm_mla` for every missing `kernel_source`. In EMPIRICAL mode the silicon-path backend validation is not run, so invalid inputs such as `torch` or an empty string now return valid estimates from the `trtllm_mla` slice; I reproduced 5.1478 ms for context and 0.1166 ms for generation, identical to the intended fallback. Preserve exact-key precedence, allow the compatibility fallback only for the supported `flashinfer`/`fa3` inputs, and keep the typed miss for unknown values. Please add tests for allowed fallback, exact-key precedence, both slices missing, and invalid/empty inputs.

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -2023,9 +2044,20 @@ def build_model_config(
             # Hardware fact, injected alongside the comm backend: the large-EP
             # ops take the comm node span at construction and would otherwise
             # have no channel to it (models.helpers.large_ep_gpus_per_node).
-            num_gpus_per_node=self._num_gpus_per_node(role),
+            num_gpus_per_node=num_gpus_per_node,
         )
         model_config._gemm_quant_mode_is_explicit = self._gemm_quant_mode_explicit_by_role.get(role, False)
+        if parallel is not None:
+            required_phases = tuple(dict.fromkeys((*self._role_phases(role), *self._required_large_ep_phases(role))))
+            resolve_model_config_moe_comm(
```
Arsene12358, line 2057: **Minor — `Task` now resolves the same tuple twice.** The `ModelConfig` constructor immediately above still seeds `moe_comm_backend` through `_resolve_moe_comm_backend`, then this added call independently probes A2A and expert-compute coverage again. On a real DeepSeek-R1/H200/SGLang task with 315 aggregate tuples, model-config construction repeated 50 A2A coverage walks and 100 expert-compute walks after the Task cache had already resolved coverage; the measured cost was about 4.3 ms, so this is not a performance blocker, but it leaves two policy seams that can drift. Please collapse Task onto the shared resolver and pass it the cached per-role coverage snapshot.

Arsene12358 (commented):
Review of `36fb28fa`. I found two blockers: the resolver does not enforce the PR's exact-topology contract for cross-node EP, and three frozen parity records were rewritten without the required post-freeze pin provenance. I also left one important correctness finding and one minor design finding. I am submitting this as a comment review, without changing the formal approval state.

src/aiconfigurator/sdk/moe_comm_resolver.py
Harrilee: Thanks — I agree that the PR description stated the wrong contract, and I also agree that Python accepting an arbitrary node-1 EP coordinate while Rust used the canonical legacy `(8, 1)` coordinate left room for drift. Both are fixed in `0749ab9c`: the PR description now documents exact-first resolution, and Python/Rust both require the canonical legacy `(8, 1)` coordinate for the fallback.

I am intentionally retaining the SGLang DeepEP node-1 substitution because that policy was explicitly established and merged in #1314: exact node-scale data is preferred; otherwise the same-shape node-1 DeepEP measurement is used with `estimated` provenance and no scaling. This is not an NCCL AG/RS fallback. If neither exact coverage nor that supported SGLang DeepEP node-1 row exists, the cross-node request still raises `PerfDataNotAvailableError`.

Removing the substitution here would regress #1314, so I think changing that production policy should be handled as a separate decision rather than implicitly reversed in this fix.

aic-core/rust/aiconfigurator-core/parity_tests/goldens/engine_step.json
Harrilee: Agreed. Fixed in `0749ab9c`. I refreshed exactly these three `::static` records through the documented `pin_goldens.py --refresh` workflow. Their `post_freeze_pins` entries now record `engine: rust` and the modeling commit SHA `36fb28fa`.

I also rebuilt the branch Rust extension and ran the complete engine-step parity suite: 377 passed.

aic-core/rust/aiconfigurator-core/src/perf_database/wideep_mla.rs
Harrilee: Agreed. Fixed in `0749ab9c`. Exact-key precedence is unchanged, and the `trtllm_mla` compatibility lookup now runs only when the requested source is `flashinfer` or `fa3`. Unknown and empty sources preserve the requested key and therefore reach the existing typed-miss path.

Added Rust coverage for all four cases you listed: supported fallback, exact-key precedence, both slices missing, and invalid/empty inputs. The focused WideEP MLA suite passes 7/7.

src/aiconfigurator/sdk/task_v2.py
Harrilee: Agreed. Fixed in `0749ab9c`. `ModelConfig` no longer seeds `moe_comm_backend` through a first per-tuple resolution. `build_model_config` now passes Task's cached per-role large-EP coverage snapshot into the shared resolver, and the remaining Task selection helper uses the same shared backend-selection function.

I added a regression test that warms the Task cache, replaces both database coverage probes with failures, and verifies that `build_model_config` succeeds without calling either probe. The targeted Cross-node EP suite passes 39/39.

Arsene12358:
Non-blocking provenance request for the retained SGLang DeepEP node-1 fallback.

After discussing the intended compatibility policy, I agree with retaining the EP8/node1 substitution from #1314 as an estimated result. The remaining issue is that users cannot determine from a particular estimate which measurement was substituted.

| Current behavior | Expected behavior |
|---|---|
| The default CLI emits no fallback notice. `--detail source` labels the context/generation MoE dispatch and combine operations only as `estimated`. | Make the topology substitution explicit, for example: `Estimated DeepEP A2A: requested EP32/node8; using EP8/node1 silicon data`. |
| Rust selects the donor `(EP8, node1)` locally, then returns only latency, energy, and the generic source tag. The SDK therefore has no donor coordinate, and `EngineHandle.last_provenance()` remains `None`. | Preserve structured fallback provenance containing both the requested and donor EP/node coordinates, and expose it through the SDK and CLI source detail. |

On the current head, I reproduced an EP32 request on GB200, which resolves to node8, while the available DeepEP coverage contained only `(EP8, node1)`. The estimate succeeded and its MoE communication operations were tagged `estimated`, but neither the CLI nor the result object identified EP8/node1 as the donor. An expert could inspect database coverage separately and infer the substitution, but that provenance is not attached to the result.

Please expose the requested and donor coordinates whenever this fallback fires and add a regression test covering that user-visible provenance. A stable CLI warning would be a reasonable minimum if extending the result schema is out of scope for this PR.

Arsene12358:
Implemented the requested follow-up in `f00f693c`.

- The canonical SGLang DeepEP EP8/node1 fallback remains supported for multi-node requests, with exact-topology data still taking precedence and missing donor data still raising.
- Fallback provenance now comes from the executed Rust lookup after the donor query succeeds. Results expose inference phase, backend, requested EP/nodes, and measurement EP/nodes through the SDK, default CLI warnings, `--detail source`, Task aggregate/disaggregate/sweep outputs, and saved top-N provenance artifacts. The affected operations remain tagged `estimated`.
- Public Rust and Python per-op APIs remain four-tuples; metadata uses private same-call paths, so provenance does not duplicate the performance query.
- `--detail time` and `--detail all` now preserve a successful primary estimate when only the optional SOL comparison is unavailable, while primary errors, explicit SOL errors, plain validation errors, and unrelated runtime errors remain fatal.

Real shipped-data verification for DeepSeek-R1 on GB200/SGLang 0.5.12 at EP32/node8 reports context DeepEP-HT and generation DeepEP-LL using EP8/node1 measurement data. Default, source, time, and all detail modes exit successfully with identical primary estimates; time/all report the unavailable SOL comparison separately.

Local validation is green: 3,837 Python unit tests, 396 focused Python tests, 525 Rust tests, 377 engine parity cases, 70 compile parity cases, downstream Rust public-API checks, installed CLI E2E, Ruff, Rustfmt, DCO, and two independent whole-branch review axes. GitHub CI is now running on the pushed head.

Arsene12358 (approved):
Approved at `ad20015b`. All previously blocking findings are resolved: the cross-node policy is exact-first with the accepted canonical SGLang DeepEP EP8/node1 estimated fallback, executed donor provenance is exposed, frozen golden provenance is pinned, WideEP MLA compatibility is restricted correctly, duplicate coverage probes are removed, optional SOL detail no longer discards successful estimates, public API contracts are preserved, and the authoritative API documentation now matches the implementation. Final Standards and Spec re-reviews are clean, and all 22 checks pass. The remaining `moe_quant_mode` validation suggestion is non-blocking: it concerns the error taxonomy for an invalid SDK input rather than supported estimation behavior.

src/aiconfigurator/sdk/moe_comm_resolver.py
```diff
@@ -0,0 +1,132 @@
...
+                    shape.hidden_size,
+                    shape.moe_inter_size,
+                    shape.topk,
+                    shape.num_experts,
+                    model_config.moe_quant_mode,
+                    phase,
+                )
+                if compute_probe is not None
+                else {moe_ep_size}
+            )
```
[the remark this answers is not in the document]
Arsene12358: Verified on the current head and treating this as non-blocking. Supported CLI construction resolves quantization strings before this internal resolver; the remaining case is a manually invalid SDK `ModelConfig` producing a less-specific typed data-availability error instead of `TypeError`. That diagnostic tightening does not affect supported estimation behavior or this PR contract and can be handled separately.

Issue #1498, opened 2026-08-06
DSV4 CSA CP: top_last loader ignores approved reuse.yaml donors + rust/python CP composition diverges 34% (unmasked by PR-2)


tianhaox:
## Summary
Two coupled defects around DeepSeek-V4 CSA context-parallel modeling, discovered
while validating PR #1496 (both pre-existing on main, but invisible until PR-2
removed the energy-compensation double-run that accidentally masked them):

1. **Python's CP top_last loader bypasses the reuse sidecar.**
   `ContextDeepSeekV4AttentionModule._load_csa_topk_top_last` reads the
   version-pinned primary path only (`resolve_op_data_path` + `os.path.exists`)
   — it predates the `reuse.yaml` machinery and never honors donors, unlike
   its own sibling DELTA consumer (`_get_dsv4_topk_calib`, reuse-aware via
   `_build_op_sources`). Concretely: `b200_sxm/sparse_attention/sglang/0.5.12/reuse.yaml`
   declares `dsv4_csa_topk_calib_perf from_version 0.5.14 (approved_by yimingl)`;
   the DELTA path uses the donor, the CP gate does not → every CSA CP config on
   0.5.12 raises `PerfDataNotAvailableError` despite approved data.

2. **The rust and python CP compositions disagree when both CAN compute.**
   The divergence reproduces WITHOUT any loader patching on
   `b200_sxm | sglang | 0.5.14` (where the sparse tables are primary, so both
   engines compute today):
   `DeepSeek-V4-Flash | tp1 ep8 cp8 | b=1 isl=8192 static_ctx`
   → python 42.430756 ms vs rust 56.943256 ms (identical numbers on a
   loader-patched 0.5.12 — same donor tables). Repro:
   `backend.run_static_latency_only(model, db_view("b200_sxm","sglang","0.5.14"), RuntimeConfig(batch_size=1, isl=8192, osl=8, engine_step_backend=...), mode="static_ctx")`
   with `ModelConfig(tp_size=1, moe_ep_size=8, cp_size=8)`. The four sparse gate lookups
   agree (mqa_full 0.210567 / mqa_perc 0.008104 / tl_full 0.048698 / tl_perc 0.0),
   so the delta sits in the base-module term and/or the all-gather terms of
   `_query_cp` vs `query_cp_with` (suspect: NCCL message-size units or the
   per-card base shape). The compositions were never differentially validated:
[...]

PR #1581, opened 2026-08-21
feat(sdk): queryable-version slots (current/previous/next) with old-version data prune

tianhaox:
## What

Two coupled changes, developed and reviewed together per maintainer direction:

### 1. Data prune (280+ parquet deleted; tree 182 MB → 118 MB)

Per-file triaged removal of superseded old-version perf tables across trtllm / sglang / vllm. Triage was computed against the **engine's actual lookup keys** (`model`/`model_name`/`tp_size`/`op_name` are metadata, sweep axes interpolate, `power` IS a value column feeding energy). Deliberately kept as **fill sources**: comm (multi-node curves), all wideep_*, sole-source tables (kda, raw MLA, l40s mla/w4afp8), dsa small-heads holders, power-measured b200 vllm 0.22, and rust-test fixture coordinates.

### 2. Queryable-version slots (`current` / `previous` / `next`)

Queryability is no longer "any directory exists". Each (system, framework) resolves through `systems/query_versions.yaml`:

- **current** — newest maintainer-completed full upgrade (authored): trtllm `1.3.0rc20`, sglang `0.5.14`, vllm `0.24.0`; frozen baselines override (a100: `1.0.0/0.5.10/0.14.0`; b60: vllm `0.20.0`).
- **previous** — the current before it (authored; empty until the first post-adoption bump).
- **next** — derived fleet-wide: highest DATA-BACKED version newer than current (dev drops for new models; today trtllm `1.3.0rc23`, sglang `0.5.16`). Ops it lacks ride backward fill — exactly what the retired reuse markers used to graft, without markers.

Aliases `current`/`previous`/`next` are accepted anywhere a version is requested; anything else fails loudly (`allow_unlisted_version=True` marks deliberate raw-data-coordinate access; `AIC_ALLOW_UNLISTED_VERSIONS` is a transition escape for two legacy suites, queued for retirement). `get_supported_databases` and the support matrix enumerate slots only: **122 → 43 combinations**. Version directories outside the slots that serve no fill purpose were deleted outright (185 dirs) — no reuse.yaml compatibility markers anywhere (an earlier iteration created 300+; all gone).

### Tests follow the outdated-test policy (delete/re-anchor, don't bypass)

- Parity matrices fully re-anchored onto slot versions (94 cases; duplicate coverage collapsed); goldens re-pinned from the live source-built engine per parity.md Rule 1 — **443/443 green**. Behavior tests moved to current-slot coordinates that still exhibit the behavior (e.g. xop borrow via sglang 0.5.14's missing MSA tables); genuine data-coordinate cases (wideep 0.5.6.post2, power 0.22, FPM fixture) declare `allow_unlisted_version=True` with reasons.
- CLI/task suites re-anchored to slots or the `current` alias; one obsolete donor-mechanism test deleted (its #1498 value pins kept on the 0.5.14 primary).
- Design doc gains **§14** (slot amendment; supersedes §6.3 forward-borrow, §12.3 support bar, answers §13 retention). New **aic-version-slots skill** documents the workflow for contributors/maintainers.

## Verification (local, committed tree)

2285 sdk/tools/cross-package unit tests · 443/443 parity (source-built engine) · collector gate R1–R7 · backend_facts --check (2,562 slices) · manifest regenerated · full-tree anomaly scan clean.

## 2026-08-25 update — recollection merged; final support-matrix state

The prune-created gaps flagged above were re-collected (aic-auto-collector, stock images, model-filtered) and merged:

- **sglang 0.5.14 moe** h100/h200: +3,564 Nemotron-3-Super rows each (inter=2688 — the shape lost with the 0.5.10 dirs).
- **vllm 0.24.0 moe** b200/h200: +1,134 Nemotron-3-Super rows each (Ultra rows duplicated existing keys and were discarded at merge).
- **trtllm 1.3.0rc23** (next slot) new native dirs on h100/h200: moe (Super+Ultra union, 10,692 rows) and dsa context/generation (DeepSeek-V3.2, bf16 + fp8_block gemm lanes).
- vllm Blackwell dsa remeasurements were NOT merged (~100% key overlap with shipped 0.24.0 tables — no coverage gain).

**Support matrix after the merge** (scoped regeneration of the 4 affected combos, 568 probes; committed CSVs updated in place):

- Exactly **8 rows moved, all FAIL → PASS**: `NVIDIA-Nemotron-3-Super-120B-A12B-FP8` agg+disagg on h200/b200 × vllm 0.24.0 and h100/h200 × trtllm 1.3.0rc23. Zero other movement.
- Final totals (6,248 rows): **PASS 3,896 (62.4%) · HYBRID_PASS 80 (1.3%) · FAIL 1,278 (20.5%) · HW✕ 978 (15.7%) · FW✕ 16 (0.3%)**.
- The remaining prune-attributed FAILs are **permanent and honest**: DeepSeek-V3.2 / GLM-5.2-FP8 × h100/h200 × trtllm (dsa-fp8) — TRT-LLM has no fp8 DSA kernels below SM100 (`collect_mla_module._get_precision_combos`); the pre-prune PASS rode deleted rc5/rc10 ladder rows that misrepresented Hopper capability.
- trtllm rc20 (current) still lacks the Nemotron Super moe shape natively; its FAIL rows resolve at the next slot bump (current → rc23), not by data backfill.
- Retirement of the vllm 0.19/0.22 moe bridging dirs was attempted and **reverted**: goldens proved they still feed llama4-scout / qwen3-xquant / int4_wo shapes into the 0.24 view via cross-version fill (one case moved 500×). True retirement needs those shapes collected natively on 0.24 first.

## Follow-ups

1. Retire `AIC_ALLOW_UNLISTED_VERSIONS` by re-anchoring the two legacy task_v2 suites.
2. ~~Rust-side slot awareness / dir-less next-slot loads~~ — DONE in this PR (375209f4, review follow-up): the advertised fleet `next` loads directory-less through backward fill (python gate + spec-carried `tolerate_dirless_version` for the Rust reload). Remaining cleanup: retire the rtx rc23 anchor reuse markers.
3. Re-collect dsa small heads + power sweeps on current versions, then retire their old holders and the rust-fixture coordinates.
4. Collect llama4-scout moe / qwen3 mxfp4-xquant / int4_wo shapes natively on vllm 0.24.0, then retire the 0.19/0.22 moe bridging dirs (see 2026-08-25 update).


aic-core/src/aiconfigurator_core/sdk/perf_database.py
```diff
@@ -284,7 +480,14 @@ def get_supported_databases(
     supported_dict = defaultdict(lambda: defaultdict(list))
     for system, backend_versions in supported_sets.items():
         for backend, versions in backend_versions.items():
-            supported_dict[system][backend] = sorted(versions)
+            # With version slots active, the queryable surface is the slot set
+            # (current / previous / derived next), not every declared directory —
+            # directories outside the slots are fill sources, not versions.
+            slots = get_version_slots(system, backend, systems_paths=systems_paths)
+            if slots is not None:
+                supported_dict[system][backend] = sorted(set(slots.values()), key=_version_sort_tuple)
```
Arsene12358, line 500: **Blocking:** this changes the supported-version surface to resolved literals, but the CLI and support paths still compare the raw request against it (`main.py:1490`, `1678`, `1691`, `1719`, `1749`; `common.py:443`). Consequently, `--backend-version current` is rejected even though `current` resolves to a value in this list; `check_support(..., version="current")` likewise reports unsupported while the equivalent literal reports supported. Please resolve the requested alias per `(system, backend)` before every membership/filter check, including backend-auto and disaggregated paths, and add end-to-end alias tests.

aic-core/src/aiconfigurator_core/sdk/engine.py
```diff
@@ -198,6 +198,36 @@ def _strict_provenance_flag(database: Any) -> bool:
...
+    if resolved:
+        return str(resolved)
+    if backend_version is None:
+        return None
+    try:
+        from aiconfigurator_core.sdk import perf_database
+
+        return perf_database.resolve_query_version(system, backend, backend_version, systems_paths=systems_path)
+    except Exception:
+        return backend_version
```
Arsene12358, line 229: **Blocking:** returning the raw input after resolution fails bypasses the new unlisted-version gate. Combined with `_maybe_load_database()` catching every exception, I could compile `b200_sxm/vllm/0.22.0` and construct `EngineHandle` without `allow_unlisted_version=True` or `AIC_ALLOW_UNLISTED_VERSIONS`; missing `previous` and an omitted version also reached the wire unresolved. Please validate/resolve before tolerant database loading, do not swallow slot-policy errors, map the omitted default to a literal `current`, and test the actual EngineHandle reload path.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
```diff
@@ -108,6 +108,202 @@ def get_systems_paths() -> list[str]:
...
+                with open(os.path.join(systems_root, entry)) as f:
+                    system_spec = yaml.safe_load(f) or {}
+                data_dir = os.path.join(systems_root, system_spec.get("data_dir", ""))
+                if not os.path.isdir(data_dir):
+                    continue
+                for version, version_path in _iter_backend_version_dirs(data_dir, backend):
+                    if _version_sort_tuple(version) > current_key and _database_version_dir_has_perf_files(
+                        version_path
+                    ):
+                        candidates.add(version)
```
Arsene12358, line 271: **Blocking against current `main`:** this fleet candidate scan includes override-governed b60. Main commit `5261c8c` adds complete b60/vLLM 0.26 data, which makes h200 and b200 advertise `next=0.26.0`; however, `get_database(..., "next")` returns `None` because those systems have no 0.26 directory, while b60 itself remains authored at `current=0.20.0`. Please update b60's slot and either exclude override fleets/partition compatibility domains during derivation or implement directory-less `next` loading. Add an invariant test that every enumerated slot can load.

Arsene12358 (changes requested):
Requesting changes for the three correctness blockers below. Each reproduces on this head; the current-main integration finding was rechecked against `dafcd61`.

- Resolve version aliases before CLI and support-surface membership checks.
- Enforce the unlisted-version gate before serializing an EngineSpec.
- Make every derived `next` slot loadable after integrating current `main`, and update the b60 slot.

The known merge conflict and the separate policy reminders are intentionally outside this blocking review.

.claude/skills/aic-version-slots/SKILL.md
```diff
@@ -0,0 +1,67 @@
+---
+name: aic-version-slots
+description: Use when adding perf data for a new framework version, deciding which backend_version to query or pin in tests, bumping the maintained (current/previous) versions, retiring old version directories, or hitting "is not a queryable version" errors from the version-slot gate.
+---
+
+# AIC Queryable-Version Slots
+
+## The model in one paragraph
+
+Each (system, framework) exposes at most **three queryable versions**,
```
Arsene12358, line 10: **Non-blocking policy reminder:** `.claude/rules/repo-guide.md:49-51` says normative policy lives once under `.claude/rules/`, while skills are procedural pointers. This skill restates the slot semantics, prohibited markers/overrides, retention rules, and test escape policy. Please consider moving those rules to a path-scoped rule file and keeping this skill focused on workflow steps and links.

Arsene12358 (commented):
**Non-blocking policy reminder:** `CONTRIBUTING.md:12-21` asks significant enhancements to start with a GitHub issue and an up-front design discussion. This PR changes public SDK/CLI behavior and has no linked issue. Please link the originating issue/design approval, or record a maintainer waiver. This reminder does not contribute to my request-changes verdict.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
tianhaox: Fixed in 9934fc5d. Added `_resolve_version_for_matching` (alias -> literal per (system, backend), raw versions pass through) and wired it into all five membership sites: the precheck, both backend-auto paths, the prefill/decode warning loop, and the disagg decode gate; `check_support` resolves per matrix-row backend. E2E alias tests: `TestBackendVersionAliasAcceptance` (CLI) and the existing `test_common` suite now covers `version="current"` == literal. One fixture that faked the enumeration with the string "current" was re-anchored to slot-resolved literals in 00bdd964 (the real surface never enumerates aliases).

aic-core/src/aiconfigurator_core/sdk/engine.py
tianhaox: Fixed in 9934fc5d. `compile_engine` now resolves the requested version to a literal BEFORE the tolerant database load; `_literal_backend_version` no longer swallows anything — slot-policy errors (unlisted versions, unpopulated aliases) propagate, an omitted version maps to the literal `current`, and trees without a slots file keep the ungated passthrough. Tests: spec-level rejection of `0.22.0` and `previous`, omitted->current literal, and the actual `EngineHandle` reload path (compile with the alias, reload from bytes, answer a static run).

aic-core/src/aiconfigurator_core/sdk/perf_database.py
tianhaox: Fixed in 9934fc5d. `_derive_fleet_next` now excludes override-governed systems from the candidate scan (frozen baselines are a separate compatibility domain), so the b60 0.26.0 drop no longer advertises a next that h200/b200 cannot load. b60's override moved to `current: 0.26.0 / previous: 0.20.0` (#1441 shipped the complete attention/gemm/moe/comm collection). Invariant test added: every combination `get_supported_databases` enumerates must `get_database` successfully (46 combos today). The slot bump moved the b60 scan-coverage goldens 11-24% with zero test edits — the refresh diff (ac2077ec) is the review surface.

.claude/skills/aic-version-slots/SKILL.md
tianhaox: Acknowledged. Rule files are human-owned policy in this repo, so I have not moved the normative content as a side effect of this PR; happy to split the slot semantics / retention / escape rules into a path-scoped `.claude/rules/` file and reduce the skill to workflow steps + links in a follow-up (or in this PR if you prefer to review it here) — maintainer sign-off on the new rule file is the gate.

aic-core/src/aiconfigurator_core/systems/data/h100_sxm/mla/trtllm/1.3.0rc15/reuse.yaml
```diff
@@ -4,18 +4,22 @@
...
+  from_version: 1.3.0rc20
+  reason: migrated from SHARED_LAYER_REUSE.txt (legacy newest-first effective donor) [retargeted 1.3.0rc10->1.3.0rc20
+    after pre-rc20 prune 2026-08-21]
   approved_by: yimingl
 - table: mla_generation_module_perf
-  from_version: 1.3.0rc10
-  reason: migrated from SHARED_LAYER_REUSE.txt (legacy newest-first effective donor)
+  from_version: 1.3.0rc20
+  reason: migrated from SHARED_LAYER_REUSE.txt (legacy newest-first effective donor) [retargeted 1.3.0rc10->1.3.0rc20
+    after pre-rc20 prune 2026-08-21]
```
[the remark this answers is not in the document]
tianhaox: Outdated finding: `h100_sxm/mla/trtllm/1.3.0rc15/reuse.yaml` no longer exists — the marker was retired when the branch moved to marker-free semantics (later commits in this PR). No reuse.yaml under any `mla/` directory remains.

tools/perf_database/check_cross_backend.py
```diff
@@ -750,6 +765,128 @@ def _check_gemm_sol(system: str, op_file: str, t: OpTable, spec: dict) -> tuple[
...
+                    "source": "engine",
+                    "moe_dtype": dtype,
+                    "points": len(hits),
+                    "worst_fraction_of_sol": float(worst[0]),
+                    "example_shape": worst[1],
+                    "example_latency": worst[2],
+                    "example_sol": worst[3],
+                }
+            )
+    return findings
```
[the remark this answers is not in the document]
tianhaox: Fixed defensively in 9934fc5d: the renderer now uses `a.get('gemm_dtype') or a.get('moe_dtype', '')`. Note the `--engine-sol` MoE path that produced `moe_dtype`-only findings was retired from this branch, so the KeyError is unreachable today — the .get keeps the renderer safe if that path returns.

aic-core/rust/aiconfigurator-core/parity_tests/goldens/engine_step.json
```diff
@@ -1913,32 +1913,32 @@
...
-          "static_gen_energy": 403.0624431434759,
-          "static_gen_power": 98.58538807567466,
-          "static_total": 24.314
+          "static_ctx": 20.235,
+          "static_ctx_energy": 0.0,
+          "static_ctx_power": 0.0,
+          "static_gen": 4.149,
+          "static_gen_energy": 0.0,
+          "static_gen_power": 0.0,
+          "static_total": 24.384
```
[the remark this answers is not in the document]
tianhaox: This is deliberate, not vacuous — the energy check is three-layered by design (see the comment block above POWER_CASES): (1) energy MATH is pinned on synthetic hand-derivable fixtures in `operators/{gemm,attention}.rs` (`energy_test_fixtures`); (2) DATA invariants over every power-carrying parquet live in `test_power_data_invariants.py`; (3) this case pins the FFI energy SURFACE on a latency-only current-slot identity, where well-typed zeros ARE the tree's honest answer (the only power-measured tables in the tree are b200/vllm/0.22 attention+moe — no complete end-to-end vehicle exists). Frozen zeros still catch NaN/garbage/shape regressions on the surface. End-to-end nonzero pins return with the power-sweep recollection (Follow-up #3 in the PR description).

aic-core/src/aiconfigurator_core/sdk/perf_database.py
Arsene12358: Thanks—the b60-specific trigger is fixed. I think the underlying default-fleet case is still open, and I would like your view on whether this behavior is intended.

Practical scenario: vLLM `current` is `0.24.0`; an h200-only development collection adds a data-backed `0.25.0`; b200 still has only `0.24.0`. Because `next` is derived fleet-wide, b200 advertises `next=0.25.0`, so this command passes the CLI precheck:

```bash
aiconfigurator cli estimate \
  --model-path Qwen/Qwen3-1.7B \
  --system b200_sxm \
  --backend vllm \
  --backend-version next \
  --estimate-mode static \
  --tp-size 1 --batch-size 1 --isl 64 --osl 2
```

I reproduced that flow against the current head with an h200-only `0.25.0` overlay. It exits before producing a prediction:

```text
no data found for system='b200_sxm', backend='vllm', version='0.25.0'
failed to get system='b200_sxm', backend='vllm', version='0.25.0'
Error: Failed to load perf database for system=b200_sxm, backend=vllm, version=next.
```

The minimal two-default-system probe has the same result: alpha has `1.0.0` + a `1.1.0` dev drop, beta has only `1.0.0`; beta derives and enumerates `next=1.1.0`, but `get_database(beta, vllm, next)` returns `None`. `get_database` currently requires an exact requested-version directory before constructing `PerfDatabase`, so channel-1 backward fill never gets a chance to serve beta from `1.0.0`. The new invariant test passes for today's checked-in tree but does not exercise this partial-drop workflow.

Do you consider `next` propagation from one default-governed system to the rest of the default fleet intentional? Section 14 says the highest data-backed version anywhere in the fleet becomes `next`, with missing ops served by backward fill, so my recommendation is to preserve that contract and implement a narrowly scoped directory-less load for a resolved/advertised `next` (without weakening raw-version or provenance gates). Please add a two-system regression test where only alpha has the dev-version directory and assert beta's advertised `next` loads successfully. If the intended contract is actually system-local `next`, the alternative is to change the derivation and Section 14 together; I prefer the directory-less implementation because it matches the documented contributor workflow.

tianhaox: Intentional — fleet-wide `next` with backward fill is the §14 contract we want to keep, and you are right that the directory gate contradicted it. Implemented your preferred option in 375209f4:

- **python `get_database`**: when the resolved version IS the advertised `next` and no version directory exists, `PerfDatabase` constructs anyway and channel-1 backward fill serves every op (verified: a dir-less load populates `supported_quant_mode` from current data). The relaxation covers exactly the advertised next — raw versions and the authored current/previous keep the loud missing-directory gate, and provenance labeling is untouched (fill rows carry their own source channel).
- **engine reload**: `EngineSpec` gains an additive-optional `tolerate_dirless_version` flag, set only by the python spec builder after a dir-less next load; the Rust reload skips its missing-directory gate for exactly that spec. Native builders never set it, so the Rust-side raw gate is unweakened.
- **regression test**: `test_dirless_next_load.py` — the two-system scenario you asked for (alpha carries the 1.1.0 dev drop, beta does not): beta advertises AND loads next, by alias and by literal, plus assertions that the raw-version gate and unpopulated-alias rejection stay loud.
- Your h200-only 0.25.0 overlay repro now produces a real prediction end to end (the `cli estimate` command from your comment).

This also retires the long-standing follow-up #2 in the PR description (dir-less next-slot loads).

Arsene12358: Thanks — I confirmed that the original full-model CLI scenario now works end to end.

One remaining design-boundary question: is a directory-less advertised `next` intended to work through the public model-less database handle as well, or only through full-model compilation/CLI prediction?

Using the committed alpha/beta fixture, this currently succeeds:

```python
db = get_database("b200_sxm", "vllm", "next", systems_paths=root)
assert db.version == "1.1.0"
assert db.dirless_next_load is True
```

But the next public API call fails:

```python
EngineHandle.for_database(db, systems_path=root)
# PerfDataNotAvailableError: perf data directory not found ... b200_sxm/vllm/1.1.0
```

The normal model-spec builder emits `tolerate_dirless_version`, while `build_database_probe_spec_json()` currently omits it, so Rust sees the default `false` and reapplies the missing-directory gate. This probe path also underpins table-view and ad-hoc op-list evaluation.

My recommendation is to support the database consistently by emitting the same narrowly scoped flag from the probe-spec builder and adding an `EngineHandle.for_database()` regression test, since `get_database()` returns the object as valid and the public handle does not document a narrower version contract. If directory-less `next` is intentionally limited to full-model prediction, could you confirm that intended boundary? In that case, I think the API should reject or document it explicitly rather than failing later during Rust reload.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
```diff
@@ -1009,6 +1237,41 @@ def get_database(
...
+                try:
+                    database = PerfDatabase(
+                        system, backend, version, systems_root, database_mode=database_mode, **extra_database_kwargs
+                    )
+                    # The engine reload (Rust) keeps its own loud
+                    # missing-directory gate; the spec builder reads this
+                    # marker to carry the dir-less-next tolerance on the wire.
+                    database.dirless_next_load = True
+                    databases_cache[cache_key][backend][version] = database
+                    return database
```
Arsene12358: **Blocking: defer directory-less fallback until every configured root has been checked for exact data.**

I reproduced this with two ordered roots:

```text
base:  b200/vllm/1.0.0 only
extra: b200/vllm/1.1.0 exact data
systems_paths = [base, extra]

slots:         {'current': '1.0.0', 'next': '1.1.0'}
selected_root: base
dirless:       True
```

The new return here fires while examining `base`, so `get_database()` never reaches the exact `1.1.0` directory in `extra`. The call succeeds but silently predicts from backward-filled `1.0.0` measurements instead of the exact `1.1.0` measurements the caller configured.

Current behavior versus required behavior:

```text
Current:  first missing root -> construct directory-less DB -> return
Required: search all roots for exact data -> if none exists, construct directory-less fallback
```

Suggested fix:

- Determine whether the resolved literal is the advertised `next` once before the root loop.
- Inside the loop, retain the first eligible `(systems_root, cache_key)` as a `dirless_next_candidate` instead of constructing or returning it.
- Preserve the existing immediate return for an exact, usable version directory in any root.
- After the loop, if no exact database was found, construct/cache the retained candidate, set `dirless_next_load = True`, and return it.

Please add a two-root regression test where the first root contains only current and the second contains exact next, asserting that the returned database uses the second root and is not marked directory-less. Keep the existing single-root test to prove that the fallback still works when no root contains exact next.

Arsene12358 (changes requested):
Requesting changes for one reproduced regression in the directory-less `next` follow-up. When multiple systems roots are configured, a missing exact version in an earlier root now triggers an immediate directory-less return, so an exact `next` collection in a later root is silently ignored and predictions use backward-filled older measurements. The inline comment includes the reproduction, the recommended control flow, and the regression-test shape. The separate `EngineHandle.for_database()` design-boundary question remains open for author clarification and is not part of this change request.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
tianhaox: Fixed in 3818bdc2 exactly per your suggestion: the directory-less fallback is now a deferred candidate — the root loop retains the first eligible (systems_root, cache_key) and only constructs it AFTER every configured root has been checked, so an exact, usable version directory in any later root keeps its immediate-return win. Added the two-root regression you asked for (base holds only current, extra holds the exact next collection: the returned database is the exact 1.1.0 and is NOT marked directory-less) alongside the existing single-root fallback tests.

aic-core/src/aiconfigurator_core/sdk/perf_database.py
tianhaox: Intended to work through the model-less handle too — a database get_database returns as valid should stay valid across the public surface. Fixed in 3818bdc2: build_database_probe_spec_json now emits the same narrowly scoped tolerate_dirless_version flag as the model spec builder, and there is an EngineHandle.for_database regression test on the committed alpha/beta fixture (dir-less next database -> probe handle constructs).

Arsene12358: I pushed d3f1404 to address this. Advertised `next` is now classified before the generic `allow_missing_data` path, and a cached missing-data database that is later promoted to advertised `next` receives `dirless_next_load` before it is returned. The new tests cover HYBRID `get_database_view()` through `EngineHandle.for_database()` and the cache-promotion sequence; a practical HYBRID CLI prediction also succeeds. Local verification: 4,196 unit tests passed, plus 336 Rust golden tests in the independent review. I will resolve this thread once the new CI run is green.

Arsene12358 (approved):
Approved after addressing the remaining HYBRID directory-less-next blocker in d3f1404. Advertised `next` now takes precedence over generic missing-data handling, including cached identities promoted to the slot. Verified with practical SILICON and HYBRID CLI predictions, 4,196 local unit tests, 336 independently rerun Rust golden tests, clean standards/spec reviews, and all 26 GitHub checks green. The two earlier repository-policy notes remain non-blocking follow-ups as discussed.

PR #1582, opened 2026-08-21
feat(sdk): declared queryable-version list (query_versions.yaml)

tianhaox:
## What

Queryability is currently defined by which version directories exist on disk. That makes every historical version a permanent liability: retiring one requires a per-directory `reuse.yaml` marker, and #1581 needed 300+ markers, 83 donor retargets and four fixture-restore rounds just to delete files.

This PR separates **queryable versions** (a declared list) from **data directories** (resolver fill sources):

- `systems/query_versions.yaml` — per (platform, framework) list, `defaults` + per-platform `overrides`, first entry = platform primary. Initial values: trtllm `{1.3.0rc20, 1.3.0rc23}`, sglang `{0.5.14, 0.5.16}`, vllm `{0.24.0}`; a100/b60 declare their real baselines (`1.0.0`/`0.5.10`/`0.14.0`, `0.20.0`) instead of pretending to be current.
- `get_database` resolves the requested version through the list: **listed** → unchanged; **unlisted, directory survives** → unchanged + deprecation warning (grace for pinned tests / fixture data); **unlisted, no directory** → soft-mapped to the platform primary with a warning instead of failing.
- Unlisted version directories become *data sources, not versions*: they still feed the fill channels, but retiring one is now a plain directory deletion — no marker needed.

Non-breaking: absent list file disables mapping entirely; all current dirs exist so today only warnings change. 380 database unit tests pass (374 + 6 new).

## Follow-ups enabled

1. Delete the ~300 transitional `reuse.yaml` markers created by #1581 (their mapping role moves here).
2. Rust-side twin for `Engine::from_spec_bytes` version resolution.
3. Test-fixture discipline: fixtures pin listed versions or checked-in synthetic trees only.

Draft until the initial list values are confirmed.

## Summary by CodeRabbit

* **New Features**
  * Added query-version metadata for supported inference backends and platforms.
  * Added platform-specific version overrides and primary-version fallbacks.
  * Preserved access to undeclared versions when corresponding data is available.
  * Added graceful fallback behavior when query-version metadata is missing or invalid.

* **Tests**
  * Added coverage for version resolution, platform overrides, fallback behavior, and missing configurations.

### DIFF
--- a/aic-core/src/aiconfigurator_core/sdk/perf_database.py
+++ b/aic-core/src/aiconfigurator_core/sdk/perf_database.py
@@ -108,6 +108,85 @@ def get_systems_paths() -> list[str]:
     return list(_SYSTEMS_PATHS)
 
 
+_QUERY_VERSIONS_BASENAME = "query_versions.yaml"
+
+
+@functools.cache
+def _load_query_versions(systems_paths: tuple[str, ...]) -> dict | None:
+    """Load the declared queryable-version list, or None when absent.
+
+    The file lives next to the ``<system>.yaml`` specs; the first search root
+    that carries one wins. Absence disables version mapping entirely (the
+    module behaves exactly as before the list existed).
+    """
+    for systems_root in systems_paths:
+        path = os.path.join(systems_root, _QUERY_VERSIONS_BASENAME)
+        if os.path.isfile(path):
+            try:
+                with open(path, encoding="utf-8") as f:
+                    doc = yaml.safe_load(f) or {}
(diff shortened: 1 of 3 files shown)


tianhaox:
Closing: direction changed — retired versions are simply deleted (no queryable-alias layer needed, see #1581's final semantics). The broader question of version lifecycle vs op-centric data design needs a deeper look at the overall op design logic before any mechanism like this; design notes retained internally.

PR #1583, opened 2026-08-21
feat(sdk): Qwen3.8-Max vLLM support at v0.27.1 — collector, 4-system silicon, serving-true lanes

Arsene12358:
Extends Qwen3.8-Max (`Qwen/Qwen3.8-2.4T-A95B`) support to the vLLM backend at **v0.27.1** (the first stock release registering `Qwen3_5MoeForCausalLM`), stacked on #1573 (sglang 0.5.17 support). Linear: AIC-1782. Draft until #1573 lands, then rebased to show only the vLLM delta.

## What's here

**Collector verification 0.24.0 → 0.27.1** (three minor versions of drift): `collector/vllm/collect_{moe,gdn,gemm}.py` audited surface-by-surface and version-branched where vLLM moved (FusedMoE → FusedMoEFactory; `fla/` relocated to `third_party/flash_linear_attention/`), with `__compat__` bumped to the two-version set and strict per-version import-surface tests whose red-proofs are self-contained (embedded pre-fix snippets — no git-history reads; see the CI note below). `collect_attn.py` deliberately stays at `vllm==0.24.0` (see gaps).

**Model-scoped runtime pins**: the three Max checkpoint ids pin vLLM v0.27.1 in `framework_manifest.yaml` (digest-pinned, default + cu129 variants), using the owner-approved model-pin mechanism from AIC-1762. The 0.24.0 default is untouched for everything else.

**Case backend maps with dispatch citations** (v0.27.1 source, file:line): moe bf16 + fp8_block resolve to FLASHINFER_TRTLLM on SM100/103; GDN context selects the flashinfer chunk-scan on CUDA-13 builds (head_k_dim=128); the RadixArk NVFP4 artifact's vLLM gate is fail-closed (`allowed_modes: []`) pending a modelopt dispatch trace. A phantom-mode bug found during tracing (unlisted backends inheriting allow-everything) is fixed with per-backend `framework_quantization` entries (−42 phantom cases).

**Silicon data, 4 Blackwell systems** (b200_sxm, b300_sxm, gb200, gb300; sparse `vllm/0.27.1` version dirs — every non-Max shape inherits 0.24.0 at query time, verified value-identical for a 397B probe): moe 2106/2106/2079/2079 rows at the Max tuple (8192/2048/topk10/512), gdn 1226 ×4, gemm 143486 ×4 (full grid + k=8192 n∈{1,16} extras). **Collected kernel_source matched the dispatch traces on every system**: bf16 → `…flashinfer_trtllm_trtllmbf16expertsmonolithic`, fp8_block → `…flashinfer_trtllm_trtllmfp8expertsmonolithic`, GDN context → `chunk_gated_delta_rule_flashinfer`. Sidecars carry full provenance (auto-collector child pipeline 63879009 for gb300/gb200/b200_sxm; JET 63873278 on aws-pdx for b300_sxm; image digest = the manifest pin; per-table collector hashes tied to the collector commit).

**Modeling fix (rust)**: the GDN physical-lane alias walk was gated version-exact to vLLM 0.24.0, so 0.27.1 context-scan queries fell through to cross-backend donor rows instead of the landed flashinfer silicon (~22% high on that op). The gate now covers 0.24.0|0.27.1 (`vllm_gdn_physical_aliases`) after verifying both versions persist the same physical-label family; older tables that persist logical labels are test-pinned outside the walk. Per rust-core parity Rule 1 the golden diff carries the change: exactly the 16 context-side values of the 2 new vllm-0271 records moved (static_ctx −1.045 ms), zero drift elsewhere.

**Golden parity cases**: `qwen38-max-gb300-vllm-0271-fp8-agg` (tp16, matching the dynamo deployment recipe) and `…-bf16-disagg` (tp16 × pp2 — the first SMOKE case exercising pp>1; bf16 at tp16/pp1 exceeds gb300 memory and tp32 is rejected by the GDN head-divisibility gate). nvfp4 cannot golden (gate fail-closed, no data) — documented divergence from the plan. Parity 463 → 471.

## Known gaps (deliberate, documented)

- vLLM attention is uncollectable at 0.27.1 (`collect_attn.py` pinned `vllm==0.24.0`); Max attention models via 0.24.0 sibling inheritance — dispatch traced identical at both versions for this shape. Needs its own verification pass.
- trtllm fp8/nvfp4 on the Max moe base row remain allowed without a trtllm-scoped dispatch trace: the allow-list defaults open for backends with no `framework_quantization` entry, and only the vllm side was closed here (with citations). Documented in the case YAML comment; out of scope for this stack, which covers sglang + vLLM only — left to future trtllm enablement of this model.
- Silicon failure records (all classified in the run logs, sidecar'd): an intermittent Grace-only IMA in the flashinfer_trtllm bf16 monolithic kernel cost one 27-point sweep task each on gb200/gb300 (bf16 tp16, interpolation-covered); gemm fp8_block n=1 k=8192 raises in the triton block-scale fallback (74 rows); GDN decode grid-y > 65535 tail (4 rows). tp32 moe fails structurally on this kernel (local inter 2048/32 = 64 < 128), both dtypes.
- `gb300/vllm/0.22.0` gdn is a reuse-pointer whose physical labels sit outside the alias gate — pre-existing miss, unchanged here.
- `collector/vllm/utils.py` docstrings still say 0.24.0 (surface verified unchanged at 0.27.1; file untouched).

## CI note

The import-surface red-proof tests originally fetched pre-fix collector code via `git show <commit>` — those pinned hashes dangle after squash-merges/rebases (this exact mechanism was breaking #1573's CI). Both branches now embed the pre-fix snippets verbatim with provenance comments; the suites pass with `GIT_DIR` pointed at an empty repository.

## Verification

Full gate set at HEAD: parity suites 471 passed; `cargo test --release --lib` 513; `pytest tests/unit -m unit` 3903 passed / 10 skipped; op-backend-facts check OK (3,387 slices); collector data checks R1–R7 OK; validate_database spot loads green (vllm@latest resolves 0.27.1 on the 4 Blackwell systems); ruff clean. Every task in this stack was implemented and independently reviewed subagent-to-subagent, with a whole-branch final review closing clean after one fix wave.


Arsene12358:
**End-to-end accuracy validation results** (serving vs prediction, per the parallel validation track; full report + data in the campaign workspace):

Both arms ran Qwen3.8-2.4T-A95B-FP8 at tp16 on GB200 NVL72 (4 nodes, segment-pinned), exact collection-pinned images, direct engine, 8k/1k fixed-length, nextn=0, concurrency ladder {1, 8, 32, 64}, predictions pre-registered before measurement.

| Stack | TPOT median abs err | TTFT @ conc=1 | Throughput abs err |
|---|---|---|---|
| vLLM v0.27.1 (this PR's stack) | **11.1%** (4/4 points, mild systematic over-prediction — the known kernel-vs-step composition bias band) | **−0.2%** | 3.6–6.7% |
| sglang v0.5.17 (base PR's stack) | 37% (3 valid points, systematic under-prediction shrinking with batch — launch-bound small-batch signature + a flagged GDN serving-lane audit; conc64 excluded as admission-saturated) | over-pred 44% | ~25% |

Serving logs confirmed every modeled lane verbatim on the vLLM side (`FLASHINFER_TRTLLM` fp8 MoE, FlashInfer GDN chunk-scan, FLASHINFER attention). Cross-stack: vLLM decodes ~2.2× faster than stock sglang at conc32 on identical hardware because sglang's stock fp8 MoE lane is triton — the dynamo recipe pins flashinfer_trtllm for this reason, and that pinned lane remains uncollected (AIC-1781, re-prioritized by this result).

Also surfaced during validation (upstream, documented in the campaign log): sglang NVFP4 serving is currently a three-way dead end at 0.5.17 (EP crash = sgl-project/sglang#31503; the dispatch-selected trtllm-gen FP4 lane is hard-blocked in serving with "use flashinfer_cutlass"; the cutlass lane hit an undiagnosed NCCL failure at 16 ranks). The trtllm-fp4 serving block also means the collected nvfp4 lane is collector-runnable but serving-blocked — the serving error message now provides the citation needed to collect the cutlass lane.

Arsene12358:
**Follow-up: the sglang accuracy gap is now root-caused** (nsys per-op decode census, same method as the 397B L3 audit; full decomposition in the campaign report).

At bs1 the 21.7ms measured vs 9.4ms predicted decode step decomposes as: **allreduce +9.1ms (74% of the gap)**, MoE composition +2.0ms, dense-GEMM composition +2.0ms, everything else ≈ neutral. Root cause verified in the packaged data: **the gb200 NCCL comm table holds world sizes {2, 4} only** — comm collection ran single-node, so tp16's 185 allreduces/step are priced at intra-node latency (~2.7µs) instead of the measured 16-rank cross-node ring-LL (~50µs). With a correct multi-node curve the bs1 error moves −54% → ~−11%, inside the acceptance bar; conc32 likewise.

This also explains the backend asymmetry cleanly: vLLM engages CUSTOM (MNNVL one-shot) allreduce cross-node — genuinely µs-class — while stock sglang falls back to pynccl ring. Same mechanism behind vLLM's 2.2× serving advantage at multi-node tp16, and a deployment note in itself.

Implied follow-ups: multi-node comm collection (world 8/16, gb200+gb300, both backends — generically valuable beyond this model), and the familiar MoE/GEMM composition-bias residual (~+4ms) on the sglang stack.

PR #1588, opened 2026-08-24
fix(cli): use the plotext plotsize API

xianlubird:
#### Overview:

Fix terminal Pareto rendering failures caused by calling a Plotext API alias that is not available in every installed environment.

#### Details:

- Use the canonical plotsize API when setting terminal plot dimensions.
- Extend the existing plain-output regression test to assert the expected 80 by 30 dimensions.

- src/aiconfigurator/sdk/pareto_analysis.py
- tests/unit/cli/test_plain_output.py

#### Related Issues:

- Relates to ai-dynamo/dynamo#13079

#### Validation:

- git diff --check
- Plotext 5.3.2 API smoke check passed.
- Full test suite was not run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed ASCII Pareto chart rendering by updating plot sizing to use the supported plotting interface.
  * Ensured consistent 80×30 chart dimensions for both ANSI and plain-text output modes.

* **Tests**
  * Added coverage verifying Pareto output sizing across supported display formats.

### DIFF
--- a/src/aiconfigurator/sdk/pareto_analysis.py
+++ b/src/aiconfigurator/sdk/pareto_analysis.py
@@ -1722,7 +1722,7 @@ def draw_pareto_to_string(
             keys "df", "label", "color", "marker" similar to ``series``.
     """
 
-    plotext.plot_size(80, 30)
+    plotext.plotsize(80, 30)
     plotext.theme("clear")
 
     palette = [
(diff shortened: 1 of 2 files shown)


natoscott:
Hi — heads up that this will still fail in CI. The pre-built venv appears to have a version of plotext with no `plotsize` either. PR #1589 adds an early-exit guard plus the size-function probe.

PR #1589, opened 2026-08-24
fix(sdk): guard draw_pareto_to_string against incompatible plotext versions

natoscott:
## Overview

\`draw_pareto_to_string\` calls \`plotext.plot_size()\` which does not exist in the plotext version installed in the CI environment, causing \`test_cli_default_build_subset\` to fail.

Changing it to \`plotsize\` (as in PR #1588) also fails — the CI has a version where neither name exists. From testing, both plotext 2.x and plotext 6.x lack both functions (3.x–5.x have both). The function naming is inconsistent across releases.

## Fix

- **Early-exit guard**: if \`theme()\` or \`build()\` are absent, return \`""\` immediately. The Pareto plot is cosmetic; callers already handle an empty return. This covers any plotext version that has undergone a full API rename.
- **Size-function probe**: try \`plotsize\` / \`plot_size\` / \`limit_size\` / \`limitsize\` in order using \`getattr\` with a default, skipping gracefully if none match.

## Test plan

- [ ] \`test_cli_default_build_subset\` passes in CI
- [ ] \`ruff check\` and \`ruff format --check\` clean
- [ ] Existing \`test_draw_pareto_to_string\` still passes (5.3.2 has \`theme\` and \`build\`)

### DIFF
--- a/src/aiconfigurator/sdk/pareto_analysis.py
+++ b/src/aiconfigurator/sdk/pareto_analysis.py
@@ -1722,7 +1722,19 @@ def draw_pareto_to_string(
             keys "df", "label", "color", "marker" similar to ``series``.
     """
 
-    plotext.plot_size(80, 30)
+    # plotext 6.x is a full rewrite with an incompatible API (no theme/plot/build).
+    # Return early so callers get an empty string rather than an AttributeError.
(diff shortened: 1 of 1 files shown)


natoscott:
This is fixed more simply by PR #1591

PR #1597, opened 2026-08-27
feat(afd) : Rebase/afd onto upstream

gehujun:
Four commits on top of 2ed278a9, all AFD-scoped. Rebased onto current
main, so the engine-routed shims from #1552 / #1555 / #1566 are already
accounted for.

## What

**1. `feat(afd)`: per-pool hardware and backend.** Each of AFD's three
pools (static prefill / A attention / F FFN-MoE) can run on its own
system and backend, mirroring the disagg `--system` / `--decode-system`
model. A pool left unset inherits the top-level value, so an AFD run
naming no pool is byte-for-byte identical to before. Cross-pool A2F/F2A
traffic is priced at the slower endpoint (bandwidth = min of the two
sides). Modeling only -- deployment artifact generation still requires a
single system and fails fast when pools disagree.

**2. `feat(afd)`: super-node fabric tiering + four FastAFD-informed
knobs.** The cross-pool transfer prices itself on the correct fabric tier
rather than hardcoding `inter_node_bw`: on GB200 NVL72 a pool pair inside
one NVLink domain is priced on the domain bandwidth, one exceeding
`num_gpus_per_rack` on the scale-out fabric. This needs a small engine
change -- see Notes. The knobs: uncap the default A:F node ratio (FastAFD
measured optima at 7:1, 11:1 and 17:1, all outside the old 4:1 bound),
`f_latency_scale` for calibrating against a specific FFN runtime,
`router_on_attn`, and `comm_hiding_tolerance`.

**3. `fix(afd)`: three modeling defects that inflated the F-pool cost.**

- `num_microbatches < 2` now uses the serial cadence. Both overlapped
  models declare N_min >= 2, but only the optimistic branch enforced a
  threshold and its fallback landed on conservative, which cannot be
  sustained at mb=1 either. Measured 27.8% under-report on an asymmetric
  topology.
- The F-node AllGather/ReduceScatter pair was gated on
  `min(n_f_workers, f_gpus_per_node) > 1`, a property of the *fabric*. A
  token-dimension collective presupposes a *TP group*: under pure expert
  parallelism every F rank owns its own experts and there is nothing to
  gather along tokens. Now gated on `f_moe_tp_size` as an existence check
  that never touches sizing.
- The model-internal MoE dispatch was billed twice. Under AFD that
  all-to-all *is* the cross-pool transfer, already priced as `t_a2f` /
  `t_f2a`. `build_afd_ops_partition` drops a bare `MoEDispatch` via its
  skip list, but one nested in an `OverlapOp` survived: the marker only
  affects that op's classification vote while `OverlapOp.query()` still
  summed it into `t_f`. It never reached `skipped_ops` either, so the
  audit trail showed nothing dropped.

**4. `docs(afd)`: reconcile four Phase-2 TODOs with what shipped.**
Comment-only.

## Notes for reviewers

**One engine-side change is included** (`aic-core/rust/.../P2POp`): an
optional `span_gpus` that selects the fabric tier. `None` means "caller
did not say" and reproduces the flat pricing verbatim, so no
pipeline-parallel caller moves -- adjacent PP ranks are one hop apart and
the deployment size is not their span. Both new fields are
`#[serde(default)]`, so an EngineSpec produced before this change still
deserializes and no schema bump is needed.

The three-tier *bandwidth* selection already existed in
`SystemSpec::get_p2p_bandwidth`; only the latency counterpart was
missing.

Per the note in `afd_transfer.py` about the modeling boundary needing a
tracking issue and maintainer sign-off, happy to split that change into
its own commit or file an issue first -- just say which you prefer.

## Verification

- AFD targeted suite: **431 passed**.
- Full unit suite: **13 failed / 4363 passed**, against a pristine `main`
  run in the same environment at **13 failed / 4137 passed**. The failure
  sets are identical name for name (all `HuggingFaceDownloadError` from a
  sandboxed network), so zero regressions.
- `ruff check` clean on `src` + `aic-core` + `tests`.
- Tiering is asserted against the real engine on gb200, not against a
  Python mirror of the formula: span unset, span=4 and span=72 all return
  0.012222 ms byte-identical, while span=144 returns 0.025000 ms off the
  inter-rack pair. Mutating the Rust tier selection back to the flat pair
  fails those tests, so the coverage is load-bearing rather than
  tautological.
- Two of the three fixes in (3) overlap with
  `yunyang1999/aiconfigurator@afd-pr1323-fixes-megamoe`; the analysis
  there matches. These are corrected unconditionally rather than behind a
  selectable comm model.

Building this branch requires `maturin build` for the
`span_gpus`-bearing engine; the published 0.11.0 wheel predates it.

## Summary by CodeRabbit

* **New Features**
  * Added heterogeneous AFD prefill, attention, and FFN/MoE pool support with separate hardware and performance settings.
  * Added configurable router placement, latency scaling, communication-hiding tolerance, rack-aware networking, and expanded topology searches.
  * Added GPU-hour pricing to report cost efficiency alongside throughput.
  * Increased the default AFD candidate search limit to 20,000.

* **Bug Fixes**
  * Improved MoE placement, communication billing, latency accounting, and low-microbatch pipeline handling.

* **Documentation**
  * Added AFD configuration guidance and examples.


src/aiconfigurator/sdk/inference_session.py
```diff
@@ -1496,10 +1741,43 @@ def _integrate_decode_phase(
...
         stride = self._AFD_DECODE_STRIDE
         verify_width = self._nextn + 1
 
+        # Embedding / logits GEMM run once per step, not once per layer, and
+        # read no ``s`` -- evaluate them once outside the stride loop and keep
+        # them out of the per-layer average the pipeline maxes over. An empty
+        # once-list is safe: ``_sum_latency`` returns ``(0.0, {})``.
+        a_layer_ops, a_once_ops = _split_once_per_step(a_partition.attn_ops, num_layers)
+        f_layer_ops, f_once_ops = _split_once_per_step(f_partition.ffn_ops, num_layers)
+        t_a_once, a_once_per_op = self._sum_once_per_step(
```
yunyang1999, line 1764: `t_once` is evaluated at micro-batch size (:2047 → :1363) but charged once (:1595).

Either reading of the intended semantics says this under-charges: if embedding / logits run per micro-batch the charge should be num_microbatches × t_once(micro); if they run once for the whole step it should be t_once(full_batch). Currently it is 1 × t_once(micro).

aic-core/src/aiconfigurator_core/sdk/config.py
```diff
@@ -243,6 +254,30 @@ class AFDConfig:
...
     comm_overhead_factor: float = 1.0
+    # Comm-hiding tolerance for the optimistic (K=3) pipeline. The strict
+    # occupancy bound requires ``num_microbatches >= 2 + t_c / max(t_a,
+    # t_f)``, which is >= 3 whenever t_c > 0 and so always demotes mb=2 to
+    # the blocking K=2 model. When the cross-pool round trip is a negligible
+    # fraction of compute (``t_c <= comm_hiding_tolerance * max(t_a, t_f)``)
+    # the session keeps the overlapped K=3 cycle instead -- the measured
+    # behavior on NVLink-class fabrics, where FastAFD reports mb=2 already
+    # hides both directions. Set 0 to restore the strict bound.
+    comm_hiding_tolerance: float = 0.1
```
yunyang1999: Where does 0.1 come from?

src/aiconfigurator/sdk/task_v2.py
yunyang1999, line 2057: The topology enumeration takes the *top-level* system's node width, while costing
takes each pool's own:

  :1922  gpus_per_node = _lookup_num_gpus_per_node(self.system_name)
  :2015  build_afd_parallel_lists(gpus_per_node=gpus_per_node, ...)
  :1932  _afd_a_gpus_per_node / _afd_f_gpus_per_node resolved per pool
  :2906  ...but only reach AFDConfig, i.e. the cost side

`build_afd_parallel_lists` budgets with `(n_a + n_f) * gpus_per_node <= total_gpus`;
`afd_pareto:1274-1278` then sizes the pools with `effective_a/f_gpus_per_node`. When
the two widths differ the candidate's real footprint is not what the enumerator
budgeted, and nothing downstream catches it — the rejection categories are oom /
fixed_batch / tpot / low_batch_oom, with no GPU-budget check on the decode row.

Concretely, system_name=gb200 (4/node), afd_f_system_name=b200_sxm (8/node),
total_gpus=72: the enumerator computes total_nodes = 72//4 = 18 and admits
(n_a=10, n_f=8); the real deployment is 10*4 + 8*8 = 104 GPUs, 44% over budget,
silently.

`afd_hetero.yaml` pairs h200_sxm with b200_sxm, both 8/node, so neither the example
nor the test suite exercises this. Any gb200/gb300 (4) <-> h200/b200/h100 (8) pairing
does — arguably the most interesting hetero combination.

Should `build_afd_parallel_lists` take the two widths separately, or should there be
a post-enumeration budget filter using the real footprint?

aic-core/src/aiconfigurator_core/sdk/system_spec.py
```diff
@@ -41,3 +42,27 @@ def get_p2p_bandwidth(self, num_gpus: int) -> float:
         if num_gpus <= num_gpus_per_rack:
             return node_spec["inter_node_bw"]
         return node_spec.get("inter_rack_bw", node_spec["inter_node_bw"])
+
+    def get_p2p_latency(self, num_gpus: int) -> float:
```
yunyang1999, line 50: This makes a latent yaml inconsistency load-bearing — nothing read `inter_rack_latency`
before this PR. On gb200/gb300 the two values are inverted:

  p2p_latency 10us ("nonofficial correction")  vs  inter_rack_latency 5us ("typical IB")

so crossing racks returns *lower* latency.

aic-core/rust/aiconfigurator-core/src/operators/communication.rs
```diff
@@ -437,17 +447,22 @@ impl P2POp {
         // term (`message_bytes / inter_node_bw * 1000`, source "sol"); every
         // other mode — including SILICON/HYBRID, which have no P2P table —
         // uses the empirical formula tagged "empirical".
-        let inter_bw = spec.node.inter_node_bw.max(1.0);
+        //
+        // `span_gpus` selects the fabric tier. Unset means "caller did not
+        // say": keep the flat inter-node pricing so every pre-existing PP
+        // call site reproduces its previous value exactly.
+        let (inter_bw, latency) = match self.span_gpus {
```
yunyang1999, line 466: The tier comes from the pool pair's total span but is applied to every cross-pool link.
On gb200 the boundary step is 900 → 100 GB/s (9x). A 76-GPU deployment is 72 in one rack
plus 4 in the next — ~4/76 of pairs actually cross, yet 100% are priced on InfiniBand.

Net effect: multi-rack AFD always looks catastrophic, so AFD is effectively capped at one
rack — exactly where this feature is meant to add value. Weight the tiers by the crossing
fraction? Fine as a follow-up, but as written the tiering can only penalise going wide.

aic-core/rust/aiconfigurator-core/src/operators/communication.rs
```diff
@@ -413,6 +413,15 @@ pub struct P2POp {
...
     pub seq_split: u32,
+    /// Number of GPUs the link spans, used to pick the fabric tier
+    /// (intra-node / intra-rack / inter-rack). `None` keeps the flat
+    /// `inter_node_bw` + `p2p_latency` pricing, which is what every
+    /// pipeline-parallel caller wants: adjacent PP ranks are one hop apart
+    /// and their span is not the deployment size. Cross-pool callers (AFD)
+    /// set it so a topology leaving the scale-up domain is priced on the
+    /// scale-out fabric instead. `serde(default)` keeps older specs loadable.
+    #[serde(default)]
+    pub span_gpus: Option<u32>,
```
jasonqinzhou, line 424: [P1] Initialize `span_gpus` in every `P2POp` literal

Adding this required field still breaks the two direct `P2POp` literals outside this module. With `PYO3_PYTHON=/absolute/path/to/.venv/bin/python cargo test -p aiconfigurator-core`, rustc reports E0063 at `src/engine/spec.rs:362` and `src/operators/fpm_sol.rs:712` because both omit `span_gpus`. The same build also reports the separately posted `NodeSpec.inter_rack_latency` omissions, so the Rust test target and rebuilt Python extension cannot compile. Please set `span_gpus: None` in both literals (or route them through `P2POp::new`) and keep the Rust suite in the validation gate.

src/aiconfigurator/sdk/pareto_analysis.py
```diff
@@ -1051,6 +1170,25 @@ def afd_pareto(
...
+    # reuse the very same objects when a pool inherits the shared value.
+    effective_a_database = a_database if a_database is not None else database
+    if not a_backend_name or a_backend_name == backend_name:
+        effective_a_backend = backend
+    else:
+        effective_a_backend = get_backend(a_backend_name)
+    effective_f_database = f_database if f_database is not None else effective_a_database
+    if not f_backend_name or f_backend_name == (a_backend_name or backend_name):
+        effective_f_backend = effective_a_backend
+    else:
```
jasonqinzhou, line 1185: [P1] Keep an omitted F override on the shared pool

The public contract above says each unset `a_*`/`f_*` argument inherits `database` and `backend_name`, but these lines make an omitted F database/backend inherit the already-overridden A side. A direct `afd_pareto(..., a_database=h200_db, f_database=None)` therefore evaluates F compute on H200 even though F was left on the shared pool. Please resolve F defaults from `database` and `backend` independently of A, and add coverage for one-sided A and F overrides.

src/aiconfigurator/sdk/task_v2.py
```diff
@@ -1047,6 +1169,28 @@ def _resolve(system: str, backend: str, current: str | None) -> str | None:
...
+                        continue  # explicit version wins
+                    pool_system = getattr(self, system_field, None)
+                    pool_backend = getattr(self, backend_field, None)
+                    if not (pool_system or pool_backend):
+                        continue  # fully inherited: afd_pool_triple falls back
+                    system = pool_system or self.system_name
+                    backend = pool_backend or self.backend_name
+                    if not (system and backend):
+                        continue
+                    setattr(self, version_field, _resolve(system, backend, None))
```
jasonqinzhou, line 1193: [P2] Preserve a pinned version for redundant pool overrides

This treats any explicit pool system/backend field as a moved pool, even when its effective `(system, backend)` pair is identical to the top level. With a pinned top-level `backend_version` and `afd_a_backend_name` redundantly set to the same backend, an unset `afd_a_backend_version` resolves to latest instead of inheriting the pin. Please compare the effective pair with the top-level pair before resolving a new version and cover redundant same-value overrides.

jasonqinzhou (changes requested):

## What this PR does

This PR expands AFD modeling with independently configurable prefill, attention, and FFN pools, topology-aware communication costs, revised pipeline and once-per-step accounting, MoE billing corrections, GPU-price efficiency output, documentation, and broad regression coverage.

The update responds substantively to prior feedback: heterogeneous node widths now participate in sweep budgeting, nested MoE dispatch removal records its audit trail, and the once-per-step and super-node changes come with focused behavioral tests rather than only plumbing coverage.

## Why this score

4.2/10 quality assessment, BLOCKED packet. The required Claude Fable lane could not be verified: `claude-fable-5` was requested, but only Claude Sonnet 4.6 and Claude Haiku 4.5 were observed, so the skill forbids a non-blocked canonical recommendation. Independently, Codex confirmed that the exact head still cannot compile and that all four prior findings remain reproducible; the heterogeneous pinned-topology path also remains inconsistent, though current GitHub feedback already covers that and several related AFD concerns. The branch additionally conflicts with live `main`, fails title validation, and has 17 unresolved review threads.

### P1 — Build static-prefill candidates from the prefill pool

[P1] Build static-prefill candidates from the prefill pool

This still calls `build_disagg_parallel_lists` with the top-level system/backend and then mirrors those values into the internal prefill role, even when `afd_prefill_system_name` or `afd_prefill_backend_name` overrides that pool. On the exact head, a task with `afd_prefill_system_name="gb200"` resolves the AFD prefill pool to GB200 but leaves `prefill_system_name="h200_sxm"` and builds the prefill candidates from H200. Please resolve the role and candidate builder from `afd_pool_triple("afd_prefill")` and add system- and backend-override regressions.

### P1 — Use per-pool widths for pinned AFD topology

[P1] Use the resolved pool widths for pinned AFD topology

The new sweep enumerator now budgets A and F with their own node widths, but the pinned path still validates `afd_tp_a`, computes `pinned_gpus`, and derives `tp_f` entirely from top-level `gpus_per_node`. Mixed-width pinned configurations can therefore reject a feasible deployment, accept an over-budget one, or size the F pool incorrectly before single-point evaluation. Please use `_afd_a_gpus_per_node` for A-side TP/budgeting and `_afd_f_gpus_per_node` for F-side TP/budgeting, and apply the same contract in the single-point runner.

PR #1598, opened 2026-08-27
feat(rust): heal kv_seed_regime="fake_fallback" latencies by in-station extrapolation at FPM load

liyuanzhe1991:
# feat(rust): heal kv_seed_regime="fake_fallback" latencies by in-station extrapolation at FPM load

## Problem

At each batch size's top KV rung, the collector's kvwarm chain cannot always
reach the target KV state; those points were measured against a fabricated
seed. The poison is bidirectional — ×2–3.7 high on 4-GPU shapes, LOW on
2-GPU (tp2 b=128: the fake top rung reads 33.24 ms while the real rung below
it reads 41.51 ms) — and once such a value enters the interpolation lattice
it corrupts the entire large-KV band. On same-node ground truth (5,524
votes, top band b≤512), serving the fake values scores MAPE tep4 29.8% /
tp4 42.6% / tp2 11.9%.

Simply dropping the rows is worse: the top band becomes unanswerable
(fail-closed no-extrapolation domain gate), and the measured cost of that
was 46–67% error on tp2 via out-of-domain misrouting.

## Fix (R17 final form, A/B-validated)

The collector records seed provenance in a `kv_seed_regime` column. At FPM
load, for every **decode** row marked exactly `"fake_fallback"`, the loader
now overwrites its in-memory `latency_ms` with a linear extrapolation from
the same station's (same cell, same batch rung) **two highest `"real_kv"`
rows strictly BELOW the fake row's kv** (below-only anchoring, matching the
Python loader's `FPM_REPLACE_FAKE_EXTRAP`; a fake rung is only ever a forward
extension of the real curve beneath it, never back-extrapolated):

```
value = l2 + (kv_fake − k2) · (l2 − l1) / (k2 − k1)
```

- Rows, coordinates, the grid, the domain gate, and the query path are
  untouched — the top band keeps full value coverage and answers normally.
- The parquet on disk always keeps the original measurement + marker
  (audit chain); replacement exists only in the loaded view.
- A station with fewer than two `real_kv` anchors below the fake row (or a
  non-finite/non-positive extrapolation) keeps its original value, tallied in
  `kept_fake_fallback_rows()` — the keep-and-warn case; zero occurrences
  across all measured shapes (minimum 7 real points per station).
- Every other marker (`real_kv`, `skip:<reason>`, prefill `n/a`,
  re-aggregated `legacy`, null) and pairs predating the column load
  bit-for-bit unchanged — old databases are a strict no-op, so this can
  land ahead of the collector-side column writer (fpm-prc chain).
- Escape hatch: `AIC_FPM_FAKE_FALLBACK_RAW=1` serves the original fake
  values (for A/B and audit runs). Default is healing on.
- Tallies are exposed as `replaced_fake_fallback_rows()` /
  `kept_fake_fallback_rows()` (the crate has no logging facade; the
  counters are the warning surface).

## Validation

- **Same-node A/B (5,524 votes, measured 2026-08-20):** healed values score
  top-band MAPE **tep4 2.34% / tp4 0.34% / tp2 2.71%** (P95 2.55/0.65/5.56)
  vs 29.79/42.61/11.91% serving fake values.
- **Wrong-row regression case** (tp2, b=128, kv=131200, truth 41.69 ms):
  answers ≈41.5 ms on the healed segment; pinned as a loader oracle test
  with the exact ladder numbers (fake 33.24 → healed ≈44.56).
- **Full-database replay** (53,483-row R17 pair, opt-in ignored test
  `r17_pair_replay_heals_every_fake_row`): all 605 fake rows healed, zero
  kept, every healed value bit-equal to the two-anchor extrapolation
  recomputed from the raw view, every other value bit-for-bit unchanged.
- Tests: 8 new loader oracle tests (incl. a mid-curve fake rung pinning
  below-only vs backward anchoring) + the opt-in replay. Full crate 499
  pass; `--all-features` embedding targets compile; Python units
  (test_fpm_forward + test_rust_engine_step, 50) pass; golden parity suites
  pass with zero golden diffs (fixtures predate the column — answer
  preserving).

## Dependencies / ordering

The column is written by the collector-side v6 chain (#1475 lineage);
`schema_version` stays 6 (additive column, strict-equality sidecar gate
unchanged). This PR is a no-op for databases without the column and can
merge in either order.

Supersedes the unopened `fpm-exclude-fake-fallback` branch (drop-rows
semantics, retired by the R17 A/B evidence).

## Summary by CodeRabbit

* **Improvements**
  * Improved performance-data accuracy by automatically estimating missing fake-fallback decode latencies from valid real-measurement data.
  * Preserved original values when suitable reference data is unavailable.
  * Added compatibility for legacy performance files without provenance metadata.
  * Added reporting for adjusted and unchanged fallback records.

* **Testing**
  * Expanded coverage for latency correction, compatibility, and raw-value handling.

### DIFF
--- a/aic-core/rust/aiconfigurator-core/src/perf_database/fpm_forward.rs
+++ b/aic-core/rust/aiconfigurator-core/src/perf_database/fpm_forward.rs
@@ -13,7 +13,10 @@
 //! ```
 //!
 //! is validated (sidecar schema/sha256/row_count, per-row workload checks,
-//! duplicate physical row keys) and grouped into cells keyed by
+//! duplicate physical row keys), healed of
+//! `kv_seed_regime == "fake_fallback"` values (fabricated-KV measurements
+//! replaced in memory by in-station extrapolation; see
+//! [`FPM_KV_SEED_FAKE_FALLBACK`]), and grouped into cells keyed by
 //! `(model_path, 15 identity columns)`. Each
 //! cell holds one nested table per phase — prefill
 //! `[batch][total_prefill][total_kv]`, decode `[batch][total_kv]` — plus the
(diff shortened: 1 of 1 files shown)


Arsene12358 (approved):
Approved. The loader implements the intended below-only healing contract: it changes only in-memory latency values for decode rows marked exactly <code>fake_fallback</code>, anchors each replacement on the two highest same-cell and same-batch <code>real_kv</code> rows strictly below the fake KV coordinate, never chains healed values as anchors, and preserves rows, coordinates, domain coverage, the query path, and the original parquet audit data. Invalid extrapolations keep the original value and are counted, the raw escape hatch is retained, and legacy or missing-column databases remain unchanged.

I found no PR-introduced functional regression or standards blocker. Focused FPM loader tests passed 25/25 with only the externally provisioned R17 replay ignored; the full Rust library passed 499 tests with two ignored, Clippy completed successfully, and the repository checks covering unit, e2e, Rust/Python parity, public API compatibility, regression snapshots, wheels, DCO, and policy are green. The remaining Rustdoc wording and external-replay reproducibility notes are nonblocking and do not undermine the implemented behavior.

Issue #1422, opened 2026-07-26
collector unit tests on torch-equipped machines: TORCH_LIBRARY re-registration aborts collection + two masked issues


tianhaox:
## Context

Follow-up from #1421 (which fixes the CI-visible torch-mock pollution). Running `pytest tests/unit/collector` on a machine **with torch installed** has three pre-existing problems, all locally verified. None of them reproduce in the unit-test CI container (it ships without torch), which is why they have gone unnoticed.

## 1. Evict→fresh-import→restore pattern can only succeed once per process

Three modules borrow the real torch by evicting a possibly-cached MagicMock, importing torch fresh, and restoring the mock afterwards:

- `tests/unit/collector/test_dsv4_megamoe_workload.py`
- `tests/unit/collector/test_sglang_moe_ep_routing.py`
- `tests/unit/collector/test_helper_moe_distribution.py` (added by #1421)

`torch/__init__.py`'s C++ `TORCH_LIBRARY` registrations are once-per-process: the second module to run the dance re-executes the init and aborts the whole collection with `RuntimeError: Only a single TORCH_LIBRARY can be used to register the namespace triton`.

**Proposed fix:** a shared per-process helper (e.g. `tests/unit/collector/_real_torch.py`) that performs the fresh import exactly once, caches the result, and returns the cached module on subsequent calls; all three modules switch to it.

**Hard constraint:** the helper MUST put the MagicMock back into `sys.modules` after the borrow. `collect.py` must never see the real torch — `test_parallel_run`'s fork-worker tests genuinely deadlock on macOS with a real torch cached (child processes hang past `--timeout`; experimentally confirmed while evaluating a conftest pre-import approach, which was retracted for exactly this reason).

## 2. Masked deadlock ordering dependency

The fork-worker deadlock above is currently masked by the collection abort. Once (1) is fixed and the full directory actually runs on torch machines, verify that `test_collect_provenance_writer.py`'s "import collect.py before anything else caches a real torch into it" mechanism holds under full-directory collection order.

## 3. Pre-existing GLM checkpoint assertion failures

`tests/unit/collector/sglang/test_collect_mla_module.py` has 3 failures that only execute with real torch... (they also reproduce without any mock interplay, in isolation):

- `test_full_dsa_sweep_canonicalizes_consumer_equivalent_checkpoints` — expects `zai-org/GLM-5-FP8` in the canonical set
[...]

PR #1606, opened 2026-09-06
fix(tests): reuse real torch across collector test modules

git-jxj:
#### Overview:

On machines with PyTorch installed, collecting the collector tests after the provenance tests imports PyTorch repeatedly while borrowing it from a mocked `sys.modules` entry. The second import fails with `Only a single TORCH_LIBRARY can be used to register the namespace triton`.

Cache the real module in one shared test helper and restore an existing caller module entry after each borrow. When PyTorch was not previously imported, retain the ordinary import behavior so subsequent plain imports reuse its native registrations.

#### Details:

- Share the borrow context across the three tensor-dependent collector test modules, including DSV4's lazy imports during test execution.
- Preserve the mock used by `collect.py` and its fork-worker tests.
- Register the routing test's dynamically loaded helper module before executing it. Once the PyTorch collection error is removed, its dataclass annotations need the module in `sys.modules` on Python 3.13.
- Exercise repeated tensor operations, mock restoration after exceptions, and an existing real module.

Validation on Linux, Python 3.13.13, PyTorch 2.13.0+cpu:

- Baseline: provenance followed by the three tensor test modules fails with two TORCH_LIBRARY collection errors.
- Fixed: the three modules, provenance writer, parallel-run fork-worker tests, and new helper tests pass together: **301 passed**.
- Fresh-process check confirms ordinary import after the first borrow works and three successive mock swaps reuse the real module.
- Applicable Ruff and Ruff-format pre-commit hooks pass. The all-hook launcher encountered an unrelated existing system Node/npm incompatibility while initializing ESLint; the changed files are Python only.
- Full collector collection remains blocked by eight missing `aiconfigurator_core._aiconfigurator_core` extension imports. The unmodified baseline has those same eight errors plus the two PyTorch errors fixed here. Tests were run with `--confcutdir=tests/unit/collector` and local source paths to avoid the root CLI fixture's dependency on that unbuilt extension.

This addresses the PyTorch import collision in #1422 and verifies the affected fork-worker combination on Linux. It does not change the independent GLM checkpoint expectations mentioned there; macOS deadlock behavior was not tested.

#### Review follow-up and CI baseline:

The first-import behavior is intentional: an originally absent `torch` entry remains registered, while an existing entry (including the collector mock) is restored. A fresh-process check with real PyTorch reproduces the native `TORCH_LIBRARY` error if the entry is removed before a subsequent ordinary import. The helper now documents this distinction explicitly, and the regression checks ordinary imports after borrowing. The related combination still passes all **301 tests**, and Ruff/format pass.

The initial PR's [Rust/Python parity job](https://github.com/ai-dynamo/aiconfigurator/actions/runs/34038559202/job/101501018917) has the same two golden-energy failures, including identical actual values, as the [unmodified main job at this PR's base](https://github.com/ai-dynamo/aiconfigurator/actions/runs/33604681100/job/100165870723): 2 failed, 62 passed. These tests execute separately from the modified collector tests. The initial PR's [unit job](https://github.com/ai-dynamo/aiconfigurator/actions/runs/34038559202/job/101501018953) likewise has the same single `test_power_columns_satisfy_energy_model_input_contract` failure as the [baseline unit job](https://github.com/ai-dynamo/aiconfigurator/actions/runs/33604681100/job/100165870879): both have 2750 passing tests (14 skips on the PR, 12 on the baseline). This PR does not modify those data or energy-model expectations.

`tests/unit/collector/_real_torch.py`, then the three migrated consumers.

#### Related Issues:

- Relates to #1422 (part 1; Linux validation of part 2).

## Summary by CodeRabbit

- **Tests**
  - Improved test reliability for scenarios requiring the native Torch module.
  - Tests now preserve and restore the prior Torch environment, including when Torch was absent, mocked, or already loaded.
  - Expanded coverage to verify cleanup after repeated usage and ensure temporary test setup does not affect subsequent tests.
  - Simplified setup across collector and routing test suites while retaining appropriate handling for environments where Torch is unavailable.

### DIFF
--- a/tests/unit/collector/_real_torch.py
+++ b/tests/unit/collector/_real_torch.py
@@ -17,7 +17,11 @@ _real_torch: ModuleType | None = None
 
 @contextmanager
 def real_torch() -> Iterator[ModuleType]:
-    """Cache torch's native registrations and restore the caller's module entry."""
+    """Borrow torch, restoring any existing module entry on exit.
+
+    A first import remains registered when there was no previous entry. Removing
+    it would make a later ordinary import repeat torch's native registrations.
+    """
     global _real_torch
     previous = sys.modules.get("torch")
     try:
--- a/tests/unit/collector/_real_torch.py
+++ b/tests/unit/collector/_real_torch.py
@@ -13,17 +13,14 @@ from unittest.mock import MagicMock
 import pytest
 
 _real_torch: ModuleType | None = None
+_MISSING = object()
 
 
 @contextmanager
 def real_torch() -> Iterator[ModuleType]:
-    """Borrow torch, restoring any existing module entry on exit.
-
-    A first import remains registered when there was no previous entry. Removing
-    it would make a later ordinary import repeat torch's native registrations.
-    """
+    """Borrow torch, restoring the original module state on exit."""
     global _real_torch
-    previous = sys.modules.get("torch")
+    previous = sys.modules.get("torch", _MISSING)
     try:
         if _real_torch is None:
             if isinstance(previous, MagicMock):
@@ -35,5 +32,7 @@ def real_torch() -> Iterator[ModuleType]:
         sys.modules["torch"] = _real_torch
         yield _real_torch
     finally:
-        if previous is not None:
+        if previous is _MISSING:
+            sys.modules.pop("torch", None)
+        else:
             sys.modules["torch"] = previous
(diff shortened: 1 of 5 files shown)


git-jxj:
Rechecked the failing CI on current head `a97c69677af2` against the latest completed main Build and Test run, [34306677905 at `f254959eb89e`](https://github.com/ai-dynamo/aiconfigurator/actions/runs/34306677905). The [unit failure](https://github.com/ai-dynamo/aiconfigurator/actions/runs/34320781446/job/102366698459) has exactly the same 17 assertion lines for invalid `power_limit` values in 15 B200/TRTLLM data files. The [parity failure](https://github.com/ai-dynamo/aiconfigurator/actions/runs/34320781446/job/102366698485) has the same four assertion lines for the two golden-energy mismatches (443.84428875568517 and 1241.060314309411). No new failure signature appears in this PR. Could a maintainer confirm the upstream data/golden repair path and rerun these checks once that baseline is repaired?

PR #1584, opened 2026-08-22
chore(data): add matched B200 TRT-LLM 1.3.0rc20 power

kaim-eng:
> [!IMPORTANT]
> ## Dependency and zero-sentinel update
>
> This pure-data PR is paired with the loader/packaging fix in #1590. Merge #1590 first, then rebase/update this branch onto the resulting `main` before merging.
>
> Current signed-off data head: `6b1f4b78941b6362f9202a554a50cc35091d2fb1`.
>
> The 40,549 unmatched rows now carry typed `0.0` / `0.0` sentinels. Any earlier diff or CI output showing blank/NULL power cells describes a pre-remediation commit and is superseded. All logical keys, stored latency values, non-power Arrow values, and physical row order are unchanged.

## Summary
- add matched `power` and `power_limit` measurements to all 18 existing B200 TRT-LLM 1.3.0rc20 parquet tables
- encode unmatched rows as typed `0.0` / `0.0` sentinels
- preserve all stock rows, identities, latency values, and non-power Arrow values
- keep this as a pure-data PR based directly on `main`

### Broad TRT-LLM campaign
- machine: x86_64, 8x NVIDIA B200
- image: `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc20@sha256:1532b38814b3faf2affdb5ef01ca91468685d314ffb7e8926a0567595355ed88`
- collector revision: `0c6d722ecbbbce39c477adceae2687eb7245d22c`
- all four collection/resume jobs completed successfully

### Focused MHC follow-up
- same pinned runtime image and x86_64 8x B200 environment
- exact recorded MHC collector revision: `8cb6658da404369d68dec6d61cb2f7bc646d8470`
- Compute Lab node: `umbriel-b200-049`; Slurm job `3883216`; exclusive 8-GPU allocation; completed `0:0` in 2m56s
- one `mhc_module` pass generated 140 cases and produced 139 valid rows
- all 139 produced rows exactly match the 139 stored MHC identities; 0 duplicates and 0 candidate-only rows
- the sole failed case was the extra `mhc_pre_hs7168_hcm4_524288` case, which OOMed during TRT-LLM autotuning and is not present in the stock table
- median absolute latency drift: 2.31%; every accepted row is within the 25% per-row gate

## Packaging policy
- exact identity match excluding `latency`, `power`, and `power_limit`
- candidate metrics must be finite with latency > 0, power >= 0, and power_limit > 0
- require power <= 1.05x power_limit
[...]

PR #1610, opened 2026-09-08
test: restore B200 power CI contract

jasonqinzhou:
## Summary

- restore the shipped B200 power-data invariant: rows must be either a finite positive `power` / `power_limit` measurement pair or the exact `0.0` / `0.0` unavailable-measurement sentinel
- refresh the two affected TRT-LLM current-slot per-op goldens with the repository pinning tool and activate their anti-vacuous energy coverage
- keep this maintenance-only under the AIC code freeze; no parquet data or runtime behavior changes

Linear: [AIC-1940](https://linear.app/nvidia/issue/AIC-1940/cidata-restore-aic-b200-power-data-invariant-and-engine-parity)

Related: #1584, #1590, [AIC-1934](https://linear.app/nvidia/issue/AIC-1934/power27data-close-collector-dataset-and-power-support-coverage-gaps)

## Why

PR #1584 added matched B200 TRT-LLM `1.3.0rc20` power data with 487,422 measured rows and 40,549 intentional `0.0` / `0.0` sentinels for unmatched rows. A later invariant test rejected every non-positive `power_limit`, while the compile-engine goldens still pinned the pre-power current slot. That merge-order drift made `main` fail despite the shipped data satisfying its documented paired-value contract.

## Validation

- `pytest tests/unit/sdk/database/test_power_data_invariants.py`: 2 passed
- compile-engine parity: 64 passed
- full engine-step parity: 276 passed
- Rust crate and public API tests: 539 passed, 2 ignored
- broader Python unit/build suite: 4,857 passed, 11 skipped
- `ruff check .`: passed
- `ruff format --check .`: 678 files already formatted
- golden audit: 29 intended `energy_wms` updates, two provenance SHA updates, zero other case-field changes, zero parquet changes

The broad local Python lane excludes `tests/unit/sdk/database/test_moe_dispatch.py` because the repository's macOS development environment does not install Torch; hosted container CI remains authoritative for that test.

## Summary by CodeRabbit

- **Tests**
  - Expanded energy-measurement validation to require complete, valid power data and support explicit zero-value cases.
  - Added focused coverage for accepted and rejected power measurements.
  - Extended performance parity checks to cover energy reporting across additional model operations and routing scenarios.

- **Chores**
  - Refreshed benchmark reference data and compatibility baselines to reflect current energy measurements and runtime versions.

### DIFF
--- a/tests/unit/sdk/database/test_power_data_invariants.py
+++ b/tests/unit/sdk/database/test_power_data_invariants.py
@@ -6,16 +6,19 @@
 Policy (2026-08): power/energy tests pin no values and bind to no specific
 backend version. They assert query-surface invariants over WHATEVER power
 data is currently shipped: every parquet that carries power columns must
-satisfy the energy model's input contract (finite, non-negative power;
-positive power limit). The energy MATH is anchored by the rust synthetic
-oracles on power-carrying fixtures (``energy_test_fixtures`` tests in
-``operators/{gemm,attention}.rs``); this test guards the shipped data plane
-those models consume. If no power-carrying parquet is shipped at all, the
-suite records that state explicitly instead of passing vacuously.
+store either a finite positive ``power`` / ``power_limit`` measurement pair
+or the exact ``0.0`` / ``0.0`` unavailable-measurement sentinel. The energy
+MATH is anchored by the rust synthetic oracles on power-carrying fixtures
+(``energy_test_fixtures`` tests in ``operators/{gemm,attention}.rs``); this
+test guards the shipped data plane those models consume. If no power-carrying
+parquet is shipped at all, the suite records that state explicitly instead
+of passing vacuously.
 """
 
 from pathlib import Path
 
+import numpy as np
+import pandas as pd
 import pyarrow.parquet as pq
 import pytest
 
@@ -36,6 +39,15 @@ def _power_carrying_files() -> list[Path]:
     return files
 
 
+def _invalid_power_pairs(frame: pd.DataFrame) -> pd.Series:
+    power = frame["power"]
+    power_limit = frame["power_limit"]
+    finite = np.isfinite(power) & np.isfinite(power_limit)
+    measured = (power > 0.0) & (power_limit > 0.0)
+    unmeasured = (power == 0.0) & (power_limit == 0.0)
+    return ~(finite & (measured | unmeasured))
+
+
 def test_power_columns_satisfy_energy_model_input_contract():
     files = _power_carrying_files()
     if not files:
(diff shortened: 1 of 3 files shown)


tests/unit/sdk/database/test_power_data_invariants.py
```diff
@@ -36,21 +39,44 @@ def _power_carrying_files() -> list[Path]:
     return files
 
 
+def _invalid_power_pairs(frame: pd.DataFrame) -> pd.Series:
+    power = frame["power"]
+    power_limit = frame["power_limit"]
+    finite = np.isfinite(power) & np.isfinite(power_limit)
+    measured = (power > 0.0) & (power_limit > 0.0)
+    unmeasured = (power == 0.0) & (power_limit == 0.0)
+    return ~(finite & (measured | unmeasured))
```
Arsene12358, line 48: **[P2] Reject nullable missing values in the invalid-pair mask**

When a Parquet file preserves pandas nullable `Float64` columns, `np.isfinite` and the comparisons above propagate `pd.NA`. This return expression then produces missing booleans, which `bad.any()` at line 66 skips. Consequently, the invariant accepts null measurement pairs instead of rejecting them.

I reproduced this through the complete invariant function, pointing `_DATA_ROOT` at a temporary directory containing a real `nullable_perf.parquet` written from:

```python
frame = pd.DataFrame(
    {
        "power": [None, 100.0, 100.0],
        "power_limit": [1000.0, None, 1000.0],
    },
    dtype="Float64",
)
frame.to_parquet(tmp_path / "nullable_perf.parquet", engine="pyarrow")
```

After the test's normal `pq.read_table(...).to_pandas()` call, the mask is `[pd.NA, pd.NA, False]` and `bad.any()` is `False`. **Base `77fd077` rejects the two invalid rows; head `7e754583` passes the same file.** With ordinary `float64`, both revisions reject it, which explains why the added synthetic examples miss the regression.

Please treat missing mask entries as invalid and add a nullable-Parquet round-trip regression test. This minimal correction makes the complete invariant reject both invalid rows while preserving the existing synthetic test results:

```suggestion
    return (~(finite & (measured | unmeasured))).fillna(True)
```

This is an introduced regression in the guard required by AIC-1940, not a claim that current shipped data contains null pairs.

Arsene12358 (changes requested):
Requesting changes for one P2 regression in the power-data invariant: nullable numeric Parquet columns can cause null measurement pairs to pass the check. This violates AIC-1940's explicit fail-closed requirement; the inline comment includes a base/head reproduction and the minimal correction.

Validation: 342 local invariant/compile-engine/engine-step tests passed, Ruff checks passed, and both refreshed per-op golden records exactly matched the repository pinning tool's live computation. Current committed power data is valid; this finding concerns the invariant's handling of invalid input.

